PR #11, opened 2018-10-12
Added return in components dock

conblem:


documentation/components.md
```diff
@@ -41,7 +41,7 @@ import { State, root } from 'solid-js'
 function Component(fn) {
   state = new State({});
   props = new State({});
-  fn({state, props});
+  return fn({state, props});
```
conblem: Wouldn't you need to return the component to later mount it?

ryansolid: Yes absolutely. Thank you.

Issue #22, opened 2019-04-17
Typescript definitions


r0skar:
Hi! Thanks for sharing your work with us! I was wondering if you have any plans to provide typescript definitions? I started to really like typescript, but I am not yet capable of writing definitions for a complex framework like yours.

ryansolid:
I've been waiting for this issue to show up.

I've known there would be a desire and I've actually attempted to rewrite my libraries in TypeScript on a couple occasions, because I believe for a library having that API surface covered is invaluable. For some reason I've really struggled with this though(unlike Flow) and admittedly I haven't invested enough time. But I also haven't not invested a good chunk of time.

I'd love to help out here but it might be a slow go. But now that there is a standing issue I will probably get to this again sooner than later, or at least put it out there to get some assistance. I think it is pretty likely I will move the whole set of libraries to TypeScript at some point if I can ever get past the continually horrendous experiences I have every time I give it a chance.

r0skar:
>  I believe for a library having that API surface covered is invaluable.

That's so very true. It is so much easier to get into a new library when the Editor/IntelliSense is doing all of the guess work about available methods and exports.

> I think it is pretty likely I will move the whole set of libraries to TypeScript at some point if I can ever get past the continually horrendous experiences I have every time I give it a chance.

IMO, the great thing about Typescript is, that basically any JS code is valid TS code as soon as you change the file extension to `.ts`. I have tried to port this repo to TS and it works without any modifications: https://github.com/r0skar/solid. The few reported type errors are minor. I´d be happy to keep hacking at this.

Since the `S` library is using Typescript, IntelliSense did already pick up a few warnings (Should this be a new Issue?):

```ts
// src/state.ts (Line 46 & 50)
// S.js requires an argument for makeDataNode()
node = nodes._self || (nodes._self = S.makeDataNode());

// src/dom/index.ts (Line 18 & 36)
// S.js requires a seed value when the passed fn has an argument.
S.makeComputationNode(element => {})
```

ryansolid:
Ok yeah I hit those warnings too. It's because I don't explicitly pass undefined in both scenarios. They are fairly trivial to address. But I see you have no type annotations yet. I was using `strict: true`.  It's also possible to start loose and then make it more strict over time but I don't really see the benefit of TypeScript in that case. So I was getting annoyed pretty quick having to cast Node to HTMLElement back and forth for like nextSibling calls and in the end still having it complain.

[...]

PR #23, opened 2019-04-22
Add Babel and TypeScript

neodon:
- Enables Babel with the TypeScript preset and integrates with Rollup through a plugin.  This results in all TypeScript being stripped out during the build--it doesn't use the TypeScript compiler.
- Adds the TypeScript compiler *only* for emitting type definitions.
- Converts some top-level modules to TypeScript with the smallest changes needed to get them to compile.


ryansolid:
Thanks

neodon:
I meant to mention the issue (#22) as well.

PR #24, opened 2019-05-06
Fix typos/spelling

nihakue:
Was reading through the docs and getting excited about the project, but got distracted by the `it's/its` typos.

It's is a contraction of 'it' and 'is', 'its' is defined as: "belonging to or associated with a thing previously mentioned or easily identified"


ryansolid:
Yes definitely. This was just sloppy. Thank you.

PR #25, opened 2019-05-06
[@types] Only wrap state props of type object

r0skar:
Objects in the state are wrapped, but simple primitives are not. The updated typings now reflect this behavior.


ryansolid:
Thanks for the update.  I'm working on some other stuff so not quite sure when I will push to npm. But I will  update this issue when I do.

PR #47, opened 2019-07-18
Improved wrapped type definitions (support for Date and functions)

bluenote10:
The `Wrapped<T>` type definition is still missing a few pieces to fully work in general. I first noticed this when working with `Date`, which always required manual casts because of the `Symbol.toPrimitive` symbol. I had to [ask for help](https://stackoverflow.com/questions/57100616/how-to-forward-well-known-symbols-in-a-type-wrapper-mapped-type) in order to come up with a solution. To make other well-known symbols work, we could either add more explicitly forwarded definitions, or wait until the [upstream TS issue](https://github.com/microsoft/TypeScript/issues/24622) gets a release (had recent activity, so may be available soon).

I also may have to refine the forwarding of `AddCallable`, but if you don't mind I'd first experiment with this change for a while and make necessary improvements step-by-step.

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -16,8 +16,8 @@ type AddSymbolToPrimitive<T> = T extends
   {[Symbol.toPrimitive]: V;} :
   {};
 type AddCallable<T> = T extends
-  {(...x: any): infer V;} ?
-  {(...x: any): V;} :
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks for looking into this. We may have clean type defs one day.

PR #48, opened 2019-07-21
Improved readability of control flow function signatures

bluenote10:
This PR is just a minor code reformatting. I was trying to read the function signatures on GitHub and it was very hard to read because of having everything in one line:

Peek 2019-07-21 11-10

Since they are public interface good readability would be nice. Also leads to better git diffs, but feel free to reject this PR if the shorter lines are just too far away from your personal taste ;).

The only functional change included in the PR is that I added a `?` to the `transform` of `Switch`. I guess it was not intended to make it mandatory, was it?

### DIFF
--- a/src/dom/index.ts
+++ b/src/dom/index.ts
@@ -13,20 +13,38 @@ export function render(code: () => any, element: Node): () => void {
   return disposer!;
 }
 
-export function For<T, U>(props: {each: T[], fallback?: any, transform?: (mapped: () => U[], source: () => T[]) => () => U[], children: (item: T) => U }) {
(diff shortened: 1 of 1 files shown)


bluenote10:
Looks like something is broken with the style checks. Some of them simply make no sense (non-arrow functions are forbidden). I agree with the complains about "`any` loses type-safety" and "expected call-signature / missing return type", but these problems existed before, but were swallowed by the long lines?

ryansolid:
Yeah don't worry about that. I setup Codacy but I haven't configured it optimally. This is an improvement. It's because I tend to write everything in naked js and then add the types at the end. So the lines aren't long initially. Thank you for taking the time to clean that up.

PR #49, opened 2019-07-21
Removed possibly unneeded anys

bluenote10:
Just trying to make sense of the code... I don't think these `any`s were needed, right?


ryansolid:
I had it that way originally but I wasn't positive that covered all cases since you know it can be almost anything. It's probably fine.

PR #52, opened 2019-07-28
more type safety in control flow elements

bluenote10:
I have adjusted the control flow type signatures based on the refined `JSX.Element` type. The compiler has already revealed a few interesting things:

- I tried to remove most of the `any` in the control flow signatures to actually make use of the compiler.
- In some cases a value `false` can be returned, which was not on the list of allowed values you mentioned. This currently requires `as any as JSX.Element` casts. If this is legal we can add it to the `JSX.Element` definition.
- The show control flow allowed transforms to undefined, which is also not supported so far. I've removed the possibility to have an undefined return value.
- The switch element would have bugs it gets arbitrary JSX children, only one or multiple match elements seem to work. I have narrowed down the allowed children to that.
- The portal element currently doesn't fit into the JSX.Element type. I have left it out so far.
- There was a compilation error in `suspense.ts` which I have fixed by adding NodeJS types (but does `NodeJS.Timeout` work in the browser?).

If you feel uneasy of narrowing down the types at this moment, feel free to keep the PR open for now. I can update it as Solid continues to develop. Just wanted to give you an impression how it could look eventually.

### DIFF
--- a/src/dom/index.ts
+++ b/src/dom/index.ts
@@ -52,9 +52,9 @@ export function Show(props: {
 export function Switch(props: {
   fallback?: JSX.Element;
   transform?: (mapped: () => JSX.Element, source: () => number) => () => JSX.Element;
-  children: JSX.MatchElement[];
+  children: JSX.MatchElement | JSX.MatchElement[];
 }): JSX.Element {
-  let conditions = props.children;
+  let conditions = props.children as JSX.MatchElement[];
   Array.isArray(conditions) || (conditions = [conditions]);
   const useFallback = "fallback" in props,
     evalConditions = createMemo(
--- a/src/dom/index.ts
+++ b/src/dom/index.ts
@@ -21,39 +21,39 @@ export function render(code: () => any, element: Node): () => void {
   return disposer!;
 }
 
-export function For<T, U>(props: {
+export function For<T, U extends JSX.Element>(props: {
   each: T[];
-  fallback?: any;
+  fallback?: U;
   transform?: (mapped: () => U[], source: () => T[]) => () => U[];
   children: (item: T) => U;
-}) {
+}): JSX.Element {
   const mapped = map<T, U>(
     props.children,
-    "fallback" in props ? () => props.fallback : undefined
+    "fallback" in props ? () => props.fallback! : undefined
   )(() => props.each);
   return props.transform ? props.transform(mapped, () => props.each) : mapped;
 }
 
-export function Show<T>(props: {
+export function Show(props: {
   when: boolean;
-  fallback?: any;
-  transform?: (mapped: () => T, source: () => boolean) => () => T | undefined;
-  children: any;
-}) {
+  fallback?: JSX.Element;
+  transform?: (mapped: () => JSX.Element, source: () => boolean) => () => JSX.Element;
+  children: JSX.Element;
+}): JSX.Element {
   const condition = createMemo(() => props.when, undefined, EQUAL),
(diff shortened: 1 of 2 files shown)


ryansolid:
Yeah, this is a tricky one. The more I think about it the weirder JSXElements as a type is. Like, have strict types around specific Components like these control flow makes a ton of sense I generally like the changes. But JSXElements are nothing. What matters is what gets inserted not what JSX Elements return since they can be intercepted by other JSX Elements. Basically, everything until insertion into an actual DOM node can be anything. A JSXElement can return anything. It's no different than if you have a Function A that has expected typed parameters. But as long as you call Function A with the right arguments it doesn't matter what Function C returns, and Function B is the adapter.

```js
a(b(c()));
```
JSXElements/Components are like that. Ultimately I will call insert on it which will have specific inputs like we've been talking about. You are right I missed Booleans. `null`, `undefined`, `true`, `false` all resolve to empty. So I missed that.

So on one hand, I'm fine with these definitions since they work, but I wonder if on the whole we aren't quite doing these right.

ryansolid:
@bluenote10 just had a thought. These methods are also used by tagged template literals and hyperscript versions. Does JSXElement make sense for those?

src/dom/index.ts
```diff
@@ -21,40 +21,40 @@ export function render(code: () => any, element: Node): () => void {
...
-export function Show<T>(props: {
+export function Show(props: {
   when: boolean;
-  fallback?: any;
-  transform?: (mapped: () => T, source: () => boolean) => () => T | undefined;
-  children: any;
-}) {
+  fallback?: JSX.Element;
+  transform?: (mapped: () => JSX.Element, source: () => boolean) => () => JSX.Element;
+  children: JSX.Element;
```
ricochet1k, line 41:
```suggestion
  children: JSX.Children;
```

src/dom/index.ts
```diff
@@ -21,40 +21,40 @@ export function render(code: () => any, element: Node): () => void {
...
-export function Switch<T>(props: {
-  fallback?: any;
-  transform?: (mapped: () => T, source: () => number) => () => T;
-  children: any;
-}) {
-  let conditions = props.children;
+export function Switch(props: {
+  fallback?: JSX.Element;
+  transform?: (mapped: () => JSX.Element, source: () => number) => () => JSX.Element;
+  children: JSX.MatchElement | JSX.MatchElement[];
```
ricochet1k, line 55: `JSX.MatchElement` doesn't seem to exist?

src/dom/index.ts
```diff
@@ -21,40 +21,40 @@ export function render(code: () => any, element: Node): () => void {
   return disposer!;
 }
 
-export function For<T, U>(props: {
+export function For<T, U extends JSX.Element>(props: {
   each: T[];
-  fallback?: any;
+  fallback?: U;
```
ricochet1k, line 26: Forcing the fallback to be the same type as children seems like a bad idea. Have you checked that this actually works? Can fallback be a single element or string while children returns multiple elements?

ryansolid: I was thinking about this and I think I'm going to remove children. Element is a possible multi-type too. I realized the types there were redundant since Element already handles arrays and functions.

ryansolid:
I'm finally here. Some stuff has moved around and updated. But I'm finally at the point to make these sort of changes. I think I've moved too far from this commit now, but it definitely helped. I'm going to close it.

PR #53, opened 2019-07-31
Don’t use optional parameters in callbacks

MarcoPolo:
See this: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks.

(May require more changes, just made this to point this out)

### DIFF
--- a/src/dom/index.ts
+++ b/src/dom/index.ts
@@ -51,7 +51,7 @@ export function Show<T>(props: {
     mapped = () =>
       condition()
         ? sample(() => props.children)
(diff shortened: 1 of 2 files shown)


ryansolid:
I did not know that I will give it a try. This pull request, breaks down stream since I use optional params there too (in the map function) so I can't merge it straight up. If you want to update that too so the build passes that would be great or I will just address this later.

MarcoPolo:
Sure thing, I'll update those

ryansolid:
Thank you!

PR #54, opened 2019-08-03
mismatch obj key

yuens1002:
the passed in todo obj uses `task` as the key. the ex. uses `title`.... a typo?


ryansolid:
Yes, you are correct. Thanks for noticing.

PR #70, opened 2019-09-06
improve typings for setState

high1:
- this should have less friction with TypeScript when using string paths for setState


high1:
Have not been able to fully test it, but should be fine - if you could create a preview build, I'll do the testing.

bluenote10:
By the way, the tests contain some dedicated tests for verifying type definitions. Would be nice to add something for setState as well (was on my to-do list ;)).

high1:
> 
> 
> By the way, the tests contain some dedicated tests for verifying type definitions. Would be nice to add something for setState as well (was on my to-do list ;)).

Will try to think of something.

PR #71, opened 2019-09-06
improve transpiled setState type safety

high1:
- using ```NestedStateSetter<unknown>``` instead of ```NestedStateSetter<any>```

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -42,15 +42,15 @@ type StatePathPart =
 
 // do up to depth of 8
 type StatePath =
-  | [string, NestedStateSetter<any>]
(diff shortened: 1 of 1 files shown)


high1:
Hopefully got this right this time... Sorry for the noise.

ryansolid:
Sounds good to me. Anyone else want to verify this makes sense? @bluenote10? It seems inconsequentially simple to swap any for unknown so I will go ahead shortly if there are no critiques. But I still not clear the difference here. It just means you have to cast the type when assigning to a different variable instead of anything going. Makes sense to me.

PR #88, opened 2019-10-06
Remove lots of type casts

bikeshedder:
I removed almost all explicit type casts (`x as Y`) as they weren't needed or could be expressed in a different way. The code still contains a lot of implicit type casts because `any` is used a lot.


ryansolid:
Thanks so much for taking the time to clean this all up.

PR #107, opened 2019-12-18
Fix selectAll not selecting nodes with a model that is a falsy value, such as 0

boogerlad:


ryansolid:
Model 0 I take it.. Thanks.

PR #124, opened 2020-03-01
Update MountableElement type so it accepts more actually-mountable types

trusktr:
Node is the common class between all mountable nodes. DocumentFragment is a direct subclass of Node for example.


ryansolid:
Sorry updated base DOM Expressions forgot to do it here.

PR #127, opened 2020-03-12
Update old link in the readme

aminya:


ryansolid:
Thanks, there are probably more, if no one else gets to them I will when I move to the 0.17.0 release of solid.

PR #145, opened 2020-04-24
Accurate type checking for setState functions

ricochet1k:
setState will now reject bad keys, or using `{}` when you're not at an array, etc.

It might actually be too restrictive, but it should at least be safe.


ryansolid:
Even if so this is a huge improvement. Thank you so much. I've spent countless hours trying to do this. I'd all but given up. If this was around 6 months ago I may have not adopted the immer style syntax.

This makes me so happy.

PR #154, opened 2020-05-09
Update solid-rx's solid-js peer dependency to the latest version

rvadhavk:
When I install solid-rx 0.18 with solid-js 0.18, yarn gives me the following warning: solid-rx@0.18.0" has incorrect peer dependency "solid-js@^0.15.5".  This change updates the solid-js peer dependency to 0.18 to fix the warning.

### DIFF
--- a/packages/solid-rx/package.json
+++ b/packages/solid-rx/package.json
@@ -26,7 +26,7 @@
     "url": "https://github.com/ryansolid/solid/issues"
   },
   "peerDependencies": {
-    "solid-js": "^0.15.5"
+    "solid-js": "^0.18.0"
   },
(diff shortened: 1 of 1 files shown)


rvadhavk:
Looks like the build failure is unrelated to my change and should be fixed by updating to the latest version of coveralls: https://github.com/nickmerwin/node-coveralls/issues/280

ryansolid:
That is a weird warning since 0.18.0 would satisfy 0.15.5 I'd think.. I guess unless it treats 0. versions differently. But I will keep a closer watch on it. Lerna auto updates other dependencies but I guess not peer dependencies.

PR #155, opened 2020-05-09
Update to latest version of coveralls

rvadhavk:
Current version of coveralls causes a build failure: 
https://travis-ci.com/github/ryansolid/solid/builds/164916703

Upgrading to the latest version should fix it: https://github.com/nickmerwin/node-coveralls/issues/280


ryansolid:
Hmm.. not the first time something has changed and broken it.

PR #156, opened 2020-05-11
make signal setter return the set value

trusktr:
Makes it easier to do things like

```js
if (setFoo(!!someValue)) { ... }
console.log(setBar(123))
```

instead of having to write

```js
setFoo(!!someValue)
if (!!getFoo()) { ... }
setBar(123)
console.log(getBar())
```

### DIFF
--- a/packages/solid/src/signal.ts
+++ b/packages/solid/src/signal.ts
@@ -230,11 +230,11 @@ function readSignal(this: Signal<any> | Memo<any>) {
 }
 
 function writeSignal(this: Signal<any> | Memo<any>, value: any) {
-  if (this.comparator && this.comparator(this.value, value)) return;
+  if (this.comparator && this.comparator(this.value, value)) return value;
   if (Pending) {
     if (this.pending === NOTPENDING) Pending.push(this);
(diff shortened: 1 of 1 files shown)


ryansolid:
Hmm.. I don't see any particular conflict here. As long as know one thinks this would get them something reactive. I don't think I could do the same with the state setter, but that seems fine.

trusktr:
> As long as know one thinks this would get them something reactive. 

You mean, as long as one knows that the set call isn't reactive? I think that should be obvious (at least for us), but perhaps it might be worth noting it with an example somewhere in the docs.

ryansolid:
No I meant more like:
```js
const [s, set] = createSignal(), 
  v = set(3);

// s !== v;
```
I don't think anyone would mistake that for signals. But say state if we ever did the same:
```js
const [s, set] = createState(), 
  v = set(3);

// s !== v;
// s is a proxy, v is an object? not even sure.. actually never would do this
```

trusktr:
> s !== v;

Ooh, riiight. Yeah. Maybe worth a quick mention of that too.

PR #168, opened 2020-06-03
chore(readme) mention command requirements upfront

basarat:
And add an explicit command for TypeScript starter :rose:


ryansolid:
Thank you.

PR #202, opened 2020-07-27
Add detailed SVG types

Gustorn:
I'm planning on doing a larget project in Solid that will use SVG rendering pretty extensively so I thought it would be a good investment to improve the typings a bit. I painstakingly went through the [SVG 1.1 spec](https://www.w3.org/TR/SVG11/) and collected all the relevant categories, presentation attributes and specific attributes for all elements that were currently in the jsx definitions. No additional elements were added. I think this should provide a better TypeScript experience.

### DIFF
--- a/packages/solid/src/rendering/jsx.ts
+++ b/packages/solid/src/rendering/jsx.ts
@@ -2517,21 +2517,7 @@ declare global {
       fontStretch?: string;
       fontStyle?: "normal" | "italic" | "oblique" | "inherit";
       fontVariant?: string;
-      fontWeight?:
-        | "normal"
-        | "bold"
-        | "lighter"
-        | "bolder"
-        | "inherit"
-        | 100
-        | 200
-        | 300
-        | 400
-        | 500
-        | 600
-        | 700
-        | 800
-        | 900;
+      fontWeight?: number | string;
       glyphOrientationHorizontal?: string;
       glyphOrientationVertical?: string;
       imageRendering?: "auto" | "optimizeQuality" | "optimizeSpeed" | "inherit";
--- a/packages/solid/src/rendering/jsx.ts
+++ b/packages/solid/src/rendering/jsx.ts
@@ -2502,25 +2502,27 @@ declare global {
         | "central"
         | "mathematical"
         | "hanging"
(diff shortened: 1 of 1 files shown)


Gustorn:
I'm not sure about the CI, when looking at Travis directly it says it passed.

ryansolid:
It's Codacy I should turn it off. Basically any change makes it complain. Thanks for the PR. I apologize ahead for slow response times on merges I'm moving from Canada to the US this week and will be driving. I should catch up on stuff early next week.

high1:
Nice effort. I see that SVG specification says:

> font-weight = "all | [normal | bold | 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900] [, [normal | bold | 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900]]*"
>     Same syntax and semantics as the ‘font-weight’ descriptor within an @font-face rule.
>     The weight of a face relative to others in the same font family. Takes on the same values as the ‘font-weight’ property with three exceptions:
> 
>         relative keywords (bolder, lighter) are not permitted
>         a comma-separated list of values is permitted, for fonts that contain multiple weights
>         an additional keyword, 'all', is permitted, which means that the font will match for all possible weights; either because it contains multiple weights, or because that face only has a single weight.
> 
>     If the attribute is not specified, the effect is as if a value of 'all' were specified.
>     Animatable: no. 

So, I guess that 

>   fontWeight?:
>         | "normal"
>         | "bold"
>         | "lighter"
>         | "bolder"
>         | "inherit"
>         | 100
>         | 200
>         | 300
>         | 400
>         | 500
>         | 600
>         | 700
>         | 800
>         | 900;

does not account for all, or for multiple font sizes, right? Was this a deliberate decision?

Gustorn:
No, it wasn't, nice catch. I'll just change it to `string`.

ryansolid:
Thank you. This is a lot of good work. I might do some more changes here before I release. It's important to me for the code to work in standard HTML form(for copy paste cases with UX designers etc) which means supporting strings for most numeric attributes because otherwise TypeScript complains.

PR #204, opened 2020-07-30
Improved Show and Match types

Gustorn:
Previous version:
image

New version:
image


ryansolid:
This was my misunderstanding of TypeScript and coercion. I had it that way originally and changed it since I thought it was effectively the same. Thanks for rectifying that.

PR #205, opened 2020-08-01
Fix a typing issue with the Dynamic component

amoutonbrady:
The `component` props of the `Dynamic` component accept a string or a Component and renders accordingly to its type. Unfortunately, Typescript is unhappy when you give it a string. This PR should fix the issue.


ryansolid:
Thank you

PR #211, opened 2020-08-10
Use mount as container if provided for <Portal />

MrFoxPro:


MrFoxPro:
Maybe there should be check if component is text or jsx element?

ryansolid:
Yeah it's a bit more complicated when you consider things like fragment reconciliation. I recently figured out how I can do so without completely taking over the parent but generally these sort of libraries need to be the only thing in their parent. I haven't checked for a while but I'd be surprised if React was any different here given it has the same render mount constraints.

ryansolid:
React does do the extra div. We can do better. I'm going come up with an approach with no container for non-shadow DOM portals that can handle top-level partial reconcilliation.

ryansolid:
There is another issue.. Event propagation. If I get rid of the container and the mount point is on the main path.. like say document.body it will screw up all the event delegation creating an infinite loop. As events will propagate from the portal to the virtual parent back up to document.body which will propagate back to the portal and so on. So instead I think we just make the head the special case and leave everything else as is.

PR #224, opened 2020-08-28
Add ComponentProps<T> type

MrFoxPro:
```ts
export type ComponentProps<T> = T extends Component<infer P>
  ? P
  : T extends keyof JSX.IntrinsicElements
  ? JSX.IntrinsicElements[T]
  : {};
```
It could be used in cases:
```ts
// type ParagraphProps  = JSX.HTMLAttributes<HTMLParagraphElement>
type ParagraphProps = ComponentProps<'p'>; 
/*
type PortalProps  = {
    mount?: Node | undefined;
    useShadow?: boolean | undefined;
    children: JSX.Element;
}
*/
type PortalProps = ComponentProps<typeof Portal>;
```
from react implementation: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L831

### DIFF
--- a/packages/solid/src/rendering/component.ts
+++ b/packages/solid/src/rendering/component.ts
@@ -2,7 +2,16 @@ import { untrack } from "../reactive/signal";
 
 type PropsWithChildren<P> = P & { children?: JSX.Element };
 export type Component<P = {}> = (props: PropsWithChildren<P>) => JSX.Element;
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks for the addition. I think it will be very useful.

PR #241, opened 2020-10-15
Change return type for renderToNodeStream

frenzzy:
...from `ReadableStream<string>` to `NodeJS.ReadableStream`, to fix the following issue:

<img width="711" alt="Property 'pipe' does not exist on type 'ReadableStream<string>'" src="https://user-images.githubusercontent.com/640669/96113356-c24f9c00-0eec-11eb-84f8-6d9758f127b5.png">

By analogy with [DefinitelyTyped/react-dom](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/server/index.d.ts)

### DIFF
--- a/packages/solid/src/server/ssr.d.ts
+++ b/packages/solid/src/server/ssr.d.ts
@@ -1,4 +1,9 @@
+// forward declarations
+declare namespace NodeJS {
+  interface ReadableStream {}
+}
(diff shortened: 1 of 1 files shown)


ryansolid:
Thank you.

PR #260, opened 2020-11-13
Expose The Asynchronous Behavior Of renderStatic

s0kil:
The function is async under the hood, and trying to post-process the generated files is difficult, due to not knowing when the static generation is complete.

### DIFF
--- a/packages/solid-ssr/static/index.js
+++ b/packages/solid-ssr/static/index.js
@@ -1,14 +1,15 @@
 const path = require("path");
-const { execFile } = require("child_process");
+const execFile = require("util").promisify(require("child_process").execFile);
 
 const pathToRunner = path.resolve(__dirname, "writeToDisk.js");
 
-function run({ entry, output, url }) {
-  execFile("node", [pathToRunner, entry, output, url], (err, stdout, stderr) => console.log(stdout));
+async function run({ entry, output, url }) {
+  const { stdout } = await execFile("node", [pathToRunner, entry, output, url]);
+  if (stdout.length) console.log(stdout);
 }
 
-module.exports = function renderStatic(config) {
+module.exports = async function renderStatic(config) {
   if (Array.isArray(config)) {
-    config.forEach(run);
-  } else run(config);
+    await Promise.all(config.map(run));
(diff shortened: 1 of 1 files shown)


packages/solid-ssr/static/index.js
```diff
@@ -1,14 +1,15 @@
...
-const { execFile } = require("child_process");
+const execFile = require("util").promisify(require("child_process").execFile);
 
 const pathToRunner = path.resolve(__dirname, "writeToDisk.js");
 
-function run({ entry, output, url }) {
-  execFile("node", [pathToRunner, entry, output, url], (err, stdout, stderr) => console.log(stdout));
+async function run({ entry, output, url }) {
+  const { stdout } = await execFile("node", [pathToRunner, entry, output, url]);
+  if (stdout.length) console.log(stdout);
```
s0kil, line 8: Do not print empty lines.

ryansolid:
Really good point. I only threw that together to figure out how SSG could work. I need to do better here. I will get this in as part of the next minor release (I'm only a day or so away from releasing). But I definitely am interested in how to make SSG better. I just wrote the simplest script that came to mind. I know this will hit walls with even medium size sites.

s0kil:
@ryansolid
I've noticed that SSG build takes a long time, up to 30 seconds and sometimes times out on Node v12.18.0
Ref: https://app.netlify.com/sites/s0kil-solid-ssg-demo/deploys/5faebc172371a10007054a47

When using the latest Node v15.2.0 the SSG generator behaves as expected.
Ref: https://app.netlify.com/sites/s0kil-solid-ssg-demo/deploys/5faebcc38dfb020008ca82c3

Any idea what it could be?

ryansolid:
Hmm.. it's fine on 14 too.. maybe it hangs. 30 seconds is the timeout time. I wonder if there are features not supported in Node 12.  `globalThis` seems fine.. Do you mean my example in the repo. I see I'm using the optional chaining operator untranspiled in the example. That would hang up Node 12.

s0kil:
Not a big deal, will stick with the latest version.

PR #269, opened 2020-12-01
useMutableState hook for React via createMutable

steve8708:


ryansolid:
Hmm.. only thing is so far I've kept the naming congruent with Solid's naming.. Ie.. if `createState` -> `useState`.  Now it's arguable that I probably should have been more explicit with `createMutable` but I think I want to keep the consistent naming pattern.

steve8708:
ah yes makes sense, updated 👍

packages/react-solid-state/src/index.js
```diff
@@ -23,7 +24,7 @@ let inSolidEffect = false;
 function trackNesting(args) {
   const fn = args[0];
   return [
-    function() {
+    function () {
```
steve8708, line 27: just realized this got added from prettier in my IDE - can revert if not desired

ryansolid: Don't worry about it.

PR #270, opened 2020-12-01
Fix state getter syntax

amoutonbrady:


ryansolid:
I thought I fixed all these. Thanks.

PR #271, opened 2020-12-02
detect the global variable

trusktr:
to support environments that aren't up to date with globalThis yet

### DIFF
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -1,3 +1,7 @@
+import { getGlobal } from "./getGlobal";
+
+export * from "./getGlobal";
+
 export {
   createRoot,
   createSignal,
@@ -40,7 +44,8 @@ declare global {
   var Solid$$: boolean;
 }
 
-if ("_SOLID_DEV_" && globalThis) {
+if ("_SOLID_DEV_") {
+  const globalThis = getGlobal() as any;
   if (!globalThis.Solid$$) globalThis.Solid$$ = true;
   else
     console.warn(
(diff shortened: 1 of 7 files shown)


packages/solid/src/index.ts
```diff
@@ -1,49 +1,54 @@
...
-if ("_SOLID_DEV_" && globalThis) {
-  if (!globalThis.Solid$$) globalThis.Solid$$ = true;
-  else
-    console.warn(
-      "You appear to have multiple instances of Solid. This can lead to unexpected behavior."
-    );
-}
+import { getGlobal } from "./getGlobal";
+
+export * from "./getGlobal";
```
trusktr, line 3: Not sure why the diff is so big. All I did in this file was add the above two lines, and see below.

packages/solid/src/index.ts
```diff
@@ -1,49 +1,54 @@
...
+export * from "./reactive/array";
+export * from "./render";
+
+// handle multiple instance check
+declare global {
+  var Solid$$: boolean;
+}
+
+if ("_SOLID_DEV_") {
+  const globalThis = getGlobal() as any;
```
trusktr, line 48: ...And I add this line and modified the conditional check

trusktr:
After chatting on discord, I agree, users can polyfill `globalThis` if they need.

```js
function getGlobal(): Window | NodeJS.Global {
  if (typeof globalThis !== "undefined") return globalThis as any;
  else if (typeof window !== "undefined") return window;
  else if (typeof global !== "undefined") return global as any;
  else return Function("return this")();
}

window.globalThis = getGlobal()
```

PR #272, opened 2020-12-03
merged dom-expressions into the repo

trusktr:
This PR looks big, but there's lots of CRLF/LF conversions. The main changes are:

- move dom-expressions into the solid repo
- update package.json depdendencies (remove dom-expressions dependencies, add them to local package's package.json as needed) so that they are linked into each dependent by Lerna
- gitignore build artefacts that are currently copied into the source `src` folder
- rollup.config now copies files from the local packages instead of the ones in node_modules.

Other than that, it's only new files from those packages, and all the line ending changes (respecting editorconfig).


trusktr:
Added a new commit. It is easier to check out any new commits individually rather than looking at the `Files changed` tab.

trusktr:
@ryansolid Oh goodness, master conflicts with this PR _everywhere_ due to the line endings.

ryansolid:
I wasn't expecting you to keep adding more stuff. I am not particularly in favor of the first commit. And in parallel started working on the other stuff you mentioned.

trusktr:
Actually, nevemind, I was being hasty, the conflicts are the deleted files only

trusktr:
> I am not particularly in favor of the first commit. 

Something that kept throwing me off is looking for code in Solid's repo that's actually coming from dom-expressions.

I think it can easier to understand if they're in the same repo. After all, they're all made by _solid_ (you and contributors).

ryansolid:
I suppose but the intent is to support multiple libraries. Yes there are only 6 currently to my knowledge and I wrote them all. But I really was intending it to be its own thing. There are many things that I do intentionally not to tie it specifically to Solid etc... Maybe it's time for the experiment to be over but that was idea anyway.

trusktr:
merged master to here, no conflicts.

trusktr:
> I suppose but the intent is to support multiple libraries

Having the packages here in this repo doesn't prevent you from supporting those libs (yours or not).

I think it is really nice stuff actually. Do you think you could (or do you?) support SSR with arbitrary reactive libs plugging into the templates (f.e. mobx-jsx)?

trusktr:
I'm still planning to make `lume-jsx` (or some name) at some point, hooking into the system like mobx-jsx, etc.

trusktr:
Btw, if there's something you like about someone's PR and want to take only that part into master, it can be nice if you let the author know, then they can get their name onto the commits (vs having them sniped and getting no credit).

ryansolid:
Yeah it takes a bit more setup so I haven't bothered but everything is done in a generic way so no reason any of those other libraries couldn't have the same SSR capability. I've done more with Solid like create a static non-reactive runtime for scenarios that don't need it. But ignoring those kind of things it's definitely doable. The reason I stopped was Solid's SSR is hinged on Suspense and I didn't bother implementing it in the other libraries. Although to be fair ignoring Transitions (Concurrent Mode type stuff) we could implement Suspense in the other libraries.

trusktr:
> Although to be fair ignoring Transitions (Concurrent Mode type stuff) we could implement Suspense in the other libraries.

That's awesome. :+1:

PR #273, opened 2020-12-04
fix state getter using `this` syntax

iFwu:
It will throw `TypeError: Cannot read property 'firstName' of undefined` if not removing `user` property.


ryansolid:
Oh.. right.. after I changed it from using external state reference(no longer works). Thank you.

Issue #310, opened 2021-01-16
<Dynamic /> does not work with svg elements


russelgal:
```jsx
const Svg = (p) => <svg {...p}/>
// works
<svg
	fill="none"
	stroke="currentColor"
	viewBox="0 0 24 24"
	stroke-width="2"
	stroke-linecap="round"
	stroke-linejoin="round"
>
	<path d="M12 2L2 7l10 5 10-5-10-5zM2 17l10 5 10-5M2 12l10 5 10-5" />
</svg>
// works
<Dynamic
	component={Svg}
	fill="none"
	stroke="currentColor"
	viewBox="0 0 24 24"
	stroke-width="2"
	stroke-linecap="round"
	stroke-linejoin="round"
>
	<path d="M12 2L2 7l10 5 10-5-10-5zM2 17l10 5 10-5M2 12l10 5 10-5" />
</Dynamic>
// not works
<Dynamic
	component={Svg}
	fill="none"
	stroke="currentColor"
	viewBox="0 0 24 24"
	stroke-width="2"
	stroke-linecap="round"
	stroke-linejoin="round"
>
	<Dynamic
		component="path"
		d="M12 2L2 7l10 5 10-5-10-5zM2 17l10 5 10-5M2 12l10 5 10-5"
	/>
</Dynamic>
// not works
<Dynamic
	component={"svg"}
	fill="none"
	stroke="currentColor"
	viewBox="0 0 24 24"
	stroke-width="2"
	stroke-linecap="round"
	stroke-linejoin="round"
>
	<path d="M12 2L2 7l10 5 10-5-10-5zM2 17l10 5 10-5M2 12l10 5 10-5" />
</Dynamic>
```
https://codesandbox.io/s/solid-dynamic-svg-bug-863ne?file=/index.jsx:123-1365

ryansolid:
Fixed in 0.24.0.

PR #311, opened 2021-01-16
Fix Dynamic component to also render SVG elements

alexmarucci:
This should fix SVG Elements not been rendered correctly in #310 

@ryansolid to review. Not sure adding a "isSVG" prop is the best solution here.
I followed a similar approach used for Portals.

Preview of the new Dynamic Component
https://codesandbox.io/s/solid-dynamic-svg-bug-forked-ttsy8?file=/index.jsx

### DIFF
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -1,4 +1,4 @@
-import { insert, spread } from "./runtime";
+import { insert, spread, SVGElements } from "./runtime";
 import {
   createSignal,
   createMemo,
@@ -26,6 +26,12 @@ export {
 
 export * from "./server-mock";
 export const isServer = false;
+const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
+
+function createElement(tagName: string, isSVG = false): HTMLElement|SVGElement {
+  return isSVG ? document.createElementNS(SVG_NAMESPACE, tagName) :
+                 document.createElement(tagName);
+}
 
 export function Portal(props: {
   mount?: Node;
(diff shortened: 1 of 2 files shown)


ryansolid:
Yeah we were discussing this on discord. The difference here is that if we want to pull in my lookup into the runtime we can cleanly solve this. Dynamic is a bit different than most things so Ive also debated making it part of the comiled output/introducing a syntax since it needs to deal with native els and is different on every platform.

alexmarucci:
Hmm, the only alternative I could  find without introducting extra properties is to look into our SVGElements Set defined in the runtime. In terms of introducing a syntax I feel this being a bit out of my grasp given my unfamiliarity with the project.

If you think this is something I could look at, I may need some guidance or else feel free to close this PR and this can go back in the queue I reckon :) np

Let me know what do you think!

```typescript
import {SVGElements} from 'runtime'

// ... Dynamic
case 'string':
  const isSvg = SVGElements.has(p.component);
  const el =  createElement(p.component, isSvg);
  spread(el, others, isSvg);
  return el;
```

ryansolid:
Yes that is what I meant by pulling in the lookup from the runtime. Let's do that for now and worry about the other stuff later. Any sort of syntax thing I do probably won't be for atleast a month. It will add some size but so be it. I think it's the expectation this would just work.

This looks good. I'm in the middle of doing some SSR stuff but when I get a chance I will merge.

alexmarucci:
Ah nice one! Yeah I wasn’t entirely sure what to look for in the runtime. Good to see I got it at last.

PR #314, opened 2021-01-24
added funding.yml

shadowtime2000:


ryansolid:
Thank you I hadn't looked into this before.

PR #324, opened 2021-01-31
Update 'on' functions typings

Jocaetano:
Changed the typings of the 'on' functions so there is no type error when someone passes more than 5 dependencies.

I also changed two things:
- `prev` now returns `T | undefined` on the single dependency version and `[...T] | []` on the multiple dependencies version.
- `prevResults` can be undefined

I think it reflects better what the js code does, but can be a breaking change for ts users.


ryansolid:
Looks good to me. I will trust you on this. Thanks.

PR #327, opened 2021-02-02
Fix some inaccurate typings and Selector bug

Jager567:
While using this library, I encountered a couple issues, mainly related to types.

- When an initial value is passed to `createComputed`, `createRenderEffect` or `createEffect`, the argument will always be of type `T`, never `undefined`. Additionally, when no initial value is given, there is no reason to restrict the return type to a defined value. The code can handle undefined values just fine. I added overloads to these methods to reflect this in the types.

- The setter returned by `createSignal()` accepts an optional argument, but returns a non-optional type. This is incorrect when passing undefined.

- `createContext()` creates a context with a default value of undefined. However, when used in `useContext()`, it will return a non-undefined type. This is possible because `lookup()` returns `any`. Typescript incorrectly assumed that `any` equals `T` and forgets that it could also be undefined. Therefore the inferred return type is `T`. I work around that by encoding the possible undefined in the `T` stored in `Context<>`.

- A good use case for selectors is using an array index as key. This means that the value `0` is likely to be used quite often. However, due to the use of a falsy check instead of an undefined check, passing 0 will not clean up correctly. I replaced the check with a check for `undefined`.

All these issues have corresponding tests that fail without these changes.

I only changed the types in `/src/reactive/`, not in `/src/static/`. I'm not sure if that's correct, but it seemed all overloads were missing in `/src/static/`, so I left it that way.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -205,7 +205,7 @@ export function createSelector<T, U>(
     (p: T | undefined) => {
       const v = source();
       for (const key of subs.keys())
-        if (fn(key, v) || (p && fn(key, p))) {
+        if (fn(key, v) || (p !== undefined && fn(key, p))) {
           const c = subs.get(key)!;
           c.state = STALE;
           if (c.pure) Updates!.push(c);
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -96,7 +96,7 @@ export function createRoot<T>(fn: (dispose: () => void) => T, detachedOwner?: Ow
   return result!;
 }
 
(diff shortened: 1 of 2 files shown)


ryansolid:
Oh wow. Thank you so much. I've been learning TS as the project progresses and relying a lot on open source contributions. My biggest challenge has been trying to figure out the definitive right way to do things because so many ways sort of work. I will review this tonight.

Jager567:
> My biggest challenge has been trying to figure out the definitive right way to do things because so many ways sort of work.

About that... My change to `createSignal()` seems to have a larger impact than I expected. The setter thinks it could return undefined even when a non-undefined value is passed.

ryansolid:
Ok I'm just going to merge locally and see what I can do. The undefined on signal come from wanting to support void signatures more so than being able to set undefined.

PR #342, opened 2021-02-09
allow import of files directly, as an alternative to the primary export aliases

trusktr:
…, in case someone needs to import `solid-js/web` in Node, or other issues with tools that don't understand Node ESM format

### DIFF
--- a/packages/solid/dev/package.json
+++ b/packages/solid/dev/package.json
@@ -2,6 +2,5 @@
   "name": "solid-js/dev",
   "main": "./dist/dev.cjs",
   "module": "./dist/dev.js",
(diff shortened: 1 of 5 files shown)


trusktr:
When I run code through Jest, for example, it does not know how to make `solid-js/web` to a particular file. So I have to import it as `solid-js/web/dist/web.js`.

That will work in Jest, but with Node ESM or Webpack, it will be forbidden because the direct file is not exposed by `exports`.

So this PR makes it possible to import files directly so that things work in both places that don't know Node ESM (like Jest) or NodeESM/Webpack.

PR #346, opened 2021-02-09
apparently the package.json:types field is needed for places with nested package.json files…

trusktr:
…, but only the exports field should be at the top level of the published package


ryansolid:
Thank you so much.

trusktr:
No prob!

PR #386, opened 2021-03-29
Add type for styled props and return

mmamedel:
Working with the styled API I missed having types in JSX code.
Example:
```ts
const PlayButton = styled('button')<{paused: boolean}>(props => ({
  background: props.paused ? 'green' : 'yellow' // <-- types work
}))

export const App = () => (
  <PlayButton
    className={`${buttonCss}`}
    paused={state.paused}
    onClick={paused.toggle}
  >
    {state.paused ? 'Start' : 'Pause'}
  </PlayButton>
) // <-- All types didn't work
```

With this change now props are typed and the return is a JSX.Element


ryansolid:
Awesome that looks great. I had some issues with global JSX types and Goobers own types so I had to hack it together a bit. Always welcome type improvements. Much appreciated.

PR #401, opened 2021-04-11
Proposal: state notifiers.

MrFoxPro:
By state notifiers I mean some callbacks, that is being executed when observable object changes with all information about change. 

Signature:

It could be very useful for: 
- optimizing exisiting web apps by reducing sent data size
- fast, live web applications with state syncronization between servers and clients (games, dashboards)

Signature: 
```js
createNotifier(<reactive state property>, (newValue, oldValue, path, targets)=>{ /**/ });
newValue: newValue
oldValue: value of property before applying
path: array of string and number that provides path to property
targets: array of values creates path to property
```

### DIFF
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -26,7 +26,7 @@ export {
 } from "./reactive/signal";
 export type { Resource } from "./reactive/signal";
 
-export { createState, unwrap, $RAW } from "./reactive/state";
+export { createState, createNotifier, unwrap, $RAW } from "./reactive/state";
 export type { State, SetStateFunction } from "./reactive/state";
 export * from "./reactive/mutable";
 
--- a/packages/solid/src/reactive/state.ts
+++ b/packages/solid/src/reactive/state.ts
@@ -166,12 +166,14 @@ export function setProperty(
     node;
   (node = nodes[property]) && node.set();
   notify && (node = nodes._) && node.set();
-  traversedPath && traversedTargets && notifiers?.forEach(notifier =>
-    notifier(traversedPath, traversedTargets, StateChangeType.Property, {
-      oldValue,
-      newValue: value
-    })
-  );
+  traversedPath &&
+    traversedTargets &&
+    notifiers?.forEach(notifier =>
+      notifier(traversedPath, traversedTargets, StateChangeType.Property, {
+        oldValue,
+        newValue: value
+      })
+    );
 }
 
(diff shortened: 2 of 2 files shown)


MrFoxPro:
I think it is worth to add disposing of Notifiers, but I don't know how to deal with it.

ryansolid:
This is interesting and a nice chunk of extra code, but I think it is possible to write this in a way that is more or less completely dead code eliminatable.

Traversed Targets. What do you use this for?

What I keep coming back to is the internal library code can be simpler and we extract out the other logic. Like that array diff is very rudimentary and pulling yet another into everyone's `setState` seems like a lot. Notifiers can do whatever with that information. As for the subscribe/dispose just have each create return a dispose function.

Still thinking about what the best API is here. It is global since it intercepts any state setter a top level API makes sense. But it would be even better if we had a way to track the single `setState` function. Maybe we could pass it as an option to `createState`.  And then close over it in the setState function. Downside of that thought it wouldn't be code eliminatable, which would be ok if it was super rudimentary, like didn't add anything additional. The positive is it could be scoped per state object. Then the API would be like:

```js
const [state, setState] = createState(init, { onChange: (value, traversedPath) => {} })
```
In this case I wouldn't even worry about disposal as it is tied to the specific state object. Thoughts? Is it worth talking about how the traversed path is in reverse? I did that to not have to dig for length but maybe that isn't intuitive.

ryansolid:
My idea won't work with state modifiers that call setProperty directly, so that's out. Too bad too since it would have been a lot cleaner. 

MobX has: https://mobx.js.org/intercept-and-observe.html. But not quite what we are looking at I think as it isn't deep.

ryansolid:
I want a solution here but I want something more on the outside. Ideally it would work with state modifiers and be per state object.

I have a crazy idea. Not necessarily suggesting this library. But what if we did something like this: https://codesandbox.io/s/observing-state-6epzp?file=/index.js

Patching setProperty in all cases requires a bunch of passing values around. But a proxy does it easily.

tomByrer:
@MrFoxPro Had any more progress on this?  I can see some use cases myself....

ryansolid:
It's more that I was hesitant to expand the core and was looking for solutions to do this without doing so. Then I put out the idea that we could do this outside of the library. We could look at having a similar module ourselves but I want to keep it outside the base store mechanism. Some sort of extension that was opt in.

PR #406, opened 2021-04-16
fix: don't polyfill Symbol.observable & export ObservableObserver type

thefliik:
This PR stops `solid-rx` from polyfilling `Symbol.observable` if it doesn't already exist. Currently, because `solid-rx` polyfills `Symbol.observable`, it can mess with other libraries (e.g. RXJS) and causes the order of imports to matter (e.g. `solid-rx` must come before RXJS to ensure they both use the same `Symbol.observable` value).

This new strategy for grabbing the value of `Symbol.observable` is the same one used by RXJS ([source](https://github.com/ReactiveX/rxjs/blob/df51b04d7ec68a72b3a4b0d69c3bb29264c72611/src/internal/symbol/observable.ts#L2)).

Separately, this PR also exports the `ObservableObserver` type.


ryansolid:
Thanks

PR #407, opened 2021-04-17
add createGlobalStyles to solid-styled-components

AequoreaVictoria:
The `glob` method currently exported is [apparently the older way](https://github.com/cristianbote/goober#createglobalstyles) of attaching global styles with goober. The `createGlobalStyles` method provided by goober directly references its own `styled`, so this patch recreates it along with a needed update to `styled`.

### DIFF
--- a/packages/solid-styled-components/README.md
+++ b/packages/solid-styled-components/README.md
@@ -137,6 +137,28 @@ glob`
 `;
 ```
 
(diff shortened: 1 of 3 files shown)


ryansolid:
Oh cool. I haven't been on top of this much recently. Thank you for the PR.

PR #415, opened 2021-04-21
docs: improve `props` description

thefliik:
This PR attempts to improve the description of `props` in the docs by adding additional examples and a lengthier introduction. I also rearranged the `props` and `children` sections. Since children is a property on `props`, I think it makes sense to introduce `children` after introducing `props` itself.

### DIFF
--- a/documentation/components.md
+++ b/documentation/components.md
@@ -1,6 +1,6 @@
 # Components
 
-Components in Solid are just Pascal(Capital) cased functions. Their first argument is an props object and return real DOM nodes.
+Components in Solid are just Pascal(Capital) cased functions. Their first argument is an props object and they return real DOM nodes.
 
 ```jsx
 const Parent = () => (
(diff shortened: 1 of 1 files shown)


documentation/components.md
```diff
@@ -11,75 +11,137 @@ const Parent = () => (
   </section>
 );
 
-const Label = ({ greeting, children }) => (
```
thefliik, line 14: I don't think "good" examples in the docs should ever use object destructuring for props.

ryansolid: I agree I honestly don't know how I let that one sneak in. I did start at points doing a lot of copy and paste from React docs.

documentation/components.md
```diff
@@ -11,75 +11,137 @@ const Parent = () => (
...
+const Label = (props) => (
+  <>
+    <div>{props.greeting}</div>
+    {props.children}
+  </>
+);
 ```
 
-## Children
+Unlike in some other frameworks (e.g. React), you cannot use object destructuring on the `props` of a component. This is because the `props` object is, behind the scenes, a Proxy object that relies on getters to lazily retrieve values. Using object descructuring breaks the reactivity of `props`.
```
thefliik, line 63: I'm not sure how accurate this description of the `props` implementation is, but I think it paints a clearer picture than the current description in the docs.

pavulon: To be precise, props are not Proxies unless you use `mergeProps`/`splitProps`. Normally they are just plain objects with regular properties (for static props) and getters (for reactive props), but that's probably too much detail for this doc.

documentation/components.md
```diff
@@ -11,75 +11,137 @@ const Parent = () => (
...
 
-// will not update on name change and pass by value
-const { name } = state;
-<DynamicComponent name={ name }/>
+  <section>
+    <Label greeting={value}>
+      <div>John</div>
+    </Label>
+  </section>
+);
```
pavulon, line 49: Small syntax mistake here:
```diff
- const Parent = () => (
+ const Parent = () => {
    let value = Math.random().toString();

+   return (
      <section>
        <Label greeting={value}>
          <div>John</div>
        </Label>
      </section>
    );
+ };
```
Also IMO this isn't a great example for dynamic props (in reactive terms), because this prop won't change, but ¯\\\_(ツ)_/¯

thefliik: I completely agree. I'll admit it was a bit of laziness on my part. I figured the point would be made that you can pass variables via props. I figure the docs are an iterative process, and at some point in the future they can be improved further.

Issue #418, opened 2021-04-22
potential bug: cannot run tests


thefliik:
Attempting to clone and run this repo's tests locally fails because [this file](https://github.com/ryansolid/solid/blob/main/packages/solid/src/index.ts#L38) cannot be found. My guess is that you have that file defined locally, but it is untracked in git?

More broadly, it would be helpful if a [`.devcontainer.json` file](https://code.visualstudio.com/docs/remote/devcontainerjson-reference) was added for this repo to make getting up and running easier, but I notice that you've used npm v7 which doesn't have a ready made `.devcontainer.json` config template available. Is npm v7 important or incidental?

ryansolid:
Run build before you run tests. Solid builds itself from my library DOM Expressions. Because TS doesn't let you bundle type declarations I manually copy the type files as part of the build process.

Run `npm run build` and you should be able to run the tests after.

thefliik:
Gotcha. I've run into some build errors, but that could be the result of my using npm v6. Regardless, I'll figure it out.

ryansolid:
I've heard there are issues with npm 7.. but I've been using 6.

thefliik:
Interesting, [the lockfile](https://raw.githubusercontent.com/ryansolid/solid/main/package-lock.json) in this repo is listed as `"lockfileVersion": 2`, which I believe is specific to npm v7.

It looks like this shouldn't matter though, as lockfile v2 is considered backwards compatible with older npm clients according to [this npm blog post](https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major).

ryansolid:
Yeah you're right I do appear to be on 7.

PR #419, opened 2021-04-23
add devcontainer

thefliik:
So I've no idea if this is a PR you'll be interested in merging, but this adds a [`.devcontainer.json` file](https://code.visualstudio.com/docs/remote/containers) to this repo to make contributing easier. It uses the standard node + typescript devcontainer docker image that Microsoft publishes. This PR also adds a very simple `CONTRIBUTING.md` file which simply provides instructions on using the devcontainer if someone is unfamiliar.

The short version is:

```md
1. Open [VSCode](https://code.visualstudio.com/) on your computer.
2. Install the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) official extension. Follow the installation instructions for the extension.
3. Clone this repo and open it on your computer. You should then be prompted to "Reopen in container" (do so).
4. After the container has been built, open the VSCode integrated terminal and run `npm install` to install all the packages.
5. After package installation is complete, try running `npm run test`. They should all pass.
```

This post also updates the `package-lock.json` file because the devcontainer uses npm `v6` and, as noted in #418, this repo is current build using npm `v7`.


ryansolid:
On the surface seems fine. My only concern is maintenance. While I have no problem adding something like this, it is highly unlikely I would look into issues that arose. So sort of split. I could just merge and say whatever but realistically I am not going to keep this up to date at this point.

Also, npm 7 is because I'm on Node 15 because I have an M1 Macbook. So I'm pretty stuck on that.

thefliik:
@ryansolid that's completely understandable. As you point out, it would be important for it to be kept up-to-date so I think you'd need to commit to using the devcontainer yourself in order for it to be useful. It's very reasonable if you're not interested in doing that. Though I'll point out that the point of a devcontainer is normalizing the dev environment. This devcontainer would allow you to use node 12 and npm v6 on your computer. At least, it will in theory.

jorroll:
For those seeing interested, now that node 16 has been released [microsoft's official `typescript-node` devcontainer](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/typescript-node) is provided with node versions 12, 14, and 16. 

Node 16 comes with npm v7 installed by default and is probably most appropriate for contributing to Solid.

PR #421, opened 2021-04-27
feat: ignore-devcontainer

thefliik:
As an alternative to #419, this PR adds `/.devcontainer.json` and `/devcontainer` to the `.gitignore` list. This seems appropriate if a devcontainer isn't going to be added to the repo at this time. In the future this can always be undone, but in the meantime it will make it easier for contributors to use `devcontainers` on their own.


ryansolid:
Thanks

PR #424, opened 2021-04-28
Update package.json

emmreonal:


ryansolid:
Thank you for making a PR.

We use Lerna to manage the monorepo. The top level package is private to prevent accidental publishing. The top level in the github never needs to be published. You will see similar pattern in Inferno(https://github.com/infernojs/inferno/blob/master/package.json) and Lerna's own repo(https://github.com/lerna/lerna/blob/main/package.json)

PR #429, opened 2021-04-29
fix some typos

modderme123:


ryansolid:
Not to be annoying but can you do this sort of PR against next branch. I need to keep merging stuff in from main and it would be better than DOC type PRs go against that.

modderme123:
Did that work?

ryansolid:
Looks like it. Thanks.

PR #449, opened 2021-05-12
Update reactivity.md to add missing import for `createEffect`

wendtcode:
Add missing import for `createEffect`


wendtcode:
Noticed that you've got a `next` branch, will submit new PR against that.

PR #450, opened 2021-05-12
Update reactivity.md to add missing import for `createEffect`

wendtcode:
add missing import for `createEffect`


ryansolid:
Thank you. Normally I would say do it against main, but just prepping docs for the 1.0 release so this is good timing. Thank you for the assist.

PR #470, opened 2021-05-27
Handle errors in batch()

romgrk:
Fixes the error shown in https://stackblitz.com/edit/solid-batch-issue?file=index.ts

I've update the code mostly as was discussed on discord. I've also made sure to rethrow the error as to allow the users to deal with errors on their side.

I've added tests, however I wasn't able to make them fail like the code example shows :| So they're crap, I'm not sure why, maybe you see what I'm doing wrong.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -238,12 +238,9 @@ export function createSelector<T, U>(
 export function batch<T>(fn: () => T): T {
   if (Pending) return fn();
   let result;
-  let error;
   const q: Signal<any>[] = Pending = [];
   try {
     result = fn();
-  } catch (e) {
-    error = e;
   } finally {
     Pending = null;
   }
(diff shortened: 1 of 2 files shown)


ryansolid:
I believe you can just skip catching the error and do the finally and not need to rethrow. It should just do the finally and continue going to where every you're catching it in your code. The code I posted with discord I actually inlined in your example and it seemed to work fine as is.

romgrk:
Ok, but that would ignore the updates that happened before the error was thrown inside `batch()`, wouldn't it? Not saying it's the right behavior, just wanna understand correcly everything that's going on.

ryansolid:
My thinking is that if an error happens things are pretty unpredictable anyway. The group of changes probably should be thrown away anyway as they were designed to change together anyway. We shouldn't commit any value. Think of it as a transaction.

The way batch works is it doesn't commit the updates until after the end user supplied function executes and then propagates all updates at the same time. So if it errors during that, I'd say it makes sense to not commit any of the changes. The idea of this being a single changeset means that I think that doing up to that point wouldn't be the desired behavior.

ryansolid:
Looks good to me. Thank you.

PR #495, opened 2021-06-19
improve typings for state setter by utilizing readonly

cynecx:
I am not that familiar with typescript's advance type system. The `DeepReadonly` utility was taken from: https://pastebin.com/Z8YSdThR.


cynecx:
Seems to have some issues (with `reconcile`)

ryansolid:
Thank you.

PR #501, opened 2021-06-22
fix: add source entry and includes src files

aminya:
See https://github.com/solidjs/solid-app-router/pull/17#issuecomment-865675780

This mostly works, but there is a `dom-expressions` dependency that is not declared anywhere in package.json of `solid-js`, and makes issues for the direct source build. I don't know how you publish the packages, so this might be an issue with the way I use `npm pack`.

```js
× Build failed.
@parcel/core: Failed to resolve 'dom-expressions/src/client' from './node_modules/.pnpm/local+C+solid+solid+packages+solid+solid-js-1.0.0-rc.7.tgz/node_modules/solid-js/web/src/client.js'
C:\test\node_modules\.pnpm\local+C+solid+solid+packages+solid+solid-js-1.0.0-rc.7.tgz\node_modules\solid-js\web\src\client.js:1:15
> 1 | export * from "dom-expressions/src/client";
>   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ERROR  Command failed with exit code 1.
```

### DIFF
--- a/packages/solid/h/package.json
+++ b/packages/solid/h/package.json
@@ -2,6 +2,7 @@
   "name": "solid-js/h",
   "main": "./dist/h.cjs",
   "module": "./dist/h.js",
+  "source": "./src/index.ts",
   "types": "./types/index.d.ts",
   "type": "module",
   "sideEffects": false
--- a/packages/solid/html/package.json
+++ b/packages/solid/html/package.json
@@ -2,6 +2,7 @@
   "name": "solid-js/html",
   "main": "./dist/html.cjs",
   "module": "./dist/html.js",
+  "source": "./src/index.ts",
   "types": "./types/index.d.ts",
   "type": "module",
   "sideEffects": false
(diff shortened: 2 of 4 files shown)


ryansolid:
It's a dev dependency because I compile it into the final output using tooling to rewrite its imports. Weird system but I basically need a build system to generate Solid's source and maintain tree-shakeability.

Michael19841984 (approved):

trusktr:
I also trip on the non-standard project layout sometimes. Instead of having a fake `import` statement importing a non-existent `rxcore` package, what if we default it to `solid-js/web` instead, and replace _that_ depending on moduleName?

I'm not sure what issue there was, but I'm guessing it is possible to make this work and keep tree shaking.

LXSMNSYC:
> I'm not sure what issue there was, but I'm guessing it is possible to make this work and keep tree shaking.

IIRC the issue is more of when packing the package and not during building.

ryansolid:
This issue is stale and not going anywhere. Parcel is the only bundler that respects source but it also doesn't transform node_modules. The build process pretty much makes this not a thing. We might revise DOM Expressions at some point but at that time we will have to come up with a plan.

PR #517, opened 2021-07-01
Minor grammar change to rendering.md

Celeo:
Adding "via" and a comma to make the sentence flow a bit better.

The "via" here could also be "with" or "by", up to the reviewer.


ryansolid:
Thanks

PR #531, opened 2021-07-02
Allow generic to be left out in PropsWithChildren type

Jakst:
Many components take only `children` as a prop, and nothing else. Defaulting the generic helper type `PropsWithChildren` to `{}` allows the consumer to use it as simply `props: PropsWithChildren` instead of the awkward-looking `props: PropsWithChildren<{}>`.


ryansolid:
Sound good.

PR #561, opened 2021-07-14
fix: Prevent hashValue from throwing when stringifying bigints

otonashixav:
hashValue uses `JSON.stringify` which throws when it encounters a bigint. This commit fixes that by checking for bigints and converting them to strings and appending `n` when stringifying.


ryansolid:
Thanks

Issue #507, opened 2021-06-28
Export a type for createSignal's setter


otonashixav:
**IMPORTANT: If you have a question or an idea for a new feature use Github Discussions instead**

**Describe the bug**
The type of createSignal's setter is not exported anywhere. This was recently made more problematic due to [https://github.com/solidjs/solid/commit/2f5311dc90f25b8a68de536d7a095112235d385b](https://github.com/solidjs/solid/commit/2f5311dc90f25b8a68de536d7a095112235d385b) which changed the setter's type from a simple setter to one that also accepted a function form, making writing types, for example when passing a setter as a prop, more tedious. 

**To Reproduce**
N/A

**Expected behavior**
Export a setter equivalent of `Accessor<T>`.

**Reproduction**
N/A

**Additional context**
N/A

ryansolid:
Trying to get this work reminds me why I am no good at TypeScript. I'm very much open to PRs.

atk:
https://github.com/solidjs/solid/pull/575

PR #575, opened 2021-07-23
expose generic setter type

atk:
See issue #507

### DIFF
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -24,7 +24,7 @@ export {
   equalFn,
   $PROXY
 } from "./reactive/signal";
-export type { Accessor, Resource, ResourceReturn, Context, ReturnTypes } from "./reactive/signal";
+export type { Accessor, Setter, Resource, ResourceReturn, Context, ReturnTypes } from "./reactive/signal";
 
 export * from "./reactive/observable";
 export * from "./reactive/scheduler";
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -4,6 +4,7 @@ import { sharedConfig } from "../render/hydration";
 import type { JSX } from "../jsx";
 
 export type Accessor<T> = () => T;
+export type Setter<T> = <U extends T | undefined>(v?: (U extends Function ? never : U) | ((prev: U) => U)) => U;
 export const equalFn = <T>(a: T, b: T) => a === b;
 export const $PROXY = Symbol("solid-proxy");
 const signalOptions = { equals: equalFn };
@@ -103,16 +104,16 @@ export function createRoot<T>(fn: (dispose: () => void) => T, detachedOwner?: Ow
 
 export function createSignal<T>(): [
   get: Accessor<T | undefined>,
-  set: <U extends T | undefined>(v?: (U extends Function ? never : U) | ((prev: U) => U)) => U
+  set: Setter<T>
 ];
 export function createSignal<T>(
   value: T,
   options?: { equals?: false | ((prev: T, next: T) => boolean); name?: string }
-): [get: Accessor<T>, set: (v: (T extends Function ? never : T) | ((prev: T) => T)) => T];
+): [get: Accessor<T>, set: Setter<T>];
 export function createSignal<T>(
   value?: T,
   options?: { equals?: false | ((prev: T, next: T) => boolean); name?: string }
-): [get: Accessor<T>, set: (v: (T extends Function ? never : T) | ((prev: T) => T)) => T] {
+): [get: Accessor<T>, set: Setter<T>] {
   options = options ? Object.assign({}, signalOptions, options) : signalOptions;
   const s: Signal<T> = {
     value,
@@ -129,11 +130,11 @@ export function createSignal<T>(
     value => {
       if (typeof value === "function") {
         if (Transition && Transition.running && Transition.sources.has(s))
-          value = (value as (p?: T) => T)(
+          value = ((value as unknown) as (p?: T) => T)(
             s.pending !== NOTPENDING ? (s.pending as T) : s.tValue
           ) as any;
         else
-          value = (value as (p?: T) => T)(
+          value = ((value as unknown) as (p?: T) => T)(
             s.pending !== NOTPENDING ? (s.pending as T) : s.value
           ) as any;
       }
--- a/packages/solid/test/signals.spec.ts
+++ b/packages/solid/test/signals.spec.ts
@@ -83,7 +83,7 @@ describe("Update signals", () => {
   });
   test("Create and update a Signal with fn", () => {
     const [value, setValue] = createSignal(5);
-    setValue(p => p + 5);
+    setValue((p: number) => p + 5);
     expect(value()).toBe(10);
   });
   test("Create Signal and set different value", () => {
@@ -199,8 +199,8 @@ describe("Batch signals", () => {
       const [a, setA] = createSignal(0);
       const [b, setB] = createSignal(0);
       createEffect(() => {
-        setA(a => a + 1);
-        setB(b => b + 1);
+        setA((a: number) => a + 1);
+        setB((b: number) => b + 1);
       });
       createComputed(() => (count = a() + b()));
       setTimeout(() => {
@@ -214,8 +214,8 @@ describe("Batch signals", () => {
       let count = 0;
       const [a, setA] = createSignal(0);
       createEffect(() => {
-        setA(a => a + 1);
-        setA(a => a + 2);
+        setA((a: number) => a + 1);
+        setA((a: number) => a + 2);
       });
       createComputed(() => (count = a()));
       setTimeout(() => {
@@ -230,11 +230,11 @@ describe("Batch signals", () => {
       const [a, setA] = createSignal(1);
       const [b, setB] = createSignal(0);
       createEffect(() => {
-        setA(a => a + b());
+        setA((a: number) => a + b());
       });
       createComputed(() => (count = a()));
       setTimeout(() => {
-        setB(b => b + 1);
+        setB((b: number) => b + 1);
         setTimeout(() => {
           expect(count).toBe(2);
           done();


ryansolid:
I mean this is is what I did and gave up. Isn't it weird to have to set the type on the setter when we already should know the type? I understand the code changes internally but there has to be a way to do this without needing to change any of the tests, right?

Like:
```ts
const [name, setName] = createSignal("John")

// should not need argument types to do
setName(s => s + 1);
```

I figured forcing this was unacceptable so I hoped some TS wizard could come up with a better solution.

atk:
We could do that, but we would have to coerce the `v` in the callback variation to `T` instead, which has the potential to cause unexpected behaviour if we receive the value as `undefined`:

```typescript
setName(undefined);
setName(name => name + "!"); // "undefined!"
```

I first thought of overloading the createSignal type if there was an initial value, but we can still call the setter with undefined. I'm not sure if we can exploit typescript's inference engine to check if there ever was the chance that the setter is called with undefined. This I see as the only chance to have a Setter type that could potentially allow what you want.

ryansolid:
Hmm.. well if they did:
```ts
const [name, setName] = createSignal("John")
setName(undefined) // this should complain it's not a string
```
And if they did:
```ts
const [name, setName] = createSignal<string | undefined>("John")
setName(s => s + "!"); // this should complain s may be undefined
```
The reason for the first overload on signal is so that we can do the undefined case without passing an argument. But now that we do equality check by default that probably isn't a thing actually. Hmm.. Since you need to do `createSignal(undefined, { equals: false })` to do the type of things where a signal could be undefined. So maybe we can get rid of the overload and clean this up.

atk:
I have one tiny problem: I want the setter to have a mandatory parameter if the type is not extending undefined. I tried

```typescript
export type Setter<T> = T extends undefined 
  ? (v?: (T extends Function ? never : (T | undefined)) | ((prev?: T) => (T | undefined))) => (T | undefined)
  : (v: (T extends Function ? never : T) | ((prev: T) => T)) => T;
```
But that results in the very strange type `((v?: ((prev?: undefined) => undefined) | undefined) => undefined) | ((v: string | ((prev: string) => string)) => string)`, when the intended type would be `(v?: (T extends Function ? never : T) | ((prev: T) => T)) => T`.

I'll have to research that a bit more.

mcjazzyfunky:
Hope I do not waste anybody's time, as frankly I cannot see a the real problem here. Isn't the following way of defining this `Setter` type more straightforward (I know, I've simplified a bit in this demo to keep focus on the real issue)?: [Link to TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBA4hzAgJwDwBUB8UC8UAUAlDlmlAPRlTAAWAlgM5QNQDGAhgDYcQAmUABgEEWLCPXoB7JPyYA7KtWgBlCR1p9JAVySiANFACSUHhNkByYFADWsiQHcod6iACwAKCiev3n7+8V8Nih6eEQkJkZOSSg2eTYRMUkkfQZzS1pgAH4oADo8mNkQOzYQAnd3AKNqNgA3aFsmHgggpxBWJGbEJVoAc1lOM0YQhGQoADNNWRZgWlMoWQheSKgO4G15Gs5NaFp5FTUeHIrKNAlg2gBbMA4xtppdnuWAIwyoRQ79I17bDoVmPCMPW24hSwEG8wk6SuUmAsUsYykELsBCObkqUEgSEk-S4bTsEk0HD4YA6Y1GwDO7C4f0YoEgAgAqmAeGwwjJbJZ+EpQsh+Po2IxXswLlJoMcFLEoAArTT0SxBeiXa7QYZhcaTaazWT6J6aeWFYogPKounKHmoTA4KAAb3cnjwmw4AC4oGgCC6ahJ1Hb8Bc2GBMS68KoeAA1Tgut3EV3uqCe71uAC+5TRlC+fVFEssEjAM1MnAEObzsnoMjYSCBFwgsks7wg+gp1kWYAU0CaIpiQyV3E7GIk4loT24qImU2L7U6EG6fU46AwhBdAG04CMLVAAD5QSZNMa7XgYfTc1foDdb2Q7vc8DAAXXco81cxYHVZU++s8weF2GVonHDHG2kaxsu5pzoeIGYLebjimgiisKYT7wDsVzcFWNaslqEQQlCYAwnC4yIvcjBqAgPbthId4auOCEvtOOJzp+sjfr+WwQJkgFLiuYSgVAR5cRBNo+jQSD2FAZjdhAqGwuOHJMNCSCwjWbzIBAZjuMmUGplAAC0Om6VpKYsKYcpQIuPTwII+jDII15WtRXRvhweBmGwZhlBpVlOWwTwsK54oeduEC7gsPBEAEKBaTSjgEkSUBPNAQSmlAyDCckUASGMcHaCE7geUE2BYGwOTAEglyEG57iGSWlgdPQhLAIIVpmfVhApgEem6QZRmWKZ8AAEKWX1Nm4HZr4zhwKByiVsg9KeAVBfuTlPL57l9Z53nLcMvV4HNl5uZteBPNGTyZEVJUXGV5CUHYGTUASlg0NAACOwLjn6SBWK2rpKHBVy0Nw4T4oSGi3Q4CXgNAyVSJ1VUrGIdW9Y1q3lZp7U6dDxk9cAADCA3Y0NE40Q5E3FQ887I8MWNrT5e3wJTO3BTT2N4Cw0YsCdJPnQQoVXTdd2fc9YiveWH0PV9P1gH9oyAzF9AgzEVDg0lSApej1Vwxw2OI0zblAA)

atk:
It's a question of developer experience. `createSignal` is the equivalent of `let variable` of the reactive handling in solid. You may not want to initialize the value immediately. It's a valid use case to call it without a value.

xdev1:
> It's a question of developer experience. createSignal is the equivalent of let variable of the reactive handling in solid. You may not want to initialize the value immediately. It's a valid use case to call it without a value.

That may be a matter of taste. I personally doubt that something like this leads to better DX. If I see something like the following with the current Solid types, I am - as developer - quite confused about the concrete type of for example `getX`. Will it really return `string` or `string | undefined`? If it returns `string` that may cause trouble without seeing explicitly that the types are currently not 100% fine because something has not been initialized yet.
 
```ts
const [getX, setX] = createSignal<string>()
const x = getX()
```

While, if I read something like the following with the current Signal types, then that question mark tells me that the developer may have the intention to suppress  a little part of the type checks for a couple of lines and will fix this quite soon (with `let` you do not have such problems as the compiler knows exactly if a variable has been initialized or not - at least in a lot of cases ... so `let variable` is not really an "equivalent").

```ts
const [getY, setY] = createSignal<string>(undefined!)
const y = getY()
```

But that was just a remark, this was not important at all. The important part was that `type Setter = ...` definition.

mcjazzyfunky:
> It's a question of developer experience. createSignal is the equivalent of let variable of the reactive handling in solid. You may not want to initialize the value immediately. It's a valid use case to call it without a value.

That may be a matter of taste. I personally doubt that something like this leads to better DX. If I see something like the following with the current Solid types, I am - as developer - quite confused about the concrete type of for example `getX`. Will it really return `string` or `string | undefined`? If it returns `string` that may cause trouble without seeing explicitly that the types are currently not 100% fine because something has not been initialized yet.
 
```ts
const [getX, setX] = createSignal<string>()
const x = getX()
```

While, if I read something like the following with the current Signal types, then that question mark tells me that the developer may have the intention to suppress  a little part of the type checks for a couple of lines and will fix this quite soon (with `let` you do not have such problems as the compiler knows exactly if a variable has been initialized or not - at least in a lot of cases ... so `let variable` is not really an "equivalent").

```ts
const [getY, setY] = createSignal<string>(undefined!)
const y = getY()
```

But that was just a remark, this was not important at all. The important part was that `type Setter = ...` definition.

atk:
The main reason is that we borrowed this pattern from react. If we now deviate from it in such a manner, we should have a really good reason to disappoint the developers' expectations. I don't think "because it makes the types simpler" would suffice, but of course, that's only my opinion.

mcjazzyfunky:
The React typing team uses proper types here, while the concrete type of `createSignal` in Solid is at least dubious:

```ts
   const [count, setCount] = React.useState<number>()
```
The type of `count` is `number | undefined` and not just `number`

While with Solid
```ts
  const [count, setCount] = createSignal<number>()
```

`count` is of type `() => number` instead of `() => number | undefined`

otonashixav:
I believe it should not with `strict: true` specified in the tsconfig. It isn't set in the solid playground, but in my own projects it does in fact report `Accessor<number | undefined>`

mcjazzyfunky:
Oh sorry, my bad ... I have indeed trusted the type hints on the Solid Playground.

mcjazzyfunky:
But the main question is still open:
Wouldn't it simplify everything if you guys just use something like

```ts

type Setter<T> = {
  (...): ...
  (....): ....
}
```

instead of type monsters like

 ```ts
type Setter<T> = undefined extends T
    ? ... some-non-trivial-type-expression ...
    : .... some-other-nontrivial-type-expression ...
```
???
In the (simplified) demo above, typing seemed quite straightforward (okay, maybe I am just missing something...)

atk:
We still want a way for developers to use `setter(old => new)` without having to type old or having to care for the undefined case – at least this is how I understood @ryansolid.

ryansolid:
Now my understanding of TS is limited, but if the end user is allowing for undefined then they should need to check for it everywhere right? Since in theory it could be undefined. But if a user initializes with a value  and never suggests it could be undefined shouldn't we be able to know the type without being explicit and trust it can't be undefined? I think the current types for `createSignal` might be more to
support void signals(ones that never need to be passed a value) rather than to allow for implicit undefined initialization. But maybe it is more nuanced than that.

ryansolid:
I tried what I was thinking and I was mistaken. There are cases for this and my simplistic attempts didn't quite get to where I wanted. I think it's convenient for nullable types to be initiated with an argument.. like lazy component, resource signal, or error signal. These only tend to be fulfilled once like a promise but they can sometimes be reset and I don't like TS getting in people's ways. So I will review these more involved solutions.

ryansolid:
On review I'm going with the other PR but thank you for spearheading this. Who knows how long I would have pushed this off. Thank you.

PR #636, opened 2021-09-03
solid-js/store: improve typings for nullish property types (strict mode)

cynecx:
Current state:

```tsx
import { createStore } from "solid-js/store";

type C = {
  num: number | null;
};

type B = {
  c: C | null;
};

type A = {
  b: B | null;
};

const [store, setStore] = createStore<A>({ b: { c: { num: 1 } } });

setStore("b", "c", "num", (x) => x);
//            ~~~
//            Argument of type 'string' is not assignable to parameter of type 'Part<B | null>'.ts(2345)
```

```json
"solid-js": "^1.1.2"
```

tsconfig.json

```json
{
  "compilerOptions": {
    "jsx": "preserve",
    "jsxImportSource": "solid-js",
    "strict": true
  }
}
```

This PR fixes this by "filtering out" `null` or `undefined` (for `Next`).

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -1,10 +1,4 @@
-import {
-  getListener,
-  batch,
-  DEV,
-  $PROXY,
-  Accessor, createSignal
-} from "solid-js";
+import { getListener, batch, DEV, $PROXY, Accessor, createSignal } from "solid-js";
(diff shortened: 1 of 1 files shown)


ryansolid:
I think this makes sense. While it is possible for these to be null and broken we trust the developer to do the right thing and there is no practical way for them to insure that things are legit purely through static typing.

PR #643, opened 2021-09-04
Improve types of Show and Match

bikeshedder:
Show and Match both make sure that the `item` is "truthy" and therefore `NonNullable` via their implementation.

This makes it easier to use resources without needing to use `?.` or `!`.


ryansolid:
Thank you

PR #648, opened 2021-09-07
feat(signal): abortable resource

atk:
As mentioned earlier on Discord, it would be helpful to be able to abort resource fetching (for fetch or anything else you intend to make abortable). A simple way to do so is to introduce an optional third argument to the fetcher that if used creates an AbortController automatically and calls it whenever the resource is fetched anew. This saves developers the hassle of managing the creation of new AbortControllers themselves, as those are not reusable once called.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -199,6 +199,10 @@ export interface Resource<T> extends Accessor<T | undefined> {
   error: any;
 }
 
+export type SimpleResourceFetcher<T, U, A> = (k: U, getPrev: Accessor<A>) => T | Promise<T>;
+export type AbortableResourceFetcher<T, U, A> = (k: U, getPrev: Accessor<A>, abort: AbortController) => T | Promise<T>;
+export type ResourceFetcher<T, U, A> = SimpleResourceFetcher<T, U, A> | AbortableResourceFetcher<T, U, A>;
+
 export type ResourceReturn<T> = [
   Resource<T>,
   {
@@ -207,43 +211,64 @@ export type ResourceReturn<T> = [
   }
 ];
 
+export type AbortableResourceReturn<T> = [
+  Resource<T>,
+  {
+    mutate: Setter<T>;
+    refetch: () => void;
+    abort: () => void;
+  }
+]
+
 export function createResource<T, U = true>(
-  fetcher: (k: U, getPrev: Accessor<T | undefined>) => T | Promise<T>,
+  fetcher: SimpleResourceFetcher<T, U, T | undefined>,
   options?: { initialValue?: undefined; name?: string }
 ): ResourceReturn<T | undefined>;
 export function createResource<T, U = true>(
-  fetcher: (k: U, getPrev: Accessor<T>) => T | Promise<T>,
+  fetcher: AbortableResourceFetcher<T, U, T | undefined>,
+  options?: { initialValue?: undefined; name?: string }
+): AbortableResourceReturn<T | undefined>;
+export function createResource<T, U = true>(
+  fetcher: SimpleResourceFetcher<T, U, T>,
+  options: { initialValue: T; name?: string }
+): ResourceReturn<T>;
+export function createResource<T, U = true>(
+  fetcher: AbortableResourceFetcher<T, U, T>,
+  options: { initialValue: T; name?: string }
+): AbortableResourceReturn<T>;
+export function createResource<T, U>(
+  source: U | false | null | (() => U | false | null),
+  fetcher: SimpleResourceFetcher<T, U, T | undefined>,
+  options?: { initialValue?: undefined; name?: string }
+): ResourceReturn<T | undefined>;
+export function createResource<T, U>(
+  source: U | false | null | (() => U | false | null),
+  fetcher: AbortableResourceFetcher<T, U, T | undefined>,
+  options?: { initialValue?: undefined; name?: string }
+): AbortableResourceReturn<T | undefined>;
+export function createResource<T, U>(
+  source: U | false | null | (() => U | false | null),
+  fetcher: SimpleResourceFetcher<T, U, T>,
   options: { initialValue: T; name?: string }
 ): ResourceReturn<T>;
 export function createResource<T, U>(
   source: U | false | null | (() => U | false | null),
-  fetcher: (k: U, getPrev: Accessor<T | undefined>) => T | Promise<T>,
-  options?: { initialValue?: undefined; name?: string }
-): ResourceReturn<T | undefined>;
-export function createResource<T, U>(
-  source: U | false | null | (() => U | false | null),
-  fetcher: (k: U, getPrev: Accessor<T>) => T | Promise<T>,
+  fetcher: AbortableResourceFetcher<T, U, T>,
   options: { initialValue: T; name?: string }
-): ResourceReturn<T>;
+): AbortableResourceReturn<T>;
 export function createResource<T, U>(
-  source:
-    | U
-    | false
-    | true
-    | null
-    | (() => U | false | null)
-    | ((k: U, getPrev: Accessor<T>) => T | Promise<T>),
-  fetcher?: ((k: U, getPrev: Accessor<T>) => T | Promise<T>) | { initialValue?: T },
+  source: U | false | true | null | (() => U | false | null) | ResourceFetcher<T, U, T>,
+  fetcher?: ResourceFetcher<T, U, T> | { initialValue?: T },
   options: { initialValue?: T; name?: string } = {}
-): ResourceReturn<T> {
+): ResourceReturn<T> | AbortableResourceReturn<T> {
   if (arguments.length === 2) {
     if (typeof fetcher === "object") {
       options = fetcher;
-      fetcher = source as (k: U, getPrev: Accessor<T>) => T | Promise<T>;
+      fetcher = source as ResourceFetcher<T, U, T>;
       source = true;
     }
   } else if (arguments.length === 1) {
-    fetcher = source as (k: U, getPrev: Accessor<T>) => T | Promise<T>;
+    fetcher = source as ResourceFetcher<T, U, T>;
     source = true;
   }
   const contexts = new Set<SuspenseContextType>(),
@@ -257,7 +282,8 @@ export function createResource<T, U>(
     initP: Promise<T> | null = null,
     id: string | null = null,
     loadedUnderTransition = false,
-    dynamic = typeof source === "function";
+    dynamic = typeof source === "function",
+    abort: AbortController | undefined;
 
   if (sharedConfig.context) {
     id = `${sharedConfig.context!.id}${sharedConfig.context!.count++}`;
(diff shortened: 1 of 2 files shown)


ryansolid:
If I take a while here please forgive me. The focus of what I'm working on currently, caching, affects new features like say adding a 3rd argument to the fetcher perhaps and I want to tread carefully here. This feature is also still questionable for me in terms of requirement of core inclusion. I think this could have merit but it also isn't something that I can move quickly on because of potential for API change.

ryansolid:
This is interesting read too: https://react-query.tanstack.com/guides/query-cancellation

atk:
I find that it usually pays off to go the standard route. You can easily adapt the AbortController to all of these use cases (as you can see in the unit test; if you don't like DOM level 4 events, you can also set `onabort` for the same effect), if not already directly supported.

At the same time, you are reinforcing standard behavior, i.e. libraries like axios will adapt to the patterns they are used in, and at some point start to support AbortController too, if they're not already doing it. Then we're in the happy place where the developer just needs to pop in the provided AbortController and all is well.

By abstracting over the standard, like suggested in this article, we'd be reducing the pressure to adapt to it, so every library will continue with their own incompatible solution, and the developer has to provide another adapter every single time again and again.

We're also reducing code size, because we get the AbortController for free in all browsers supporting it.

LXSMNSYC:
I think for caching, the solution might be to perform reference-counting, and cleanup/abort the request when there's no potential consumers, otherwise, the promise still goes through if one of the consumers decides to cancel.

atk:
> I think for caching...

I'm trying to understand the problem you want to solve here. You have an API and two or more potential consumers. If each consumer has its own resource handling, then yes, the promise will only be aborted for the resource of the consumer and caching will just be performed on request level. If both consumers use the same resource and you only want to abort if both of them cancels, then you need some reference counting. Luckily, the AbortController.abort is a happy little function, ready to be abused as a memo reference for your ref counting, i.e.

```js
const Consumer1 = (props) => {
  const [data, { abort }] = useMyResource();
  abort.consumers = (abort.consumers ?? 0) + 1;
  useEffect({
    props.abort;
    if (--abort.consumers === 0) {
      abort();
    }
  })
}
// and a similar code in Consumer2
```

In any case, until the data() accessor is called, we do not know if there are any consumers. Also, automated ref counting would slow us down in those cases where we don't need it and I don't see any option to detect the requirement before it is actually required.

LXSMNSYC:
> In any case, until the data() accessor is called, we do not know if there are any consumers.
`createResource`, according to Ryan, won't fetch until a read happens.

> Also, automated ref counting would slow us down in those cases where we don't need it and I don't see any option to detect the requirement before it is actually required.

It's just the potential feature Ryan is thinking about. If somehow `createResource` utilizes an internal shared cache in the future, we may want to not abort on the component-level, but rather on the cache-level.

atk:
> It's just the potential feature Ryan is thinking about. If somehow createResource utilizes an internal shared cache in the future, we may want to not abort on the component-level, but rather on the cache-level.

Thanks for the clarification. It would be even more important then to manage the abort functionality inside createResource. The AbortController provided to the request need not be directly connected to the abort function that we give out to the consumers, so that would be an ideal entry point for ref counting, should we need any.

ryansolid:
> createResource, according to Ryan, won't fetch until a read happens.

No .. fetcher runs immediately. It is interesting that during Hydration it's the server-side promise that we are using not the one in the client fetcher. In any case, from a caching perspective, we have a consumer as soon as there is a createResource call that matches the key.

> It's just the potential feature Ryan is thinking about. If somehow createResource utilizes an internal shared cache in the future, we may want to not abort on the component-level, but rather on the cache level.

This is true. Although I intend to build this on top and not part of `createResource` itself. What I was thinking was the fetcher would be
```js
(key, prev, { refetch, mutate }) => ...
```
So you have access to the mutators before the resource creation completes. In a sense argument 2/3 are the returned value of `createResource`. It's an API awkwardness I'm thinking about mostly. Which is why I liked React Queries cancel. I also was thinking having an onCancel passed to the options object.
```js
const [s] = createResource(src, fetch, { onCancel() {} });
```
But certain disconnect if you want a re-useable fetcher function.

LXSMNSYC:
> No .. fetcher runs immediately.

Wasn't this contrary to our discussion regarding the idle state? I thought the `createResource` doesn't fetch until `data` is called?

ryansolid:
`idle` is when src/key is null initially(or maybe any time it is). When there's no active query.

LXSMNSYC:
> idle is when src/key is null initially(or maybe any time it is). When there's no active query.

Ahh, that makes sense now.

atk:
I think it is more awkward to have to connect onCancel to the fetcher manually.

```js
const fetchOpts: { onCancel?: () => void } = {};
const fetcher = (uri) => {
  const ac = new AbortController();
  fetchOpts.onCancel = () => ac.abort();
  return fetch(uri(), { signal: ac.signal });
};
const [s] = createResource(src, fetcher, fetchOpts);

// as compared to


const fetcher = (uri, _, abort) => fetch(uri(), { signal: abort.signal() });
const [s, { abort }] = createResource(src, fetcher);
```

And if you are using something not supporting AbortController by default:

```js
const fetchOpts: { onCancel?: () => void } = {};
const fetcher = (uri) => {
  const cancelTokenSource = axios.CancelToken.source();
  fetchOpts.onCancel = () => cancelTokenSource.cancel('request cancelled');
  return axios.get(uri(), { cancelToken: cancelTokenSource.token });
};
const [s] = createResource(src, fetcher, fetchOpts);

// as compared to

const fetcher = (uri, _ abort) => {
  const cancelTokenSource = axios.CancelToken.source();
  abort.onabort = () => cancelTokenSource.cancel('request cancelled');
  return axios.get(url(), { cancelToken: cancelTokenSource.token );
};
const [s, { abort }] = createResource(src, fetcher);
```

So even in this "worst case" scenario, the DX is still improved.

ryansolid:
I meant fetcher signature. Having mpre than 3 arguments seems awkward. So thinking about extensibility. The 3rd argument could be a mixed bag although to be consistent it'd also have the abort that would be returned and couldn't use length. In fact using length might be awkward for this in general.

atk:
How about a different solution then:

```
interface createResource<T, U, A> {
  (
    src: Accessor<U>,
    fetcher: (src: Accessor<U>, getPrev: Accessor<T>, { abort?: A extends true ? AbortController : A /* room for future additions */ }) => T | Promise<T>,
    options: { abort?: A | true, initialValue?: T, name?: string, ... }
  ) => ResourceReturn<T, A extends true ? AbortController : A>
}
```

This way we would both map the simple case, allow for custom cancellation functions and provide a simple solution to use AbortControllers. Would that be acceptable to you, @ryansolid?

ryansolid:
I see. Seems reasonably from an API standpoint I think.

The option would be `abortable` or something to indicate it can be aborted vs is aborted. We may also need to disambiguate `abort` the action (we return in the tupple) and `abort` the controller if it makes sense for both to be passed in to the 3rd argument. Mostly that I'm thinking the 3rd argument just to forward the second return value in the tuple.

Also do we need more consideration around error handling. Since abort rejects which would put an error in our resource and cause it throw up to the next error boundary. I assumed cancellation to this point would not error. So I suspect we need to special case this error.

atk:
The distinction makes sense, so "abortable" is a good choice.

I just thought that we need to specify the API of the custom abortable constructs. Contenders are [token, trigger] or { token, trigger } (maybe with different names). Preferences?

ryansolid:
> I just thought that we need to specify the API of the custom abortable constructs. Contenders are [token, trigger] or { token, trigger } (maybe with different names). Preferences?

How abstracted do we want this to be I guess is a fair question. I sort of gagged when I saw the interface for AbortController the first time, but I imagine there are reasons that it to be so verbose.

> Lastly, about the special handling of AbortErrors: first, a custom abortable method night yield different results and second, devs expect these specific errors, so all I would do is to document them sufficiently.

So internally we don't abort for race conditions then? What I mean is if you navigate to one page and then the next before it loads we handle race conditions properly today, but we also don't abort. I feel this is super common. Having this error that throws when read in the UI would be unexpected I think to say the least. Is abort only called when the end-user chooses to abort? With the returned abort method? We also tend to hold the previous value by default so meaning of inflight sort of differ with our model in terms of expectation. Reading a stale resource is fine. And Suspense handles it if that is your perogative. I guess you can use the resource.error guard to prevent read to prevent throwing but I think that isn't preferred generally.

I'm still trying to understand the full value of this. I've used promise chaining and promise.all with resources before without building mechanisms internally, that's the benefit of using promises here. So I want to understand all the scenarios under which abort makes sense.

ryansolid:
I read up a bit more and AbortController only applies to fetch requests. I wonder if this belongs more in our equivalent to React Query (the thing I'm working on) than the core createResource mechanism which takes any promise.

LXSMNSYC:
> I read up a bit more and AbortController only applies to fetch requests. I wonder if this belongs more in our equivalent to React Query (the thing I'm working on) than the core createResource mechanism which takes any promise.

I can think of some other use cases for `AbortController` in a non-fetch promise. For instance, 3P libraries may allow some promise cancellations (e.g. file uploader libraries, etc.) which `AbortController` may help. There's also potential side-effects we can cleanup with `AbortController`.

atk:
> I read up a bit more and AbortController only applies to fetch requests.

That's not true; you can also abort any event listeners. Since we can easily map events to AsyncIterators, we'd be able to stream events (which might not make sense for DOM events, but for web socket events, it might be highly useful).

Also, since the standard exists, it is rather likely to be used in new versions of existing APIs as well as new APIs. For example, one could make media stream requests or the resulting streams abortable.

atk:
The whole structure of AbortController would read like this if implemented for modern browsers:

```js
export class AbortSignal extends EventTarget {
  readonly [Symbol.toStringTag] = 'AbortSignal';
  private _aborted = false;
  public onabort: (ev: Event) => void | undefined;
  constructor() {
    super();
    this.addEventListener('abort', (ev: Event) => {
      this._aborted = true;
      this.onabort?.(ev);
    })
  }
  get aborted() { return this._aborted; }
  toString() { return '[object AbortSignal]' }
}
export class AbortController {
  readonly [Symbol.toStringTag] = 'AbortController';
  public signal = new AbortSignal();
  abort() { this.signal.dispatchEvent(new Event('abort')); }
  toString() { return '[object AbortController]'; }
}
```
And ~40% of the lines are just due to native class behavior (toString, toStringTag). For a reusable standard, this is quite concise IMO.

atk:
The interesting feature about abortable promises is that you can auto-abort your fetcher calls on the next load, thus freeing resources for your resources. If we want an automated handling of that, we need to know the API, ie. the token/signal and the aborting call. Having a pattern similarly to what's used elsewhere makes sense.

One way would be the react way that you've shown, but that will add unneccessary extra dev work on the most common use case: the fetch request. In any case, I would like to hear more about what you're planning with the react-query-like API before I continue.

atk:
After some consideration, I arrived at a more extendable pattern:

```typescript
type ResourceOpts<S, R= {
  src?: Accessor<S>,
  fetcher: (src: S | true, getPrev: Accessor<R>) => R | Promise<R>,
  initialValue?: R,
  name?: string,
  compose?: ((opts: ResourceOpts, return: ResourceReturn) => void)[]
};
```

The compose functions can make simple changes to every aspect of the resource handling.

The only thing I'm afraid of is that this is too powerful a pattern and leads to basically reimplementing rxjs, which is something I'd like to avoid.

atk:
After another round of consideration, it's probably better to move this part into solid-primitives and help this become an official package.

Issue #608, opened 2021-08-19
Proposal: `data` with better type elimination from `createResource`


LXSMNSYC:
```js
const [data] = createResource(...);
```

Currently, `data` returns an accessor that returns the current result of the resource. To check if it is still loading, we use `data.loading` and use `data.error` when checking for failures. This is great but not for Typescript. 

https://github.com/solidjs/solid/blob/main/packages/solid/src/reactive/signal.ts#L216-L233

The accessor returns `T | undefined` but that is incorrect for 3 reasons:
- `Suspense` should guarantee the type is exactly what is returned for `data`.
- `undefined` is a valid value for a Promise to result into.
- The presence of `data.loading` and `data.error` is not exclusive enough to perform type elimination.

There is a way to fix this but would introduce a breaking change:

```ts
interface Pending {
  status: 'pending';
}
interface Success<T> {
  status: 'success';
  value: T;
}
interface Failure {
  status: 'failure';
  value: any;
}
type Result<T> = Success<T> | Failure | Pending;

// Either the bottom one or a proxy for Result<T>
type ResourceData<T> = () => Result<T>;
```

This way, the result is deterministic, exclusive and can be type-eliminated.

```ts
const result = data(); // Result<T>;
switch (result.status) {
  case 'pending': // Pending
    return null;
  case 'success': // Success<T>
    return <h1>Success! {result.value}</h1>;
  case 'failure': // Failure
    return <h1>Something went wrong. {result.value.message}</h1>;
}
```

ryansolid:
I see. That's interesting. I guess with that API it doesn't even need to be a function accessor. It just has 2 getters. My instant hesitation was I liked the whole feels like a Signal thing.. with Suspense you rarely use the loading or error states.

We can't do that union trick the other way I guess. Like this won't fly?

```ts
interface Pending {
  loading: true;
  error: false;
  (): undefined;
}
interface Success<T> {
  loading: false;
  error: false;
  (): T;
}
interface Failure {
  loading: false;
  error: any;
  (): undefined;
}
```
I'm gathering not or you would have suggested it.

Hmm, it is breaking though which ties my hands I think for now.

LXSMNSYC:
I think that should work as well! Although I'm not sure if we can correctly eliminate the types (for instance, checking for `data.error` doesn't really eliminate into `Failure`. Is there a merit on adding a similar field to `data.status`? `status` helps brand the types.

ryansolid:
[...]

PR #660, opened 2021-09-19
improve createResource types

atk:
+ JSDoc comments for better IDE integration

Exporting the internal types of createResource makes extending it by tertiary primitives (e.g. https://github.com/davedbase/solid-primitives/tree/main/packages/fetch) easier.

JSDoc comments allow us to improve the developer experience by helping them directly in their IDE. To try it, hover over a function signature e.g. of `createSignal` or `createResource` in visual studio code or similarly advanced IDEs.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -126,7 +126,7 @@ export type SignalOptions<T> = { name?: string, equals?: false | ((prev: T, next
  *  options?: { name?: string, equals?: false | ((prev: T, next: T) => boolean) }
  * )
  * ```
- * @param value inital value of the state; if empty, the state's type will automatically extended with undefined; otherwise you need to extend the type manually if you want setting to undefined not be an error
+ * @param value initial value of the state; if empty, the state's type will automatically extended with undefined; otherwise you need to extend the type manually if you want setting to undefined not be an error
  * @param options optional object with a name for debugging purposes and equals, a comparator function for the previous and next value to allow fine-grained control over the reactivity
  * 
  * @returns ```typescript
@@ -301,6 +301,8 @@ export interface Resource<T> extends Accessor<T> {
 
 export type ResourceActions<T> = { mutate: Setter<T>; refetch: () => void; };
 
+export type ResourceReturn<T> = [Resource<T>, ResourceActions<T>];
+
 export type ResourceSource<S> = S | false | null | (() => S | false | null);
 
 export type ResourceFetcher<S, T> = (k: S, getPrev: Accessor<T>) => T | Promise<T>;
@@ -334,8 +336,8 @@ export type ResourceOptions<T> = T extends undefined
  * 
  * @description https://www.solidjs.com/docs/latest/api#createresource
  */
-export function createResource<T, S = true>(
-  fetcher: ResourceFetcher<S, T | undefined>,
+export function createResource<T extends any, S = true>(
+  fetcher: ResourceFetcher<S, T>,
   options?: ResourceOptions<undefined>
 ): [Resource<T | undefined>, ResourceActions<T | undefined>];
 export function createResource<T, S = true>(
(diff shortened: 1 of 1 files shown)


LXSMNSYC:
- Any particular reason as to why `initialValue` has a different type from the success value?
- There's also some potential `Resource` type change to handle idle state (source = null), please check this [issue thread](https://github.com/solidjs/solid/issues/608).

ryansolid:
I'm pretty stoked to have the JSDocs in.

TypeScript need to look at a bit more to understand. I know work was done to make sure that the type could be inferred correctly from initial value. And it's important that resources that don't start with an initial value are nullable, since it is possible for them to not have a value. That being said it going null afterwards I'm fine being explicit about in end user types and not including here as it would ever go null unless you set the source that way. But just want to ensure that is retained, the type inference and the fact when not initially set it can be nullable.

rturnq (commented):
I started this review with some specific code changes in mind but after further examination it looks like these changes have regressions which could be resolved in a number of ways depending on questions like does allowing having the type `T` and `I` be different outside of the undefined `initialValue` case.

Here are a number of different cases that work as expected today. This PR causes `r2` and `r7` to fail:
```ts
const [r1] = createResource(() => Promise.resolve(1))
const [r2] = createResource(() => Promise.resolve(1), { name: 'me' })
const [r3] = createResource(() => Promise.resolve(1), { initialValue: 0 })
const [r4] = createResource(() => Promise.resolve(1), { initialValue: undefined })
const [r5] = createResource(true, () => Promise.resolve(1))
const [r6] = createResource(true, () => Promise.resolve(1), { name: 'me' })
const [r7] = createResource(true, () => Promise.resolve(1), { initialValue: 0 })
const [r8] = createResource(true, () => Promise.resolve(1), { initialValue: undefined })
```
I will add my remaining thoughts to the general comments.

rturnq:
Looking primarily at the `createResource` type changes

- I think it makes sense to create named types for the argument and return types. Like you said this will make it easier for someone to build on top.
- I see little value in adding a second generic to `ResourceReturn` and `Resource` to specify the initial value type. In `Resource` they ultimately get combined in a union for the function return type and the individual types are never needed separately. Therefore, the upstream type user can just pass the type union in as a single generic.
- Same question as @LXSMNSYC, do we want to allow the initial value to have a different type as the fetcher return type outside of the undefined `initialValue` case?

atk:
Thanks for your feedback. While the idea was that `I` either extends `T` or `T | undefined`, this does not seem to work as intended. I have refactored the types to fix the failing cases and removed `ResourceReturn` as you suggested.

rturnq:
I actually thought `ResourceReturn` was fine to keep, I just wanted to remove the second generic like you have done with `Resource`. It might make sense to at codify the actions object to make it easier for things to build on top and tighten the other definitions.

```ts
interface ResourceActions<T> { mutate: Setter<T>; refetch: () => void; }
```

I think createResource can be simplified from 6 overloads down to 4 be combining the cases where `options` is defined but `initialValue` is not defined and where `options` is not defined. This means overload 1 is removed and 3 has its `options` parameter made optional with a `?`. Similarly overload 4 is removed and 6 has `options` made optional. As far as I can tell we don't lose anything by doing this.

atk:
Thanks again for your time and your valuable feedback. I have implemented all the suggested changes. Yes, it makes sense to bundle the ResourceActions and this makes it easier to extend, too.

atk:
I'll keep the return types as is, because you'll see easier that the result is an array then.

rturnq:
@atk Your latest updates look good to me. Those test-cases I provided are all working as expected and the types seem succinct. From a `createResource` types standpoint, I think this is in a good state to merge now.

LXSMNSYC:
Hmmm I was thinking, can we consider this? https://github.com/solidjs/solid/issues/608

atk:
@LXSMNSYC let's not draw this PR out even longer, shall we? Incremental improvements is the way to go.

@rturnq We forgot about fixed generics.

ryansolid:
Thank you so much. Thank you for all the effort put in here.

Issue #656, opened 2021-09-17
TypeScript error when update store use functioin filter


weijarz:
**To Reproduce**

Code snippet from: <https://www.solidjs.com/docs/latest/api#updating-stores> 

```typescript
const [state, setState] = createStore({
  todos: [
    { task: 'Finish work', completed: false },
    { task: 'Go grocery shopping', completed: false },
    { task: 'Make dinner', completed: false }
  ]
});
setState('todos', todo => todo.completed, 'task', t => t + '!')
```

**Expected behavior**

The example code copied from Solid document will cause typescript error on filter function.

**Reproduction**

<https://playground.solidjs.com/?version=1.1.1#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9mjOFFxwAyrm5xeAX14AzRrRi8AOjVoAbNABMAtACsyAejJThGgNxqIt5rQjnewc6LhZeZOLknuAurwAvILC7pLSABQgtry8Urq0ZIgusXH88VBkANYpAOQAYhhoZAAWvADu3Nl5ng6cWj5wuikKUFreclhpcQK4Wbm8eQDitLwA5sqsjPhepbSEhBjjtYIqhI1iLYrtnbLdEOkZ-Tn5ALJQ2TK6GBBwjKv1G03bbR0ysmn+trIAlDZOHx+MSRPIJJKrcHBAB88VoiQAdE9Ns1PGCBpCYfFeABqIYAQjyv1sYFk-iAA>

fictitious:
For TypeScript, all the functions passed to `setState`, except the last one, need explicit types for their parameters.

There's no error if `todo` parameter type is spelled out:

```typescript
 setState('todos', (todo: {task: string; completed: boolean}) => todo.completed, 'task', t => t + '!')
```

The code in Solid examples is javascript. TypeScript type inference has its limits. Here there are multiple parameters with inter-dependent types which it can't figure out without some help.

ryansolid:
As far as I know @fictitious is correct. These are very difficult types and have been the effort of many of the community to get right. There are still gaps though. So I'd welcome anyone who has the knowledge to look at this but realistically this is probably beyond my or any of the team member's current TypeScript knowledge to address.

PR #665, opened 2021-09-21
Improve types for set state

rturnq:
This resolves issue #656 by making the `Part` type conditional on whether the state it represents is an array or object.

There was a comment left on the `Part` type
> changing this to "T extends any[] ? ArrayFilterFn<T> : never" results in depth limit errors

I tested this with a depth of 8 arguments/path parts which is the maximum supported by the set state function and didn't run into any errors, but I didn't try a lot of different permutations.


ryansolid:
Thanks

PR #706, opened 2021-10-31
dehydrate the Solid reactivity type definitions

trusktr:
This

- makes the code more DRY, less WET,
- and allows Solid ecosystem lib authors the opportunity to import and re-use the types (the main motivation I had for making these changes)

On my end, the `suspense.spec.tsx` test fails, but I have not modified anything about the runtime, I've modified only the type annotations.

Furthermore, I do not get any type errors in VS Code, but _sometimes_ the `signals.memo.spec.ts` file fails with a type error. It seems as if maybe sometimes this test runs before the `suspense.spec.tsx` test, or vice versa.

Not sure what's up with that yet. But while ironing that out, I'd like feedback on the changes in general. 

UPDATE: upon further investigation, the type error in the signals.memo test makes sense. I'm not sure why the error does not appear in VS Code though.


atk (changes requested):
This breaks the disambiguation between creating a signal with undefined as a valid value; in this case, calling the setter without an argument will fail the type check, but is valid.

trusktr:
Ah yeh, I noticed that for that. Working on fixing that now.

atk:
It's a requirement that could appear more often; perhaps we can create a helper type that makes entries extending undefined optional.

trusktr:
Hmmm, weird. I'm not getting type errors in VS Code, yet inside of reactive.ts I don't see the ambiguation problem, but I do see it inside of the spec file (with no error regardless. Something is funked up.

trusktr:
Here's the proof of the weirdness I'm seeing, which led me to believe the types were fine because I was assuming what I was seeing in reactive.ts in VS Code, but not in the spec file, was correct:

This is inside reactive.ts:

Screenshot from 2021-10-31 22-58-37

And this is inside of signals.spec.ts:

Screenshot from 2021-10-31 22-59-14

Note both of them call `createSignal<number>()` but only one of them has the value type `number | undefined`.

Phewf, not going crazy. :)

trusktr:
Oh, I wonder if the tests, because they're excluded from tsconfig, don't have `strict` as `true`, they just use whatever VS Code defaults to (strict:false)? I think that explains the missing `undefined`.

trusktr:
Fixed the ambiguity with `Setter`

packages/solid/src/reactive/signal.ts
```diff
@@ -141,39 +151,45 @@ export type SignalOptions<T> = { name?: string; equals?: false | ((prev: T, next
...
     comparator: options.equals || undefined
   };
+
   if ("_SOLID_DEV_" && !options.internal)
     s.name = registerGraph(options.name || hashValue(value), s as { value: unknown });
 
   return [
     readSignal.bind(s),
-    ((value: T extends Function ? never : T | ((p?: T) => T)) => {
+    value => {
```
atk, line 172: Removing the type annotation here means that `value` is typed as `any`, which obviously results in an error when building the types.

trusktr: Yes indeed! On a walk. Will get it in just a bit.

atk:
I helped creating the current state of the types. Ryan requested the following behavior:

```ts
// No initial value, typed T: undefined and parameter omission allowed
const s1 = createSignal<number>();
// Initial value, typed T not extending undefined: undefined and parameter omission not allowed
const s2 = createSignal<number>(1);
// initial value, typed T | undefined: parameter omission allowed
const s3 = createSignal<number | undefined>(1);
```

One could argue that parameter omission was easily misunderstood, though. I think it's Ryan's decision which way to proceed here.

LXSMNSYC:
@atk  I believe the design came from React since React's `useState` allow omission of initial state, although unlike React, `createSignal` does not have lazy initialization

trusktr:
@atk parameter omission was preserved in the `Setter` type I made. In my last comment, I was referring only to the _return type_, which is what changed as follows:

The difference between the previous `Setter` and the one I made (assuming `T` with a concrete type of `number | undefined` for sake of example) is that the previous `Setter` returns type `number` if a number value is passed in, otherwise returns the type `undefined` if the value `undefined` is passed in, while the new version always returns the type `number | undefined` regardless of the input values.

So what I'm curious about is, how can we use the new version of `Setter`, but restore the splitting of `number | undefined` into `number` and `undefined` for the _return type_ depending on input value?

If we can do this, then we will solve the issue that the previous `Setter` has, which is that when `strict` mode is set to `false` in `tsconfig.json`, that previous version of `Setter` returns type `any`. For example, with `"strict": false`, this happens:

```js
    const signal = createSignal<number>();
```

where here `signal` is inferred to have a type of `any`. With this new `Setter` definition, `signal` will have an inferred type of `number | undefined` regardless of the value of `strict` inside of `tsconfig.json`.

The only thing is, the return value splitting is not preserved in the new version.

That's what I'm curious about, but I just pushed the original `Setter` definition back. We can figure this out in a following pull request (if it is possible).

packages/solid/src/index.ts
```diff
@@ -30,7 +30,7 @@ export type {
   Accessor,
   Setter,
   Resource,
-  ResourceReturn,
+  ResourceState,
```
trusktr, line 33: This is the only breaking change (not covered by the tests). The `ResourceReturn` and `Resource` types were not being used in the source code, but only exported. I'm not sure if anyone was relying on these unused-yet-exported types.

I renamed them to align with the API, f.e. `createResource()` now returns the type `Resource`, which happens to be a tuple.

Is this ok? I didn't like the name of `ResourceReturn`.

I did similar with renaming the internal (this time not exported) `Signal` type to `SignalState`, so that the `createSignal` function could return the new `Signal` type (a tuple) while `SignalState` is used only internally.

I think the names make more sense like this.

Another issue, that I didn't change, is that the word "accessor" in JavaScript tends to mean a getter/setter pair, and the name of the current `Accessor` type does not align with this. I think `Getter` would be a better name for it, but this would probably definition be a breaking change.

Does Solid intend to freeze at 1.0 for a long time? Or will the plan be to move it along major version bumps and keep it fresh and new?

I'm a fan of the latter, and one such library example is [three.js](https://threejs.org/) which keeps progressing for the better and not allowing itself to become stale. The three.js core team makes sure to write a migration guide in every release note, and they often break things by first deprecating old APIs for a while while the new APIs are put in place, then eventually removing them. Another method they have is that they might disable a feature, leave it behind a flag for a while, then remove it, and people who migrate it need to only set the flag to true so that they can keep using the feature for a while but eventually they know they will have to stop using it if they wish to keep upgrading.

Based on this, what we might be able to do is add the `@depracted` JSDoc comment to `Accessor`, and add a new `Getter` type. Then at some point (like after a year?) remove it.

atk: Yes, the unused Resource types were meant to be used in solid-primitives.

trusktr:
I undid the resource type change (we can do that later if still desired), and improved the EffectFunction type to make it more clear what the intent is.

This PR so far only DRYs up src/reactive/signal.ts, but not src/server/reactive.ts which is now outdated (but it should re-use the same types).

LXSMNSYC:
I'll probably just take over the resource types if https://github.com/solidjs/solid/issues/608#issuecomment-901576636 was considered, on a separate PR at the least

trusktr:
That sounds good, I didn't see that one.

packages/solid/test/type-tests.ts
```diff
@@ -0,0 +1,308 @@
+import { createEffect, createComputed, createRenderEffect, createMemo, Accessor, on } from "../src";
+
+createEffect(
+  (v?: number) => {
+    return 123;
+  },
+  123,
+  {}
+);
```
trusktr, line 9: I added a bunch of type tests. Still a bunch more can be added.

trusktr:
Last thing before solidifying this (no pun intended!): is the naming ok? I'm just thought maybe `EffectFunction` should be just `Effect`.

Wdyt?

packages/solid/src/reactive/signal.ts
```diff
@@ -592,10 +654,10 @@ export function createSelector<T, U>(
  *
  * @description https://www.solidjs.com/docs/latest/api#batch
  */
-export function batch<T>(fn: () => T): T {
+export function batch<T>(fn: Accessor<T>): T {
```
otonashixav, line 657: Not sure if `Accessor` makes sense in this case, since I don't think you'd ever pass an accessor to batch. It might make more sense to leave this as `() => T`.

otonashixav (changes requested):
Mainly see the comments regarding the effect functions' overloads; add this test to the three of them:
```ts
// @ts-expect-error undefined second arg is not assignable to the number parameter
createEffect((v: number) => 123);
```

trusktr:
> add this test to the three of them:
> 
> ```ts
> // @ts-expect-error undefined second arg is not assignable to the number parameter
> createEffect((v: number) => 123);
> ```

Ah, good point. First value is undefined, and no error currently. :+1:

packages/solid/src/reactive/signal.ts
```diff
@@ -56,25 +53,21 @@ interface Owner {
...
-  fn: (v?: T) => T;
+export interface Computation<T extends U, U = T> extends Owner {
+  fn: EffectFunction<T, U>;
   state: number;
   tState?: number;
-  sources: Signal<T>[] | null;
+  sources: SignalState<T>[] | null;
   sourceSlots: number[] | null;
-  value?: T;
+  value?: U;
```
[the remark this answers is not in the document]
trusktr: The reason for `U` is that value is going to be passed to the function's input parameter, which the return type `T` is assignable to, and the function may transform it into a `T` for the next iteration.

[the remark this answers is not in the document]
trusktr: In my previous comment this is true about value passed into the effect functions. What about `Computation`s? If those hold only the returned values, then I think what you said makes sense.

packages/solid/src/reactive/signal.ts
```diff
@@ -189,9 +200,22 @@ export function createSignal<T>(
  *
  * @description https://www.solidjs.com/docs/latest/api#createcomputed
  */
-export function createComputed<T>(fn: (v?: T) => T | undefined): void;
-export function createComputed<T>(fn: (v: T) => T, value: T, options?: { name?: string }): void;
-export function createComputed<T>(fn: (v?: T) => T, value?: T, options?: { name?: string }): void {
+export function createComputed<T extends U, U = T>(fn: EffectFunction<T, U>): void;
```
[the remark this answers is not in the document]
trusktr: Noted about the test case you mentioned to add. :+1: 

> `createComputed((v: number) => 123)` should not be deemed correct (since `v` is undefined on first execution). I think the effect functions should drop this first overload and `value` should be optional in the second overload i.e.:
> 
> ```ts
> // export function createComputed<T extends U, U = T>(fn: EffectFunction<T, U>): void; remove
> export function createComputed<T extends U, U = T>(
>   fn: EffectFunction<T, U | undefined>,
>   value?: undefined, // make optional
>   options?: EffectOptions
> ): void;
> export function createComputed<T extends U, U = T>(
>   fn: EffectFunction<T, U>,
>   value: U extends U ? U : never,
>   options?: EffectOptions
> ): void;
> ```
> 
> The third overload handles the `U = undefined` case as well, but this might be less confusing.
> 
> I'm unsure the second generic is needed for the first two overloads, are they for usability? `EffectFunction<T, U | undefined>` always returns T and can only be supplied with `T | undefined`, so `U` is redundant otherwise.

I was aiming for things like `(v: number | string) => number`, where `U` is `number | string`. Or is that different what you're talking about?

packages/solid/src/reactive/signal.ts
trusktr: Does it expect any function, not necessarily a reactive getter?

packages/solid/src/reactive/signal.ts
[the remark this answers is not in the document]
trusktr: Just took a look, and it has to be `U` because the initial value is passed to that location, then after that it becomes `T` on second iteration.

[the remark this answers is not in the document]
trusktr: Oh, oops, I misread what you said. You meant just to remove `?`.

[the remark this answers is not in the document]
trusktr: I tried removing `?`, and that introduced 5 new errors.

trusktr:
Once more issue in signals spec. Coming back tomorrow.

packages/solid/src/reactive/signal.ts
otonashixav: I think usually it would be a void function containing a few set calls, possibly returning a value if convenient to do so. Either way, passing a reactive getter i.e. `batch(getter)` doesn't make sense compared to something like `untrack(getter)`.

packages/solid/src/reactive/signal.ts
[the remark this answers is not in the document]
otonashixav: My bad, I should've phrased that better. I think it makes sense to make it not optional; if it's a type only returned by `createComputation` and `createComputation` always creates it with some value, then it should always exist, even if undefined. 

There are probably other instances like this where optional properties could be typed ever so slightly better, but they're usually not paid attention to as it's not much of an issue, and in this case especially so since it's on a type meant to be used internally.

trusktr:
@webstrand @tjjfvi this is a tough one. There's one case that might make people scratch their heads, because it's not so edgy:

```js
const value = createMemo(_prev => {
  return "hello";
}, "init");
const test1: string = value(); // Type 'unknown' is not assignable to type 'string'.
```

ryansolid:
Thanks for the incredible work here. Really excited what this means for the types in Solid.

PR #739, opened 2021-11-26
feat: add `enableExternalSource` for reactivity interoperability

3Shain:
This pull request brings a new API `enableExternalSource` which provides more flexible reactivity interoperability (i.e. it's possible to use external reactive source __directly__ in solid primitives (createMemo, createEffect, etc.))

For a concrete example please check out `packages/solid/test/external-source.spec.ts`.

Potential issue: 
* `enableExternalSource` could be piped, but this feature seems not to be reflected well from the naming?

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -25,6 +25,7 @@ const [transPending, setTransPending] = /*@__PURE__*/ createSignal(false);
 export var Owner: Owner | null = null;
 export let Transition: Transition | null = null;
 let Scheduler: ((fn: () => void) => any) | null = null;
+let ExternalSourceFactory: ExternalSourceFactory | null = null;
 let Listener: Computation<any> | null = null;
 let Pending: Signal<any>[] | null = null;
 let Updates: Computation<any>[] | null = null;
@@ -83,6 +84,8 @@ interface Transition {
   running: boolean;
   cb: (() => void)[];
 }
+
+type ExternalSourceFactory = <T>(fn: (v?: T) => T, trigger: () => void) => (v?: T) => T;
 
 /**
  * Creates a new non-tracked reactive context that doesn't auto-dispose
@@ -955,6 +958,16 @@ export function getSuspenseContext() {
   return SuspenseContext || (SuspenseContext = createContext<SuspenseContextType>({}));
 }
 
+// Interop
+export function enableExternalSource(factory: ExternalSourceFactory) {
+  if (ExternalSourceFactory) {
+    const original = ExternalSourceFactory;
+    ExternalSourceFactory = (fn, trigger) => original(factory(fn, trigger), trigger);
+  } else {
+    ExternalSourceFactory = factory;
+  }
+}
+
 // Internal
 export function readSignal(this: Signal<any> | Memo<any>) {
   const runningTransition = Transition && Transition.running;
@@ -1125,6 +1138,14 @@ function createComputation<T>(
         `${(Owner as Computation<any>).name || "c"}-${
           (Owner.owned || (Owner as Memo<T>).tOwned!).length
         }`;
+  }
+  if (ExternalSourceFactory) {
+    const [track, trigger] = createSignal<void>(undefined, { equals: false });
+    const patchedFn = ExternalSourceFactory(c.fn, trigger);
+    c.fn = x => {
+      track();
+      return patchedFn(x);
+    };
   }
   return c;
 }
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -15,6 +15,7 @@ export {
   batch,
   on,
   enableScheduling,
+  enableExternalSource,
   startTransition,
   useTransition,
   createContext,
--- a/packages/solid/test/external-source.spec.ts
+++ b/packages/solid/test/external-source.spec.ts
@@ -1,0 +1,62 @@
+import { createRoot, createMemo, onCleanup, enableExternalSource } from "../src";
+
+import "./MessageChannel";
+
+class ExternalSource<T = any> {
+  listeners: Set<() => void> = new Set();
+
+  constructor(private value: T) {}
+
+  update(x: T) {
+    this.value = x;
+    this.listeners.forEach(x => x());
+  }
+
+  get() {
+    this.listeners.add(listener);
+    sources.get(listener).add(this);
+    return this.value;
+  }
+
+  removeListener(listener: () => void) {
+    this.listeners.delete(listener);
+  }
+}
+
+let listener: (() => void) | null = null;
+
+let sources: Map<() => void, Set<ExternalSource>> = new Map();
+
+enableExternalSource((fn, trigger) => {
+  sources.set(trigger, new Set());
+  onCleanup(() => {
+    sources.get(trigger).forEach(x => x.removeListener(trigger));
+    sources.delete(trigger);
+  });
+  return x => {
+    const tmp = listener;
+    // trigger could play the role of listener，as it has stable reference
+    listener = trigger;
+    try {
+      return fn(x);
+    } finally {
+      listener = tmp;
+    }
+  };
+});
+
+enableExternalSource(fn => fn); // do nothing, make sure multiple factories be piped.
+
+describe("external source", () => {
+  it("should trigger solid primitive update", () => {
+    createRoot(() => {
+      const e = new ExternalSource(0);
+      const memo = createMemo(() => {
+        return e.get();
+      });
+      expect(memo()).toBe(0);
+      e.update(1);
+      expect(memo()).toBe(1);
+    });
+  });
+});


LXSMNSYC:
Interesting. However I would like to see a demo other than the test and its use cases.

The concept reminds me of React's external sources however I'm not sure they solve the same problem.

3Shain:
> Interesting. However I would like to see a demo other than the test and its use cases.
> 
> The concept reminds me of React's external sources however I'm not sure they solve the same problem.

It's primarily designed for reactive library creators rather than normal users. The result is: users are able to use other _framework-agnostic_ reactivity implementations seamlessly.
```ts
// before
function Component() {
    const reactiveObject = createThirdPartyReactiveObject();
    const [signal, setSignal] = createSignal(reactiveObject.value);
    createEffect(()=>reactiveObject.subscribe(setSignal))
    return <>{signal()}</>
}

// after
function Component() {
    const reactiveObject = createThirdPartyReactiveObject();
    return <>{reactiveObject.value}</>
}
```
The principle is: the third-party library will be able to track its own reactive dependencies and notify solid if there is a change.

@ryansolid suggested the name `enableExternalSource`, I think it has nothing to do with react's concept.

3Shain:
Mobx integration example (haven't tested yet, but I believe it's almost there)

```ts
import { Reaction } from 'mobx';

let id = 0;

enableExternalSource((fn,trigger)=>{
  const reaction = new Reaction(`externalSource@${++id}`, trigger);
  onCleanup(()=>reaction.dispose());
  return x => reaction.track(()=>fn(x));
});
```

LXSMNSYC:
> I think it has nothing to do with react's concept.

Interestingly, it is. React's external sources provides a way to bridge non-React third-party reactive objects (e.g. Observables, event emitters, etc.) into React components.

I think we can probably simplify this into something that just provides a special kind of signal?

3Shain:
> > I think it has nothing to do with react's concept.
> 
> Interestingly, it is. React's external sources provides a way to bridge non-React third-party reactive objects (e.g. Observables, event emitters, etc.) into React components.
> 
> I think we can probably simplify this into something that just provides a special kind of signal? much like how `from` works

I'm not sure what do you mean by react's external source (google provides nothing useful) so I assume you are talking about `useMutableSource`. This API has gone (becomes `useSyncExternalStore`) and the its primary use case is to make reading from external mutable data __safe in concurrent mode__, not just about triggering re-rendering when data changes (otherwise `useSubscription` is enough).

I'm not sure what do you mean by `from`, I think this is already the simplest solution. Could you elaborate it?

LXSMNSYC:
> I'm not sure what do you mean by react's external source (google provides nothing useful) so I assume you are talking about useMutableSource. This API has gone (becomes useSyncExternalStore) and the its primary use case is to make reading from external mutable data safe in concurrent mode, not just about triggering re-rendering when data changes (otherwise useSubscription is enough).

Yes I got the terminologies mixed up, sorry.

> I'm not sure what do you mean by from, I think this is already the simplest solution. Could you elaborate it?

This was a new feature shipped in Solid 1.2.0. I transforms an Observable source into a signal.

3Shain:
The big advantage of this API is, the end user can use external reactive object __directly__ in any place a signal could be placed, and it's almost imperceptible.

```ts
// example taken from https://mobx.js.org/README.html

class Timer {
    secondsPassed = 0

    constructor() {
        makeAutoObservable(this)
    }

    increase() {
        this.secondsPassed += 1
    }

    reset() {
        this.secondsPassed = 0
    }
}

const myTimer = new Timer();

render(()=><button onClick={() => timer.reset()}>Seconds passed: {timer.secondsPassed}</button>, document.body)

// Update the 'Seconds passed: X' text every second.
setInterval(() => {
    myTimer.increase()
}, 1000)
```

LXSMNSYC:
Hmmm yeah I think this make sense, however personally I think doesn't this hide any implementation detail. Well I guess this is just by preference

ryansolid:
I am super stoked about the direction here. It does add an overhead to every computation but for people basically looking to swap the reactive system I think it's pretty cool. While most of it can be very closely done by just wrapping Solid's APIs to the point I'd say that is the preferred method if it is a syntax thing. Some libraries have subtly different execution. Being able to use Kairo or like MobX State Tree essentially natively is really cool. This may even finally be the thing that lays stuff like mobx-jsx to rest. Really cool stuff.

That being said the form of the API makes sense to me from my knowledge of MobX and even how the React Solid State works. But would be interested in opinions.

3Shain:
@ryansolid  Another thing worth to notice: transition (scheduling) will not work (I mean, not effective, but user will still get the expected output/behavior) with external computations. This is expected and so far I'd say there is no solution to perfectly tackle it. Even react hasn't figured it out.
>  There are 3 levels of implementations in React Concurrent Feature:
> 1. Make it work: useSubscription
> 2. Make it right: useSyncExternalStore -> __all external state management libraries' the best situation.__ Essentially level 1 without tearing, to guarantee consistency. But de-opts still occur (sync re-rendering).
> 3. Make it fast: use react hooks only


In solid we don't need to worry about consistency. This feature is fully built upon existing mechanisms. But the scheduling won't produce any effect (unless there are HUGE amounts of updates).
If the scheduling is INDEED necessary, the best solution is: __external scheduling + solid effects__. So it's up to external reactive library to implement the scheduling stuff. 99.9% (I made up) users don't need to worry about this.

ryansolid:
Let me think. Would it work? Yes the third party would update on its change, but wouldn't the internal computation still only rerun as scheduled. We are just proxying the signal.  We are leveraging the og computations to synchronize.

Let's say we are under Suspense without transitions being read under an effect. 3rd party triggers update, sets signal, then Effect gets scheduled and held. Then at the end runs again.

I guess the problem with concurrency is we can't hold setting the external source if it is the trigger of the concurrency. So for example you change the tab it blurs both realities. Maybe the side effects get held but any new recalculation reads the updated value. Also any external to Solid side effect runs immediately.  The second is less concerning as there is nothing to be done for that. But the first hmm.. i guess without providing another wrapper where it is read. Or atleast where it enters the app. Hmmm.. it's clunkier but it would potentially be more correct.

3Shain:
I prefer to leave it over for the present...... It's a rare edge case (and likely to be over-engineered). Not a bug but could be enhanced (for better or worse). I may continue investigating on interruptible scheduling.

ryansolid:
It's not interruptible or at least what I call interruptible that I'm concerned about. We have the `enableScheduling` but it's the transitions without that I'm more concerned with. The typical async transitions. Those start a change, wait on async, and then commit. Those are way more common (ie.. baked into the router/all routing?). While I agree for navigation it will probably never get you in trouble it is not an edge case. To be fair I have a couple of holes still there in Solid proper but I intend to resolve those. And introducing something like this we should at least consider if it has an answer. I'm fine as-is for today, assuming there is a way out. We don't have the legacy of something like React, so it's beneficial to at least consider if a new feature can play along.

3Shain:
The critical issue is, should referential transparency be holden in (transition) pure computation? (by the name it should, of course) (reading a signal is seen as input, while `untrack(signal)` is the same as reading an external source). This could be a common pitfall, not only for this. 
It's not likely to be solved by solid solely <del>and could be a pain in the ass forever</del>.

3Shain:
After digging into the current implementation and future roadmap of solid transitions(concurrency),  I believe this feature is not compatible with it. The reasons:
* There are multiple realities, but externals have no clue in which reality they are
    *  and some of them dynamically collect dependencies. There will be multiple versions of dependency graph, not just multiple versions of value
* Referential transparency will be broken but no (easy) way to rescue. This is vital for interruptible scheduling, as it guarantees a computation will produce the same result with the same input no matter when it is evaluated.

But it doesn't mean this feature is not valuable. Without transition enabled it works well. And thanks to the low coupling between reactivity and rendering, it is possible for third party libraries to provide equivalent/similar mechanisms. (so it doesn't means it's impossible to have transition/concurrent scheduling with `enableExternalSource` anymore,  but this capability is turned over to externals)

That's to say, `enableExternalSource` will make solid transition noneffective.

3Shain:
I think I need some help. No clue how to test out transition-related stuffs . Seems `Transition` has been lack of tests

packages/solid/src/reactive/signal.ts
```diff
@@ -1211,6 +1242,19 @@ function createComputation<Next, Init = unknown>(
         }`;
   }
 
+  if (ExternalSourceFactory) {
+    const [track, trigger] = createSignal<void>(undefined, { equals: false });
+    const ordinary = ExternalSourceFactory(c.fn, trigger);
+    onCleanup(() => ordinary.dispose());
+    const triggerInTransition: () => void = () =>
+      startTransition(trigger, () => inTransition.dispose());
+    const inTransition = ExternalSourceFactory(c.fn, triggerInTransition);
```
3Shain, line 1251: These are just work-arounds and will be refactored once ryan 'fill the holes'

PR #762, opened 2021-12-13
refactor(store): improve store types

### LINKED ISSUE #732: produce causes TS error
**Describe the bug**
Valid usage of `produce` causes TS error. 

**To Reproduce**
Provide a code snippet and steps to reproduce the behavior:
1. Go to [playground](https://playground.solidjs.com/?version=1.2.2#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr2ZCouOAGVc3OFl6F5I+qxSsvKKKnQa2nr6ZB5CFlZWMvQQzLhotBC8AMIMEC6MABQAlPxWvA4ZMbzAMZ7eZHC47p4AurwAvA5OLs1CBSDMubiIvAAMkkWWmRUQVRiOcPB5HbzFHQB8pdPlDU2xcAW+tP6sBWQbW+VXvGQAdIPJuADUT2VXE5Nvkm9CuEyZBTe5QAPCZ6LgPJkCMR2iowRCMipeBkshpmABrdogeZCJa4STrIFXEC1IT3Ibfba8YH6eGQwnTT4QSlWISicQFNbtTbAnKPcS8fTrbwiWjMeh424Ac0aAFE1IthLgAEL4ACSIkBYCghEIKiKTLAkhaQA)
2. Wait for code to be type checked 
3. See the error

**Expected behavior**
No error

**Reproduction**
See above

**Additional context**
The error is silenced in the tutorial by `// @ts-nocheck` directive: https://www.solidjs.com/tutorial/stores_mutation?solved

otonashixav:
This PR focuses more on the performance and inference issues the store types have.
It is not meant to be a complete typing overhaul.

Added `bigint` to `NotWrappable`.

`Store` was quite complicated and often confusing as it contained properties specific to the Store implementation.
However, most internal functions use `StoreNode`. As such `Store<T>` has been repurposed to simply be `DeepReadonly<T>`.

`DeepReadonly` has been simplified.

The Setter type is mostly unchanged externally, except:
- Only `number` is allowed to index Array types (before, for example Array methods and properties were allowed).
- The fallback overload is now much better and correctly constrains types but requires type annotations (they can't be inferred).
- All functions passed via the setter have parameters typed readonly as they should(?) be.

I tried to optimize the Setter as much as possible, and I think it's better at the same number of overloads but don't have any definitive proof of that.

`produce` is now able to infer types correctly from the `setStore` call.
~~It currently uses a `DeepMutable` type to circumvent `DeepReadonly` being part of the inferred input; the DeepReadonly part shouldn't be inferred.~~

fixes #732

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -11,28 +11,8 @@ export type StoreNode = {
   [k: number]: any;
 };
 
-// well-known symbols need special treatment until https://github.com/microsoft/TypeScript/issues/24622 is implemented.
-type AddSymbolToPrimitive<T> = T extends { [Symbol.toPrimitive]: infer V }
-  ? { [Symbol.toPrimitive]: V }
-  : {};
-type AddSymbolIterator<T> = T extends { [Symbol.iterator]: infer V }
-  ? { [Symbol.iterator]: V }
-  : {};
-type AddSymbolToStringTag<T> = T extends { [Symbol.toStringTag]: infer V }
-  ? { [Symbol.toStringTag]: V }
-  : {};
-type AddCallable<T> = T extends { (...x: any[]): infer V } ? { (...x: Parameters<T>): V } : {};
-
-export type NotWrappable = string | number | boolean | Function | null;
-// Intersection for missing fields https://github.com/microsoft/TypeScript/issues/13543
-export type Store<T> = {
-  [P in keyof T]: T[P] extends object ? Store<T[P]> & T[P] : T[P];
-} & {
-  [$RAW]?: T;
-} & AddSymbolToPrimitive<T> &
-  AddSymbolIterator<T> &
-  AddSymbolToStringTag<T> &
-  AddCallable<T>;
+export type NotWrappable = string | number | bigint | boolean | Function | null;
+export type Store<T> = DeepReadonly<T>;
 
 function wrap<T extends StoreNode>(value: T, name?: string): Store<T> {
   let p = value[$PROXY];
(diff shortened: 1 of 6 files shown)


otonashixav:
Some areas I'd like help on:

- I've left `DeepReadonly` in, but it might make sense to just remove it. Should I? I'm trying not to mess with the exported types too much, since this PR is more focussed on improving performance and functionality, and I would rather the types be properly overhauled later on, with more tests to go with it. 
- The order of the overloads isn't very nice, but vscode won't give suggestions otherwise. This isn't the case with the existing types. I'd like to reorder them if possible while still having suggestions, but don't really know what's causing it.

packages/solid/store/src/store.ts
```diff
@@ -11,28 +11,8 @@ export type StoreNode = {
...
-// Intersection for missing fields https://github.com/microsoft/TypeScript/issues/13543
-export type Store<T> = {
-  [P in keyof T]: T[P] extends object ? Store<T[P]> & T[P] : T[P];
-} & {
-  [$RAW]?: T;
-} & AddSymbolToPrimitive<T> &
-  AddSymbolIterator<T> &
-  AddSymbolToStringTag<T> &
-  AddCallable<T>;
+export type NotWrappable = string | number | bigint | boolean | Function | null;
```
atk, line 14: Shouldn't (Weak)?(Map|Set) also be on that list? I know they are rarely used, but I don't think we'll map them to a proxy in a store.

otonashixav: I think all classes are on that list, but I'm not sure how to type "any class instance". Though I think there was some talk on discord about allowing user classes to be wrappable somehow. That said, listing all standard classes doesn't seem like the right way to go.

otonashixav: So it turns out if I define `type Wrappable = readonly unknown[] | Record<string, unknown>` and use it instead of `NotWrappable`, nothing breaks and only plain objects and arrays are allowed. Two errors appear [here](https://github.com/solidjs/solid/blob/main/packages/solid/store/test/store.spec.ts#L478):
```ts
    setStore("inner", "a", 10);
    //                ^ string is not assignable to never
    expect(sum).toBe(11);
    setStore("inner", "b", 5);
    //                ^ string is not assignable to never
    expect(sum).toBe(11);
```
...but since these setStore calls don't trigger updates, the error is possibly desirable? 

IThere might also be a slight naming issue here: `wrap` is used to wrap more than `Wrappable` and what passes `isWrappable`. Not sure if that will cause any confusion later on.

otonashixav:
Realised a little late that this is now broken:
```ts
const [store, setStore] = createStore({a: {b: 1}, c: {d: 2}});
// should be a type error but isn't
setStore("a", "d", 3)
```

PR #763, opened 2021-12-16
refactor(store): improve store types

### LINKED ISSUE #732: produce causes TS error

otonashixav:
store.ts
`NotWrappable` now consists of all non-object/array types.
Simplified `DeepReadonly<T>`.
Simplified `StoreNode`.
`Store<T>` is now an alias for `DeepReadonly<T>`.
Refactored `Part`, `Next`, and `SetStoreFunction` to improve their performance.
Changed `NullableNext` to `Next`, and `Next` to `WrappableNext`. `WrappableNext` excludes all non-wrappable types instead of just non-nullable types, so `{a: number | {b: number}}` now allows `setStore("a", "b", 1)`.
Improved the rest overload for `SetStoreFunction<T>`: it now correctly infers all types in the correct order, except for `ArrayFilterFn`s which need to be explicitly typed.

store.spec.ts
Added a few type test cases for store types.

modifiers.ts
Improved types of `reconcile` and `produce`; the latter now correctly infers types.

modifiers.spec.ts
Allow `produce` calls to have their types inferred.

server.ts
Import types from store.ts instead of redefining them.

mutable.ts
Return `T` instead of `Store<T>`, as the latter is no longer writable.

index.ts
Removed the `Readonly` export. Added `WrappableNext`.

fixes #732

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -3,36 +3,17 @@ export const $RAW = Symbol("store-raw"),
   $NODE = Symbol("store-node"),
   $NAME = Symbol("store-name");
 
-export type StoreNode = {
-  [$NODE]?: any;
-  [$PROXY]?: any;
-  [$NAME]?: string;
-  [k: string]: any;
-  [k: number]: any;
-};
-
-// well-known symbols need special treatment until https://github.com/microsoft/TypeScript/issues/24622 is implemented.
-type AddSymbolToPrimitive<T> = T extends { [Symbol.toPrimitive]: infer V }
-  ? { [Symbol.toPrimitive]: V }
-  : {};
-type AddSymbolIterator<T> = T extends { [Symbol.iterator]: infer V }
-  ? { [Symbol.iterator]: V }
-  : {};
-type AddSymbolToStringTag<T> = T extends { [Symbol.toStringTag]: infer V }
-  ? { [Symbol.toStringTag]: V }
-  : {};
-type AddCallable<T> = T extends { (...x: any[]): infer V } ? { (...x: Parameters<T>): V } : {};
-
-export type NotWrappable = string | number | boolean | Function | null;
-// Intersection for missing fields https://github.com/microsoft/TypeScript/issues/13543
-export type Store<T> = {
-  [P in keyof T]: T[P] extends object ? Store<T[P]> & T[P] : T[P];
-} & {
-  [$RAW]?: T;
-} & AddSymbolToPrimitive<T> &
-  AddSymbolIterator<T> &
-  AddSymbolToStringTag<T> &
-  AddCallable<T>;
+export type StoreNode = Record<keyof any, any>;
+export type NotWrappable =
+  | string
+  | number
+  | bigint
+  | symbol
+  | boolean
+  | Function
+  | null
+  | undefined;
+export type Store<T> = DeepReadonly<T>;
 
 function wrap<T extends StoreNode>(value: T, name?: string): Store<T> {
   let p = value[$PROXY];
@@ -95,7 +76,7 @@ export function getDataNodes(target: StoreNode) {
   return nodes;
 }
 
-export function proxyDescriptor(target: StoreNode, property: string | number | symbol) {
+export function proxyDescriptor(target: StoreNode, property: keyof any) {
   const desc = Reflect.getOwnPropertyDescriptor(target, property);
   if (
     !desc ||
@@ -108,7 +89,7 @@ export function proxyDescriptor(target: StoreNode, property: string | number | s
     return desc;
   delete desc.value;
   delete desc.writable;
-  desc.get = () => target[$PROXY][property as string | number];
+  desc.get = () => target[$PROXY][property];
   return desc;
 }
 
@@ -130,7 +111,7 @@ const proxyTraps: ProxyHandler<StoreNode> = {
   get(target, property, receiver) {
     if (property === $RAW) return target;
     if (property === $PROXY) return receiver;
-    const value = target[property as string | number];
+    const value = target[property];
     if (property === $NODE || property === "__proto__") return value;
 
     const wrappable = isWrappable(value);
@@ -145,7 +126,7 @@ const proxyTraps: ProxyHandler<StoreNode> = {
       node();
     }
     return wrappable
-      ? wrap(value, "_SOLID_DEV_" && target[$NAME] && `${target[$NAME]}:${property as string}`)
+      ? wrap(value, "_SOLID_DEV_" && target[$NAME] && `${target[$NAME]}:${property.toString()}`)
       : value;
   },
 
@@ -164,7 +145,7 @@ const proxyTraps: ProxyHandler<StoreNode> = {
   getOwnPropertyDescriptor: proxyDescriptor
 };
 
-export function setProperty(state: StoreNode, property: string | number, value: any) {
+export function setProperty(state: StoreNode, property: keyof any, value: any) {
   if (state[property] === value) return;
   const array = Array.isArray(state);
   const len = state.length;
@@ -188,7 +169,7 @@ function mergeStoreNode(state: StoreNode, value: Partial<StoreNode>) {
   }
 }
 
-export function updatePath(current: StoreNode, path: any[], traversed: (number | string)[] = []) {
+export function updatePath(current: StoreNode, path: any[], traversed: (keyof any)[] = []) {
   let part,
     prev = current;
   if (path.length > 1) {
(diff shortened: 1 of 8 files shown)


packages/solid/store/src/store.ts
```diff
@@ -235,141 +215,208 @@ export function updatePath(current: StoreNode, path: any[], traversed: (number |
...
-  index: number
-) => boolean;
+export type ArrayFilterFn<T> = (item: T, index: number) => boolean;
 
-export type Part<T> = T extends any[]
-  ? keyof T | Array<keyof T> | ArrayFilterFn<T> | StorePathRange
-  : T extends object
-  ? keyof T | Array<keyof T>
-  : never;
+export type Part<T> = [T] extends [NotWrappable]
```
bgoscinski, line 230: Out of curiosity: do you have any article or could you tell me what should I search for to understand this, emm... thing?
`[T] extends [U]` 🤯

bgoscinski: Nvm. I think I found it https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types :)

packages/solid/store/src/store.ts
```diff
@@ -235,141 +215,208 @@ export function updatePath(current: StoreNode, path: any[], traversed: (number |
...
-    K6 extends Part<Next<Next<Next<Next<Next<T, K1>, K2>, K3>, K4>, K5>>,
-    K7 extends Part<Next<Next<Next<Next<Next<Next<T, K1>, K2>, K3>, K4>, K5>, K6>>,
-    K8 extends Part<Next<Next<Next<Next<Next<Next<Next<T, K1>, K2>, K3>, K4>, K5>, K6>, K7>>
-  >(
-    ...args: [K1, K2, K3, K4, K5, K6, K7, K8, ...(Part<any> | StoreSetter<any>)[]]
-  ): void;
-}
+export type WrappableNext<T, K extends Part<T>> = Exclude<Next<T, K>, NotWrappable>;
+
+type DistributeRest<T, P extends Part<T>> = P extends P ? [P, ...Rest<Next<T, P>>] : never;
```
bgoscinski, line 249: Okay this one is a mystery for me. Can you point me where should I look for clues what `P extends P` does?

otonashixav: It forces the type to be distributed. For example if `P = A | B`, the resultant type would be `[A, ...Rest<Next<T, A>>] | [B, ...Rest<Next<T, B>>]` instead of `[A | B, ...Rest<Next<T, A | B>>]`.

packages/solid/store/src/store.ts
otonashixav: I've removed this particular one. Since we're never going to pass a non-wrappable `T` to `Part`, there's no need to check and it improves performance which was previously a major concern.

non25:
Hi, trying this out: https://playground.solidjs.com/?hash=2095550606&version=1.2.5
Can't think of a way to fix it :thinking: 
Works better in the playground I guess...

atk:
The playground disables strict type checks, which breaks a few types that rely on them, unfortunately. Maybe we can enable strict in the playground, but selectively disable no-implicit-any and a few other convenient options that don't break strict types.

otonashixav:
I think it isn't possible to avoid a type error in this case; the existing types also error here:
```ts
// interactions with generics
<T extends string>(v: T) => {
  type A = { a?: T; b?: Record<string, string>; c?: Record<T, string> };
  const [, setStore] = createStore<A>({});
  // should allow
  setStore("a", v);
  setStore("b", "a", "c");
  setStore("c", v, "c"); // <-- here
};

```

atk:
Could `<T extends string = string>` work to narrow down the cases and still allow type overwrites?

otonashixav:
It doesn't seem to work. Even if it did I'm not sure if it would be expected to the end user.

atk:
The problem here is that T extends string can be an arbitrary union that doesn't fully overlap with keyof {} that is string | number | symbol. I don't think this is overly useful.

otonashixav:
`T extends string` should be able to index `Record<T, string>` though, which is what's going on here. I suspect `Store<T>` is causing issues since I can index a non-store just fine:
```ts

  type A = { a?: T; b?: Record<string, string>; c?: Record<T, string> };
  const a = {} as A;
  const [store, setStore] = createStore<A>(a);
  // should allow
  setStore("a", v);
  setStore("b", "a", "c");
  // TODO
  setStore("c", v, "c"); // 
  store.c?.[v]; // <-- Expression produces a union type that is too complex to represent.ts(2590)
  a.c?.[v]; // <-- works
```

atk:
Obviously, you cannot set a value that is readonly, so that error is to be expected. Maybe what we need is not DeepReadonly, but DeepStructuralReadonly, which only makes properties immutable that are arrays or objects.

otonashixav:
It is not readonly; I'm not setting the value on the readonly store, but on a variable the same type as the value at the end of the store i.e. `let a: Readonly<Record<string, string>> = "1"` works, so this should too.

otonashixav:
If no one is able to solve the type errors caused by [generics](https://github.com/solidjs/solid/pull/763/files#diff-718d3786c3e47b99ae8f43e84bf9d68788a8a193ac46f3e5e1d649d0963f6655R662), I believe no more improvements can be made until https://github.com/microsoft/TypeScript/issues/43122 is resolved.

ryansolid:
I don't understand this stuff too well and there have been a few version.. but what is the state of this PR? Candidate for 1.3? What does it do?

otonashixav:
I don't believet there's anything more to add to the PR, other than possibly a type for the return value of `createStore`. The main post includes most of the changes; the most noticable changes are likely `Store`, which now is fully readonly and simplified to allow ts to show better error messages, and `SetStoreFunction`, which has its types inferred much faster. 

I realise the one change I've left out is the replacement of instances of `property` typed as `string | number` with `keyof any`, resulting in a runtime change [here](https://github.com/solidjs/solid/pull/763/files#diff-86ed9ba88e2ceb505ab412a1b54f475b36555ad70a84af10685b8df8fa52f001R129), to avoid a runtime error in dev mode when wrapping a value whose property is a symbol. 

```ts
  const s = Symbol();
  const [store] = createStore({ [s]: { a: 1 } });
  const c = store[s].a; // <-- TypeError: Cannot convert a Symbol value to a string
```

ryansolid:
Hmm.. I noticed type tests are off in the store package. I probably did that because I was having headaches with it. But a bunch of stuff is failing it might be worth using that to double-check the work here. This list should have "./store/test" added to it:
https://github.com/solidjs/solid/blob/main/packages/solid/tsconfig.test.json#L11

otonashixav:
Some of the errors are expected, since there are some edge cases which cannot be represented with typescript currently. I suppose the tests shouldn't be failing though, so I'll change that. The other errors were missed due to `strict: false`, so those need fixing.

yonathan06:
Could it be that this changes broke types for `setStore` using `produce`?
Getting this type check error:
Screen Shot 2022-02-01 at 10 40 46

[playground example](https://playground.solidjs.com/?hash=-187295914&version=1.3.3)

atk:
I think the type for produce needs to unwrap the store.

otonashixav:
I figured it was fine since there wasn't any issue with the test cases but the types are definitely wrong for the top level.

otonashixav:
I see the issue: readonly on arrays isn't being removed by inference. See https://tsplay.dev/mL42vm for an example. Fix is in #824

PR #769, opened 2021-12-23
fix: Change ordering for module exports field

Drevoed:
This PR fixes the ordering of exports field to comply with the node spec and how https://github.com/lukeed/resolve.exports 
works.

Also adds exports field for subpackages.

```
Within the "exports" object, key order is significant. 
During condition matching, earlier entries have higher priority and take precedence over later entries. 
The general rule is that conditions should be from most specific to least specific in object order.
```

### DIFF
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -44,13 +44,17 @@
         "import": "./dist/solid.js",
         "require": "./dist/solid.cjs"
       },
+      "node": {
+        "development": {
+          "import": "./dist/dev.js",
+          "require": "./dist/dev.cjs"
+        },
+        "import": "./dist/server.js",
+        "require": "./dist/server.cjs"
+      },
       "development": {
         "import": "./dist/dev.js",
         "require": "./dist/dev.cjs"
-      },
-      "node": {
-        "import": "./dist/server.js",
-        "require": "./dist/server.cjs"
       },
       "import": "./dist/solid.js",
       "require": "./dist/solid.cjs"
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -44,13 +44,13 @@
         "import": "./dist/solid.js",
         "require": "./dist/solid.cjs"
       },
-      "node": {
-        "import": "./dist/server.js",
-        "require": "./dist/server.cjs"
-      },
       "development": {
         "import": "./dist/dev.js",
         "require": "./dist/dev.cjs"
+      },
+      "node": {
+        "import": "./dist/server.js",
+        "require": "./dist/server.cjs"
       },
       "import": "./dist/solid.js",
       "require": "./dist/solid.cjs"
--- a/packages/solid/store/package.json
+++ b/packages/solid/store/package.json
@@ -20,13 +20,17 @@
         "import": "./dist/store.js",
         "require": "./dist/store.cjs"
       },
+      "node": {
+        "development": {
+          "import": "./dist/dev.js",
+          "require": "./dist/dev.cjs"
+        },
+        "import": "./dist/server.js",
+        "require": "./dist/server.cjs"
+      },
       "development": {
         "import": "./dist/dev.js",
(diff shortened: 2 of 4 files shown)


Drevoed:
This also fixes integration with vitest and probably anything that has to do with bundlers that are dependent on https://github.com/lukeed/resolve.exports.

ryansolid:
Thank you. Actually made a very similar fix on the `next` branch. Development needs to come after node since there is no server dev build and ssr builds will be wrong. It's probably best to test new stuff against the beta tag. This will all get merged pretty soon.

Drevoed:
> Thank you. Actually made a very similar fix on the `next` branch. Development needs to come after node since there is no server dev build and ssr builds will be wrong. It's probably best to test new stuff against the beta tag. This will all get merged pretty soon.

Yeah, I thought about it and moved top-level `development` after `node` instead, adding extra `development` fields to `node`

ryansolid:
I meant the server builds don't have dev and shouldn't use dev because that is client. So they shouldn't have their own nested development exports. It should be browser, node, development, import/require. And only browser should have the nested development.

atk:
Wouldn't a server-dev variant be useful, too?

Drevoed:
But if the exports were specified as you said then the tests wouldn't work properly. As in, the node test runner will try to resolve server.(js/cjs) instead of dev.(cjs/js)

LXSMNSYC:
> But if the exports were specified as you said then the tests wouldn't work properly. As in, the node test runner will try to resolve server.(js/cjs) instead of dev.(cjs/js)

have you also tested this in esbuild?

Drevoed:
> > But if the exports were specified as you said then the tests wouldn't work properly. As in, the node test runner will try to resolve server.(js/cjs) instead of dev.(cjs/js)
> 
> have you also tested this in esbuild?

Haven't tested, but glancing at the source code it *should* work.

https://github.com/evanw/esbuild/blob/434bbc62091e918a235b6f3755d1c13b14216dc3/internal/resolver/package_json.go#L874

LXSMNSYC:
I had problems with this before. With conditions `node` and `development` existing, `resolve.exports` resolves to `server` builds while `esbuild` resolves to `dev` builds.

ryansolid:
> Wouldn't a server-dev variant be useful, too?

Possibly, but it doesn't exist currently. There is no server dev build, so giving it the client one will just break things.

atk:
We can still use web instead of dev in vitest, so let's remove node/development for now.

ryansolid:
Hmmm.. also the export maps on the nested package.json's shouldn't be needed. Any modern ESM module will read off the root one, and the nested ones are just for older CJS aliasing.

Drevoed:
> Hmmm.. also the export maps on the nested package.json's shouldn't be needed. Any modern ESM module will read off the root one, and the nested ones are just for older CJS aliasing.

In case of vite it does resolve each module and uses that nested module package.json.

I might debug it and check if that's the intended behaviour, but this is exactly the case.

ryansolid:
I haven't merged mostly because I don't see most of this as necessary with 1.3 release. Can anyone confirm that?

Drevoed:
I think this can wait a patch release later? I mean 1.3.x

ryansolid:
I mean is it necessary at all with changes made in 1.3?

atk:
I've tested it again. Yes, at least the reordering will fix the resolution in vitest.

ryansolid:
I am saying current 1.3 did reordering without this PR. Is it currently fine?

I'm trying to evaluate whether we need to merge the subpackage export maps which shouldn't be necessary but never know with certain tooling.

atk:
No, the updates in version 1.3 are not sufficient.

ryansolid:
Ok subpackage export maps it is. There is one issue here. Every development under node needs to be removed because there is no development node build. Only client. That just breaks would be development server environments.

PR #784, opened 2022-01-06
include test files in tsconfig.json so that IDEs will type check using the same settings as with source files

trusktr:
Without this change, test files are not in strict mode, while Solid's
source is designed to work in strict mode, so intellisense and type
checking within IDEs will not show the full picture. For example, in
signals.type-tests.ts, without this change there will be errors on many
of the @ts-expect-error comments because non-strict mode won't trigger
the errors we've designed the types to produce in strict mode.

Namely, now you'll be able to clone the repo, npm install && npm run build, and opening signals.type-tests.ts won't be full of red squigglies.

Previously, one would have to modify tsconfig files to get tests to have proper type checks.

I'm not sure if modifying tsconfigs like this has any undesirable effect on build output, which is why I didn't do this before, and what I'm now investigating.


trusktr:
Ah, that's right. The build fails because including the test files changes the structure of the output in the `types` folder.

We need a solution that allows us proper intellisense and type checking in our IDE, but does not include test files in the output. I think project references is the way to do go, but I'd like to see if there are any other options.

fictitious:
> I believe project references is the way to do go, but I'd like to see if there are any other options

Yes project references is the "official" way.

Another possible solution is to have three `tsconfig` files: one with common compiler settings, and two others - for test and for build - which  [extends](https://www.typescriptlang.org/tsconfig#extends) the common one. But you need to repeat (and maintain in sync) the part that includes non-test files in two tsconfigs, and, in my experience, some IDEs become confused when they see that the same source file is included in two different tsconfig files.

trusktr:
Just pushed a commit that goes with the first non-project-references option. It was the simplest way to do it, for now.

PR #790, opened 2022-01-10
Refactor store to remove readonly

otonashixav:
After changing `Store` to be readonly, @high1 pointed out that non-readonly types might be more desirable e.g. in this scenario:

```ts
function getFirst(list: string[]): string {
  return list[0];
}

const [store, setStore] = createStore({ list: ["hello"] });

getFirst(store.list) // <-- readonly string[] not assignable to string[]
setStore("list", (list) => [getFirst(list)]) // <-- readonly string[] not assignable to string[]
```

It's fairly common to encounter immutable functions that do not type their parameters as `readonly` outside of native functions/classes like `Object.fromEntries` or `Map`. This PR makes it so none of these are readonly. However, I'm not sure if `setStore` should be kept as readonly. 

I imagine that the same reasoning can be applied to both `store` and `setStore` here, but it is much less likely for users to pass the `list` in `setStore` to a function typed with `string[]`, and much more likely for users to try to directly modify `list`.

### DIFF
--- a/packages/solid/store/src/index.ts
+++ b/packages/solid/store/src/index.ts
@@ -9,8 +9,7 @@ export type {
   ArrayFilterFn,
   Part,
   Next,
-  WrappableNext,
-  DeepReadonly
+  WrappableNext
 } from "./store";
 export * from "./mutable";
 export * from "./modifiers";
--- a/packages/solid/store/src/modifiers.ts
+++ b/packages/solid/store/src/modifiers.ts
@@ -1,4 +1,4 @@
-import { setProperty, unwrap, isWrappable, Store, StoreNode, $RAW } from "./store";
+import { setProperty, unwrap, isWrappable, Store, StoreNode, $RAW, DeepReadonly } from "./store";
 
 export type ReconcileOptions = {
   key?: string | null;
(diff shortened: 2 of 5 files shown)


trusktr:
I have a feeling that keeping it correct (all readonly) is the way to go.

It is similar in practice to writing `let foo` for a variable that never changes, when we should be using `const` instead.

In a similar vein, if a function is only reading, it _should_ be defined with `readonly` for it to be correct (how many of us forget to do that :), let's add lint rules).

[playground example](https://playground.solidjs.com/?hash=-420939416&version=1.3.3) that shows the "right" and "wrong" ways to write code.

> It's fairly common to encounter immutable functions that do not type their parameters as `readonly` outside of native functions/classes like `Object.fromEntries` or `Map`.

I think it is an artifact from a time before `readonly` was in place.

---

I think it is better to keep people accidentally trying to modify state and things not working, or passing state to some APIs and expecting it to perform state management and it not working, etc.

high1:
The issue here is the usage of the store - this forces the user to unbox every value that he reads from the store, and that's visible from the example, right? Although the value is only read and not mutated, you are forced to unbox it from readonly.

trusktr:
If you can post playground examples, that'd be helpful.

non25:
I guess it's better to use `_.cloneDeep`, spread or override some library types globally in the project.

[Sad times...](https://tsplay.dev/WkMEPN)

non25:
https://tsplay.dev/WvVkBW this is actually great and will prevent shooting ourselves in the foot in case we try to do pure functions without immer.
I've seen `.sort()` and deep object spreads with nested properties mutation in "pure" functions.
DeepReadonly would shout on them.

atk:
I feel this is more breaking than making the types. Can we instead provide an `Unwrap<Store>` helper type for those rare cases when you really need to abuse stores in such a way?

PR #796, opened 2022-01-16
fix: fix type of `ResourceOptions<T>`

### LINKED ISSUE #795: `createResource` + `T` generic from a parent function throws a TS error.
Doing something like this:

```ts
function useSomeResource<T>(value: T) {
  const [data] = createResource<T>(source, fetchData, {
    initialValue: value,
  });
}
```

Raises a TS error `Expected 1-2 arguments, but got 3.ts(2554)` for the options field.
Omitting `<T>` from `createResource` raises the TS error

```
No overload matches this call.
  Overload 1 of 4, '(source: ResourceSource<true>, fetcher: ResourceFetcher<true, T>, options?: { initialValue?: undefined; name?: string | undefined; globalRefetch?: boolean | undefined; onHydrated?: (<S, T>(k: S, info: ResourceFetcherInfo<...>) => void) | undefined; } | undefined): ResourceReturn<...>', gave the following error.
    Type 'L' is not assignable to type 'undefined'.
  Overload 2 of 4, '(source: ResourceSource<true>, fetcher: ResourceFetcher<true, T>, options: ResourceOptions<T>): ResourceReturn<T>', gave the following error.
    Argument of type '{ initialValue: T; }' is not assignable to parameter of type 'ResourceOptions<T>'.ts(2769)
```

otonashixav:
Changed `T extends undefined` to `undefined extends T`
Fixed one other line that errored as a result

Fixes #795

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -420,7 +420,7 @@ export type ResourceFetcher<S, T> = (k: S, info: ResourceFetcherInfo<T>) => T | 
 
 export type ResourceFetcherInfo<T> = { value: T | undefined; refetching?: unknown };
 
-export type ResourceOptions<T> = T extends undefined
+export type ResourceOptions<T> = undefined extends T
   ? {
       initialValue?: T;
       name?: string;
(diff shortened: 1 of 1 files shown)


otonashixav:
Note this only fixes the case with generics unspecified, since it isn't possible to have typescript partially infer generics, which is required in this case to fix the issue when calling createResource with 3 parameter and specifying only 1 generic. The solution if the types are not correctly inferred is to provide both generics.

atk:
Looks good to me.

PR #814, opened 2022-01-28
Export noShadowDOM from component-register in solid-element

ghinda:
* Export the `noShadowDOM` method from `component-register` in `solid-element`.

The `noShadowDOM` method is available in `component-register`, but not directly available from `solid-element`.
You can work-around this by importing it from `component-register`. 

Since we already have `hot` and `getCurrentElement` exported, it would be great to also have `noShadowDOM`.


ryansolid:
Yeah that makes a ton of sense. Thanks.

PR #824, opened 2022-02-01
fix(store): fix inferrence of array types

otonashixav:
Previously produce would wrongly infer all arrays from a store as readonly.
This fix makes the return type of produce StoreSetter instead, fixing the issue.
This change has a few caveats; namely, produce is less strictly typed: it should return a function that takes and returns `DeepReadonly<T>`, but now it returns all types contained in `StoreSetter<T>`. This has no consequence for normal usage other than better type inference.
Additionally to facilitate this change, DeepReadonly is no longer applied to the entire SetStoreFunction, but only where it is consumed (the setter and filter functions).

#790 would also fix this issue.

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -223,11 +223,11 @@ export type DeepReadonly<T> = {
 export type StoreSetter<T> =
   | T
   | Partial<T>
-  | ((prevState: T, traversed?: (keyof any)[]) => Partial<T> | void);
+  | ((prevState: DeepReadonly<T>, traversed?: (keyof any)[]) => Partial<DeepReadonly<T>> | void);
 
 export type StorePathRange = { from?: number; to?: number; by?: number };
 
-export type ArrayFilterFn<T> = (item: T, index: number) => boolean;
+export type ArrayFilterFn<T> = (item: DeepReadonly<T>, index: number) => boolean;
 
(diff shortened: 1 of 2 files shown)


packages/solid/store/src/store.ts
```diff
@@ -223,11 +223,11 @@ export type DeepReadonly<T> = {
 export type StoreSetter<T> =
   | T
   | Partial<T>
-  | ((prevState: T, traversed?: (keyof any)[]) => Partial<T> | void);
+  | ((prevState: DeepReadonly<T>, traversed?: (keyof any)[]) => Partial<DeepReadonly<T>> | void);
```
trusktr, line 226: I haven't gotten to test inference here yet. But if there's a way to fix the broken inference you mentioned, while keeping the "correct" readonly types, that'd be nice, so as to prevent people doing the wrong thing.

otonashixav: This PR changes two things:

* The types of `SetStoreFunction` and `StoreSetter` such that `DeepReadonly` isn't applied directly at the root of the former, and instead only where the user would "interact" with the store in the latter. 
* `produce` now accepts all types that are part of the union of `StoreSetter` so that it can use the `T` type included in that type to properly type its own parameters, avoiding the inference issue. 

Do you mean that `StoreSetter` should also accept `DeepReadonly` variants of `T | Partial<T>`?

PR #843, opened 2022-02-09
fix(store): fix traversed returning duplicate values

### LINKED ISSUE #804: Solid JSX HTMLElement types do not allow `on:event` props
**Describe the bug**
When using the `on:event` syntax on HTML elements in JSX in TypeScript the props are considered invalid:

```typescript
<div on:click={() => null} />
```

```
(JSX attribute) on:click: () => null
Type '{ "on:click": () => null; }' is not assignable to type 'HTMLAttributes<HTMLDivElement>'.
  Property 'on:click' does not exist on type 'HTMLAttributes<HTMLDivElement>'. Did you mean 'onclick'?ts(2322)
```

**Expected behavior**

The behaviour of `on:[event]` handlers should be typed correctly. I've tried extending the `HTMLElement` interface to allow this, but I'm stuck with being unable to correctly:

```typescript
interface HTMLElement {
  [prop: `on:${keyof HTMLElementEventMap}`]: <
    T extends keyof HTMLElementEventMap
  >(
    ev: HTMLElementEventMap[T]
  ) => void;
}
```

Unfortunately, the error with this type seems to suggest that I'd have to write out each individual `on:${event}` JSX handler, unless someone can suggest how to create a mapped type from the keys presented above. Before I attempt that, would this type extension belong in `solidjs` or `dom-expressions`?

### LINKED ISSUE #840: `traversed` in `setStore` duplicates traversed keys when passed an array, function or object
**Describe the bug**
When passing arguments to `setStore`, passing an array of keys, or function/object to filter an array, the traversed key is added twice: once on initial pass, and a second time by the recursive call of `updatePath`. 

The traversed array is also in reverse order from what I expected, but I'm not sure if that's intentional. 

**To Reproduce**
[Playground Link](https://playground.solidjs.com/?version=1.3.5#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9mjOFFxwAyrm5xeAX14AzRrRi8AOjVoAbNABMAtACsyAejJThGgNxqIt5rQjnewLLzJxck6QF1eAXkFhUQkLOAAKAR1zRBcABjcARh83WgAjQzhmXFiBKFjEtzTYgCY5OQBKGwgTE156MgwAc14oRkYofF5aBV4Aazh8MlsPLzDwjWjcDTdgOJTecIB9N1wKgIA+QUc6LTgAOi1aJvC1qpHPb2EJsHTM7JmXDSgNBeXV9f8thydtA6OTmdqrV6o0IC1mFAtFo0lBmH0LmNpBMILx1GApjNbGj3rw0J8tmgAv5AglsYsVrw1pttr89odjqcKrZzjU6g1moplDB9FJ9Gl8IirhFJmhzI8BEoVLEElTaDK5G5cdSvrTdv9GUDbGBZD4gA)

**Expected behavior**
Each traversed key is only added once.

otonashixav:
Fixes #840

~~Also made the type of `traversed` reflect the keys traversed e.g.~~ Will make a PR after the other store type PRs are done
```ts
setStore("list", [0, 1], (v, traversed) => {
  traversed[0]; // <-- number
  traversed[1]; // <-- "list"
})
```

### DIFF
--- a/packages/solid/store/src/server.ts
+++ b/packages/solid/store/src/server.ts
@@ -40,21 +40,20 @@ export function updatePath(current: any, path: any[], traversed: (number | strin
     if (Array.isArray(part)) {
       // Ex. update('data', [2, 23], 'label', l => l + ' !!!');
       for (let i = 0; i < part.length; i++) {
-        updatePath(current, [part[i]].concat(path), [part[i]].concat(traversed));
+        updatePath(current, [part[i]].concat(path), traversed);
       }
       return;
     } else if (isArray && partType === "function") {
       // Ex. update('data', i => i.id === 42, 'label', l => l + ' !!!');
       for (let i = 0; i < current.length; i++) {
-        if (part(current[i], i))
-          updatePath(current, [i].concat(path), ([i] as (number | string)[]).concat(traversed));
+        if (part(current[i], i)) updatePath(current, [i].concat(path), traversed);
       }
       return;
     } else if (isArray && partType === "object") {
       // Ex. update('data', { from: 3, to: 12, by: 2 }, 'label', l => l + ' !!!');
       const { from = 0, to = current.length - 1, by = 1 } = part;
       for (let i = from; i <= to; i += by) {
-        updatePath(current, [i].concat(path), ([i] as (number | string)[]).concat(traversed));
+        updatePath(current, [i].concat(path), traversed);
       }
       return;
     } else if (path.length > 1) {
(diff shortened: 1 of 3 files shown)


otonashixav:
I left #824 in this by accident, going to just keep it to fixing traversed and leave the types to later

ryansolid:
Thanks. Looks good.

PR #848, opened 2022-02-12
refactor(store): Refactor store types again

otonashixav:
Overall:
- Changed all instances of property keys typed with `keyof any` or some subset of `string | number | symbol` to the built-in `PropertyKey` type to improve readability.
- Changed all instances of `Store<T>` with `DeepReadonly<T>` except those that explicitly accept a store e.g. `create(Store|Mutable)`.
- Marked parameters `readonly` or `DeepReadonly` where possible to enforce const-correctness (is this the correct term in TS?).
- Fixed interactions with `any` (ugh).
- Removed more instances of unneeded casting.
- Added a `DeepMutable` type. This is probably the easiest way of solving mutability issues, including `produce`. Users can now cast stores as `DeepMutable<typeof store>` if desired (not recommended).

SetStoreFunction:
- Fixed the type of `traversed` in `setStore`.  It now correctly types the traversed path.
- Removed the "intermediate" types in SetStoreFunction as they were causing issues with inference for `produce` and `reconcile` in the absence of a proper `NoInfer` (`T & { [K in keyof T]: T[K] }` only delays inference, but doesn't completely disable it, causing issues when providing a generic function directly to a generic parameter). This drastically slows down inference.
- To combat the slowdown described above, further optimized SetStoreFunction by not using `Part` as the generics to avoid needing to transform the types back from `Part` to access a property's type. This is strictly an improvement afaict, also improving readability when calling `setStore`.
- As a result, `WrappableNext` and `Next` are removed. There is a `W` (for `Wrappable`) type, but it is a simple shortcut for (`Exclude<T, NotWrappable>`). I don't believe these types were necessary to export, but if desired they can be readded (albeit they are unnecessary now). Added a `KeyOf<T>` as `keyof T` is inadequate.

produce:
- Now correctly strips arrays of their readonly type in the mutation function. Uses the new `DeepMutable` type, as this is the least complicated way to correctly type a produce function called on a nested store.

reconcile:
- Now disallows passing a value with missing non-optional properties which resulted in the store being wrongly typed after being reconciled with a value with missing properties.

runtime changes:
- The server `reconcile` function was changed to more closely match the client's version of the function in both functionality and typing. Previously it returned `(state: T) => void`, but this was incompatible with the client version which always returned `(state: T) => T` (or now (state: U) => T).
- Added `toString` to a `string | symbol` used in a template literal in `mutable.ts` to avoid a dev mode error. A similar fix was in https://github.com/solidjs/solid/pull/763 for stores which also had the same issue.

### DIFF
--- a/packages/solid/store/src/server.ts
+++ b/packages/solid/store/src/server.ts
@@ -1,4 +1,4 @@
-import { SetStoreFunction, Store } from "store";
+import type { DeepMutable, DeepReadonly, SetStoreFunction, Store, StoreSetter } from "store";
 
 export const $RAW = Symbol("state-raw");
 
@@ -14,7 +14,7 @@ export function unwrap<T>(item: any): T {
   return item;
 }
 
-export function setProperty(state: any, property: string | number, value: any, force?: boolean) {
+export function setProperty(state: any, property: PropertyKey, value: any, force?: boolean) {
   if (!force && state[property] === value) return;
   if (value === undefined) {
     delete state[property];
@@ -29,7 +29,7 @@ function mergeStoreNode(state: any, value: any, force?: boolean) {
   }
 }
 
-export function updatePath(current: any, path: any[], traversed: (number | string)[] = []) {
+export function updatePath(current: any, path: any[], traversed: PropertyKey[] = []) {
   let part,
     next = current;
   if (path.length > 1) {
@@ -81,8 +81,8 @@ export function createStore<T>(state: T | Store<T>): [Store<T>, SetStoreFunction
   return [state as Store<T>, setStore];
 }
 
-export function createMutable<T>(state: T | Store<T>): Store<T> {
-  return state as Store<T>;
+export function createMutable<T>(state: T | Store<T>): T {
+  return state as T;
 }
 
 type ReconcileOptions = {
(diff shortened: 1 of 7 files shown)


otonashixav:
Seeing as the discussion in #790 and on discord steered towards readonly-correctness, this PR enforces that and fixes remaining issues/issues caused by #763 (the produce bug). 

This PR does not include a helper for removing `DeepReadonly`. If possible I'd like to accomplish it without a `DeepMutable`, although I doubt it is possible without attaching the original type somehow. `$RAW` which already exists on the object can be used to contain it without adding a "fake" property, but it's somewhat confusing to see the symbol on all stores.

Additionally `reconcile` is unreconcilably wrongly typed -- for now I've left it untouched but it's possible to do:

[Playground](https://playground.solidjs.com/?hash=1468163508&version=1.3.7)
```ts
const [store, setStore] = createStore({v: 1});
setStore(reconcile({}));
store.v // this is undefined now, but store.v is of type number
```

I can't figure out how to force TS to infer from strictly the return type and not the parameter, while also allowing `NotWrappable`s to be passed in/accepted by the returned function.

packages/solid/store/src/store.ts
```diff
@@ -215,171 +215,150 @@ export function updatePath(current: StoreNode, path: any[], traversed: (keyof an
...
-  ? [...(keyof any)[], any]
-  : [StoreSetter<T>] | (T extends NotWrappable ? never : DistributeRest<T, Part<T>>);
+// specially handle keyof to avoid errors with arrays and any
+type KeyOf<T> = number extends keyof T
+  ? 0 extends 1 & T // if it's any just return keyof T
+    ? keyof T
+    : [T] extends [never] // keyof never is PropertyKey which number extends; return never
+    ? never
+    : number // it's not any or never so it's an array or tuple; exclude the non-number properties
+  : keyof T;
```
otonashixav, line 255: This solves a lot of issues, as originally using `keyof T` would break with arrays, as it included other undesired array properties. By using a custom `KeyOf` instead, we can simplify most of the other types greatly. Not sure if this should be exported.

packages/solid/store/src/server.ts
```diff
@@ -91,27 +91,27 @@ type ReconcileOptions = {
...
-export function reconcile<T>(
-  value: T | Store<T>,
+export function reconcile<T extends U, U>(
+  value: T,
   options: ReconcileOptions = {}
-): (state: Store<T>) => void {
+): (state: U) => T {
   return state => {
-    if (!isWrappable(state)) return value;
+    if (!isWrappable(state) || !isWrappable(value)) return value;
```
otonashixav, line 99: I added a wrappable check to `value` to match the client version of `reconcile`, and afaict it makes sense to return if `value` is not wrappable since the whole object will be replaced. Also `Object.keys` will fail below if `value` is nullable. @ryansolid need your input on this change, and the added return statement at the end of this function to make it match the client which always returns either `state` or `value`.

ryansolid: Looks good. Historically I have missed making updates in both places when I should have.

otonashixav: Thanks!

PR #849, opened 2022-02-13
Fix incompatibility with RxJS

### LINKED ISSUE #554: observable -> Cannot read property 'isStopped' of undefined
**Describe the bug**
try to use solidjs's `observable` and rxjs's `from` operator for converting a signal to stream, but got a little error。
It looks like solid’s `observable` cant adapt rxjs's observable interface。

image
image

**To Reproduce**
just run the following code
```tsx
import { from, Observable } from "rxjs";
import { observable, createSignal, onMount } from "solid-js";

function App() {
  const [getText, setText] = createSignal("Hello Solid");
  onMount(() => {
    setInterval(() => {
      setText(new Date().toLocaleTimeString());
    }, 1000);
  });

  // workaround
  new Observable((sub) => {
    observable(getText).subscribe((v) => {
      sub.next(v);
    });
  }).subscribe((v) => {
    console.log("new Observable", v);
  });

  // bug
  const text$ = from(observable(getText));
  text$.subscribe((v) => {
    console.log("from operator", v);
  });

  return <div>{getText()}</div>;
}

export default App;
```

**Expected behavior**
work without any errors

**Reproduction**
repo
https://github.com/Gavin-Gong/solidjs-observable-reporting 

**Additional context**
maybe we could add some description of `observable`' `subscribe` method in documentions?

weltenwort:
## :memo: Summary

This fixes an incompatibility of `observable()` with RxJS's `from()`.

fixes #554 

## :female_detective: Review notes

The cause for the incompatibility seems to be the following combination of circumstances:

- The `subscribe()` function created inside of Solid's `observable()` stores a reference to the observer's `next` function and calls it without explicitly supplying `this`.
- The `next` function of RxJS's `Subscriber` is not statically bound to the subscriber instance.

These two combine to the effect that a call to `next` originating from Solid's `observable` sees an incorrect value for `this`. This fix explicitly `.bind()`s the `next` handler to the observer to avoid that.

### DIFF
--- a/packages/solid/src/reactive/observable.ts
+++ b/packages/solid/src/reactive/observable.ts
@@ -1,4 +1,4 @@
-import { createComputed, untrack, Accessor, createSignal, Setter, onCleanup } from "./signal";
+import { Accessor, createComputed, createSignal, onCleanup, Setter, untrack } from "./signal";
 
 function getSymbol() {
   const SymbolCopy = Symbol as any;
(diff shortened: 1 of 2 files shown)


ryansolid:
Thank you.

PR #868, opened 2022-02-24
Enable blank GitHub issues

edemaine:
Oren mentioned on Discord that he couldn't report a general issue/concern, only fill out the bug report template. This change should allow starting from a blank template as an option, while still making the bug report template prominent at the top.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: 🤔 Long question or idea
     url: https://github.com/solidjs/solid/discussions


ryansolid:
I see.. Biggest reason I went this way is I want people to use discussions more and try to keep the issues to things that we acknowledge as actual issues. Bugs are the obvious one. But sure lets see how this goes.

edemaine:
The alternative would be to add some more templates, I guess, but I feel like it's pretty hard to find the blank template at the bottom after all the other options. So this should still cut down on extraneous issues. Anyway, as you said, we'll see what happens!

cliffordfajardo:
> Biggest reason I went this way is I want people to use discussions more and try to keep the issues to things that we acknowledge as actual issues.

The user probably could have directed the issue or question into the discussion's forum?

CleanShot 2022-02-28 at 02 25 43@2x



> The alternative would be to add some more templates - @edemaine 

You could also create other types of forms that will appear when the user clicks "new issue"
For example a `feature_request.yml`  file
   - https://github.com/cliffordfajardo/foam/blob/master/.github/ISSUE_TEMPLATE/feature_request.yml

The options would then be displayed as an option (not shown in image below, simply illustrating the idea) 

CleanShot 2022-02-28 at 02 23 07@2x

ryansolid:
Yes I like this even better.

PR #877, opened 2022-03-07
Revised Component types (children, ref)

### LINKED ISSUE #778: Add type for PropsWithRef to define components that accept refs
This type will be helpful and will the serve the purpose that forwardRef in react did. It allows us to easily define the ref type. The ref prop should be a function should that people don't use it by mistake and know that they either have to pass it or call it with the reference.

eg.
```typescript
type PropsWithRef<T, Props> = Props & { ref: (val: T) => void 

function Box(props: PropsWithRef<HTML.DivElement, { color: string }) {
	return <div ref={props.ref} />
}
```

edemaine:
Based on recent discussion on #typescript. This change summary is longer than the diff. 🙂 

* Rename past `Component` type to ~~`ComponentWithChildren`~~ `ParentComponent`.
  * Motivation: Most components don't actually support children.  This causes a type error upon accidental passing of children / forces you to think about which components support children.
  * Also, many users on #help/#typescript have asked for how to type such a component.
  * This is a slight breaking change (users may need to change some instances of `Component` type to `ParentComponent` to avoid type errors) so should probably wait for Solid 1.4.
* Two-parameter `FlowComponent` requires `children` and makes it easy to give a custom type for `children`, as in many flow control components.
* New `Component` type does not have automatic `children` property (like React's preferred `VoidFunctionalComponent`),  offering another natural way to type `props.children`:

  ```ts
  Component<{children: JSX.Element}>
  ```
* New `VoidComponent` forces not having `children` prop, to make it extra clear how to type this.
* `ParentProps`, `FlowProps`, and `VoidProps` type helpers for the `props` argument to `ParentComponent`, `FlowComponent`, and `VoidComponent`.
* Sadly none of these are great when you need to parameterize by a generic type, but still good starting points for people typing simple components.
* ~~`props` argument in both `Component` and `ComponentWithChildren` automatically cast to `readonly` (shallow one level only), to avoid accidental assignment to `props.foo` (usually a getter) while still allowing passing mutables in props.~~ Removed; [see below](https://github.com/solidjs/solid/pull/877#issuecomment-1060928619).
  * ~~Add `Props<T>` helper for this transformation.~~
  * Inspired by [this React plan](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/).
* Add @LXSMNSYC's [`Ref<T>` type](https://github.com/solidjs/solid/issues/778#issuecomment-1003236081) so it's easy to type `props.ref`:

  ```ts
  Component<{ref: Ref<Element>}>
  ```

  Fixes #778.

[An alternative](https://discord.com/channels/722131463138705510/783844647528955931/950117186456920074) would be to leave `Component` as is, and rename the new `Component` type to `VoidComponent`.  This would preserve backward compatibility and [match React](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/). But see [this discussion](https://github.com/facebook/create-react-app/pull/8177) for good arguments why (in React) `VoidComponent` is better than straight `Component`. And [React 18 is changing the type of `FunctionalComponent` to what was `VoidFunctionalComponent`](https://twitter.com/reactjs/status/1512453969490108418?t=-y2TJVyb_AmlYgA6ts3j1Q&s=19).


JoshuaHall:
This Playground link shows why this is an issue well: https://playground.solidjs.com/?hash=865026612&version=1.3.9

There are two main problems:
1. The children are then compiled but never used, remaining in the build output, even after optimization/minification.
2. It's easy to not realize that the misplaced markup you're writing won't be rendered, the TypeScript compiler doesn't make any mention of it. This is in general against the spirit of TypeScript, since TypeScript should bubble up the underlying outwards facing API details as accurately as possible, in order to increase developer productivity and deliver a smooth user experience.

edemaine:
Based on further discussion on #typescript, I've changed this in two ways:
* Removed the shallow `Readonly` constraint on props
* Removed `Props` helper as it wasn't doing anything anymore.

One motivation for this change is that it would not play well with existing explicitly typed `props`, for example:
```ts
function MyComponent(props: {name: string}): JSX.Element { ... }
```
This type is not consistent with `Component<any>`, so in particular would not be a valid argument to `lazy`, for example, as illustrated in [the previous need for this commit](https://github.com/solidjs/solid/pull/877/commits/64d8af13d32c9f88a950020983bfd6ae39fac0e3).

Another motivation for this change is that `props` might not be shallow `Readonly`: via spread props, it's possible to add a setter:
```ts
<A {...{ set a(){} }} />
```

ryansolid:
> Another motivation for this change is that props might not be shallow Readonly: via spread props, it's possible to add a setter

In my opinion this is a stretch. And while obviously it isn't stoppable I wouldn't give it any consideration or motivation here. 99.999% case shallow readonly is what you want. I can't speak to other implications so if it makes sense not to type that way I'm all for it. But the setter shouldn't be a consideration.

edemaine:
@high1 has proposed removing `ComponentWithChildren` (and `PropsWithChildren`) types, instead encouraging people to just list `children?: JSX.Element` in the argument to `Component`. The main motivation is that this is simpler, helping learn. Does anyone else have preferences either way?

JoshuaHall:
In my opinion, since children are something that not all components will need (really, the majority of them, mainly wrapper components that apply some styling or have their own local state), it makes more sense to write the names in an additive way, which would mirror the natural usage, such as `ComponentWithChildren`. The `VoidComponent` name doesn't seem as immediately obvious, it's not clear from the name what you're taking away from the regular `Component` type

otonashixav:
> I still think that component should be the default use case, which includes the composability of having children; like with HTML you get void elements which have no children. So I would prefer to have VoidComponents and VoidProps for this special case.


If `Component` implicitly adds `children?: JSX.Element`, then `VoidComponent` must be allowed to have children (similar to `React.VFC`), otherwise it is impossible to type components accepting non-`JSX.Element` children without a third `Component` type. So they wouldn't be comparable to void elements which have no children (i.e. `children?: never`). Then, it may be confusing to have `Component` for components that accept `JSX.Element` as children, and `VoidComponent` for components that accept nothing, or non-`JSX.Element` children if not already familiar with the react types.

atk:
> ... if not already familiar with the react types.

If you know JSX, chances are you are already familiar with the react types. And if not, we're not exactly clashing with preconceived expectations, are we?

edemaine:
FWIW, I know React well, but never used React with TypeScript. I started using TypeScript at roughly the same time I started using Solid. So it's all new to me, and I find React's types pretty confusing (and React seems to regret them). So I'm not convinced "same as React" is the right goal... By contrast, there's a natural notion of `Component` type that this PR adheres to.

But I also originally proposed adding `VoidComponent` (or a better name) &mdash; see "An alternative" in the PR description &mdash; before being convinced that this PR is the right way to go. So I'm not totally opposed to the idea, but I think we'd need a better argument than "same as React". (Perhaps "backwards compatibility is too important"? But it's just potential type errors, not functionality.)

LXSMNSYC:
> @high1 has proposed removing `ComponentWithChildren` (and `PropsWithChildren`) types, instead encouraging people to just list `children?: JSX.Element` in the argument to `Component`. The main motivation is that this is simpler, helping learn. Does anyone else have preferences either way?

tbf I've never used the built-in types since I can just define my own props interface. I'm down with the `Ref` type though

edemaine:
@LXSMNSYC I suppose another option is to just deprecate `Component` (e.g. discourage it in the documentation). But it's unfortunate that there's no type then for "any component", which is what the existing Solid codebase means where it currently uses `Component<any>` (incorrectly without this PR).

high1:
@edemaine I don't think we should deprecate it. The main question is - should the Component type receive children by default, as a number of people expects, and I would say that's what the main purpose of a Component is, and if that's the case, VoidComponent type makes sense as a type that does not receive children. I don't think that there should be any assumptions which type is used more often. I also see the argument which comes from @otonashixav, that by unifying everything in one Component type which would require the user to be explicit with children declaration, and TBH, I'm starting to be more and more in favor of the latter, but I'm OK with any of these two. I don't have a very strong case against ComponentWithChildren, I just don't like the name and I feel that we would be more in line with what users are expecting by using Component and VoidComponent, favored by @atk, or be more on the edge, and simplify the typings by using one Component type which requires explicit children.

edemaine:
Based on more yet more discussion on #typescript with @atk and @otonashixav, I revised further to this plan:
* `Component<T>` type is just `(props: T) => JSX.Element` (no implicit `children`).
  * All component types below extend this
  * `Component<T>` extends `Component` which is equivalent to `Component<{}>` (no `Component<any>` needed).
* `VoidComponent<T>` further asserts that there's no `children` in the `props` type. (Not sure this is crucial but it's nice for symmetry.)
* `ParentComponent<T>` is like our old `Component`, and has an **optional** `children?: JSX.Element`
* `FlowComponent<T, C>` has **required** `children: C`. This can be useful for building components that require a single-argument callback, for example.
* `VoidProps`, `ParentProps`, and `FlowProps` are helpers for typing just the `props` of such components.

I also tried to clean up some use of types in the code, and add some type exports to server code (I assume this is necessary to repeat here?). I'm not sure I found all places where these types could be used in the code, but I think it's an improvement.

otonashixav:
> * `Component<T>` extends `Component` which is equivalent to `Component<{}>` (no `Component<any>` needed). [Right? Why is CI failing?]

My bad, when I said that I only quickly tested in solid's playground (not a good place to test typescript) and assumed it would work the same as `{}`, but that's clearly not the case with function parameters, so `Component<any>` is still needed if `<T>(props: Component<T>) =>` isn't viable. 

```ts
function lazy<T>(
  fn: () => Promise<{ default: Component<T> }>
): Component<T> & { preload: () => Promise<{ default: Component<T> }> 
```

should work.

timothyallan:
Something just to add that I've seen throughout my React/Typescript years is the need to specify the Type of children. Will something like the `FlowComponent` cater for a `<TabManager>` component which does accept optional children, as long as they are of type `<Tab>`?

edemaine:
> Will something like the `FlowComponent` cater for a `<TabManager>` component which does accept optional children, as long as they are of type `<Tab>`?

Yes, that's exactly the intent.

otonashixav:
> Something just to add that I've seen throughout my React/Typescript years is the need to specify the Type of children. Will something like the `FlowComponent` cater for a `<TabManager>` component which does accept optional children, as long as they are of type `<Tab>`?

When you invoke a component using JSX (`<Tab>`), the return type is always `JSX.Element`. See https://github.com/microsoft/TypeScript/issues/21699. As such, it isn't possible to narrow the components allowed as children. 

`FlowComponent` is used where the children can be something other than `JSX.Element`, for instance `<For>` and `<Show>` which accept functions which can benefit from the explicit typing.

high1:
@edemaine @otonashixav Is VoidComponent needed at all, then? All components would be void by default - and that seems to deprecate VoidComponent, doesn't it? I see no reason to have it as a separate type.

otonashixav:
> Is VoidComponent needed at all, then? All components would be void by default - and that seems to deprecate VoidComponent, doesn't it? I see no reason to have it as a separate type.

While you cannot call either directly with children, the former is still allowed to assigned to a type that may accept children, whereas the former cannot since it never accepts children. 

```ts
let MyComponent!: Component
let MyVoidComponent!: VoidComponent

let A: ParentComponent = MyComponent;
let B: ParentComponent = MyVoidComponent;
/*  ^ Type 'VoidComponent<{}>' is not assignable to type 'ParentComponent<{}>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type '{ children?: Element; }' is not assignable to type '{ children?: undefined; }'.
      Types of property 'children' are incompatible.
        Type 'Element' is not assignable to type 'undefined'.
          Type 'null' is not assignable to type 'undefined'.(2322) */
```

Practically, this affects functions which accept a component e.g.: https://playground.solidjs.com/?hash=-853165744&version=1.3.13

high1:
@otonashixav I've looked at the playground, and it seems weird that the Dynamic accepts children property, when a plain component is referenced. Isn't the problem maybe that Dynamic uses current definitions which have implicit children?

otonashixav:
That's probably why that specific example works in solid's playground. See https://tsplay.dev/wOPxpN instead.

high1:
The children will not be silently accepted on the Component - if not explicitly defined. I don't see a lot of value in explicitly forbidding them, when they first have to be opt-in. On the other hand, maybe having a convenience type is not bad...

atk:
It was meant as a convenience type for library developers to make sure users were not confused by components that don't render children. Since types usually don't influence the bundle size of resulting pages and apps, it shouldn't be an issue to provide them.

high1:
@atk @edemaine @otonashixav My main concern here is the user viewpoint - now we have Component  which doesn't have implicit children,  VoidComponent which explicitly forbids them, ParentComponent which brings back implicit children. How should a user type his component which doesn't accept children? Should he type it with Component or a VoidComponent - and is there a real difference in the most prevalent usage case?

atk:
We need to document these types (JSDoc; especially VoidComponent and ParentComponent need to be mentioned in Component's description). They could basically build VoidComponent themselves, but using a dedicated type which reuses the pattern of Void Elements in the HTML living standard is much more obvious and simple.

otonashixav:
Not explicitly typing children as never only means losing a type error in a single unlikely scenario. For the most common case there isn't a difference. 

Users are free to choose whatever suits them. They can define the types themselves without using any of the helpers, use only Component, or use all of the helpers. The extra types are named such that it is clear they are a special subset of components with extra semantic value, so I don't see it being difficult to decide which to choose. That said, if I had to argue for their removal:
- Libraries using the types force users to learn what the types do, if the names aren't self-descriptive to the person reading them
- VoidComponent may get misinterpreted as a component with a void return if you are not familiar with the definition, since that's what a void function is

high1:
I see that React has deprecated VoidFunctionComponent - and while I'm OK with keeping it here, adding these additional types only to the documentation would also serve the purpose, wouldn't it? The user could just copy the definitions for these if he feels he needs them, and we would avoid adding utility types to the core - types which are not used anywhere inside the codebase itself.

atk:
As I already remarked, it's a question of DX: if we already provide the type and the documentation, we save the developers some work while at the same time helping to make their types be as sound and useful as possible.

These types don't increase our bundle size noticably and will not even appear in the resulting app bundles, so what's not to like?

I also believe the react type was deprecated, because it lacked sufficient documentation, which led to confusion they sought to clean up by doing so. We can avoid the confusion to begin with, so that shouldn't be an issue.

high1:
image
The reason for deprecation is because it's equivalent with the FunctionComponent. As I said, I don't mind either solution, it's just that I don't see any notable benefit from adding Types which can be easily derived from Component.

packages/solid/src/reactive/signal.ts
```diff
@@ -1063,7 +1064,7 @@ export interface Context<T> {
  * ```typescript
  * interface Context<T> {
  *   id: symbol;
- *   Provider: (props: { value: T; children: any }) => any;
+ *   Provider: ParentComponent<{ value: T }>;
```
otonashixav, line 1067: You probably forgot to change this to `FlowComponent`

edemaine:
From a "compatible naming" perspective, it does bother me a bit that `VoidComponent` in this PR is different from React's `VoidComponent` (which should really have been called `Component` -- and soon it will be).  Given the disagreement, perhaps it's better to remove `VoidComponent` (easier to add than remove)?  I'm happy to go either way; I'd probably never need to use `VoidComponent` myself because `Component` (without a `children` prop) does the job when defining actual components.

high1:
I'm all on the side of adding only what's needed, maybe removing ParentComponent also, and having children optional in FlowComponent would be the best course?

edemaine:
Oops, didn't mean to open the can of worms that far. 😄 

I feel like `ParentComponent` (which is the current `Component`) is the most used in practice today, so we should at least keep that one. I guess `FlowComponent` could be merged into it, adding a second parameter for the `children` type like the current `FlowComponent`.

On the other hand, I think `ParentComponent` and `FlowComponent` are each useful in different ways/contexts, and have very different semantic meaning too. (Whereas `Component` vs. `VoidComponent` seems like a less useful distinction.) In particular, forcing the existence of (nonoptional) children for `FlowComponent` seems useful for avoiding bugs.

atk:
> I'm all on the side of adding only what's needed, ...

I would agree to this sentiment if we were talking about API surface – however this is about types, which are more documentation than code and in documentation, you should cover everything that is helpful to most people.

ryansolid:
Thank you all for all your effort to get this one across the finish line.

novacrazy:
I'm quite a bit late on this, but since SolidJS doesn't seem to follow semantic versioning, it's worth noting the change from `Component` to `ParentComponent` has broken some of my code and some from libraries I use, as Typescript no longer thinks they accept children.

ryansolid:
Yeah TypeScript has been very difficult to keep with semantic versioning. This is partly because I don't think in TS and it takes some time for understanding the use cases in practice. Also TypeScript has been making updates that significantly improve typing story. In general these have been deemed worthwhile because of their impact and we've tried to keep them to minor versions.

PR #881, opened 2022-03-09
feat: throw error when using ssr render functions in browser

### LINKED ISSUE #878: Throw error when SSR rendering the app in the browser
### Describe the bug

If possible, when using the `render*` functions that are aimed for SSR / SSG in the browser context, we should thow an error to indicate they should not be used on it. Instead, the current behaviour is to return undefined. This is misleading because typescript expects a string as well (and it should always be a string on the server anyway).

No need

```tsx
import { renderToString } from "solid-js/web";

import App from "./App";

const result = renderToString(() => <App />);
console.log(result);
```

### Expected behavior

To throw an error when used in a browser context.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 99.0.4844.51 (Official Build) (x86_64)

### Additional context

Screenshot 2022-03-07 at 19 29 24

Aslemammad:
Resolves #878 

This PR tries using throwInBrowser, so we can log an error that tells the user that using such functions is not possible in the browser, though it does not break the user's code (logging instead of throwing)!

### DIFF
--- a/packages/solid/web/src/server-mock.ts
+++ b/packages/solid/web/src/server-mock.ts
@@ -1,11 +1,19 @@
 //@ts-nocheck
+function throwInBrowser(func: Function) {
+  const err = new Error(`${func.name} is not supported in the browser, returning undefined`);
+
+  console.error(err);
+}
+
 export function renderToString<T>(
   fn: () => T,
   options?: {
     nonce?: string;
     renderId?: string;
   }
-): string {}
+): string {
+  throwInBrowser(renderToString);
+}
 export function renderToStringAsync<T>(
   fn: () => T,
   options?: {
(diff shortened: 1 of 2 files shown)


packages/solid/web/src/server-mock.ts
```diff
@@ -1,19 +1,29 @@
 //@ts-nocheck
+function throwInBrowser(func: Function) {
+  const err = new Error(`${func.name} is not supported in the browser, returning undefined`);
+
+  console.error(err);
```
Aslemammad, line 5: This is debatable, we can try throwing, but sometimes may the user's making something env agnostic, so they don't want the app breaks while importing/using such functions!

ryansolid: I'm ok with the warning. Knowing what is wrong is the most important.

ryansolid:
Thank you.

PR #888, opened 2022-03-16
feat: allow adding types to stores' `NotWrappable`

otonashixav:
Classes are not distinguishable from types in typescript which can be troublesome for stores, which don't wrap classes.

This adds an interface `Unwrappable` which can be used to add unwrappable types via declaration merging, like how custom directives are currently handled. 

For example, to exclude `Element`:
```ts
declare module "solid-js/store" {
  namespace SolidStore {
    interface Unwrappable {
      anyUniqueKey: Element
    }
  }
}
```

### DIFF
--- a/packages/solid/store/src/index.ts
+++ b/packages/solid/store/src/index.ts
@@ -3,6 +3,7 @@ export type {
   Store,
   SetStoreFunction,
   NotWrappable,
+  SolidStore,
   StoreNode,
   StoreSetter,
   StorePathRange,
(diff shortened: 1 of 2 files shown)


otonashixav:
I'm not sure if this is the best way to approach this problem, but seeing as it is the solution for directives, there should be some level of familiarity with this method at least. However, this approach is very powerful, allowing you to exclude just about whatever type you want; it's also possible to split this into more types: "DeepReadonly-able", "DeepMutable-able", and "Unwrappable".

PR #892, opened 2022-03-21
Improve the typing of parameters of unwrap

TiagoCavalcante:
The type of the 1st parameter was any, but we can assume that the
user won't pass as the template type a type that is "wrapped"
(contains properties that unwrapping will remove), this way the
user doesn't needs to pass a type to unwrap to it works well.

### DIFF
--- a/packages/solid/store/src/modifiers.ts
+++ b/packages/solid/store/src/modifiers.ts
@@ -110,7 +110,7 @@ export function reconcile<T extends U, U>(
   const { merge, key = "id" } = options,
     v = unwrap(value);
   return state => {
-    if (!isWrappable(state) || !isWrappable(v)) return v;
+    if (!isWrappable(state) || !isWrappable(v)) return v as T;
     applyState(v, { state }, "state", merge, key);
     return state as T;
   };
--- a/packages/solid/store/src/mutable.ts
+++ b/packages/solid/store/src/mutable.ts
@@ -79,7 +79,7 @@ function wrap<T extends StoreNode>(value: T, name?: string): T {
 }
 
 export function createMutable<T extends StoreNode>(state: T, options?: { name?: string }): T {
-  const unwrappedStore = unwrap<T>(state || {});
+  const unwrappedStore = unwrap(state || {});
   if ("_SOLID_DEV_" && typeof unwrappedStore !== "object" && typeof unwrappedStore !== "function")
     throw new Error(
       `Unexpected type ${typeof unwrappedStore} received when initializing 'createMutable'. Expected an object.`
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -47,13 +47,13 @@ export function isWrappable(obj: any) {
   );
 }
 
-export function unwrap<T extends StoreNode>(item: any, set = new Set()): T {
+export function unwrap<T extends StoreNode>(item: T, set = new Set()): T {
   let result, unwrapped, v, prop;
   if ((result = item != null && item[$RAW])) return result;
   if (!isWrappable(item) || set.has(item)) return item;
 
   if (Array.isArray(item)) {
-    if (Object.isFrozen(item)) item = item.slice(0);
+    if (Object.isFrozen(item)) item = item.slice(0) as unknown as T;
     else set.add(item);
     for (let i = 0, l = item.length; i < l; i++) {
       v = item[i];
@@ -68,7 +68,7 @@ export function unwrap<T extends StoreNode>(item: any, set = new Set()): T {
       prop = keys[i];
       if ((desc as any)[prop].get) continue;
       v = item[prop];
-      if ((unwrapped = unwrap(v, set)) !== v) item[prop] = unwrapped;
+      if ((unwrapped = unwrap(v, set)) !== v) item[prop as keyof T] = unwrapped;
     }
   }
   return item;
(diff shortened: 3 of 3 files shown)


packages/solid/store/src/store.ts
```diff
@@ -47,13 +47,13 @@ export function isWrappable(obj: any) {
   );
 }
 
-export function unwrap<T extends StoreNode>(item: any, set = new Set()): T {
+export function unwrap<T extends StoreNode>(item: T, set = new Set()): T {
```
otonashixav, line 50: The return type should be `DeepMutable<T>`, otherwise it will remain readonly despite being unwrapped.

packages/solid/store/src/modifiers.ts
```diff
@@ -110,7 +110,7 @@ export function reconcile<T extends U, U>(
   const { merge, key = "id" } = options,
     v = unwrap(value);
   return state => {
-    if (!isWrappable(state) || !isWrappable(v)) return v;
+    if (!isWrappable(state) || !isWrappable(v)) return v as T;
```
TiagoCavalcante, line 113: TypeScript type guards fail here, probably in the next 10 years they'll fix this.

packages/solid/store/src/mutable.ts
```diff
@@ -79,7 +79,7 @@ function wrap<T extends StoreNode>(value: T, name?: string): T {
 }
 
 export function createMutable<T extends StoreNode>(state: T, options?: { name?: string }): T {
-  const unwrappedStore = unwrap<T>(state || {});
+  const unwrappedStore = unwrap(state || {});
```
TiagoCavalcante, line 82: And this is the objective: don't specify the type of unwrap.

packages/solid/store/src/store.ts
```diff
@@ -39,21 +39,33 @@ function wrap<T extends StoreNode>(value: T, name?: string): DeepReadonly<T> {
   return p;
 }
 
-export function isWrappable(obj: any) {
-  return (
+function isUnwrappable(obj: {} | undefined | null): obj is NotWrappable {
```
TiagoCavalcante, line 42: This is a type guard, this will work when (if) the TS team decide to improve type guards (these days I also discovered that they don't work in arrays with non-const index).

packages/solid/store/src/store.ts
```diff
@@ -39,21 +39,33 @@ function wrap<T extends StoreNode>(value: T, name?: string): DeepReadonly<T> {
...
+      !(obj as any).__proto__ ||
+      (obj as any).__proto__ === Object.prototype ||
+      Array.isArray(obj)
+    )
   );
 }
 
-export function unwrap<T extends StoreNode>(item: any, set = new Set()): T {
+export function isWrappable<T>(obj: T) {
+  return !isUnwrappable(obj);
```
TiagoCavalcante, line 56: Also, there isn't a `is not Type` so we use this (NonNullable introduces the `never` and we don't want that).

packages/solid/store/src/store.ts
```diff
@@ -219,17 +231,14 @@ export function updatePath(current: StoreNode, path: any[], traversed: PropertyK
   } else setProperty(current, part, value);
 }
 
-export type DeepReadonly<T> = 0 extends 1 & T
```
TiagoCavalcante, line 222: Probably this was necessary in an older version, but isn't anymore (at least not in my tests).

packages/solid/store/src/store.ts
otonashixav: You can do `export function isWrappable<T>(obj: T | NotWrappable): obj is T` for `is not T`.

packages/solid/store/src/store.ts
otonashixav: Thanks

packages/solid/store/src/store.ts
```diff
@@ -219,17 +231,14 @@ export function updatePath(current: StoreNode, path: any[], traversed: PropertyK
...
-  ? T
-  : T extends NotWrappable
-  ? T
+export type DeepReadonly<T, W = any> =
+  T extends DeepMutable<W> ? W
+  : T extends NotWrappable ? T
```
otonashixav, line 236: I am quite certain that this causes `T` to be `any` in the default case. There are type tests in `store/test/*.spec.ts`, check that those are fine. I'm not sure if this needed to be changed bar removing the extra `any` check.

packages/solid/store/src/store.ts
```diff
@@ -39,21 +39,33 @@ function wrap<T extends StoreNode>(value: T, name?: string): DeepReadonly<T> {
...
 
-export function unwrap<T extends StoreNode>(item: any, set = new Set()): T {
+export function isWrappable<T>(obj: T) {
+  return !isUnwrappable(obj);
+}
+
+export function unwrap<T extends StoreNode>(
+  item: T,
+  set = new Set()
+): T extends NotWrappable ? T : DeepMutable<T> {
```
otonashixav, line 62: `DeepMutable<T>` should be fine.

otonashixav (changes requested):
Here's how I would approach the same changes: https://github.com/otonashixav/solid/tree/improve-store-helper-types; note the test changes as well, since returning a specific type for `unwrap` means that accessing `$RAW` will be a type error.

ryansolid:
Closing as this was addressed in #994. Thank you for the PR.

PR #906, opened 2022-03-31
add type test cases for createSignal/Setter

trusktr:
~~I think this makes the `Setter` type easier to read and understand. I am thinking that in this form, the cases are explicitly laid out, easier to read.~~

~~What are your opinions?~~

~~If we end up preferring the previous form, we can merge just the added test cases.~~

I removed the type changes, so this now only adds new type test cases to cover signal function values well.

### DIFF
--- a/packages/solid/test/signals.spec.ts
+++ b/packages/solid/test/signals.spec.ts
@@ -626,8 +626,8 @@ describe("runWithOwner", () => {
     });
 
     runWithOwner(owner, () => {
-      createEffect(() => effectRun = true);
-      onCleanup(() => cleanupRun = true);
+      createEffect(() => (effectRun = true));
+      onCleanup(() => (cleanupRun = true));
       expect(effectRun).toBe(false);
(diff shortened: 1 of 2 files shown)


packages/solid/test/signals.spec.ts
```diff
@@ -160,9 +160,9 @@ describe("Update signals", () => {
   });
   test("Set signal returns argument", () => {
     const [_, setValue] = createSignal<number>();
-    const res1: undefined = setValue(undefined);
+    const res1: number | undefined = setValue(undefined);
     expect(res1).toBe(undefined);
-    const res2: number = setValue(12);
+    const res2: number | undefined = setValue(12);
```
trusktr, line 165: Oh, um, this change. Hmmmmm. Maybe it wasn't a bug fix like my commit message thought.

trusktr: converting back to draft

trusktr: fixed

PR #908, opened 2022-04-01
re-export all things from signal.ts?

trusktr:
they are available indirectly via the path to the file anyway

### Use case:

I am extending types from signal.ts in new primitives outside of Solid, and I have to import types directly, f.e.:

```js
import type {SignalOptions, etc} from 'solid-js/types/reactive/signal'
```

So I figured, why not just re-export to simplify the code. Then end user can more easily do the following, although it's not critical:

```js
import type {SignalOptions, etc} from 'solid-js'
```

Does this have any significant cons?

### DIFF
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -1,47 +1,4 @@
-export {
-  createRoot,
-  createSignal,
-  createEffect,
-  createRenderEffect,
-  createComputed,
-  createReaction,
-  createDeferred,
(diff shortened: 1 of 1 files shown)


ryansolid:
There are a number of APIs that are not external APIs if we are missing exporting any types we should just do that.

ryansolid:
As mentioned not doing this I think. We should probably look at better breaking things apart.

PR #910, opened 2022-04-01
fix: fix the signal setter type when used with generics

otonashixav:
This makes it easier to create generic wrappers around `createSignal` e.g.
```ts
function createGenericSignal<T>(): Signal<T | undefined> {
  const [generic, setGeneric] = createSignal<T>();
  const customSet: Setter<T | undefined> = (v?) => setGeneric(v);
  return [generic, (v?) => setGeneric(v)];
}

function createInitializedSignal<T>(init: T): Signal<T> {
  const [generic, setGeneric] = createSignal<T>(init);
  const customSet: Setter<T> = (v?) => setGeneric(v);
  return [generic, (v?) => setGeneric(v)];
}
```
where previously such an implementation would not be assignable to `Setter`, and would need to be cast.

In the `createSignal` implmentation, this still doesn't allow `value` to be callable after the `typeof value` type guard, as `Exclude<U, Function> & Function` doesn't get reduced to `never` due to how generics work. As such (and also because the result of the call is assigned back to `value`, which might not be able to hold it), I've left it as is instead of using inference.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -124,7 +124,7 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: Owner): T {
 
 export type Accessor<T> = () => T;
 
-export type Setter<T> = (undefined extends T ? (value?: undefined) => undefined : {}) &
+export type Setter<T> = (undefined extends T ? () => undefined : {}) &
   (<U extends T>(value: Exclude<U, Function> | ((prev: T) => U)) => U);
 
 export type Signal<T> = [get: Accessor<T>, set: Setter<T>];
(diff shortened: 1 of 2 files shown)


otonashixav:
Interestingly this also makes the return type of function setters more specific if they can contain `undefined` e.g. before
```ts
const [s, set] = createSignal<string>();
// before:
const v = set(() => "literal");
//    ^? - string
// after:
const v = set(() => "literal");
//    ^? - "literal"
```

Going to modify test cases to take this into account.
This is strange since I'd expect the first overload to have no effect as its signature is `(v?: undefined) => undefined` and clearly a function is not undefined, yet somehow it affects something. I love typescript.

PR #913, opened 2022-04-01
refactor: simplify/improve effect types

otonashixav:
- Simplified `createEffect` etc. such that they no longer use rest params.
- Added test cases for effect functions failing partial generic inference (not supported by typescript)
- Removed extra params from effects' first overload

Edit: removed the changes not related to effect types, it would be better to clean them up all at once in another PR.


otonashixav:
@trusktr Just to let you know, I made another change to remove `value` and `options` from the first overloads of these functions, to bring them closer to how they were originally, and also make them consistent with the other instance of overloading, `createSignal`. AFAICT there's no issue with the existing tests, this just makes cases where `undefined` is explicitly passed use the second overload instead of the first.

PR #927, opened 2022-04-08
Warn about unhandled errors (fix #919)

### LINKED ISSUE #919: Default error handling behavior
Consider [this buggy code](https://playground.solidjs.com/?hash=-286161745&version=1.3.13) (based on some actual code I wrote and had a hard time debugging). The actual bug is that `createEffect` isn't `import`ed, so attempting to call it causes an error. But the only error that gets output to the console is `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'style')` which is caused by the `onMount` code, which assumed the `ref` would be set by this point, but it wasn't because the component didn't get to running the JSX. This makes debugging hard.

@otonashixav tracked this down to [runUpdates's `try` block](https://github.com/solidjs/solid/blob/9cd3f260fc6c7f39d69c752f3d6b3ffb7cd35891/packages/solid/src/reactive/signal.ts#L1392-L1398). When there are no error handlers installed, `handleError` just rethrows the exception, and then the `finally` block runs, which includes the `onMount`, which throws its own error, which takes precedent.

**Question:** Could the `completeUpdates` be moved outside the `finally` block, so that it only runs if there are no uncaught errors? Or is it intentional that scheduled effects run even when there's an uncaught error?

Assuming the behavior is intended, it would be really nice to have a default error handler, perhaps at least in dev mode, that outputs all errors that occur, something like ``onError((e) => console.error(`Uncaught error in Solid component: ${e}`))``.  When there are error handlers, we [actually get the correct error message](https://discord.com/channels/722131463138705510/722131463889223772/960268191815307344). @mdynnl suggested maybe wrapping all `render` calls in an `ErrorBoundary` but I think `onError` is more appropriate if the goal is to write to `console`. Actually, to make this error handler disappear when the user provides error handlers, maybe it would be better to add it directly to [handleError](https://github.com/solidjs/solid/blob/9cd3f260fc6c7f39d69c752f3d6b3ffb7cd35891/packages/solid/src/reactive/signal.ts#L1578) just before re-throwing the error.

Thanks also to @orenelbaum for helping track this down.

edemaine:
Here's one approach to fixing #919: in dev mode, warn about unhandled errors. The last exception in `try...finally` chain will end up getting thrown as an actual error, and thus duplicated as a warning, but it means that no errors will be unreported.

(Sadly this wouldn't have helped me because I still can't get Solid dev mode working in Meteor, but it would help most people...)

Feel free to edit the message or reject the approach altogether.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1575,7 +1575,10 @@ function reset(node: Computation<any>, top?: boolean) {
 
 function handleError(err: any) {
   const fns = ERROR && lookup(Owner, ERROR);
-  if (!fns) throw err;
+  if (!fns) {
+    "_SOLID_DEV_" && console.warn(`Error during Solid code without onError or ErrorBoundary: ${err}`);
+    throw err;
+  }
   fns.forEach((f: (err: any) => void) => f(err));
 }
 


edemaine:
I tried this a bit in practice. `console.warn(err)` or `console.error(err)` is better in some ways because it seems to print the whole stack track, instead of just the error message. Maybe there's a way to do that and still prepend a message about where the message came from?

ryansolid:
I managed to fix this issue without doing the extra logging. I pulled the finalize code into the try, and added a bit more cleanup in the finally. Seems to look good in the original example. There is something about warning people if there is no Error Boundary in dev, but then again they will get the error. So yeah I think I won't go this way. But thank you for the PR.

edemaine:
That seems cleaner indeed! Now Solid avoids triggering the second error.

https://github.com/solidjs/solid/blob/34628200f7c8e0a329905688a455848096a3e429/packages/solid/src/reactive/signal.ts#L1381-L1391

PR #931, opened 2022-04-10
refactor: improve mergeProps types

otonashixav:
- Make `mergeProps` require at least 1 source
- Make the types of properties correct when merging types that would otherwise return `never` when intersected e.g. merging `{ a: number }` and `{ a: string }` is now `{ a: string }` instead of `{ a: never }`
- Add tests for the above

Edit: stopped trying to preserve the original names of types, so now:
```ts
type MyType = { a: number };
mergeProps({} as MyType); // { a: number } instead of MyType as this PR initially tried to do 
// to match the existing MergeProps type. 
```


thetarnav:
Could the `MergeProps` type be exported as well?

PR #934, opened 2022-04-13
Replace for/push loop with Array of initial size

edemaine:
I noticed this bit of code which builds an array via `push` incrementally instead of just calling `map`. I believe `map` is faster because it can pre-allocate an array of the right size. [This jsbench](https://jsbench.me/qwl1y4xyud/1) confirms my hypothesis. The code is also shorter and cleaner.

Or is `map` being avoided on purpose for compatibility? It's ECMAScript 5. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#browser_compatibility); [Can I use](https://caniuse.com/?search=array.map).

[[discussion on Discord](https://discord.com/channels/722131463138705510/963927238405935174)]

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -841,9 +841,12 @@ export function on<S extends Accessor<unknown> | Accessor<unknown>[] | [], Next,
   let prevInput: ReturnTypes<S>;
   let defer = options && options.defer;
   return (prevValue: Init | Next) => {
-    const input: ReturnTypes<S> = isArray
-      ? deps.map((d) => d()) as ReturnTypes<S>
-      : deps() as ReturnTypes<S>;
+    let input: ReturnTypes<S>;
+    if (isArray) {
+      input = Array(deps.length) as ReturnTypes<S>;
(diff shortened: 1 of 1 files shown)


ryansolid:
Hmm.. I tested this a lot before. For wasn't only faster it was like 30% faster in my benchmarks. When I was struggling to get the fastest SSR this one of the blockers that led me to creating a separate runtime for server. Then I looked at what Marko was doing and using for + push was the key. I'm sort of surprised `map` is faster. It never has been historically.

edemaine:
liquid [on Discord](https://discord.com/channels/722131463138705510/963927238405935174) pointed out that
* The map version is slower for large arrays (for me, the crossover point is around 20 items), and
* There's a for-loop version that is the best of both worlds: preallocate an Array of the correct size. [Here's the new benchmark](https://jsbench.me/qwl1y4xyud/2).  It's faster for arrays of size 1, 4, 20, etc.

I pushed code for the new version.

ryansolid:
Thank you.

Issue #932, opened 2022-04-10
Improve code accessibility


orenelbaum:
I think that we can do a better job with making the code more accessible to developers who read the code for the first time as well as developers with disabilities and beginners.
We probably need to do some more research unless someone with more experience in this field wants to weigh in but three things I have in mind right now:
- Add comments.
- Reformat the whole codebase and limit line length to 78 characters. This should help people programming on braille displays and will probably also be more accessible to people coding on mobile. Some people can't afford a computer and I think that it's good to keep them in mind as well.
- Make the code more self documenting. Currently IMO the code is not very self documenting, I personally struggled a lot with trying to read the code the first time, I felt like I was some archeologist trying to decipher an ancient language. It might be partly on me, but at least I feel like we shouldn't have one letter variables.

I'll try doing more research about this topic but we can start implementing those things for now. If anyone else wants to add ideas or comment on the ideas I presented here that would be very helpful!

fabiospampinato:
Some random ideas:

- Have mostly one file per exported function: having many many functions in a single file makes it harder to find the one you are looking for and to see all the exported functions at a glance.
- Fight the type system less: there are too many type assertions IMO, that means you kind of have to rely on hidden assumptions to understand what's going on, and those assumptions may or may not be obvious, rather than relying on what the type system tells you.
- Improve discoverability and code navigability of `solid/solid` & `ryansolid/dom-expressions`: currently most of the important code is split between two completely separate repositories under different users/organizations, and code in those repositories often imports things that don't exist because some imports are rewritten at build time, that makes it hard to discover where things are defined and to jump from thing to thing.
- Some code is not particularly idiomatic, like assignments inside `if` should never happen IMO, those kind of things should probably get rewritten: e.g. `if ((result = item != null && item[$RAW]))`

fabiospampinato:
Some more:

- Have all the `export`s at the end of files, rather than at random rows in files, otherwise one needs to scan the entire file to find all the exports. Plus some exported functions are found between non-exported functions, that also makes it harder to find what functions are named because their names won't be found at a single predictable column index.
- Define all types and interfaces in files containing only types and interfaces, that makes for cleaner separation of concerns in my opinion.
- Use `type` instead of `interface` for things that won't need to be extended through interface merging, currently the two seem to be used somewhat interchangeably.

orenelbaum:
Thanks for adding more suggestions:
[...]

PR #938, opened 2022-04-17
more accessible formatting

orenelbaum:
- Change line width from 100 to 78
- Use tabs instead of spaces
- Add prettier to devDependencies

See #932


devinrhode2:
There's a few setup things I'd recommend
.gitblameignore setup via prepare npm script
Ci checks
Pre-commit checks

We could even recommended gitlens in .vscode/settings.json but I guess that's tangential

devinrhode2:
We might also want a script to help people rebase across the mass formatting change.

I care about git blame a lot :)

PR #953, opened 2022-04-23
Respect Promises in resource first.

MrFoxPro:
Before:
image

After:
image

Btw, I'm not sure `createResource` handles `() => {}` and `() => Promise` in same manner.
upd: Seems it could harm someone by passing hidden ```lookup, {
          value: s(),
          refetching
        }``` as params to function with optional args.


ryansolid:
I don't believe this is the direction we will be taking with the resource types. Thank you for the PR.

Issue #972, opened 2022-05-10
`null` and `undefined` are missing in `createSignal` return type


julvo:
### Describe the bug

When creating a signal with a union type that includes `null` or `undefined`, the resulting signal doesn't include `null` or `undefined`. 

E.g. if we create a signal that can be `null` or a `number` using `createSignal<null | number>(null)`, the return type is `Signal<number>` instead of `Signal<number | null>`

The issue is that the value of the signal is actually `null`, but the type signature doesn't reflect that. That means we loose type-safety.

### Your Example Website or App

https://playground.solidjs.com/?version=1.3.13#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr2ZCouOAGU0Ac2hqps+YpU6DW09CysrGXoIZlw0WgheAGEGCBdGAAoASn4rXgd4sj5gZhTcLF4yOFxkqNwAXV4AXgcnF3cvKDUAHgh6GBNxXgAfXl61NQA+dLG1TMsE-IhC3gxHOHhUpt4spomKqprU9JLanYBqXgBGOdzlBf19XgADKCeVsl5aGV4CYmfe-riN4meh8MgACwYahEvAG-z6AwkIxmT1uJSWfCgWxOR0y4QWQlwTAS6VueS6INwuHiPyIcEaKkp1IgKk+EESGmYAGtGiBVkINrhJBMyXl+DjcFlJKK8iAoNKFuT9Ez4iKFjcIAqrEJROJ0jtGnsuoc0rx9BNyiJaMw+sJcAA6DxVACianWdoAQvgAJIiUlgKCEQgqTIasCSOpAA

### Steps to Reproduce the Bug or Issue

1. Create a signal with a nullable type, e.g. `number | null`
2. Observe that resulting signal type doesn't include null, e.g. only `number` instead of `number | null`

### Expected behavior

When using `createSignal<T | null>(null)`, I'd expect the resulting signal to be of type `Signal<T | null>`

### Screenshots or Videos

Example from the Solid playground with `createSignal<number | null>()`
<img width="580" alt="image (18)" src="https://user-images.githubusercontent.com/18549792/167723920-a397e523-b192-4fcb-8db6-abd8ec7e4179.png">

[...]

PR #978, opened 2022-05-12
fix: setter type incorrectly erroring with unions

otonashixav:
- Added two overloads to preserve existing functionality while also fixing the issue
- Added a test for this case

Fixes the issue described in https://github.com/solidjs/solid/issues/972#issuecomment-1124337736

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -127,6 +127,8 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: Owner): T {
 export type Accessor<T> = () => T;
 
 export type Setter<T> = (undefined extends T ? () => undefined : {}) &
+  (<U extends T>(value: (prev: T) => U) => U) &
+  (<U extends T>(value: Exclude<U, Function>) => U) &
   (<U extends T>(value: Exclude<U, Function> | ((prev: T) => U)) => U);
 
 export type Signal<T> = [get: Accessor<T>, set: Setter<T>];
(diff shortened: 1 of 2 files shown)


otonashixav:
To be noted is that this isn't a perfect solution, it doesn't cover certain scenarios:

```ts
// for a signal with string | number
// with the third overload, function return type has priority
// this will error as number cannot be assigned to string
set(1 as string | number | () => string)
// can't handle unions of functions; this one is difficult due to how distributing functions works
// if the signal contains a union of 3 types, it is even harder, since you would need to support
// all subsets of 3 unions, where using only either distributing or not distributing is insufficient
set((() => 1) as (() => string) | (() => number))
set((() => 1) as (() => string) | (() => 1) | (() => 2))
```

But these are cases which you should never encounter in normal usage.

otonashixav:
Here is another solution which does solve the above cases, however it doesn't really "make sense", basically having to struggle against typescript:
```ts
export type Setter<T> = (undefined extends T
  ? <U extends undefined & Exclude<T, Function>>() => U extends (prev: T) => infer I ? I : U
  : {}) &
  (<U extends Exclude<T, Function> | ((prev: T) => T)>(
    value: U
  ) => U extends (prev: T) => infer I ? I : U);
```

It also 
- breaks backwards compatibility when specifying generics on setters (`setCount<number>(count() + 1)`), but basically no one does that (at least I'd hope so).
- makes certain inferred values less specific: (`const c = setCount(() => 1)` would have `c` inferred as `1`, but with the above solution it would be inferred as `number` unless the return type was made more specific e.g. with `1 as const`.

Issue #982, opened 2022-05-14
`mergeProps` return type is broken when using generics and intersections types


riccardoperra:
### Describe the bug

I'm trying to upgrade my application to Solid 1.4.0 but I am having troubles related to the new  `mergeProps` return type which broke some types inside my ui components library
https://github.com/riccardoperra/codeimage/runs/6428736073?check_suite_focus=true

The result of the `mergeProps` function is ok for the most cases, but when using types that depends on generics, some properties are missing or completely broken. To me this seems related to a typescript core issue, but I think some of solid types can be improved to try to handle these errors, even if not all

### Your Example Website or App

https://playground.solidjs.com/?hash=-238626028&version=1.4.1

### Steps to Reproduce the Bug or Issue

Here's a playground

https://playground.solidjs.com/?hash=-238626028&version=1.4.1

It consists of:
- main.tsx: small example copied from my repo
- text-solid-1.3.4-types.tsx: an example of mergeProps with old solid types which are not throwing error
- test-solid-1.4.0-fixed-types: an example with a possible fix for solid 1.4.0

### Expected behavior

Solid 1.3.4 was not throwing that error.

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/37072694/168440198-8976319d-62fc-47f1-bdef-dda9bb2b5030.png)

### Platform

- Solid 1.4.0

### Additional context

A pr with the possible fix #983

ryansolid:
Should be fixed in 1.4.2. Thanks for the PR.

PR #983, opened 2022-05-14
fix: MergeProps is not inferring the types correctly

riccardoperra:
This pr tries to fix partially the issue related of mergeProps return type issue with generics types. #982
There are some issues still not resolved and I would like your advice / opinion to continue

I think the Override type has something missing but I didn't manage to get a definitive fix😢, currently with the extra inference the returned object mostly return the correct type, but fails when using Omit

Playground
https://playground.solidjs.com/?hash=-238626028&version=1.4.1

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -136,30 +136,30 @@ const propTraps: ProxyHandler<{
   }
 };
 
-type Simplify<T> = T extends object ? { [K in keyof T]: T[K] } : T;
-type UnboxLazy<T> = T extends () => infer U ? U : T;
-type RequiredKeys<T> = keyof { [K in keyof T as T extends { [_ in K]: unknown } ? K : never]: 0 };
-type Override<T, U> = {
-  // all keys in T which are not overridden by U
-  [K in keyof Omit<T, RequiredKeys<U>>]: T[K] | Exclude<U[K & keyof U], undefined>;
-} & {
-  // all keys in U except those which are merged into T
-  [K in keyof Omit<U, Exclude<keyof T, RequiredKeys<U>>>]:
-    | Exclude<U[K], undefined>
-    | (undefined extends U[K] ? (K extends keyof T ? T[K] : undefined) : never);
-};
-export type MergeProps<T extends unknown[], Curr = {}> = T extends [infer Next, ...infer Rest]
-  ? MergeProps<
-      Rest,
-      Next extends object
-        ? Next extends Function
-          ? Curr
-          : Next extends infer TNext
-          ? Override<Curr, UnboxLazy<TNext>>
-          : Curr
-        : Curr
-    >
-  : Simplify<Curr>;
+type Override<T, U> = T extends any
+  ? U extends any
+    ? {
+        [K in keyof T]: K extends keyof U
+          ? undefined extends U[K]
+            ? Exclude<U[K], undefined> | T[K]
+            : U[K]
+          : T[K];
+      } & {
+        [K in keyof U]: K extends keyof T
+          ? undefined extends U[K]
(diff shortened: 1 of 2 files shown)


otonashixav:
A lot of the issues seem to be caused by it now excluding non-objects (which should be only null and undefined; I forgot to change that). If we try to compromise and ignore those instead, we get somewhat closer to what we want: https://tsplay.dev/w2aMbW

However I'd consider reverting as well.

riccardoperra:
I think the issue now is that you are always merging T and U they are not object. This will cause a conflict imho.

With a conditional all issues seems resolved https://tsplay.dev/mAyMvm

otonashixav:
In that case it can be simplified further, since `extends object` will already distribute the generics as desired: https://tsplay.dev/WKqMpN

However `test1.prop` is still not `"a" | "b"`, currently it passes because `"a"` was assigned to it so typescript narrows it down to `"a"`.

riccardoperra:
so the problem now is Omit which is not distributing the types correctly?

https://tsplay.dev/WJqMVm

otonashixav:
I think it looks good. I think the tests have to pass though, so maybe `ts-expect-error` the todos, and comment out or remove the test for non-object types since it's not applicable anymore. 

Some observations: 

- Even without using `object` and instead just distributing with `any`, it seems to work identically: https://tsplay.dev/N9EM3N 
- `never` instead of `U & T` is what seems to affect mutability: https://tsplay.dev/mMM2Zm, though as far as I can tell using `U & T` doesn't seem to cause any unwanted behaviour either. Considering the else branch seems to be unioned with the rest of the type, I find it weird that `never` seems to affect the type somehow.

riccardoperra:
Which "non-object" test types do you mean?

Anyway, I'm thinking about this possible case which is currently failing. 

I don't know if I understand mergeProps correctly, but shouldn't the type be "number" now instead of { a: 1 } & number?
```typescript
const m4 = mergeProps(
  { a: 1 },
  1,
);
type M4 = typeof m4;
// this fails either with number or number & { a: number }
type TestM4 = Assert<IsExact<M4, number>>;
```

otonashixav:
The test was written as `mergeProps` would ignore all non-object types. Currently it ignores only nullable types. The test should probably pass with `number & { a: number }` but it passing is probably not that important since it's unlikely that a number would be a props object. Regardless it looks like using `any` instead of `object` in `Override` solves this, though I'm not sure why exactly. 

Ideally it should also treat nullable props as if all the properties were optional e.g. `mergeProps({a: 1} as {a: number} | null) -> {a?: number}`, but trying to do so seems to break generics.

riccardoperra:
If mergeProps ignore all non object types, shouldn't be something like this?
https://tsplay.dev/WoGpLW

even if I couldn't manage all cases with this signature

otonashixav:
It should only ignore null and undefined. Other than that the way it interacts with non-object types is not really important.

riccardoperra:
Currently it doesn't skip 

> It should only skip null and undefined as described above. Other than that the way it interacts with non-object types is not really important, though the implementation will not exclude them e.g. `mergeProps(1)` includes all number methods.

Currently this test is always failing

```typescript
const m4 = mergeProps({ a: 1 }, 1, null, undefined, () => 1, "", 3, { a: 1 });
type M4 = typeof m4;
type TestM4 = Assert<IsExact<M4, { a: number }>>;
```

m4 is never if we pass a `null`, `undefined`, or an `object | null` type, is it correct?

otonashixav:
It should always fail with the current changes. The test should not test for anything but null, undefined and unions of those two with objects. However I doubt it is possible to make it pass while not breaking generics which has a higher priority than correctly typing nullables.

ryansolid:
Thank you

PR #999, opened 2022-05-22
chore(solid): fix typings for createRessource when initialValue is given

### LINKED ISSUE #995: Wrong types for `latest` resource field
### Describe the bug

At the moment latest is

`latest: T | undefined`

However, If we have the case, when we have an initial value (`someFetcher` returns list):
`const [listResource] = createResource(() => someFetcher, { initialValue: [] });`

We have never have the case when `listResource.latest` is `undefined`

-

-

### Expected behavior

`latest` should have type `T` when there's `initialValue`

### Screenshots or Videos

_No response_

### Platform

-

### Additional context

_No response_

cstrnt:
## Summary
Fixes #995 

TL;DR: `Resource` was always `T | undefined` even if an initialValue was passed. This is false because if `initialValue` is set, `Resource` should be `T`

Testing is quite hard for just types, therefore I did some local tests with https://github.com/SamVerschueren/tsd.
Maybe we should add some tests to the repo using this tool to ensure no regression?

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -429,9 +429,9 @@ export type ResourceReturn<
   K extends T = T
 > = [
   Resource<
-    O extends undefined
+    O extends undefined | null
       ? T | undefined
-      : undefined extends NonNullable<O>["initialValue"]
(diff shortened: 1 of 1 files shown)


elite174:
Nice! Thank you!

PR #1017, opened 2022-05-28
fix(solid): safeguard sharedConfig.load in Suspense

dolsem:
Should the `sharedConfig.load()` call be safeguarded in Suspense? I see it safeguarded in other places. I am trying to use Suspense in an Astro app and line 143 throws an error because `load` is undefined.


ryansolid:
I see. Astro I believe only does Solid's synchronous rendering currently which doesn't do Suspense on the server so it doesn't need load. I assumed it didn't need the hydration script whereas all my examples that aren't Astro have it. So I didn't notice this. Thanks.

PR #1043, opened 2022-06-07
Replace stores' usage of `__proto__` with `Object.getPrototypeOf`

otonashixav:
`__proto__` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto) and Deno does not support it.


ryansolid:
Sigh, pretty sure I benched this one. Ok it is what it is.

PR #1048, opened 2022-06-10
fix typo, consistent .

tomByrer:
At first I thought "libraru' was French?

no red squiglies


ryansolid:
Thanks

PR #1052, opened 2022-06-11
fix(types): remove `void` as a valid return type for a store setter and improve generic interactions

otonashixav:
## Summary

- Slightly improve interactions with generic types
- Remove `void` as a possible return type for a store setter (not sure why it was there to begin with)

Changes in where errors occur are marked with `// !!! changed`:

Old: https://tsplay.dev/WPRZ5N
New: https://tsplay.dev/NlEvxm

(Some `SetStoreFunction` overloads are commented out to improve loading times)

For reference here is why `void` should not be valid: https://playground.solidjs.com/?hash=-1986110315&version=1.4.1


ryansolid:
Yeah this is probably from when we had auto `produce` behavior.

PR #1058, opened 2022-06-14
feat(types): allow setting unions in stores

otonashixav:
## Summary

This makes it so that:
```ts
const [store, setStore] = createStore([{ a: 1 }, { b: 2 }]);
setStore(0, "a", 1); // currently error, now ok
```
Which is potentially unsafe, but is already allowed if e.g. `{ b: 2 }` was a `NotWrappable` instead.

There are performance concerns with this implementation.


shams96sami (approved):
Thanks

PR #1067, opened 2022-06-21
add `_$afterCreateRoot` hook to handle new roots in dev

thetarnav:
## Summary

Solid's internal reactivity graph has this problem where roots created with `createRoot` don't have to be added to its detachedOwner object — it's enough that they are able to get to the context walking upwards.
But not for devtools, all the `createRoot` calls are effectively invisible for a debugger that wants to walk downwards to analyze the tree. And roots are very common — components such as `<For>`, `<Index>` and `<Suspense>`; primitives that want to control disposal like `createLazyMemo`; and even `render` function that is creating the main branch of the reactive tree.

Having a special hook that would fire with every newly created root would let us handle all of those cases.

Then it could be used easily for tracking:
```ts
import { registerRoot } from "solid-devtools";
window._$afterCreateRoot = (root) => {
   registerRoot(root);
}
```

Related discussion: https://github.com/solidjs/solid/discussions/860
And devtools issue: https://github.com/thetarnav/solid-devtools/issues/15

Added a new test case to [`dev.spec.ts`](https://github.com/solidjs/solid/compare/main...thetarnav:handle-new-roots?expand=1#diff-e56c67a2f45b2acf11609b6dc85426c42257396fda3a4ef7ac9a703741a40304)

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -35,6 +35,7 @@ let rootCount = 0;
 
 declare global {
   var _$afterUpdate: () => void;
+  var _$afterCreateRoot: (root: Owner) => void;
 }
 
 export interface SignalState<T> {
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks.

PR #1092, opened 2022-06-30
feat: export ResourceOptions from solid

rasendubi:
## Summary

`ResourceReturn` is half-unusable without `ResourceOptions` as it is a required second argument of `ResourceReturn`.

Export `ResourceOptions` so users don't have to redefine it in their code.

Sorry, I edited/submitted this via github interface as change is rather small. I can go via full round of testing if that is required.


ryansolid:
Thanks this looks like an oversight.

PR #1096, opened 2022-07-02
Export `Owner` type

nksaraf:
Export the `Owner` type so that people can consume the return type of `getOwner` and `runWithOwner` properly.


ryansolid:
Yeah I noticed,... and I added this in a branch I was working on but no need to wait for that.

PR #1109, opened 2022-07-10
Add `has` trap for accurate `in` with stores and mutables

### LINKED ISSUE #1107: [feature] Stores don't have `in`/`has` trap
### Describe the bug

Feature request: It'd be nice for the `in` operator to be reactive (`has` and `deleteProperty` proxy traps). Example:

https://playground.solidjs.com/?hash=-210189711&version=1.4.1

https://playground.solidjs.com/?hash=-210189711&version=1.4.1

### Expected behavior

The effect should run when `n` is deleted in the example

### Screenshots or Videos

_No response_

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

### Additional context

_No response_

edemaine:
## Summary

Fixes #1107 by adding `has` trap to Stores and Mutables, so that code like `if ('prop' in store) store.prop()` is properly reactive. In Discord it's also been requested that `$TRACK in store` should return `true`, so I also added support for that (and other node types, to match `get` — notably missing `$NAME` though).

I haven't tested this yet, sorry, in particular because I still can't build Solid and run tests locally. But hopefully it's still useful, at least as a proof of concept? Feel free to reimplement it...

## Update: read before merging

[As discussed in #reactivity channel](https://discord.com/channels/722131463138705510/751355413701591120/996345879621537792), this might need special handling for `delete`d properties. Depends how we want to interpret `setStore(..., undefined)`...

### DIFF
--- a/packages/solid/store/src/mutable.ts
+++ b/packages/solid/store/src/mutable.ts
@@ -21,7 +21,7 @@ const proxyTraps: ProxyHandler<StoreNode> = {
     if (property === $TRACK) return trackSelf(target);
     const nodes = getDataNodes(target);
     const tracked = nodes[property];
-    let value = tracked ? nodes[property]() : target[property];
+    let value = tracked ? tracked() : target[property];
     if (property === $NODE || property === "__proto__") return value;
 
     if (!tracked) {
@@ -37,6 +37,13 @@ const proxyTraps: ProxyHandler<StoreNode> = {
     return isWrappable(value)
       ? wrap(value, "_SOLID_DEV_" && target[$NAME] && `${target[$NAME]}:${property.toString()}`)
       : value;
+  },
+
+  has(target, property) {
+    if (property === $RAW || property === $PROXY || property === $TRACK ||
+        property === $NODE || property === "__proto__") return true;
+    getDataNodes(target)[property]?(); // track
+    return property in target;
   },
 
   set(target, property, value) {
--- a/packages/solid/store/src/mutable.ts
+++ b/packages/solid/store/src/mutable.ts
@@ -42,7 +42,8 @@ const proxyTraps: ProxyHandler<StoreNode> = {
   has(target, property) {
     if (property === $RAW || property === $PROXY || property === $TRACK ||
         property === $NODE || property === "__proto__") return true;
-    getDataNodes(target)[property]?.(); // track
+    const tracked = getDataNodes(target)[property];
+    tracked && tracked();
     return property in target;
   },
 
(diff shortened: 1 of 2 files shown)


packages/solid/store/src/mutable.ts
```diff
@@ -39,6 +39,13 @@ const proxyTraps: ProxyHandler<StoreNode> = {
       : value;
   },
 
+  has(target, property) {
+    if (property === $RAW || property === $PROXY || property === $TRACK ||
+        property === $NODE || property === "__proto__") return true;
+    getDataNodes(target)[property]?(); // track
```
TiagoCavalcante, line 45: I guess it should be `?.`

ryansolid:
Hmm.. `in` not tracking is something we've used to our advantage in the past I believe. I'd need to check. Also things like the JSX heuristic should take this into consideration perhaps.

packages/solid/store/src/mutable.ts
```diff
@@ -42,7 +42,7 @@ const proxyTraps: ProxyHandler<StoreNode> = {
   has(target, property) {
     if (property === $RAW || property === $PROXY || property === $TRACK ||
         property === $NODE || property === "__proto__") return true;
-    getDataNodes(target)[property]?(); // track
+    getDataNodes(target)[property]?.(); // track
```
ryansolid, line 45: In general, I've avoid optional chaining in the code base as we don't do any transforms on the code base and in some environments it die. Every time one slips in something breaks something. As annoying as it is, that's why all the others are written the way they are.

edemaine: Oops, right; I forgot.

edemaine:
> Hmm.. `in` not tracking is something we've used to our advantage in the past I believe. I'd need to check.

I would have expected `unwrap` or `untrack` would be the way to avoid tracking. But should definitely check!

In core, I see that `For`, `Index`, `Suspense`, `observable`, `from`, and `applyState` use `in`, but I don't think the second argument can ever be a Store (e.g. it's `props`, an observer/producer, or an array).

> Also things like the JSX heuristic should take this into consideration perhaps.

Could you elaborate? Does the JSX compiler output `in` tests in some cases? Or are you suggesting it should?

ryansolid:
Im suggesting right now this would never be considered reactive according to our heuristic.

```js
<Show when={"value" in store}>...</Show>
```

So if `in` is we need to change it.

ryansolid:
Yeah that is all fine, just means it's probably a 1.5 sort of thing.

ryansolid:
Thank you

PR #1114, opened 2022-07-13
export store symbol $NODE

LiQuidProQuo:
## Summary

useful for more advance use cases.

1. make a prop readonly  / prop guards
2. extend to support a property with get/set  accessors 

for example(requires patching, so wont work in playground)
https://playground.solidjs.com/?hash=-1235554887&version=1.4.1

https://playground.solidjs.com/?hash=657365322&version=1.6.0 (with [thetarnav] extractor)

### DIFF
--- a/packages/solid/store/src/index.ts
+++ b/packages/solid/store/src/index.ts
@@ -1,4 +1,4 @@
-export { createStore, unwrap, $RAW } from "./store";
+export { createStore, unwrap, $RAW, $NODE } from "./store";
 export type {
   Store,
   SetStoreFunction,


3imed-jaberi (commented):
It's better to think about additional API(s) to solve all expected use cases or trade around safe API to be more extensible

thetarnav:
Looks like I'm gonna need access to those symbols for the debugger, to track stores.
not sure yet though, just experimenting for now
But this is what I did to get to them:
```ts
const { $NODE, $NAME } = (() => {
  const obj = {}
  ;(createMutable(obj) as any).v
  let $NODE!: symbol
  let $NAME!: symbol
  for (const s of Object.getOwnPropertySymbols(obj)) {
    if (s.description === 'store-node') $NODE = s
    else if (s.description === 'store-name') $NAME = s
  }
  return { $NODE, $NAME }
})()
```

ryansolid:
Yeah this is too low level I think. Necessary for Dev tools not really desirable as an API. We should focus on the gaps rather than expose this implementation detail at this time.

PR #1115, opened 2022-07-13
fix: handle `observable()` w/o getOwner

### LINKED ISSUE #1110: Using `observable()` outside of a `createRoot()` or `render()` call doesn't behave as expected
### Describe the bug

When using `solid-js` primatives outside of solidjs (e.g. in React), attempting to use the [`observable()` function](https://github.com/solidjs/solid/blob/f9fe3e498d247fc84012602de644937420ef9a3c/packages/solid/src/reactive/observable.ts#L25) will warn that you are attempting to `createComputed()` outside of a `createRoot()` call or a `render()` call because of this line:
https://github.com/solidjs/solid/blob/f9fe3e498d247fc84012602de644937420ef9a3c/packages/solid/src/reactive/observable.ts#L34-L38

Is there any downside to fixing this warning with the following:

```diff
export function observable<T>(input: Accessor<T>) {
  const $$observable = getSymbol();
  return {
    subscribe(observer: ObservableObserver<T>) {
      if (!(observer instanceof Object) || observer == null) {
        throw new TypeError("Expected the observer to be an object.");
      }
      const handler = "next" in observer ? observer.next.bind(observer) : observer;
      let complete = false;

+      const dispose = createRoot((disposer) => {
+        createComputed(() => {
+          // Would we need the `complete` variable here? When the `disposer` is called
+          // will that clean up this Computed?
+          if (complete) return;
+          const v = input();
+          untrack(() => handler(v));
+        });
+
+        return disposer;
+      });

      return {
        unsubscribe() {
          complete = true;
+          dispose();
        }
      };
    },
    [$$observable]() {
      return this;
    }
  };
}
```

none

This issue seems simple enough as to not need a reproduction. 

### Expected behavior

See the "describe this bug" section

### Screenshots or Videos

_No response_

### Platform

N/a

### Additional context

Would this change also eliminate the need for the `completed` variable?

jorroll:
Updates the `observable()` function to support usage outside the Solidjs context. Also updates `ObservableObserver` to make the "next" property in an observer object optional to align with RXJS's [`Subscribable` interface](https://rxjs.dev/api/index/interface/Subscribable) (which is what RXJS's `from()` expects). At the moment, `from(observable(mySignal))` results in a type error (assuming `from()` is imported from RXJS).

## Summary

Fixes #1110 

I'll note that the "Before submitting a pull request" instructions include "Format your code with prettier" yet the Solidjs repo doesn't include a formatting script nor is prettier listed as a devdependency? I attempted to run `yarn prettier` but since prettier wasn't available I gave up.

I reimplemented the `observable()` function in my React app using this solution.

### DIFF
--- a/packages/solid/src/reactive/observable.ts
+++ b/packages/solid/src/reactive/observable.ts
@@ -1,17 +1,26 @@
-import { Accessor, createComputed, createSignal, onCleanup, Setter, untrack } from "./signal";
+import { Accessor, createComputed, createRoot, createSignal, getOwner, onCleanup, Setter, untrack } from "./signal";
+
+// Note: This will add Symbol.observable globally for all TypeScript users,
+// however, we are not polyfilling Symbol.observable. Ensuring the type for
+// this global symbol is present is necessary for `observable()` to be
+// properly typed for 3rd party library's like RXJS.
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
+  }
+}
 
 function getSymbol() {
-  const SymbolCopy = Symbol as any;
-  return SymbolCopy.observable || "@@observable";
+  return Symbol.observable || "@@observable";
 }
 
 export type ObservableObserver<T> =
   | ((v: T) => void)
   | {
-      next: (v: T) => void;
-      error?: (v: any) => void;
-      complete?: (v: boolean) => void;
-    };
+    next?: (v: T) => void;
+    error?: (v: any) => void;
+    complete?: (v: boolean) => void;
+  };
 /**
  * creates a simple observable from a signal's accessor to be used with the `from` operator of observable libraries like e.g. rxjs
  * ```typescript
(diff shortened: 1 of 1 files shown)


packages/solid/src/reactive/observable.ts
```diff
@@ -29,16 +29,29 @@ export function observable<T>(input: Accessor<T>) {
...
-      let complete = false;
-      createComputed(() => {
-        if (complete) return;
-        const v = input();
-        untrack(() => handler(v));
-      });
+
+      const handler = typeof observer === 'function'
+        ? observer
+        : observer.next?.bind(observer);
```
ryansolid, line 35: Small nit. We don't use optional chaining in Solid core as it breaks some of the platforms we support. And we won't transpile the source code. Don't always trust it to do the most performant thing.

jorroll: Fixed 👍. Was able to make all the changes inside a [Codespace](https://github.com/features/codespaces) which is kinda cool.

packages/solid/src/reactive/observable.ts
```diff
@@ -1,17 +1,26 @@
-import { Accessor, createComputed, createSignal, onCleanup, Setter, untrack } from "./signal";
+import { Accessor, createComputed, createRoot, createSignal, getOwner, onCleanup, Setter, untrack } from "./signal";
+
+// Note: This will add Symbol.observable globally for all TypeScript users,
+// however, we are not polyfilling Symbol.observable. Ensuring the type for
+// this global symbol is present is necessary for `observable()` to be
+// properly typed for 3rd party library's like RXJS.
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
```
jorroll, line 9: As the comment above this interface suggests, ensuring that a type for `Symbol.observable` exists is necessary for Solid's `observable()` function to be properly typed for RXJS. This is because RXJS's `from()` function expects an object with `{ [Symbol.observable](): Subscribabble<T> }`. Previously, `getSymbol()` didn't return `Symbol.observable` but instead returned `any` which breaks type interoperability.

ryansolid:
Looks good. Thank you.

ryansolid:
I do have one more thought here. Disposal is tied to the context where the subscription happens being cleaned up, but I'm sort of wondering if the disposal should tied to the observable wrapper itself. Very much like we were talking about it maybe should be a single computed. It really could go either way because it's the hot versus cold thing. But Solid is Hot, and perhaps we should treat our Observables here as hot too, basically they are Subjects rather than Observables. 

@jorroll do you have an opinion.

ryansolid:
I agree with pretty much everything you've said here. Yeah doing creation lazily is better in terms of creation. I use the hot/cold pretty loosely I suppose, similar to [Andre Staltz' justification for creating XStream](https://staltz.com/why-we-built-xstream.html). In any case I think that approach is probably the correct approach. And the direction this should go ultimately. Thank you for confirming this.

jorroll:
Ah, from the article:

> RxJS Observables are generally cold. This means that two different calls to `subscribe()` will generate two separate side effect free executions of the Observable. Hot, on the other hand, is when different calls to `subscribe()` may share the same execution of the Observable.

In this case, and with deciding to capture the `Owner` at observable creation time rather than subscription time, moving towards hot observables sounds like the right idea.

> In any case I think that approach is probably the correct approach. And the direction this should go ultimately. Thank you for confirming this.

👍

PR #1116, opened 2022-07-14
Fix `Dynamic` types

LXSMNSYC:
## Summary

`Dynamic` accepts `T` as a prop, but this is wrong, we should infer the prop based on `component`.

The original implementation is based on [`solid-headless`](https://github.com/LXSMNSYC/solid-headless/blob/main/packages/solid-headless/src/utils/dynamic-prop.ts#L4-L14)

Example:
image


packages/solid/web/src/index.ts
```diff
@@ -99,9 +99,20 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
...
+type ValidConstructor = ValidElements | ValidComponent<unknown> | string;
+
+type DynamicBaseProps<T extends ValidConstructor> =
+  T extends ValidElements
+    ? JSX.IntrinsicElements[T]
+    :
+  T extends ValidComponent<infer U>
+    ? U
+    : never;
+
```
otonashixav, line 113: This is basically [`ComponentProps`](https://github.com/solidjs/solid/blob/main/packages/solid/src/render/component.ts#L75-L80), can we use that instead? 

I guess we still need to correctly parse `string & {}` though (should it return `Record<string, unknown>`? Web Components will probably end up there if not added to `JSX.IntrinsicElements`). The question is whether `ComponentProps` should be changed to accept it, or if it should be handled here.

packages/solid/web/src/index.ts
```diff
@@ -99,9 +99,20 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   return marker;
 }
 
-type DynamicProps<T> = T & {
-  children?: any;
-  component?: Component<T> | string | keyof JSX.IntrinsicElements;
+type ValidElements = keyof JSX.IntrinsicElements;
+type ValidComponent<P> = (props: P) => JSX.Element;
+type ValidConstructor = ValidElements | ValidComponent<unknown> | string;
```
otonashixav, line 104: I imagine `string` would need to be `string & {}` in order for autocomplete to suggest `ValidElements`. https://tsplay.dev/WY4Mdm (works for `a` but not for `b` where it suggests all tokens).
```suggestion
type ValidConstructor = ValidElements | ValidComponent<unknown> | (string & {});
```

packages/solid/web/src/index.ts
```diff
@@ -99,9 +99,20 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   return marker;
 }
 
-type DynamicProps<T> = T & {
-  children?: any;
-  component?: Component<T> | string | keyof JSX.IntrinsicElements;
+type ValidElements = keyof JSX.IntrinsicElements;
+type ValidComponent<P> = (props: P) => JSX.Element;
```
otonashixav, line 103: This is equivalent to [`Component`](https://github.com/solidjs/solid/blob/main/packages/solid/src/render/component.ts#L22), can we use that instead?

otonashixav (commented):
The failing tests should probably be changed. In [both](https://github.com/solidjs/solid/blob/main/packages/solid/web/test/dynamic.spec.tsx#L12) [cases](https://github.com/solidjs/solid/blob/main/packages/solid/web/test/dynamic.spec.tsx#L61) `keyof JSX.IntrinsticElements` should be removed, since it includes SVG elements which do not accept `title` (so in fact the error is desired).

packages/solid/web/src/index.ts
LXSMNSYC: The autocomplete works properly (as shown above) so we don't really need that intersection.

packages/solid/web/src/index.ts
LXSMNSYC: Yes, I'll make that change.

packages/solid/web/src/index.ts
LXSMNSYC: We can probably just change it to `Record<string, unknown>`

packages/solid/web/src/index.ts
otonashixav: The autocomplete for `component`, not the rest of the props.

LXSMNSYC: Makes sense, I'll do that

packages/solid/web/src/index.ts
```diff
@@ -99,9 +100,8 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   return marker;
 }
 
-type DynamicProps<T> = T & {
-  children?: any;
-  component?: Component<T> | string | keyof JSX.IntrinsicElements;
+type DynamicProps<T extends ValidComponent> = ComponentProps<T> & {
+  component?: T;
```
otonashixav, line 104: As mentioned on discord I would like to know why `component` is optional. If the intent is to allow `undefined` to unmount the dynamic component then `T | undefined` would be better as omitting `component` completely doesn't seem very useful.

LXSMNSYC: @ryansolid

ryansolid: Yeah there is no reason to omit it. In terms of unmounting it, I'm pretty sure wasn't my original intent but someone probably suggested it as something they might do. It probably makes sense to allow using it as a conditional I suppose. But yeah I think the `component` prop should be present even if it is allowed to be undefined.

LXSMNSYC: I have applied the change, thank you

Issue #1111, opened 2022-07-12
createMutable - delete key with undefined value does not trigger reactivity


LiQuidProQuo:
### Describe the bug

delete key with undefined value does not trigger reactivity

```js
let mutable = createMutable({a:'somevalue', b:undefined})

// reactive
delete mutable.a

// non reactive
delete mutable.b
```

minimal patch direction ( mutable proxy handler)
```diff
- function setProperty(state, property, value) {
- if (state[property] === value)
+ function setProperty(state, property, value, deleted=false) {
+ if (state[property] === value && !deleted)


    deleteProperty(target, property) {
-      setProperty(target, property, void 0);
+      setProperty(target, property, void 0, true);
+
```

### Your Example Website or App

https://playground.solidjs.com/?hash=-641634001&version=1.4.1

### Steps to Reproduce the Bug or Issue

see playground
1. click "delete B"
expect 2nd item to be removed from the list , and effect to log

2. click "delete A"
 expect the 1st item to be removed and effect to log

### Expected behavior

it is expected that when deleting B, it will be removed from the list
and trigger the effects 

### Platform

playground

paoloricciuti:
The reason because it's not reactive is because the value does not change: a deleted property is simply undefined so the Proxy does not rerun its dependencies.

LiQuidProQuo:
there is a difference between a defined property with an undefined value. (`{k:undefined}`)
and an object without that property.`{}` 

```js
Object.keys({k:undefined}).length //  1
Object.keys({}).length //  0
```
 
deleting a key with an undefined  value should trigger reactivity like with any other value.

paoloricciuti:
Yeah but your reactive object is not Object.keys (which effectively changes)...with createMutable you are creating a Proxy in which every field is a reactive variable. Each field will trigger its dependencies when the value change.

So the check Solid is actually doing is 

```typescript
Object.is(target[field], newValue);
```

and given that both values are undefined it will not trigger the rerender.

What we could do is tap into the deleteProperty trap of the Proxy to rerun the effects but this could lead to double running the effects when the value is not undefined. I'll take a look into this.

ryansolid:
fixed in 1.5

PR #1121, opened 2022-07-15
Fix createMutable to handle reactivity in case element is undefined

paoloricciuti:
## Summary

Fix for issue #1111 

I was unable to run the test in local (i'll try again) but i've added a test in the test suite that should be run by the CI i guess.

### DIFF
--- a/packages/solid/store/src/modifiers.ts
+++ b/packages/solid/store/src/modifiers.ts
@@ -1,6 +1,6 @@
 import { setProperty, unwrap, isWrappable, StoreNode, $RAW } from "./store";
 
-const $ROOT = Symbol("store-root")
+const $ROOT = Symbol("store-root");
 
 export type ReconcileOptions = {
   key?: string | null;
@@ -139,7 +139,7 @@ const setterTraps: ProxyHandler<StoreNode> = {
   },
 
   deleteProperty(target, property) {
-    setProperty(target, property, undefined);
+    setProperty(target, property, undefined, true);
     return true;
   }
 };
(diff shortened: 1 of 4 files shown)


ryansolid:
Thank you. I know why this is failing but it does beg the question if everything goes through that check we probably don't need the signal to have the comparator and we can trim code and execution. That change has farther reaching implications. So this one might be better to merge into the next branch which I should be publishing today or tomorrow as part of pre-release for 1.5.

paoloricciuti:
What I meant is that I was unable to run any test in local, but now I'm seeing that the CI is failing...I'll take a look and maybe go a little deeper if that's something that you are interested into fix.

ryansolid:
It is and your fix is mostly right, but there is another change needed that I need to test repercussions on. So if you don't mind I will probably merge this as is, into next branch after I re-evaluate the other required changes.

paoloricciuti:
You are the boss...out of curiosity: why is failing? Also...is there something I need to do to make test work locally? I had "unexpected token" errors.

ryansolid:
Did you run `npm run build` before testing the first time?

The problem is the internal signal has a comparison function which prevents it from triggering anyway. undefined === undefined. But given we already do the equality check there I am thinking we might be able to remove the comparison function.

paoloricciuti:
> Did you run `npm run build` before testing the first time?
> 
> The problem is the internal signal has a comparison function which prevents it from triggering anyway. undefined === undefined. But given we already do the equality check there I am thinking we might be able to remove the comparison function.

Yes I did run `npm run build` but it still gave me the error...also tu build I had to npm install into every single package.

Yeah I understand and I feel totally stupid now 😃

ryansolid:
Thank you.

PR #1129, opened 2022-07-20
update and rework build system

modderme123:
## Summary

* Switch from lerna and npm to pnpm and turbo
* Update jest, typescript, rollup, babel
* Rework tsc build scripts to separate build and types

## How did you test this change?

* `pnpm run build`
* `pnpm run test`
* `pnpm run types`
* `pnpm run test-types`
* `pnpm run bump 1.5.0-beta.1` then `pnpm run bump 1.5.0-beta.0`

PR #1136, opened 2022-07-26
Add build instructions to CONTRIBUTING

edemaine:
## Summary

It's not trivial to build Solid, especially if you're not familiar with Lerna, and especially if you're on Windows. This adds the basics to CONTRIBUTING.md (somewhat similar to the PR template, but that's kind of late in the game for someone getting started with tweaking Solid). The Windows workaround is something we found on Discord for being able to build (though tests still don't quite work for me, but it's a starting point).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,7 +87,7 @@ We can't wait to see what you build!
 ## Building Solid
 
 This repository uses [pnpm](https://pnpm.io/) and
-[Turbo](https://turbo.hotwired.dev/).
+[Turborepo](https://turborepo.org/).
 If you want to build Solid from scratch, use the following steps:
 
 1. `pnpm install` (install all dependencies)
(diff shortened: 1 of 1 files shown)


ryansolid:
Only comment is we completely changed the build process in the next release thanks to #1129. That needs to be documented definitely.

edemaine:
Ah, great, `pnpm`! Do you want to merge this in `main` and then modify in `next`, or should I/you modify into a commit just for `next`?

ryansolid:
Yeah I'm thinking we just target next with updated instructions and maybe not worry about this for the next week or so before we are ready to merge into main.

edemaine:
I just tested the new build system and it works flawlessly (no more Windows workaround). I still have the same test failure as before, but it's progress at least. I've revised the instructions accordingly.

CONTRIBUTING.md
```diff
@@ -83,3 +83,14 @@ Contributing to ecosystem projects is just as important as contributing to Solid
 If you haven't found any interesting information on this page then we encourage you to start hacking at a Solid related utility or package that does. Building useful tools for fellow OSS ecosystem and Solid users enhances the whole platform.
 
 We can't wait to see what you build!
+
+## Building Solid
+
+This repository uses [pnpm](https://pnpm.io/) and
+[Turbo](https://turbo.hotwired.dev/).
```
ryansolid, line 90: Wrong turbo.. should be this one: https://turborepo.org/

edemaine: Whoops, indeed! I wasn't sure whether links were worth adding, but this kind of confirms that they are. 😅 Fixed!

ryansolid:
Thanks

PR #1137, opened 2022-07-28
Update Readme

mathieuprog:
Babel is not installed as a dev dependency


developersharif (changes requested):
it should be accept.

3imed-jaberi (approved):

ryansolid:
This doesn't really matter. But sure why not?

PR #1139, opened 2022-07-31
fix: add explicit extensions and exports fields

Drevoed:
## Summary

Following the trace of https://discord.com/channels/722131463138705510/783844647528955931/999968111937736764, I have found out the culprit of solid-js's problem with typescript and specifically NodeNext setting. The extensions in the .d.ts files must be defined explicitly, as well as have the correct `exports.import.types` being set. This pull request does exactly that. 

Adding extensions won't break existing typescript codebases.
See this comment: https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937

This PR depends on: https://github.com/ryansolid/dom-expressions/pull/154

Published a new version locally and tested pet projects using solid and checked the type output and inference of exports from 'solid-js' when using NodeNext `module` and `moduleResolution`.

### DIFF
--- a/packages/solid/jest.config.cjs
+++ b/packages/solid/jest.config.cjs
@@ -1,4 +1,7 @@
 module.exports = {
+  moduleNameMapper: {
+    "(.+)\\.js": "$1"
+  },
   verbose: false,
   collectCoverageFrom: ["src/**/*.ts", "store/src/**/*.ts", "web/src/**/*.ts", "!**/*.d.ts", "!src/server/*.ts", "!store/src/**/server.ts"],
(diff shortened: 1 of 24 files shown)


Drevoed:
FFS, I forgot Jest exists.

ryansolid:
This looks good. Any chance I can get you to rebase and make the PR against `next` branch. I usually would do it myself but we just completely changed the build system and are gearing towards a new release. Probably makes sense to add module changes there

Drevoed:
Sure, will do

ryansolid:
Thank you.

PR #1176, opened 2022-08-24
Solid 1.5

ryansolid:
As it says. Check out the branch and the changelog. 

Test it with solid-js@1.5.0-beta.6


Kerrick:
According to the release notes, "While higher-level APIs are still being determined we now have a way to plugin the internal storage by passing something with the signature of a signal to the new Experimental storage option."

One year and three releases later, is this still considered Experimental, or are we safe to bake this into our applications at a deep level? (Also, have any higher-level APIs been determined?)

ryansolid:
Yes. Although in the sense that I don't see this changing in Solid 1.x but it will likely change in 2.0. The reason I put experimental on it was that Solid's current reactivity system isn't capable of what we need to have a better solution here, so change here is inevitable.

PR #1177, opened 2022-08-25
untrack JSON.stringify to avoid reactive side-effects on serialization.

LiQuidProQuo:
**untrack**  JSON.stringify  

In dev mode, `hashValue()`  uses `JSON.stringify` to create a hash of a signal  object value.
if the value happen to contain a solid store, the nested serialization, may register reactive dependencies. 

it may also cause side effects, for non solid code. but this only guard the solid reactivity case.


ryansolid:
I think it needs to be a function passed to untrack.

LiQuidProQuo:
This is an odd error, the Listener is declared and initialized in `reactive/signal.ts`
before the  `untrack` function 

if I change the declaration of `Listener` from `let` to `var`  this error no long shows. 

https://github.com/solidjs/solid/blob/a9658f4e2fb8156ce645c82c7fc25466ccada133/packages/solid/src/reactive/signal.ts#L29
```js
solid-js: FAIL test/signals.spec.ts
solid-js:   ● Test suite failed to run
solid-js:     ReferenceError: Cannot access 'Listener' before initialization
solid-js:       789 | export function untrack<T>(fn: Accessor<T>): T {
solid-js:       790 |   let result: T,
solid-js:     > 791 |     listener = Listener;
solid-js:           |                ^
solid-js:       792 |
solid-js:       793 |   Listener = null;
solid-js:       794 |   result = fn();
solid-js:       at untrack (src/reactive/signal.ts:791:16)
solid-js:       at hashValue (src/reactive/signal.ts:1031:11)
solid-js:       at createSignal (src/reactive/signal.ts:190:[44](https://github.com/solidjs/solid/runs/8010904224?check_suite_focus=true#step:6:45))
solid-js:       at Object.<anonymous> (src/reactive/signal.ts:24:55)
solid-js:       at Object.<anonymous> (src/index.ts:1:1)
solid-js:       at Object.<anonymous> (test/signals.spec.ts:1:1)
```

ryansolid:
Thank you

PR #1182, opened 2022-08-26
fix: wrong `mutate` callback types of createResource with initialValue

riccardoperra:
As the documentation says, the prev() acessor of a createResource with initialValue should never return undefined.
```
Setting an `initialValue` in the options will mean that both the prev() accessor and the resource should never return undefined (if that is wanted, you need to extend the type with undefined)
```

Using solid-js@1.5, while using a `createResource` with an initial value, the `mutate` callback previous value is inferred as T | undefined instead of T. 

Here a small playground comparing old types which didn't have this type issue:

image
https://stackblitz.com/edit/vitejs-vite-8em7ud?file=src%2FApp.tsx,src%2Fsolid-1.4.tsx

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -447,7 +447,7 @@ export type Resource<T> = Unresolved | Pending | Ready<T> | Refreshing<T> | Erro
 export type InitializedResource<T> = Ready<T> | Refreshing<T> | Errored;
 
 export type ResourceActions<T, R = unknown> = {
-  mutate: Setter<T | undefined>;
+  mutate: Setter<T>;
   refetch: (info?: R) => T | Promise<T> | undefined | null;
 };
 
@@ -476,7 +476,10 @@ export type InitializedResourceOptions<T, S = unknown> = ResourceOptions<T, S> &
   initialValue: T;
 };
 
-export type ResourceReturn<T, R = unknown> = [Resource<T>, ResourceActions<T, R>];
+export type ResourceReturn<T, R = unknown> = [
+  Resource<T>,
+  ResourceActions<T | undefined, R>
+];
 
 export type InitializedResourceReturn<T, R = unknown> = [
   InitializedResource<T>,
--- a/packages/solid/test/resource.type-tests.ts
+++ b/packages/solid/test/resource.type-tests.ts
@@ -1,4 +1,10 @@
-import { createResource, ResourceReturn, createSignal, Resource } from "../src";
+import {
+  createResource,
+  ResourceReturn,
+  createSignal,
+  Resource,
+  Setter
+} from "../src";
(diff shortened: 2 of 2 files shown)


otonashixav:
I made this change because at the time it could be `undefined` (up until [this change](https://github.com/solidjs/solid/commit/116c584bf1177612b79161c883a1ca33cfb119f2#diff-252a63336451e12211c93f2ea84f2264b37ced5332297fc947994101cde3726dL585-R585)), though now it can be anything that is thrown: https://stackblitz.com/edit/vitejs-vite-tfbhih?file=src/App.tsx.

ryansolid:
Yeah this is tricky because the other behavior was bugged. We need to change and hence it is a different value.. However, because it is errored it will throw so you will never be able to read this other value. Except admittedly from here which I missed. I thought it was safe but it isn't. 

Hmm.. maybe we can fake undefined in this case. Like intercept the setter and if errored make it undefined here. We know when we are in an errored state.. I guess the problem is the error clearing timing. Still solvable.

riccardoperra:
> Yeah this is tricky because the other behavior was bugged. We need to change and hence it is a different value.. However, because it is errored it will throw so you will never be able to read this other value. Except admittedly from here which I missed. I thought it was safe but it isn't.
> 
> Hmm.. maybe we can fake undefined in this case. Like intercept the setter and if errored make it undefined here. We know when we are in an errored state.. I guess the problem is the error clearing timing. Still solvable.

Should the error be cleared because the mutate() will not update the `state`? I'm trying to understand the code, if I understood the 'errored' state might change only if you call the refetch function. Is there a specific reason for this behavior? 

```typescript
data.state // errored
mutate((data: Error) => true)
data.state // errored
```

Could this code be a possible implementation of a mutate()? Assuming it should update the state

https://github.com/solidjs/solid/blob/eed39f1e140671ae83ecf60edb1cd030ddfd6aac/packages/solid/src/reactive/signal.ts#L560-L565

```typescript
const mutate: Setter<T | undefined> = (value?: unknown) => {
  const currentState = untrack(state);
  return setValue(latestValue => {
    let newValue, threwError: unknown = undefined;
    //  typescript doesn't like the return type then declaring the newValue as any will not break the Setter signature
    newValue = value;
    if (typeof value === "function") {
      const resolvedValue = currentState === "errored" ? undefined : latestValue;
      try {
        newValue = value(resolvedValue);
      } catch (e) {
        threwError = e;
        newValue = e as T | undefined;
      }
    }
    // should these updates be wrapped in runUpdates()? 🤔 
    // This allows to update the state without calling refetch() 
    err = thrownError;
    const newState = err 
      ? "errored" 
      :  resolved ? "ready" : currentState
    setState(newState);
    return newValue;
  });
};
```

ryansolid:
Yeah state was tagged on after and has this flaw right now. Mind you I can't mess with the mutate function for different reasons. Thanks for pointing this out.

PR #1192, opened 2022-08-29
perf: with createMemo typings

Highnesslin:
As in the type definition of `createMemo`, 
```typescript
createMemo<T>(fn: (v?: T) => T, value?: undefined, options?: {
  equals?: false | ((prev: T, next: T) => boolean);
  name?: string;
}): () => T
```
we can automatically infer the type of `cache` directly using the paradigm


Highnesslin (commented):
use typings

ryansolid:
Thank you

PR #1213, opened 2022-09-08
fix: createResource null check failing on load

lxsmnsyc:
When `createResource` is trying to resolve, it checks with the condition `(typeof p !== "object" || !("then" in p))` so that it knows if it should await for the value. However, `typeof null` is equal to `"object"` and so the condition fails for the promise check.


ryansolid:
Thank you.

PR #1247, opened 2022-09-25
add <rb> element to jsx-runtime

Yukaru-san:
## Summary

Added the &lt;rb> element to the jsx-runtime. 

The &lt;rb> HTML element is used to delimit the base text component of a &lt;ruby> annotation and was missing in the jsx-runtime resulting in the error "Property 'rb' does not exist on type 'JSX.IntrinsicElements'" when used.


ryansolid:
To my knowledge this element is deprecated. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rb. I checked the types for Preact and Inferno and they both also didn't have `<rb>`.

Generally, we don't add deprecated elements. Although it should be possible to extend JSX.IntrinsicElements yourself in the same way we can register Web Components.

PR #1248, opened 2022-09-26
fix `Dynamic` type

mdynnl:
## Summary

Currently, `Dynamic` can't infer props from `component` and will fail to type check them
```
<Dynamic component="a" href={2} />
```
This PR tries to fix this with [`Expand`](https://stackoverflow.com/a/57683652/6808116)

I also didn't format with prettier as I'm not sure it's intentional

trailing comma here, (might help with git diff probably)
https://github.com/solidjs/solid/blob/dcfd8ceeedb3e5d98e548a83cf2063c16b18926a/packages/solid/web/src/index.ts#L15

but not here
https://github.com/solidjs/solid/blob/dcfd8ceeedb3e5d98e548a83cf2063c16b18926a/packages/solid/web/src/index.ts#L29

missing semi
https://github.com/solidjs/solid/blob/dcfd8ceeedb3e5d98e548a83cf2063c16b18926a/packages/solid/web/src/index.ts#L115

### DIFF
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -100,9 +100,7 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   return marker;
 }
 
-type Expand<T> = T extends infer O ? { [K in keyof O]: O[K] } : never;
-
-type DynamicProps<T extends ValidComponent> = Expand<ComponentProps<T>> & {
+type DynamicProps<T extends ValidComponent, P = ComponentProps<T>> = { [K in keyof P]: P[K] } & {
   component: T | undefined;
 };
 /**
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -100,7 +100,9 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   return marker;
 }
 
-type DynamicProps<T extends ValidComponent> = ComponentProps<T> & {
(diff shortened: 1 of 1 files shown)


Brendan-csel:
While I can't claim to understand exactly how this works, I can confirm it did resolve the issues I was seeing.  With this change in place, Typescript does correctly complain if the props don't match the component passed to <Dynamic />

atk (approved):
While `Expand<T>` will only work on objects without an additional type guard, this should suffice in this special case, since props are always objects. I would maybe consider narrowing `T` to `ExpandObject<T extends object>` to make sure this is not misused, just in case someone exports and reuses it, but that's up to you.

mdynnl:
```ts
type DynamicProps<T extends ValidComponent, P = ComponentProps<T>> = { [K in keyof P]: P[K] } & {
  component: T | undefined;
};
```
I'm thinking if inline is better as it doesn't introduce a new type and `DynamicProps` is also not exported (currently).

mdynnl:
Actually the problem is 
https://github.com/solidjs/solid/blob/dcfd8ceeedb3e5d98e548a83cf2063c16b18926a/packages/solid/src/render/component.ts#L70

mdynnl:
But I believe `Component<any>` is necessary so that `ValidComponent` can accept any kind of component e.g `Component<SomeInterface>`. And the current fix seems to work and no tests failed (locally). I think this can be merged.

ryansolid:
Thank you

PR #1255, opened 2022-09-27
DEV: remove autogenerated names

thetarnav:
- remove autogenerated owner and signal names
- add unique id to owners and signals
- remove `componentName` to instead use the `name` prop (component owners can be identified by the fact that `props` field exists)

My goal is to separate the autogenerated (`"c-1-1-1-1-1..."`, `"s-951856283"`) names from explicit user names, as the user names are the only ones I want to display to users. But the generated names occupy the same field.
The individual owners can be identified by a simple id, which I'm doing now, but it can be added here as a replacement for the generated names.
Alternatively id and name could still be merged. e.g. with a format: `<id>:<name>`. But that's just serialization, so probably a task for the devtools.

The `hashValue` and `serializeGraph` functions are interesting but do not have a use now.
They can still be called from the `DEV` object if necessary.

Not sure if I'm not missing something here... especially around stores, as I haven't gotten to them in the devtools so I don't know yet what's important there.
So it'll be useful for me, but the details probably will change in the future.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -29,7 +29,8 @@ let Listener: Computation<any> | null = null;
 let Updates: Computation<any>[] | null = null;
 let Effects: Computation<any>[] | null = null;
 let ExecCount = 0;
-let rootCount = 0;
+let ownerCount = 0;
+let signalCount = 0;
 
 // keep immediately evaluated module code, below its indirect declared let dependencies like Listener
 const [transPending, setTransPending] = /*@__PURE__*/ createSignal(false);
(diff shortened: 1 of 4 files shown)


ryansolid:
How pressing is this? I feel like I should do this at a clean break like 1.6 version rather than sneak it in right now. Maybe nothing here is too destructive. Just it would be best to point at a specific minor than than some random patch version.

thetarnav:
Not pressing at all, those are just names
tbh I just wanted to get my thinking out and see how it fits in the solid code

thetarnav:
I'm closing this for now so that the state persistance for HMR can be implemented first. I'll look at the names again after that.

PR #1259, opened 2022-09-30
fix exactOptionalPropertyTypes when using css modules

PodaruDragos:
This fixes typesscript's `exactOptionalPropertyTypes` when using css modules

```
Types of property 'class' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

I just added a `| undefined` to check for class property


ryansolid:
This file is actually part of the dom-expressions file and then gets imported during build. If you want you can make a PR against it. Or I can just copy the changes across when I get a chance.

https://github.com/ryansolid/dom-expressions/tree/main/packages/dom-expressions/src

ryansolid:
Ended up adding myself.

PR #1260, opened 2022-10-01
feat: call components with `.call` to allow libs to provide class components with static `.call` methods

trusktr:
Also formatted with prettier, some unformatted code got by in previous commits.

## Summary

Thanks @fabiospampinato for ideating and ending with this idea.

Calling components as `Comp.call(Comp, props)` allows for classes to be used as components if they have a `static call` method. Existing function components work the same, and this is what a class could look like:

```js
class ShowCount {
  static call(Comp, props) {
    return new Comp().template(props)
  }

  foo = 123

  template(props) {
    return <div>Count is {props.count} and foo is {this.foo}</div>
  }
}

function NonClassComp() {
  return <ShowCount count={456} />
}
```

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -65,10 +65,7 @@ export type FlowComponent<P = {}, C = JSX.Element> = Component<FlowProps<P, C>>;
 /** @deprecated: use `ParentProps` instead */
 export type PropsWithChildren<P = {}> = ParentProps<P>;
 
-export type ValidComponent =
-  | keyof JSX.IntrinsicElements
-  | Component<any>
-  | (string & {});
+export type ValidComponent = keyof JSX.IntrinsicElements | Component<any> | (string & {});
 
 /**
  * Takes the props of the passed component and returns its type
@@ -77,13 +74,11 @@ export type ValidComponent =
  * ComponentProps<typeof Portal> // { mount?: Node; useShadow?: boolean; children: JSX.Element }
  * ComponentProps<'div'> // JSX.HTMLAttributes<HTMLDivElement>
  */
-export type ComponentProps<T extends ValidComponent> =
-  T extends Component<infer P>
-    ? P
-    :
-  T extends keyof JSX.IntrinsicElements
-    ? JSX.IntrinsicElements[T]
-    : Record<string, unknown>;
+export type ComponentProps<T extends ValidComponent> = T extends Component<infer P>
+  ? P
+  : T extends keyof JSX.IntrinsicElements
+  ? JSX.IntrinsicElements[T]
+  : Record<string, unknown>;
 
 /**
  * Type of `props.ref`, for use in `Component` or `props` typing.
@@ -99,7 +94,7 @@ export function createComponent<T>(Comp: Component<T>, props: T): JSX.Element {
       setHydrateContext(nextHydrateContext());
       const r = "_SOLID_DEV_"
         ? devComponent(Comp, props || ({} as T))
-        : untrack(() => Comp(props || ({} as T)));
+        : untrack(() => Comp.call(Comp, props || ({} as T)));
       setHydrateContext(c);
       return r;
     }
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1031,7 +1031,7 @@ export function devComponent<T>(Comp: (props: T) => JSX.Element, props: T) {
     () =>
       untrack(() => {
         Object.assign(Comp, { [$DEVCOMP]: true });
-        return Comp(props);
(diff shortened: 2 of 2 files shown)


trusktr:
Maybe it can be just `Comp.call(null, props)`, and a class can use `this`, as in

```js
class ShowCount {
  static call(_, props) {
    return new this().template(props)
  }
  ...
}
```

which leads to the same result.

trusktr:
I can't build or test locally, I get

```
 ERROR  workspace configuration error: package.json: no workspaces found. Turborepo requires npm workspaces to be defined in the root package.json
```

Also GitHub actions reports the tests to have passed, but if you take a look, it had a bunch of errors.

packages/solid/src/render/component.ts
```diff
@@ -77,13 +74,11 @@ export type ValidComponent =
...
-    ? P
-    :
-  T extends keyof JSX.IntrinsicElements
-    ? JSX.IntrinsicElements[T]
-    : Record<string, unknown>;
+export type ComponentProps<T extends ValidComponent> = T extends Component<infer P>
+  ? P
+  : T extends keyof JSX.IntrinsicElements
+  ? JSX.IntrinsicElements[T]
+  : Record<string, unknown>;
```
trusktr, line 81: just formatting applied by prettier

packages/solid/src/render/component.ts
```diff
@@ -65,10 +65,7 @@ export type FlowComponent<P = {}, C = JSX.Element> = Component<FlowProps<P, C>>;
 /** @deprecated: use `ParentProps` instead */
 export type PropsWithChildren<P = {}> = ParentProps<P>;
 
-export type ValidComponent =
-  | keyof JSX.IntrinsicElements
-  | Component<any>
-  | (string & {});
+export type ValidComponent = keyof JSX.IntrinsicElements | Component<any> | (string & {});
```
trusktr, line 68: just formatting applied by prettier

LiQuidProQuo:
this change may have performance implications, where the common case of "function" has to go 
through the slower path of "class".

`func.call()` vs  `func()`




----

alternatively, the class can be wrapper in a "functional factory".
with minimal overhead (<1%), and without any internal changes.

https://playground.solidjs.com/?hash=437964492&version=1.4.1

comparing wrapped/direct instantiating of a class.
image

trusktr:
That's interesting that a JS engine doesn't simply optimize away what are essentially identical calls.

> alternatively, the class can be wrapper in a "functional factory". with minimal overhead (<1%), and without any internal changes.

What would the code look like? If we use `new`, then it means we must also write code to detect if a function is `new`able.

LiQuidProQuo:
> What would the code look like? If we use `new`, then it means we must also write code to detect if a function is `new`able.

I have attached a playground to demonstrate what it looks like practically. 
there is no magic on the framework side, no auto detecting. it is just a way to comply with the current contract of component as a function.

ryansolid:
Performance implications means it needs to be tested, considered, and not a quick merge.

Exelord:
As I love classes as components I have no idea how I missed this PR 😅 But it seams like the issues stopped being relevant as it can be simply solved by splitting class state from template:

```js
class State {
  foo = 123
}

export function ShowCount(props) {
  const state = new State(props);
  return <div>Count is {props.count} and foo is {state.foo}</div>
}
```

the flexibility of current solution is strong enough to cover all possible cases and abstract layers without creating new API and deal with perf choices. It also allows for state / template separation for those against SFC. I agree it could be simplified with some utils but I believe that's a role for external package to come up with things like:

```js
import { ComponentState, componentTemplate } from "solid-class-components"

class State extends ComponentState {
  foo = 123;
}

export const ShowCountState = componentTemplate(State, (props) => (
  <div>
    Count is {props.count} and foo is {this.foo}
  </div>
));
```

Ofc, this would affect performance as well, but it is fully controllable by the user without general cost to all components.

ryansolid:
Thanks for the PR, and for the work you’ve done exploring this model in Classy Solid and the Lume element stack. I can see why this came up: class-oriented authoring maps naturally to Custom Elements, and Classy Solid is a real exploration of that direction rather than just a hypothetical use case.

I’m going to close this one as stale for Solid 2, though. Solid core’s component contract is still intentionally function-based, and changing component creation from `Comp(props)` to `Comp.call(...)` would affect every component invocation. That has potential performance cost, and it also changes observable `this` semantics for ordinary function components.

I also don’t think this patch alone would be enough to make class components a supported core API. The component types, JSX expectations, SSR/dev behavior, docs, and tests would all need a broader design.

For now, I think Classy Solid / Lume is the better place for this authoring model to evolve. That keeps the class and Custom Element ergonomics available for people who want them without adding extra semantics or overhead to Solid’s base component path.

PR #1262, opened 2022-10-03
Fix mutable descriptor

trusktr:
## Summary

Similar to a store proxy's descriptor trap, except this adds the `set`ter, assuming that mutables are mutable.

I'm not sure if we should convert a value descriptor to an accessor though, but that may be a different topic. For example, maybe `getOwnPropertyDescriptor` on a value descriptor just counts as a read (tracked), nothing more; and `defineProperty` with a value descriptor would be a write and would trigger.

Not tested yet.


trusktr:
tests failed for some odd reason

ryansolid:
This PR has your call stuff in it too? It's all entwined.

trusktr:
Oops, didn't mean to put call stuff in it too. Thanks for getting that in.

PR #1266, opened 2022-10-07
Add option to name context nodes

thetarnav:
Context nodes should be able to be named, just as signals and computations.
Without names, context nodes don't have anything besides their value to identify different context nodes.
Luckily since context is provided using a render effect, the `createContext` API could accept the same options argument and pass it to `createRenderEffect`.
Creating context involves creating a global variable that—similarly to createSignal and createMemo—could be used by a babel plugin to automatically fill the name option.


ryansolid:
Thanks

PR #1304, opened 2022-10-23
add code format

shengxinjing:
## Summary

Added base configuration for eslint, no fixes performed yet(TODO)


ryansolid:
I've never used or had much use for stuff like this but I think there could be interest to consolidate contribution to Solid. Would love for people to weigh in since this is a place I have no experience.

maxmilton:
It would have been nice to see some discussion about this first.

Linting is a good idea, especially as the project grows in the number of contributions. Not just for code quality but also as a way to guide contributors towards preferred coding patterns. Eventually, there should probably be a shared eslint preset used across repos in the solidjs org.

I like that this PR has kept the direct dependencies small (although I've always wished there was a lightweight alternative to eslint, which has [a large dependency tree](https://npm.anvaka.com/#/view/2d/eslint)). It's nice you've started with a limited rule set so we don't overwhelm contributors with lint errors/warnings.

All that said, linting setup can be pretty personal. The rules provided by `eslint:recommended` and `plugin:@typescript-eslint/recommended` are a great baseline I use in all my projects. Those set a lot of rules though! Rule setup might be something @ryansolid will need to do to get the feel right. I wouldn't want you, Ryan, to be bummed by linting and spend less time on solid as a result.

shengxinjing:
Thanks for reply，linting is necessary for new contributors like me ,  this rules extends `plugin:@typescript-eslint/recommended` and set some strict rules off , it's popular in most typescript projects 

we can fix the code at the end of the rules discussion with `eslint --fix `, and add eslint format into build system and git hooks

shengxinjing:
after the rules discussion finished,  we can move this rules in a repo like `solid-eslint-config` for all solidjs org projects

shengxinjing:
reference links about eslint in other projects
React    https://github.com/facebook/react/blob/main/.eslintrc.js
Vue       https://github.com/vuejs/core/blob/main/.eslintrc.js

vladnicula:
A consistent format across the codebase could help as it will reduce the amount of back and forth regarding coding styles when reviewing PR.

One counter argument to eslint is that it is configurable, and therefore people will start complaining about what rules to adopt in the ESLINT, so the conversation moves from each PR's code review into a separate thread, but it still tends to go on and on forever. The solution to situations like these is to use prettier to keep the code consistent. Prettier is not configurable. And while that might seem wired, it is no configurable because that way developers won't start to complain about which coding style to pick. 

Regardless of choice, I'd say any linting and formatting is better than no formatting.

thetarnav:
imo having prettier would be great
not sure how eslint would help formatting in this particular codebase where common "rules" are often broken to push js and bundle size as much as possible 🤔
The additions here are slow and deliberate by nature anyway
But maybe there are a couple of rules that are actually worth enforcing, just so they can be caught in ci, like no console logs or debugger statements

wjq990112:
In my opinion, eslint is not only the linter but also the syntax checker.
I'd like to use prettier at the first to ensure our code is in a common style at first, and add eslint rules progressively to check our syntax.
Maybe adding all the eslint rules at once will cause the code to change a lot, or there will be a lot of syntax errors.
Changing something progressively is more important than breaking it into an existing repository.

lxsmnsyc:
I don't think this repo is desirable to have code corrections (for potential code-smells) since Solid has a lot of code golfs. If anything, eslint would definitely correct a lot of "mistakes" here but the corrections may impact the code performance overall.

quantuminformation:
code golf?

Brian-Pob:
I also learned of this term recently. https://en.wikipedia.org/wiki/Code_golf
Code golfing refers to shortening code for the sake of shortening it. Some believe shorter code is always better for readability while others disagree. For the case of Solid, some code could be made shorter but might be less performant.

shengxinjing:
I still believe that appropriate eslint code specifications are necessary, and if possible, I can be invited to join the organization to complete this task

shengxinjing:
> @quantuminformation I also learned of this term recently. https://en.wikipedia.org/wiki/Code_golf Code golfing refers to shortening code for the sake of shortening it. Some believe shorter code is always better for readability while others disagree. For the case of Solid, some code could be made shorter but might be less performant.


Great concept, learned. eslint is configurable and can customize rules and even plugins according to our needs

Issue #1301, opened 2022-10-22
Cannot redefine property


juanrgm:
### Describe the bug

Can't override a property returned by `splitProps`.

### Your Example Website or App

https://playground.solidjs.com/?version=1.6.0#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr3iMA5nAAK8wmSy8yhDbndaTylZeUUVOg1tPQsrKzgKWz4ZeghmXDRaCF4AYQYIXHEACgxAz0R7ZnVuAH4KslxGDCcpAEp+K15eKogG3mBff0KRMrIAXV4AXh8-NACPMhKIUe9gFSq1bhUx1sts3n19XlwAC3lDMmUwAFU0qHonE74AFSI4AFFGeUYKnKgICC0PhCMQyDBwXiEDziAgVDZbMCdbpZPqOFwjBZTBziFyjIqDObDFYdfZdeE-K4glRYJGSXZIoS4JjZAA8AD4QGi4BigmQAHTkyQs-RsvaSOIQISiYpFdqTNm8Fl5VKFCQi7wiWjMejwAp8ly4d5qOC63AAIXwAEkREUVFBCIQVK0AIT0zCSMZAA

### Steps to Reproduce the Bug or Issue

```ts
import { render } from "solid-js/web";
import { mergeProps, splitProps } from "solid-js";

export function Counter(inProps: { color?: string }) {
  const [splittedProps] = splitProps(inProps, ["color"]);
  // throws "Uncaught TypeError: Cannot redefine property: color"
  const mergedProps = mergeProps(splittedProps, {
    color: "red",
  });
  return <>{mergedProps.color}</>;
}

render(() => <Counter />, document.getElementById("app")!);
```

### Expected behavior

[...]

PR #1305, opened 2022-10-23
splitProps - make prop configurable by default (fix:#1301)

LiQuidProQuo:
## Summary
fix regression with the non-proxied implementation. (Introduced in 1.6)

consistency between when the props is provided or not.
https://playground.solidjs.com/?hash=-720374730&version=1.6.0

will also address the case of `splitProps` `mergeProps` #1301 
when working with default props setup.


ryansolid:
Thanks

PR #1307, opened 2022-10-23
DEV: Add a way to observe stores

thetarnav:
- ~~Adds a dev-only ability to observe stores by adding a new `$ON_UPDATE` symbol.~~
- the dev APIs are bundled into a `DEV` object, as in the core module (not sure if it should be called `STORE_DEV` to avoid conflicts, but it's probably whatever)
- Some typings got cleaned up too (nothing in public APIs though)
- Added a jsdoc comment to `unwrap` because why not

~~Draft because I still want to confirm if this is enough, and maybe add some unit tests.~~

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -35,8 +35,8 @@ let rootCount = 0;
 const [transPending, setTransPending] = /*@__PURE__*/ createSignal(false);
 
 declare global {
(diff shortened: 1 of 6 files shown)


thetarnav:
Alright, this seems to be ready.
I've changed the API to use a global listener, similar to the existing afterUpdate and afterRootCreate hooks.
This will allow for more flexibility and storing updates without pre-patching anything.

Issue #1309, opened 2022-10-24
Resource's error won't be set if used in createRenderEffect


w1ndy:
### Describe the bug

If a resource is used in createRenderEffect, it's error will not be set if the resource fetcher throws:
```javascript
const [res] = createResource(async () => {
  throw 'test'
})
createRenderEffect(() => console.log(res()))
```

Wondering if this is an issue or an expected behavior... Thanks!

### Your Example Website or App

https://playground.solidjs.com/?version=1.4.1#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr2ZCouOACVhYxgFEZMuM1wsBycXdzomVilZeUUVOg1tPQsrKxl6CAC0WgheAGEGCBdGAAoASn4rXgdssj5gITIAXV4AXmC4Zzc4cJY4YqgyfAzeMtaAPgqcqt5cAAt5Q14AchdapcqpUo3HDtCPcR8-AOLRlonmGvU4ADo1WgBzYoay0q2poVwmHOKNqoAeERoABuY1+0z+AGVDGhcMxZqCptNwQBZZxw3iGWbCFoCBq3WhQQEQe5SMZ3QkYe5-fSo2HwsFIv609GY7G47rXcTyCSSMbssicxjcqTU5n0xHg-RQmFwhGM-SAkEbN6SFIQISicQncpnXh-fLpIq8fRjIIiWjMejwQrXe5wXBeNRwa24ABC+AAkiIfmAoIRCCpSgBCUqWTCSRpAA

### Steps to Reproduce the Bug or Issue

Run the example

### Expected behavior

The result panel should show "test" instead of "loading"

### Platform

- OS: macOS
- Browser: Firefox
- Version: 106.0.1

mdynnl:
`Switch` operates top-down and currently `loading` is not set to `false` on resource error (maybe it should) so either `error` needs to be read first or be caught by `ErrorBoundary` wrapping the component reading the resource

ryansolid:
Yeah loading should be set to false I think. We should fix that.

w1ndy:
> `Switch` operates top-down and currently `loading` is not set to `false` on resource error (maybe it should) so either `error` needs to be read first or be caught by `ErrorBoundary` wrapping the component reading the resource

[...]

PR #1315, opened 2022-10-28
fix resource state

mdynnl:
fixes https://github.com/solidjs/solid/issues/1309


ryansolid:
Looking at the fix I'm worried something more serious is going on. These are in a `runUpdate` batch so they should run together and the order not matter. @mdynnl can you speak to this at all?

mdynnl:
The write order actually affects the order updates run. It wouldn't matter if `res` is accessed before `state` (`res.loading`) in the same computation stopping further execution. But if `state` also has other computations that doesn't depend on `res`, this would allow them to run first.

Issue #1316, opened 2022-10-28
Solid 1.6 mergeProps results in undefined entries


martinpengellyphillips:
### Describe the bug

Updated to Solid 1.6 and found that Solid Select now [breaks](https://github.com/thisbeyond/solid-select/issues/28) when relying on default options. It seems `mergeProps` is the cause.

For this code:

```jsx

const createSelect = (props: CreateSelectProps) => {
  const config = mergeProps(
    {
      multiple: false,
      disabled: false,
      optionToValue: (option: Option): SingleValue => option,
      isOptionDisabled: (option: Option) => false,
    },
    props
  );
...
```

In 1.5, `config` would have the defaults for `isOptionDisabled` etc correctly defined. In 1.6 they are now all `undefined` resulting in breaking logic later.

```
CONFIG: 
Object { options: (5) […], … }
   isOptionDisabled: undefined   // <---- should be defined default function
​   options: Array(5) [ "apple", "banana", "pear", … ]
   ...
```

Note that `props` passed to `createSelect` came from this code in the `Select` component:

```jsx

const Select: Component<SelectProps> = (props) => {
  const [selectProps, local] = splitProps(
    mergeProps(
      {
        format: ((data, type) => data) as CommonProps["format"],
        placeholder: "Select...",
        readonly: typeof props.options !== "function",
        loading: false,
        loadingPlaceholder: "Loading...",
        emptyPlaceholder: "No options",
      },
      props
    ),
    [
      "options",
      "optionToValue",
      "isOptionDisabled",
      "multiple",
      "disabled",
      "onInput",
      "onChange",
      "onBlur",
      "onFocus",
    ]
  );
  const select = createSelect(selectProps);
```

### Your Example Website or App

https://github.com/thisbeyond/solid-select/issues/28

### Steps to Reproduce the Bug or Issue

1. Install Solid Select with Solid 1.6
2. Create a plain `<Select options={['a', 'b', 'c']}/>`
3. Attempt to interact with the select by clicking on it. Error occurs.

### Expected behavior

No error.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

mdynnl:
workaround <https://playground.solidjs.com/?hash=-1845137203&version=1.6>

LiQuidProQuo:
in a way this looks like it is as "intended", `SplitProps`, returns all the requests keys, even if they don't exist
[...]

PR #1317, opened 2022-10-28
mergeProps - non proxy sources lookup for better competability with t…

LiQuidProQuo:
A. SplitProps avoid getter/setter for non existing props at time of call. 
 - assumes props are static and cannot get updated after the split
 - may fix: https://github.com/solidjs/solid/issues/1316
 - https://github.com/solidjs/solid/pull/1317/files#diff-2dfe8baae4d70b8ddc9b85568d1fbbb2b9626d9fe11154d0c4f902d43fee5d1eR314

-----

B. mergeProps2 
proof of concept - sources lookup for the non-proxied version 
assumes props are dynamic. 
https://github.com/solidjs/solid/pull/1317/files#diff-2dfe8baae4d70b8ddc9b85568d1fbbb2b9626d9fe11154d0c4f902d43fee5d1eR209-R252

A and B are independent and may solve different issues or create new ones.

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -178,6 +178,81 @@ function resolveSource(s: any) {
   return (s = typeof s === "function" ? s() : s) == null ? {} : s;
 }
 
+
+function mergeProps2(...sources) {
+  if (sources.some((s) => s && ($PROXY in s || typeof s === "function"))) {
+    return new Proxy({
+      get(property) {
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks for the inspiration.

ryansolid:
I took the gist of this with the latest implementation. I am not stoked on how much code I needed to add but it's probably best to take this fairly generic approach to merging where we don't worry about being configurable etc or any splitProps/mergeProps specific interplay.

PR #1320, opened 2022-10-30
splitProps - mark undefined keys in non proxy splitting.

LiQuidProQuo:
mergeProps - skip undefined keys from splitProps.

- non proxy only. 
- splitProps, will track keys that were specified, but were not defined in the object, when splitting. 
- `mergeProps`, skip merging undefined marked keys (returned by `splitProps`)

alternative fix(pr): https://github.com/solidjs/solid/pull/1317

may fix(issue): https://github.com/solidjs/solid/issues/1316

only tested in patched playground

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -178,6 +178,9 @@ function resolveSource(s: any) {
   return (s = typeof s === "function" ? s() : s) == null ? {} : s;
 }
 
+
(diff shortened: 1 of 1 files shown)


ryansolid:
This is PR is out of date since I went with a different fix. I think we go with that fix or submit a new PR that reconciles against that. Thanks.

Issue #1321, opened 2022-10-30
Spread operator custom function called multiple times


jackzard:
### Describe the bug

when call spread with custom function, the function is called multiple times, when using version before 1.6.0 it's only run once.

### Your Example Website or App

https://playground.solidjs.com/?version=1.5.9#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQiM4EACZxGAX14AzRrRi8AOjVoAbNMIC0AKzIB6AO5wARkoUQOXHv3iMA5nAAKswmUky5i5Ws07T5ztx8IMyCULhwAMq43HBusvJKdN7aemTRgn4WgfwOgVAqcR4A5EnqKQbGRWZmUvQQzLhotBC8AMIMEOGMABQAlPxmvEO8zM1pvMBpMQC6vAC8I6HhUTHdwSpQZGQAMmhpiPxQB7iM9LHivYPDoxDjtEZa8wMtw69khKHCB33zAHzPr0B1zGqjgADoVLRbN0ilIWsx8io4MIipcXkDXoJcEwWjZ7E5aC5ulNBFgARiKYIpAc4H8RiCkRCod04GiKa8LlcgeIubweRBeVicbxurzXgAeYRoABu-DB8vuWjB70+fXEul+vLR-LMghEYm6Pzm-3F7TqXQ1ZOEtGY9HgnTB9lwAFEkfbcAAhfAASWEorAUEIhCUvQAhGiwOJpkA

### Steps to Reproduce the Bug or Issue

1. Visit playground
2. Check console

### Expected behavior

expected function only run once

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

Brendan-csel:
Can confirm we're seeing this too ...with our `Box` component that is used *everywhere*.

lxsmnsyc:
before 1.6.0 I think similar behavior also happened (since this was already reported before).

Brendan-csel:
Here is a more mimimal reproduction

For 1.5.9 https://playground.solidjs.com/?hash=237697647&version=1.5.9

For 1.6.1 https://playground.solidjs.com/?hash=2093492072&version=1.6.1

lxsmnsyc:
> Here is a more mimimal reproduction
> 
[...]

PR #1326, opened 2022-11-01
mergeProps wrap functional sources in a memo. #1321

LiQuidProQuo:
wrap functional sources, provided to `mergeProps()` in a memo. 

```js
<div {...functionalSource} />
```

prevent "over calling" as shown in #1321, sources get evaluated 4 times on mounting alone.

The memo guards the result of the function, and ensure user provided function is only called once, until a reactive trigger.

Downsides:
1. create a memo for every function source (more memory, overhead)
2. double wrapping:  if a source function is already a memo it will be wrapped too. (solid does not provide a way to introspect its primitives, to prevent that)
 

fixes: #1321

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -179,6 +179,7 @@ function resolveSource(s: any) {
 }
 
 export function mergeProps<T extends unknown[]>(...sources: T): MergeProps<T> {
+  sources = sources.map(source => typeof source === "function" ? createMemo(source) : source)
   if (sources.some(s => s && ($PROXY in (s as T) || typeof s === "function"))) {
     return new Proxy(
       {


lxsmnsyc:
Personally I would prefer this to have been part of the compiler so that it can immediately decide when to wrap memo or not

LiQuidProQuo:
performance wise, adding memo is not significantly slower (for a very basic test case)

per run 
 - 10k elements, 
 - 4 attributes `[ {}, {}, store, ()=>store]`  (at least one function)

,
https://6360c92c9708da584168e820--majestic-dodol-3f071b.netlify.app/?version=1.6.1

image

ryansolid:
> Personally I would prefer this to have been part of the compiler so that it can immediately decide when to wrap memo or not

Good point slightly larger code gen but we skip iterating over things again. I agree compiler probably makes more sense.

LiQuidProQuo:
compiler solution will be interesting, though, if transforming to memo in runtime has no impact of performance. and lower code gen. what is the advantage of compiler solution.

packages/solid/src/render/component.ts
```diff
@@ -179,6 +179,7 @@ function resolveSource(s: any) {
 }
 
 export function mergeProps<T extends unknown[]>(...sources: T): MergeProps<T> {
+  sources = sources.map(source => typeof source === "function" ? createMemo(source) : source) as T
```
LiQuidProQuo, line 182: we can actually move this into the proxy branch.  

```js
  if (sources.some(s => s && ($PROXY in (s as T) || typeof s === "function"))) {
     __HERE__
    return new Proxy(
```

LiQuidProQuo, line 182: alternatively, we can merge into a single loop, 
which will detect and transform in one go(full scan). 
```js
export function mergeProps(...sources) {
  let proxy = 0;
  for (let i = 0; i < sources.length; i++) {
    const s = sources[i];
    proxy ||= s && $PROXY in s;
    sources[i] = s === 'function' ? ((proxy = 1), createMemo(s)) : s;
  }

  if (proxy) {
    return new Proxy(
      {
        get(property) {
          for (let i = sources.length - 1; i >= 0; i--) {
            let v = resolveSource(sources[i])[property];
            if (v !== void 0) return v;
          }
        },
```

ryansolid: Yeah probably this. Some needs iterate everything on a miss anyway.

LiQuidProQuo: This PR was merged, and I was about to make another one with the optimized loop as in the comments above.

and I see that @ryansolid  took care of that   👍 
https://github.com/solidjs/solid/commit/a42a5f6e7150a981a979f84c2745a4a4ba03f6ec

LiQuidProQuo: we should look into lazy memo for this case

Issue #1323, opened 2022-10-31
untrack does not restore listener when callback throws an exception


LiQuidProQuo:
### Describe the bug

```js
  createEffect(() => {
    try {
      untrack(() => {
        throw new Error();
      });
    } catch (e) {}
    console.log(count()); // this will not be reactive, listener is still null internally
  });
```

https://github.com/solidjs/solid/blob/454ed51f139219630a80d60570ed455a9b0fdc41/packages/solid/src/reactive/signal.ts#L807-L816

### Your Example Website or App

https://playground.solidjs.com/?hash=1765991341&version=1.4.1

### Steps to Reproduce the Bug or Issue

see repro

### Expected behavior

untrack should also restore listener when throwing.

### Platform

playground

reported by: fabiospampinato

ryansolid:
Yeah we should atleast try/finally any end user provided code.

ryansolid:
Fixed by #1335

PR #1335, opened 2022-11-04
improve error handling in `untrack`

mdynnl:
addresses https://github.com/solidjs/solid/issues/1323


mdynnl:
what about https://github.com/solidjs/solid/blob/763adf442c898adc4e6d6ae4a94e840338f17e7c/packages/solid/src/reactive/signal.ts#L971-L1000

ryansolid:
That one's probably fine. If the Transition is running it means it is running in a runUpdate further up. The problem with untrack is it messes with Listener/Owner and misses restoring it if it throws. Where this is handled higher up.

PR #1343, opened 2022-11-08
refactor: tidy up solid/bench/libraries

Simon-He95:


ryansolid:
I see what you did and some of these changes are probably merited. But the idea here is the code is the code as is from some of these libraries. Since I'm not going to benchmark every little change I'm inclined not touch this code and assume it is as intended.

PR #1368, opened 2022-11-17
refactor: tidy up reactive

Simon-He95:

### DIFF
--- a/packages/solid/src/reactive/scheduler.ts
+++ b/packages/solid/src/reactive/scheduler.ts
@@ -49,7 +49,9 @@ function setupScheduler() {
   };
 
   if (
-    (navigator as NavigatorScheduling)?.scheduling?.isInputPending
+    navigator &&
+    (navigator as NavigatorScheduling).scheduling &&
+    (navigator as NavigatorScheduling).scheduling.isInputPending
   ) {
     const scheduling = (navigator as NavigatorScheduling).scheduling;
     shouldYieldToHost = () => {
@@ -103,7 +105,7 @@ export function requestCallback(fn: () => void, options?: { timeout: number }): 
   let startTime = performance.now(),
     timeout = maxSigned31BitInt;
 
-  if (options?.timeout) timeout = options.timeout;
+  if (options && options.timeout) timeout = options.timeout;
 
   const newTask: Task = {
     id: taskIdCounter++,
--- a/packages/solid/src/reactive/scheduler.ts
+++ b/packages/solid/src/reactive/scheduler.ts
@@ -49,9 +49,7 @@ function setupScheduler() {
   };
 
   if (
-    navigator &&
-    (navigator as NavigatorScheduling).scheduling &&
-    (navigator as NavigatorScheduling).scheduling.isInputPending
+    (navigator as NavigatorScheduling)?.scheduling?.isInputPending
   ) {
     const scheduling = (navigator as NavigatorScheduling).scheduling;
     shouldYieldToHost = () => {
@@ -105,7 +103,7 @@ export function requestCallback(fn: () => void, options?: { timeout: number }): 
   let startTime = performance.now(),
     timeout = maxSigned31BitInt;
 
-  if (options && options.timeout) timeout = options.timeout;
+  if (options?.timeout) timeout = options.timeout;
 
   const newTask: Task = {
     id: taskIdCounter++,
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -240,7 +240,7 @@ export function createComputed<Next, Init>(
   options?: EffectOptions
 ): void {
   const c = createComputation(fn, value!, true, STALE, "_SOLID_DEV_" ? options : undefined);
-  if (Scheduler && && Transition && Transition.running) Updates!.push(c);
+  if (Scheduler && Transition && Transition.running) Updates!.push(c);
   else updateComputation(c);
 }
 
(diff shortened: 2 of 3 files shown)


packages/solid/src/reactive/scheduler.ts
```diff
@@ -49,9 +49,7 @@ function setupScheduler() {
   };
 
   if (
-    navigator &&
-    (navigator as NavigatorScheduling).scheduling &&
-    (navigator as NavigatorScheduling).scheduling.isInputPending
+    (navigator as NavigatorScheduling)?.scheduling?.isInputPending
```
lxsmnsyc, line 52: Optional chaining is actually verbose when compiled down. I don't think it's worth the change.

Simon-He95: Optional chaining is to help you reduce the judgment on the code, and you can do some code refactoring better. why Optional chaining is verbose when compiled down?

lxsmnsyc: >  why Optional chaining is verbose when compiled down?
Compilers does multiple steps but the code above is basically compiled as:

```js
var x, y;

if (((x = navigator.scheduling) == null ? void 0 : ((y = x.isInputPending) == null ? void 0 : y)))
```

LiQuidProQuo: "compile down", means for when you make a build that targets a lower version of JS
for example ES2019 which does not support optional chaining will compile this.
```js
(navigator as NavigatorScheduling)?.scheduling?.isInputPending
```
to this:
```js
var _a;
(_a = navigator === null || navigator === void 0 ? void 0 : navigator.scheduling) === null || _a === void 0 ? void 0 : _a.isInputPending;

```

if you target a higher version of JS, then optional chaining might be the most compact.

lxsmnsyc: > if you target a higher version of JS, then optional chaining might be the most compact.
Although that's highly unlikely so in which case the previous code is actually much more build-friendly.

maxmilton:
Signals and the scheduler are critical to performance so it would be nice to see the impact of these changes. Given your use of the optional chaining operator, I would expect it to have a negative impact on runtime performance. A benchmarking suite you could use is <https://github.com/krausest/js-framework-benchmark>.

packages/solid/src/reactive/scheduler.ts
Simon-He95: thanks for your explanation :)

packages/solid/src/reactive/signal.ts
```diff
@@ -698,7 +698,7 @@ export interface DeferredOptions<T> {
  */
 export function createDeferred<T>(source: Accessor<T>, options?: DeferredOptions<T>) {
   let t: Task,
-    timeout = options ? options.timeoutMs : undefined;
+    timeout = options && options.timeoutMs;
   const node = createComputation(
```
atk, line 702: This change will lead to a different behavior, as timeout is subsequently checked if it is `undefined`, but here, it would be `false`.

atk (changes requested):

packages/solid/src/reactive/signal.ts
```diff
@@ -752,14 +752,13 @@ export function createSelector<T, U>(
...
-          for (const c of val.values()) {
-            c.state = STALE;
-            if (c.pure) Updates!.push(c);
-            else Effects!.push(c);
-          }
+      for (const [key, val] of subs.entries()) {
+        if (fn(key, v) === fn(key, p!)) continue
+        for (const c of val.values()) {
+          c.state = STALE;
+          c.pure ? Updates!.push(c) : Effects!.push(c)
```
atk, line 759: I would argue that this is less readable and debuggable. Also, the minification will result in the same output anyways.

packages/solid/src/reactive/signal.ts
Simon-He95: Thanks, I reset

packages/solid/src/reactive/signal.ts
```diff
@@ -1327,7 +1325,7 @@ function runComputation(node: Computation<any>, value: any, time: number) {
   try {
     nextValue = node.fn(value);
   } catch (err) {
-    if (node.pure) Transition && Transition.running ? (node.tState = STALE) : (node.state = STALE);
+    if (node.pure) node[(Transition && Transition.running) ? 'tState' : 'state'] = STALE;
```
ryansolid, line 1328: Most of Transition checks are setup in a way that bundlers like rollup do dead code elimination if people don't use the feature. So we need to verify these still work.  A lot of the conditionals that opt into features, things like dev mode are laid out in a way to consider what the code looks like when the branches are completely removed by the bundler.

I don't know that string array accessors vs property accessors have much performance overhead since engine optimizations are pretty good. But we probably should investigate if we want to do this sort of change.

ryansolid:
A few overall notes in addition. Probably worth running things through Prettier. A lot of the spacing stuff resets when you do that.  We don't use optional chaining or null coalescing at this point for platform compatibility. That may change in the future but it breaks stuff last I checked.

The comments above are correct. `signals.ts` is probably the most critical file in the whole project so we are really careful with that one. So it is very difficult for me to merge any changes that I haven't thoroughly tested for performance.

ryansolid:
Yeah the likelihood I micro-test each of these changes is unlikely in 1.x lifespan. The code is already moving away from here. I appreciate the effort but I'm going to close this one.

PR #1393, opened 2022-12-04
feat: add jsx-dev-runtime

manucorporat:
## Summary

JSX-automatic is missing the jsx-dev-runtime, esencially breaking solid when using jsx-runtime in development mode. External tooling will expect this file to exisit.


ryansolid:
thanks

PR #1403, opened 2022-12-08
fix mergeProps breaks with function source

orenelbaum:
## Summary
When passing a function source into `mergeProps`, it broke it due to a change in v1.6.3. I fixed it.
Solves this issue https://github.com/solidjs/solid-start/issues/523

I tested on a reproduction. I didn't write a test since there are no tests set up for ssr.


ryansolid:
Small oversight.. thank you

PR #1411, opened 2022-12-10
Export SignalOptions

kcirtaptrick:
## Summary

Make this available will make it easier to create custom signals that pass options to createSignal. There is currently no way to access this as a generic with NodeNext module resolution.


ryansolid:
Thank you

PR #1430, opened 2022-12-16
fix: delete lazy contexts one by one as they are completed

katywings:
## Summary

When one lazy component is rendered multiple times in parallel, suspense will wait until every load is finished. The fixed bug was, that the first notifySuspense call cleared contexts of still ongoing loadings, which never got a chance to finish. This resulted in a `renderToString timeout` error during the first server render.

## Reproduction

Bildschirmfoto vom 2022-12-13 22-16-36

1. Generate a new solid-start v0.2.7 project 
2. Insert the bottom code in `src/routes/index.tsx`
3. `pnpm build && pnpm start`
4. Open localhost:3000
5. Check the server log

Code:
```tsx
import { For, lazy } from "solid-js";

const LazyCounter = lazy(async() => import('~/components/Counter'))

export default function Home() {
  return (
    <main>
      <For each={[1,2,3]}>
        {() => {
          return <LazyCounter />
        }}
      </For>
    </main>
  );
}
```

- Essentially I patched my solid-start projects with the fix and they worked afterwards. The `renderToString timeout` went away with the fix and the first request rendered properly. Also the test spec at the bottom only succeeds with the fix.
- Existing solid tests also passed.

## Test spec

After some experimentation I came up with the following test which functionally is working correctly. But I am not very happy with it, imho the test setup is too large and I had to make many assumptions which should not be in a test. Let me know if you still want to include it in the code, or if you have better ideas how to write the test :).

```ts
import { createComponent, createRoot, lazy, sharedConfig, Suspense } from "../src/server";

describe("Server rendering a lazy component twice in parallel", () => {
  test("should not indefinitely time out", async () => {
    const render = new Promise(res => {
      createRoot(() => {
        let finishedLoadingCount = 0;
        let suspenseResult: string | undefined = undefined;
        sharedConfig.context = {
          id: "",
          async: true,
          registerFragment: id => result => {
            suspenseResult = result;
            return true;
          },
          suspense: {},
          writeResource: () => {},
          block: async p => {
            await p;
            finishedLoadingCount++;
            if (finishedLoadingCount === 2) {
              res(suspenseResult);
            }
          }
        };
        const Comp = lazy(
          async () =>
            await new Promise(res => setTimeout(() => res({ default: props => props.number }), 100))
        );
        createComponent(Suspense, {
          children: () => [
            createComponent(Comp, { number: 1 }),
            createComponent(Comp, { number: 2 })
          ]
        });
      });
    });

    const result = await render;

    // Result would be undefined, if the suspense didnt finish properly
    expect(result).toBe("12");

    sharedConfig.context = undefined;
  });
});
```

### DIFF
--- a/.changeset/thick-ways-hope.md
+++ b/.changeset/thick-ways-hope.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix: delete lazy contexts one by one as they are completed
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -496,9 +496,12 @@ function suspenseComplete(c: SuspenseContextType) {
 
 function notifySuspense(contexts: Set<SuspenseContextType>) {
   for (const c of contexts) {
-    if (suspenseComplete(c)) c.completed();
+    if (!suspenseComplete(c)) {
+      continue;
+    }
+    c.completed();
+    contexts.delete(c);
   }
-  contexts.clear();
 }
 
 export function enableScheduling() {}


ryansolid:
Thank you

Issue #1359, opened 2022-11-13
Store setter type signature allows setting readonly properties


JorrenH:
### Describe the bug

While creating a store using a getter property I encountered a runtime error trying to update the store. For example when I define the following store:

```ts
const [ store, setStore ] = createStore({
    get value() { return 'value' },
});

setStore('value', 'modified');
```

I would expect a type error on the last line, instead it results in a runtime error:
`Uncaught TypeError: setting getter-only property "value"`

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-9qbvhq?file=src/App.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the linked [stackblitz](https://stackblitz.com/edit/solidjs-templates-9qbvhq?file=src/App.tsx).
2. The app will compile
3. At runtime it will throw an error

### Expected behavior

As a user, the given code should produce a compile time type-error. Instead you get a runtime error.

### Platform

- OS: Windows (but affects others)
- Browser: Firefox (but affects others)
- Version: 106.0.5

ryansolid:
Yeah makes sense. Honestly Store types are so complicated I have no idea how to even start approaching this one. But very much open to someone figuring it out.

otonashixav:
I did start on this some time ago, though it breaks the type test cases that use generics. As such I don't intend to make a PR for this. Maybe it will be of help to someone else though.

https://github.com/otonashixav/solid/commit/72bff0da65fe19ec71cc923b29080b5fb14ffe97

JorrenH:
Solved in #1444

PR #1444, opened 2022-12-24
Typescript: disallow setting readonly store props

JorrenH:
## Summary

Resolves (part of) the issue posed in #1359. Limits store setters to non-readonly properties in the store. This fix only addresses usage of the store setter by path (`setStore("a", "b", value)`). Setting by object (`setStore({ a: { b: value }})`) cannot be checked without impact.

If we would type check setting by object, setters like `setStore("el", {} as Element)` would stop working because the `Element` interface contains readonly props.

I added a test for a store setter by path and two TODO's which address the remaining issue when setting by object.

```bash
packages/solid $ pnpm test
> Test Suites: 23 passed, 23 total
> Tests:       378 passed, 378 total
> Snapshots:   0 total
> Time:        14.132 s
> Ran all test suites.
```

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -380,20 +380,22 @@ type KeyOf<T> = number extends keyof T // have to check this otherwise ts won't 
 
 type MutableKeyOf<T> = KeyOf<T> & keyof PickMutable<T>;
 
-type Rest<T, U extends PropertyKey[], K extends KeyOf<T> = KeyOf<T>> =
-  K extends keyof PickMutable<T>
+// rest must specify at least one (additional) key, followed by a StoreSetter if the key is mutable.
+type Rest<
+  T,
+  U extends PropertyKey[],
+  K extends KeyOf<T> = KeyOf<T> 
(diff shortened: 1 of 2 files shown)


otonashixav:
This wrongly disallows e.g. `setStore("readonlyProp", { mutableProp: "value" })`. In the context of property getters it doesn't make sense to try and modify the return value of a getter, but it should be allowed in the case where it isn't a getter (but is still typed `readonly`).

I'm not sure if that would be enough reason to not merge though.

JorrenH:
I updated the type-checking to be done for the store keys instead of the StoreSetter. This approach results in better developer feedback on the readonly key. I also added tests for every interface of the SetStoreFunction.

@otonashixav Fair point, and I agree with your argument. Conceptually using the merge syntax on a readonly prop doesn't make much sense to me, but it's a limitation to consider.

otonashixav (approved):
If the issue mentioned in my previous comment is acceptable, I am okay with these changes.

ryansolid:
Thanks

Issue #1446, opened 2022-12-25
`splitProps` is broken in 1.6.6


juanrgm:
### Describe the bug

`splitProps` returns undefined properties in SolidJS v1.6.6 (https://stackblitz.com/edit/angular-aejisc-h1nlkh?file=src%2FApp.tsx).

The same example it works in SolidJS v1.6.5 (https://stackblitz.com/edit/angular-aejisc-czn2cd?file=src%2FApp.tsx).

### Your Example Website or App

https://stackblitz.com/edit/angular-aejisc-h1nlkh?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

```ts
import { mergeProps, splitProps } from 'solid-js';

export default function App() {
  const inProps: { a?: any; b?: any } = mergeProps(() => ({}));
  const [props] = splitProps(inProps, ['a', 'b']);
  const keys = Object.keys(props);
  return <div>{JSON.stringify(keys)}</div>;
}
```

renders

`['a', 'b']`

### Expected behavior

`[]`

### Platform

- OS: Windows
- Browser: Chrome 108

orenelbaum:
I think that the only way to solve this is by turning `splitProps` into a proxy

PR #1447, opened 2022-12-26
make splitProps with dynamic source return proxies

orenelbaum:
- fixes https://github.com/solidjs/solid/issues/1446


ryansolid:
Thanks

PR #1454, opened 2023-01-02
Fix `For` and `Index` type definition (fixes #1453)

### LINKED ISSUE #1453: Array type accepted by `<For/>` is too narrow
### Describe the bug

Currently, the type for `<For/>` looks like this:

```ts
declare function For<T, U extends JSX.Element>(props: {
  each: readonly T[] | undefined | null | false;
  fallback?: JSX.Element;
  children: (item: T, index: Accessor<number>) => U;
}): Accessor<U[]>;
```

However, this means `each` can't accept types like `A[] | B[]` even when it extends an array.

https://playground.solidjs.com/anonymous/f1046d10-9e07-421d-a883-2de0567e44e1

check error

### Expected behavior

No errors occur. `val` has type `string | number` like how `forEach()` works.

Possible solution:
https://playground.solidjs.com/anonymous/78fe7041-1915-4df9-8588-6a1bdf01a857

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Browser: All
- Version: latest

### Additional context

https://discord.com/channels/722131463138705510/722131463889223772/1059366511187349504

thetarnav:
Fixes #1453 

In playground:
https://playground.solidjs.com/anonymous/3747093e-d640-43d7-a386-c9bc21679447

### DIFF
--- a/packages/solid/src/render/flow.ts
+++ b/packages/solid/src/render/flow.ts
@@ -26,11 +26,11 @@ import type { JSX } from "../jsx.js";
  *
  * @description https://www.solidjs.com/docs/latest/api#%3Cfor%3E
  */
-export function For<T, U extends JSX.Element>(props: {
-  each: readonly T[] | undefined | null | false;
+export function For<T extends readonly any[], U extends JSX.Element>(props: {
+  each: T | undefined | null | false;
   fallback?: JSX.Element;
-  children: (item: T, index: Accessor<number>) => U;
-}) {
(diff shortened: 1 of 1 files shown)


otonashixav:
If this can be fixed upstream it would be ideal, considering this bug also affects the types in typescript's definitions: https://tsplay.dev/WG40Kw.

ryansolid:
So is this our responsibility to fix? Feels like it's a TypeScript-ism. Only asking because I don't know if we should set precedence for this sort of thing.

PR #1456, opened 2023-01-02
Fix unowned roots having owner in dev

thetarnav:
Suppose you don't provide a callback parameter for the dispose function when using `createRoot`. In that case, it should create an "unowned" (top-level) root that has no owner so there is no way to get the context from the owner it was theoretically called under.
But in development, a new owner is created every time. The UNOWNED owner is never used.

[So in dev it'll have access to context](https://playground.solidjs.com/anonymous/1ae21915-a79b-4177-9c0a-035d15fea029)

[While in prod it doesn't](https://solid-playground.netlify.app/?version=1.6.2#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr2ZCouOAGFaEFxVxYHTlwBKtLQ+vPSejFDMANa+9GRuHl580nIKyqoa2noWVlYy4cy4aB68ALLoEAAUAJT8VrwOHmR8rrgUvAC8fnDOiZ5w3gA8UBD4AHxVKsxJg3wyUGpqJlHRKjWWEA2CcLhMELxTB428Q20UAHQACvIAbprivLeL9HCdKmmKMwPeKuPbJ344VwkRiVVqXXG9WOgMa3zoajgFzUtAA5lMwABGC4qOIJdw-XBVc41DZ5GGwxy9QLBIlVAD6dU6UJAANhcOa6iRKPRKgATDjsGF8bNvMT2qTNuzGpIyVsKYCqX0giFwUyWWz2fCuci0RiAMyCvH9ZLiiiSzUnWWbS2NIR7RgHFS5BVSGqSNlDfTna53B6Mf7HOUe+VCUTiNWQ04VDC8fTjXwiWjMejwTwXVG7ACiiLTuAAQvgAJIiDFQQiEdYAQjlYEkAF0gA) (browser console needs to be opened)

### DIFF
--- a/.changeset/khaki-ducks-yawn.md
+++ b/.changeset/khaki-ducks-yawn.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix unowned roots having owner in dev
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -108,10 +108,11 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: Owner): T {
   const listener = Listener,
     owner = Owner,
     unowned = fn.length === 0,
-    root: Owner =
-      unowned && !"_SOLID_DEV_"
-        ? UNOWNED
-        : { owned: null, cleanups: null, context: null, owner: detachedOwner || owner },
+    root: Owner = unowned
+      ? "_SOLID_DEV_"
+        ? { owned: null, cleanups: null, context: null, owner: null }
+        : UNOWNED
+      : { owned: null, cleanups: null, context: null, owner: detachedOwner || owner },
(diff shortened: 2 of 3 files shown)


ryansolid:
We should look when this was changed. I'm positive there was a very specific reason for this I just can't remember what it was.

thetarnav:
This was charged 2 years ago when the dev version of the runtime was initially added.
So it might just have been a mistake.

ryansolid:
Yeah I was misreading it. This was an oversight. I don't use UNOWNED probably for HMR but it shouldn't have an owner

Issue #1322, opened 2022-10-31
DEV build seems to have O(n^2) in registerGraph name resolution for createSignal in onMount when using signals from props


tuukkapuranen:
### Describe the bug

When invoking `createSignal` multiple times in `onMount`, there is a noticeable performance degradation in DEV build, when certain conditions are met. Performance recording call tree indicates that the `registerGraph` function is taking majority of the time.

The `registerGraph` function creates for a unique name for the signal in DEV builds, and does this in a loop, so I presume the issue is that in some cases the same name is given to the `registerGraph` for a large number of signals, and the loop has to try all of them every time a new signal is added.

### Your Example Website or App

https://stackblitz.com/edit/solid-vite-mmnbja?file=src/index.tsx

### Steps to Reproduce the Bug or Issue

1. Go to https://stackblitz.com/edit/solid-vite-mmnbja?file=src/index.tsx
2. Wait for component rendering
3. Observe the duration the `onMount` function took

### Expected behavior

Expected `onMount` duration is around 10ms, but actual `onMount` duration is around 3000ms.

### Screenshots or Videos

![Untitled picture](https://user-images.githubusercontent.com/5907626/198989148-d6fdc00b-f463-4286-a671-00285c0c7526.png)

### Platform

- OS: Windows
- Browser: Chrome, Edge

### Additional context

Note that this issue is only present in DEV build, as the `registerGraph` is not used in production.

There are several workarounds for this issue. The first seems to be not to use signals from `props` in `onMount`, as in https://stackblitz.com/edit/solid-vite-pgsggu?file=src/index.tsx which apparently alters the name generation for the signals in such a way that the issue disappears (in this example, removing the use of `props.rows` was enough). The second is to use `createSignal(false, { internal: true })`, which suppresses the name generation for the signal. The third option is to provide unique names for the signals in question, for example ``createSignal(false, { name: `selectedCells[${row}][${col}]` })``.

LiQuidProQuo:
For Dev when a signal name is generated it uses a rather expensive method(serialization) to generate a hash which  may also increase in complexity as the object grows

https://github.com/solidjs/solid/blob/454ed51f139219630a80d60570ed455a9b0fdc41/packages/solid/src/reactive/signal.ts#L1048-L1078

then if the hash already exists it will loop with an increment until a free name is found 
https://github.com/solidjs/solid/blob/454ed51f139219630a80d60570ed455a9b0fdc41/packages/solid/src/reactive/signal.ts#L1085


[...]

PR #1458, opened 2023-01-03
DEV: Minor additions and change the API of dev hooks

thetarnav:
### DevHooks

The main point o this PR is to change the API of dev hooks from using the `globalThis` namespace to storing them in a new `DEV.hooks` object that has to be explicitly imported.
The number of dev hooks will probably only increase and I don't like how they are polluting the global namespace with something that concerns almost none of the users.

image
Here I get autocomplete to `_$afterCreateRoot` hook, instead of `createRoot` in the playground...

So the behavior is the same, it just has to be imported.
```ts
import { DEV } from "solid-js"

if (DEV) {
  DEV.hooks.afterUpdate = () => {
    console.trace("after update")
  }
}
```

---

### afterCreateOwner

I've removed the previous `afterCreateRoot` hook, which enabled me to attach the debugger to any root in the app without the user needing to do anything, and replaced it with a more general `afterCreateOwner` which extends on `afterCreateRoot` to allow me to:
- monkey-patch computations right as they are created—which means that doing a flame chart of creation/execution time of computations as the app loads will be possible.
- listen to computation updates from the very beginning, instead of waiting for the next time, I walk the tree. This means that I won't miss any updates that could happen in that time window.
- capture effects or memos created outside of roots (which some seem to like doing for some reason, and this also seems like something semi-planned for 2.0)

---

### registerGraph `graph` property

Since signals have no concept of ownership, they have no owners. They are added to the `soureceMap` property if created synchronously under an owner, so there is a way to get signals from the owner, but there is no way to get an owner from a signal. Well, at least not without keeping a separate map after walking the tree.
So I've added a new `graph` property to the value object that is added to `sourceMap`. Which adds it to anything passed to `registerGraph`, signals included.


ryansolid:
Since this is mainly for you and the dev tools.. I will ask. This is breaking right? Is this a wait for 1.7 thing?

thetarnav:
Yup, it'll break existing devtools.
Changes to hooks can wait.
I can create a separate PR for adding the `graph` property to values, because this one I'll need for visualizing the dependency graph and is not breaking.

ryansolid:
We can get rid of this in 1.7 too right? https://github.com/solidjs/solid/issues/1322

thetarnav:
Sure, I can try to tackle this again.
I was waiting because there was a innitiative from @lxsmnsyc to achieve state preservation with HMR, and I didn't want to prevent it with my changes. But it seems that now he has doubts about the reliability of the approach.
My original plan was to just use a counter incremented id for signals/computations, so that it is unique rather then "regeneratable": https://github.com/solidjs/solid/pull/1255

thetarnav:
What's the timeline for 1.7 btw?

ryansolid:
I see.. Yeah I remember there being an issue. Didn't realize using what I had done was desired.

I have no timeline for 1.7 yet. Not quite sure what goes in yet. I'm sure there will be one as work on 2.0 has started and I'm sure we will learn stuff we want to get into at least one 1.x version.

thetarnav:
> Didn't realize using what I had done was desired.

I don't know if it's desired either, so I wanted to see the implementation to see what's necessary.

So if I'm to proceed with the change, what exactly can I remove?
Using `hashValue` produces names that aren't unique, too random to give any insight for the user and can take a lot of cpu time if a lot of them are queued.
As you know `serializeGraph` has no use because devtools would have their way of traversing the tree to get the data relevant to them.
If `serializeGraph` is gone there is no need to even have ids/names for each signal/computation at all as the only place they are used at is `serializeGraph`.
User-provided names would remain as they are.
The `sourceMap` property could just be an array of values (`SignalMapValue[]`) instead of a record.
So creating ids would be in toolings domain entirely.

But if `serializeGraph` and `hashValue` were to be kept exported, even if not used, all signals, owners, roots would need an id to be serializable. Which means there still needs to be a method of generating them. But generated ids should probably kept separate from user-provided names.

PR #1496, opened 2023-01-23
Add generic to onCleanup

thetarnav:
I think that the types for `onCleanup` could be better.
Since it returns the `fn` param as is, a generic could be used to keep the type of its return value.

```ts
const cleanup = () => 123

// before
onCleanup(cleanup) // T: () => void

// after
onCleanup(cleanup) // T: () => number
```

Probably has almost no value, but it's "more correct" 🤷‍♂️

But it could make it a little bit clearer that the returned function is the one passed in, and not a function to unsubscribe from the cleanup as it is also a common API pattern

### DIFF
--- a/.changeset/fast-ligers-argue.md
+++ b/.changeset/fast-ligers-argue.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Add generic to onCleanup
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks.

PR #1506, opened 2023-01-28
Export `EffectOptions` and `OnOptions` from main module

thetarnav:
These should probably be exported with the rest from the main module.


ryansolid:
Thanks

PR #1509, opened 2023-01-29
fix: `<Show keyed>` with a function child not keyed when the function does not have an argument

### LINKED ISSUE #1508: `<Show keyed>` with a function child is not keyed when the function does not have an argument.
### Describe the bug

it is expected that when passing `keyed` property to `<Show/>`
the component will be keyed, by the value of the `when` condition. 
and that the callback will be called every time the value changes.

however solid correctly "key" a callback **only**  if it has an argument.

**I work as expected**
https://playground.solidjs.com/anonymous/c1601c39-7e7f-4695-ad4f-6ce46b8bbb41
```js
<Show when={cond()} keyed> 
   { (i_have_an_argument) =><b>{Date.now()}</b> }
</Show>
```
**I don't work 😟**
https://playground.solidjs.com/anonymous/c9f3f381-b96c-479b-a1ec-9b43ef3d7338
```jsx
<Show when={cond()} keyed>
  { ()=><b>{ Date.now() }</b> }
</Show>
````

### Note (root cause):
<Show> need to support both a callback and a Signal Getter.
the design decision was, that a callback, **is a callback only if it has an argument**, 
if a function without an argument is provided it is assumed to be  **a reactive getter / signal.**  

this may look like magic to the average developer but this is as intended 

issue is likely going to be closed as  **wontfix**, and maybe a better  handling will come in  solid 2.0 which can introduce "breaking change" to fix this.

na

na

### Expected behavior

it is expected that in the future solid will find a way that when `keyed` is provided, it assumes that a callback is provided
when the children is a function, and that the function is not a signal.
and then always call the callback when the key value changes  (`when={key()}`)

### Screenshots or Videos

_No response_

### Platform

playground

### Additional context

_No response_

MoustaphaDev:
## Summary

Fix #1508

No tests added as I don't think this needs tests, but LMK if otherwise

### DIFF
--- a/.changeset/modern-readers-applaud.md
+++ b/.changeset/modern-readers-applaud.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix skipped rerun of a child function of `Show` component when it doesn't a parameter.
--- a/packages/solid/src/render/flow.ts
+++ b/packages/solid/src/render/flow.ts
@@ -108,7 +108,7 @@ export function Show<T>(props: {
       const c = condition();
       if (c) {
         const child = props.children;
-        const fn = typeof child === "function" && child.length > 0;
+        const fn = typeof child === "function";
         strictEqual = keyed || fn;
         return fn ? untrack(() => (child as any)(c as T)) : child;
       }


MoustaphaDev:
Just realized this is the intended behavior

Issue #1513, opened 2023-01-31
Breaking change in runWithOwner


Exelord:
### Describe the bug

In this commit https://github.com/solidjs/solid/commit/54f3068dee208d6b37fa5f72a9571b5443a7b81b error handling has been added which caused a change in the interface, breaking backwards compatibility, and causing users to fail.

### Your Example Website or App

https://github.com/solidjs/solid/commit/54f3068dee208d6b37fa5f72a9571b5443a7b81b

### Steps to Reproduce the Bug or Issue

```
runWithOwner(owner, () => throw new Error("error")); // may or may not throw an error
```

### Expected behavior

`runWithOwner` should stay as a primitive to run code with different owner. Currently it is doing much more than that, like handling errors and managing Listener. This make it impossible to use it for the purpose described in the docs.

I think we should create another helper, and leave runWithOwner as a primitive to allow handle advanced use cases.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

thetarnav:
What have you been doing with `runWithOwner` before that is not possible now?
It not owning errors was causing `batch` to break the ability to throw errors to the ErrorBoundry. https://github.com/solidjs/solid/issues/1452
And Listener not being reset meant that you could run a callback with different Owner and Listener. https://github.com/solidjs/solid/issues/1495
It still does exactly what it's meant to do. It's just less buggy.

Exelord:
I was using it to execute a code in scope of other owner. However now, it is also changing the Listener as well us not throwing an error is case there is an error boundary - which is not what I expected from `runWithOwner`.  It is doing much more than that.

The breaking change does not allow anymore to relay on the result as in case of error it might return undefined. There is no way now to distinguish between an error and actually returned `undefined` value.

I would suggest reverting it to a primitive, and build a higher level helper to support those cases to bring back the original behaviour.

thetarnav:
Isn't this the same behavior as any other reactive primitive? instead of throwing and freezing the app, errors will be caught by the error boundary.
If you want something to throw, you shouldn't do that when running something with owner. I think. But could you schedule a microtask?
```ts
const result = runWithOwner(owner, () => {
  return { v: init() }
})

result // undefined - error thrown, { v: T } - ran without errors
```
I consider the previous behavior a bug, but you may have a different usecase for these.
[...]

PR #1515, opened 2023-02-01
Fix `runWithOwner` breaking change

Exelord:
## Summary

Fixes: https://github.com/solidjs/solid/issues/1513

Here is a PR, that will be fixing the mentioned issues as well as preserving error handling and reseting the Listener.

`runWithOwner` will now work exactly the same way as `untrack/runWithListener`, so using it outside of a root or batch (runUpdates) will trigger effects immediately.

`pnpm test`

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -967,8 +967,6 @@ export function runWithOwner<T>(o: Owner, fn: () => T): T | undefined {
   Listener = null;
   try {
     return runUpdates(fn, true)!;
-  } catch (err) {
-    handleError(err);
   } finally {
     Owner = prev;
     Listener = prevListener;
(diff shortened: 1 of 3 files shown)


packages/solid/src/reactive/signal.ts
```diff
@@ -1482,7 +1480,13 @@ function runTop(node: Computation<any>) {
 }
 
 function runUpdates<T>(fn: () => T, init: boolean) {
-  if (Updates) return fn();
+  if (Updates) {
+    try {
```
Exelord, line 1484: I am not sure about the consequences of doing that, so I would be grateful for guidance here.

ryansolid: Did you need to do this? The idea in general here is that the only times Updates exists is if part of a parents update. I guess that is the concern because injected owner should handle it rather than the starter of the updates. You start see why I like doing it in runWithOwner.

Exelord: You are right Ryan :) I can see it now. Let me think about it a little more.

PR #1518, opened 2023-02-02
Support null for detachedOwner in createRoot

Exelord:
## Summary

To create a fully detached root, we need to make sure the owner is not inherited. Currently `null` value was treated the same as `undefined` leading to use of current owner. To be able to control that behavior, now we can support `null` as detachedOwner which is a valid Owner type.

Added unit test (pnpm test)


ryansolid:
Thank you makes sense.

PR #1519, opened 2023-02-03
Fix JSX runtime

wooorm:
## Summary

This practically fixes the JSX development runtime, which passes several other parameters, which currently results in Solid not rendering nodes.

It also fixes a case where a key parameter might be passed as `undefined`. This does not happen normally when using a compiler that compiles JSX to function calls, because Solid as far as I understand does not support keys. However, as the key is defined in the API of `jsx`, I think it’s better to cut it off.

With the following code in a browser:

```js
/* eslint-env browser */
import h from 'https://esm.sh/solid-js@1/h'
import {jsx, jsxDEV} from 'https://esm.sh/solid-js@1/h/jsx-runtime'
import {render} from 'https://esm.sh/solid-js@1/web?dev'

function jsxFixed(type, props) {
  return h(type, props)
}

render(ComponentA, document.getElementById('root'))
// Renders `<div>false</div>`
// Fixed: `<div>asd</div>`

render(ComponentB, document.getElementById('root'))
// Renders `<div></div>`
// Fixed: `<div>asd</div>`

function ComponentA() {
  return jsxDEV(
    'div',
    {children: 'asd'},
    undefined,
    false,
    {
      fileName: 'example.js',
      columnNumber: undefined,
      lineNumber: undefined
    },
    undefined
  )
}

function ComponentB() {
  return jsx('div', {children: 'asd'}, undefined)
}

function ComponentAFixed() {
  return jsxFixed(
    'div',
    {children: 'asd'},
    undefined,
    false,
    {
      fileName: 'example.js',
      columnNumber: undefined,
      lineNumber: undefined
    },
    undefined
  )
}

function ComponentBFixed() {
  return jsxFixed('div', {children: 'asd'}, undefined)
}
```

### DIFF
--- a/packages/solid/h/jsx-runtime/src/index.ts
+++ b/packages/solid/h/jsx-runtime/src/index.ts
@@ -6,5 +6,9 @@ function Fragment(props: { children: JSX.Element }) {
   return props.children;
 }
 
+function jsx(type: any, props: any) {
+  return h(type, props);
+}
+
 // support React Transform in case someone really wants it for some reason
-export { h as jsx, h as jsxs, h as jsxDEV, Fragment };
+export { jsx, jsx as jsxs, jsx as jsxDEV, Fragment };


remcohaszing (approved):

ryansolid:
Oh so the new transform doesn't do the 3rd and n... arguments as children. I didn't realize that. Thanks for pointing this out.

wooorm:
Yeah, the new transform moves a lot of the lifting to the compilation phase, so you don’t need to do that at runtime anymore.

For example, if there is zero or one child, it will now be passed as `{children: Child | undefined}, to `jsx`. But if there are multiple, you’ll get them as `{children: Array<Child>}` to `jsxs`.
You could used that for potential performance improvements.

In the dev runtime, you’ll also get as shown above useful extra values, such as in which file and at what line/column the thing was authored, which could be used to improve devtools.

PR #1530, opened 2023-02-06
Cast string into error while handling errors

Exelord:
## Summary

I am proposing to cast strings into errors, to make sure `castError` will always return the error. This will unify the behaviur and simplify errors handling by users. Also the original error, whether string or object, will be now attached as `cause` of the error.

`pnpm test`


ryansolid:
This is good change probably. I did the casting mostly for undefined/null cases. I'm not sure it is expected if you throw a string to not get a string back but this does make things more homogenous. In any case this will require a more significant release than a patch to merge it.

Exelord:
Throwing anything other than Error is considered an anti-pattern as it does not contain a stack trace (not possible to track it down).

I agree that this change may cause breaking changes, however I believe it will point out existing issues rather than create them. 

But sure, let's leave it for bigger release :)

Exelord:
I have also strongly typed `onError` hook, as we will be able to guarantee now the error.

Exelord:
@ryansolid also a question. Maybe we should only cast error in case we are passing it to onError hooks?

```ts
function handleError(err: unknown): void {
  const fns = ERROR && lookup(Owner, ERROR);
  if (!fns) throw error;
  for (const f of fns) f(castError(err));
}
```

and in case they are not handled, we should throw them they way they are? This way we can fully control its type, without breaking change, and without enforcing the cast.

ryansolid:
I needed it for resources as well. That was actually where a lot of issues were happening. Especially how it affected SSR and communicating across the wire.

ryansolid:
> @ryansolid also a question. Maybe we should only cast error in case we are passing it to onError hooks?
> 
> ```ts
> function handleError(err: unknown): void {
>   const fns = ERROR && lookup(Owner, ERROR);
>   if (!fns) throw error;
>   const error = castError(err);
>   for (const f of fns) f(error);
> }
> ```
> 
> and in case they are not handled, we should throw them they way they are? This way we can fully control its type, without breaking change, and without enforcing the cast.

Re-reading. This is a good idea.

packages/solid/src/reactive/signal.ts
```diff
@@ -1681,11 +1681,16 @@ function castError(err: unknown): Error {
...
+function runErrors(fns: ((err: any) => void)[], err: any) {
+  for (const f of fns) f(err)
+}
+function handleError(err: unknown) {
   const error = castError(err);
   const fns = ERROR && lookup(Owner, ERROR);
   if (!fns) throw error;
-  for (const f of fns) f(error);
+  if (Effects)
+    Effects!.push({ fn() { runErrors(fns, error); }, state: STALE } as unknown as Computation<any>);
```
Exelord, line 1692: This is interesting. Im curios to learn more why we need to run them in effect 🤔

Exelord:
Cool! Let me prepare a new PR then to cast only when handling with hooks.

kapilpipaliya:
when `ctx.measureText` error is not properly handled:

```
  let canvasElement:HTMLCanvasElement = document.getElementById("myCanvas");
  let ctx = canvasElement.getContext("2d");
  ctx.font = args.font;
  ctx.measureText(args.text).width // when this line throw error I am getting castError
```
image

PR #1534, opened 2023-02-07
Add fast track for `untrack` in case of `null` listener

Exelord:
## Summary

Now, we will check the Listener before untracking optimising execution in case of nested untrack usage.

```ts
const api = {
  run() {
     untrack(() => setName("fast"))
  }
}

untrack(() => api.run(())
```

`pnpm test`

### DIFF
--- a/.changeset/shaggy-pears-think.md
+++ b/.changeset/shaggy-pears-think.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Add fast track for `untrack` in case of `null` listener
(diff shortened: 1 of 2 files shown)


ryansolid:
Seems reasonable.

novacrazy:
Has this been benchmarked? A branch + inability to inline `fn` may be worse than just copying a reference + try-catch, especially for most cases it’s not nested.

LiQuidProQuo:
image

ryansolid:
> Has this been benchmarked? A branch + inability to inline `fn` may be worse than just copying a reference + try-catch, especially for most cases it’s not nested.

It crossed my mind especially because nesting is less common. I wasn't sure I was going to get anything definitive here in microbenchmarks and I don't think it will make a sizeable difference either way. 

I mean for completion we probably should try it unnested as well @LiQuidProQuo

LiQuidProQuo:
same code as above ( untrack - early exit   , untrack2 - original ) 
image

PR #1540, opened 2023-02-10
fix: set store not working when too deep

otonashixav:
## Summary

Fixes https://playground.solidjs.com/anonymous/f595df62-5514-4986-b012-4894b8b02e92. I think it's broken because strings have a key (`number`) which has a type of `string` again, causing infinite recursion. Preventing non-wrappable types from being recursed into fixes that. Checking `[T] extends [never]` fixes infinite recursion in the case where `T` becomes `never`.

`pnpm test`.

~~I can't run the tests for `solid-js` on windows, did this break recently?~~

Nevermind, nuking `node_modules` fixed it.

### DIFF
--- a/.changeset/clean-months-scream.md
+++ b/.changeset/clean-months-scream.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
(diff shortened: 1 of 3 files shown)


ryansolid:
Thank you

PR #1545, opened 2023-02-12
Fix order of onError callbacks execution

Exelord:
## Summary

I would expect that errors callbacks will be executed in reversed order ie. From latest to last defined.

```ts
function doSth() {
  onError(() => console.log("child"))
  throw new Error("error")
}

onError(() => console.log("parent"))
doSth(); // logs: "child", "parent"
```

## Why is it important?

We should be able to handle error as close to the error itself as possible. In the above example I would expect the child to have a chance to catch the error first before raising it to parent.

Similar like with `<ErrorBoundry />` the errors are handled in reversed order:
```tsx
<ErrorBoundary fallback={err => { console.log('parent'); throw err}}>
  <ErrorBoundary fallback={err => { console.log('child'); throw err}}>
    <Broken />
  </ErrorBoundary>
</ErrorBoundary>
```

`pnpm test` & `pnpm build`


Exelord:
I have revised the error handling more. Closing in favor of: https://github.com/solidjs/solid/pull/1551

PR #1551, opened 2023-02-13
Change onError mechanism to bubbling

Exelord:
## Summary

I would like to propose changing the current way of `onError` handling:

current

into bubbling mechanism:

bubbling

## Motivation

Current we are not able to handle an error without making sure that none of `onError` hooks won't throw an error. However this seams to be impossible as in a context of specific function we cannot be sure if any of the parents are doing that or not. With bubbling mechanism, the error will be first handled as close to the origin as possible, and the handler will be able to decide whether to rethrow it or not. In the first case the error will be propagated further to the next nearest handler, in the second case, it will be treated as handled error, and no other `onError` hooks will be triggered. 

This is the same way how try/catch is working, so it is more intuitive, and delivers more familiar DX 

`pnpm test && pnpm build`

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1681,11 +1681,19 @@ function castError(err: any) {
   return new Error("Unknown error");
 }
 
-function handleError(err: any) {
+function handleError(err: unknown): void {
   err = castError(err);
   const fns = ERROR && lookup(Owner, ERROR);
   if (!fns) throw err;
-  for (const f of fns) f(err);
+  for (const f of fns) {
+    try {
+      f(err);
+      return;
+    } catch (e) {
+      err = e;
+    }
+  }
+  throw err;
 }
 
 function lookup(owner: Owner | null, key: symbol | string): any {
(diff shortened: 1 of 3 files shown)


ryansolid:
I'm 50/50 on this. Bubbling is good. But I view onError like onCleanup. At the same level it should get equal priority. This is a UI concern. Registering additional isn't overriding but additive.

ryansolid:
It should be inner. But I think it is interesting that throwing could walk up to the next one. It's possible `onError` is poorly designed.. Like it should be a Provider type where there is only one at each level. But that is the intention. Closest parent. Unfortunately `onError` makes no guarantees there.

Exelord:
I see `onError` hooks the same as try/catch:

```ts
onError((e) => console.log(e))
onError((e) => {throw e})

throw "error"
```

should work in my eyes the same as:

```ts
try { // first onError
  try { // second onError
    throw "error"
  } catch(e) {
    throw e;
  }
} catch(e) {
  console.log(e)
}
```

Even though they are defined at the same level, the order should matter, as that allow to handle errors as close the the root as possible. Currently, working with it is super unintuitive and there is 99% chance for misuse, as not many will realize that none of the onError hooks can throw in the entire reactive scope (unknown) in order to catch the error. It behaves more like  a callback than a handler.

Going with that thinking further onCleanup should also execute in reversed order while inside of reactive scope, and in natural order while outside.

I was also thinking about the API design of onError. It seams to be off for its purpose but after hours of thinking I could't came up with alternative. It too easy right now to abuse it.

LiQuidProQuo:
started a discussion https://github.com/solidjs/solid/discussions/1557
for an alternative model.

ryansolid:
Yeah. I'd like to continue this in the discussion. I'm pretty much on an explicit catchError primitive that forces nesting and prevents siblings. Having subsequent `onError`'s nest and yet exist after the code they would catch is odd. I'm like 95% on deprecating `onError`. So @Exelord I'd love to get your feedback.

PR #1571, opened 2023-02-22
Fix `isServer` const type

thetarnav:
Because the type for the const `isServer` is taken from the client runtime version where it was just assigned `false`, it's type was inferred as `false` as well. This lead to some warnings when using stricter eslint rules:

image

This will make it into `boolean`

### DIFF
--- a/.changeset/brown-tips-remain.md
+++ b/.changeset/brown-tips-remain.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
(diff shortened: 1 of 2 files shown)


thetarnav:
Also unrelated but related question:
Should we add an `isDev` boolean const to `"solid-js/web"` as well?
It would make checking development/production envs clearer than using the `DEV` object.

PR #1592, opened 2023-03-03
Add `ClientOnly`

lxsmnsyc:
## Summary

Finally took the initiative. This one is the missing piece to SSR components, as it has some common use cases (e.g. rendering a component with web-only dependency). `solid-start` has one but I'm just hoping that core is the best place to put this so that other SSR frameworks (like Astro) can also use this. The fact that this is one of the most requested solution in `solid-start` and possibly other SSR applications should be enough to motivate this to be in the core.

The code is pretty basic, based on my 2-year old snippet: https://discord.com/channels/722131463138705510/722167424186843267/1022433627000291348

I just added some hydration checks + server code should be optimized to just return `undefined` directly

### DIFF
--- a/packages/solid/src/render/flow.ts
+++ b/packages/solid/src/render/flow.ts
@@ -7,7 +7,8 @@ import {
   Accessor,
   Setter,
   onCleanup,
-  MemoOptions
+  MemoOptions,
+  onMount
 } from "../reactive/signal.js";
 import { mapArray, indexArray } from "../reactive/array.js";
 import { sharedConfig } from "./hydration.js";
@@ -257,4 +258,24 @@ export function ErrorBoundary(props: {
     undefined,
     "_SOLID_DEV_" ? { name: "value" } : undefined
   ) as Accessor<JSX.Element>;
+}
+
+export function ClientOnly(props: {
+  children?: JSX.Element,
+}): JSX.Element {
+  if (sharedConfig.context) {
+    const [flag, setFlag] = createSignal(false);
+
+    onMount(() => {
+      setFlag(true);
+    });
+
+    return createMemo(() => {
+      if (flag()) {
+        return createMemo(() => props.children);
+      }
+      return undefined;
+    });
+  }
+  return createMemo(() => props.children);
 }
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -6,7 +6,8 @@ import {
   devComponent,
   $PROXY,
   $DEVCOMP,
-  EffectFunction
+  EffectFunction,
+  onMount
 } from "../reactive/signal.js";
 import { sharedConfig, nextHydrateContext, setHydrateContext } from "./hydration.js";
 import type { JSX } from "../jsx.js";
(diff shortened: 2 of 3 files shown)


packages/solid/src/render/flow.ts
```diff
@@ -258,3 +259,23 @@ export function ErrorBoundary(props: {
     "_SOLID_DEV_" ? { name: "value" } : undefined
   ) as Accessor<JSX.Element>;
 }
+
+export function ClientOnly(props: {
+  children?: JSX.Element,
+}): JSX.Element {
+  if (sharedConfig.context) {
```
lxsmnsyc, line 266: iirc this is the way to check if the app is currently hydrating. This check is needed so that UI correction can be applied, otherwise we can just render the children directly (since the app is just in CSR)

packages/solid/src/render/flow.ts
```diff
@@ -258,3 +259,23 @@ export function ErrorBoundary(props: {
     "_SOLID_DEV_" ? { name: "value" } : undefined
   ) as Accessor<JSX.Element>;
 }
+
+export function ClientOnly(props: {
+  children?: JSX.Element,
```
emdede, line 264: Is there a use-case where you'd want to use a self-closing `<ClientOnly />`?

lxsmnsyc: Maybe 🤷 🤣

ryansolid:
My only concern with this and why I did this with dynamic imports/compiler macro in SolidStart is if this will be sufficient for bundlers to drop the code. We told people to do this and they were having issues still since they used libraries that touched global APIs (like window, document) so this alone wasn't enough. At which point I was less into this approach as it could be considered false advertisement.

lxsmnsyc:
yeah I could probably do a `lazy`-like HOC. This wrapping component still has its place IMO

thetarnav:
Combining `<ClientOnly>`  with `lazy` should be enough to drop the code, right? As a component, it can be easily composed.
```tsx
const MyLazyComponent = lazy(() => import("..."))

<ClientOnly>
  <Suspense>
    <MyLazyComponent />
  </Suspense>
</ClientOnly>
```

lxsmnsyc:
@thetarnav yes it should be. I provided `clientOnly` just in case it wasn't enough and as stated by Ryan, users might mistake on using `ClientOnly` first before the lazy version

RodrigoProjects:
Is this still not merged for some reason? Just for context! :)

ryansolid:
We have it in SolidStart so there hasn't been pressure to bring it back into core yet. Still trying it out so to speak.

packages/solid/src/render/flow.ts
RodrigoProjects: If there is a play, there is a player 😆

ryansolid:
Thanks for the PR. I’m going to close this as stale for Solid 2, but the use case is real.

The main concern is still the one from the earlier discussion: a runtime `<ClientOnly>` can prevent a subtree from rendering on the server, but it does not by itself guarantee that browser-only dependencies are excluded from the server bundle or never evaluated. The cases that usually motivate this often need framework/compiler/dynamic-import support, which is why this has lived in SolidStart rather than core.

Solid has long had lower-level tools like `<NoHydration>` / `<Hydration>` and `isServer`, and Solid 2 adds more hydration control through client-sourced primitives such as `ssrSource: "client"`. Given that direction, I don’t think this exact core API is the right thing to merge now.

If we revisit this, I’d want it to be a fresh Solid 2 design that clearly defines what problem it solves: runtime render gating, hydration behavior, dependency bundling, or some combination of those. This PR mostly covers the first part, while the harder and more requested cases tend to require the latter pieces too.

PR #1620, opened 2023-03-13
feat(store): deep watch on dependencies

iosamuel:
## Summary

I come from a Vue Background and I have used a lot the [`deep` watcher](https://vuejs.org/guide/essentials/watchers.html#deep-watchers) into reactive objects to make sync-like operations, e.g save to localStorage, send information to APIs, etc. So I would love to have something similar here.

I used a similar algorithm as Vue, as the reactive systems are similar, we just need to deepTraverse through the dependency and call any value in the Proxy object so that it triggers the effect once anything changes.

- I created a UnitTest show-casing the behaviour of a deep traversed dependency object.
- Added the failing case when `deep` is off.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -951,8 +951,7 @@ export function on<S, Next extends Prev, Prev = Next>(
 /**
  * deepTraverse - traverse an object
  * @param obj an object to traverse
- * @returns a new object with all functions called
- * @description https://www.solidjs.com/docs/latest/api#deeptraverse
+ * @returns an object with all values traversed
  * @private
  * */
 function deepTraverse<S>(obj: S): S {
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -875,6 +875,7 @@ export type OnEffectFunction<S, Prev, Next extends Prev = Prev> = (
 
 export interface OnOptions {
   defer?: boolean;
+  deep?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


iosamuel:
I know this is a very naive and basic implementation, but I wanted to see if this is of some interest so I can work on it to improve and capture different scenarios.

thetarnav:
This would be an interesting addition to solid-primitives, I wanted to add it for a while but never did.
I think that adding it as a separate helper - rather than an option to `on` - would have more sense as it's tree-shakeable and it would be weird for `"solid-js"` core to include features for `"solid-js/store"`.
But as a separate helper, it's basically just `clone`.
So why not use [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone), lodash.cloneDeep or any other lib for cloning nested objects?

iosamuel:
> This would be an interesting addition to solid-primitives, I wanted to add it for a while but never did. I think that adding it as a separate helper - rather than an option to `on` - would have more sense as it's tree-shakeable and it would be weird for `"solid-js"` core to include features for `"solid-js/store"`. But as a separate helper, it's basically just `clone`. So why not use [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone), lodash.cloneDeep or any other lib for cloning nested objects?

- Yeah, I was unsure where to put it, I did iterations on `createEffect` and `on`, but still not sure. Probably it is better living inside the same `createStore` primitive, where we also set the traps for the Proxy Objects?
- This is the naive implementation. Ideally, we won't add space-complexity into it. So that we just only traverse the object instead of cloning it.

iosamuel:
The other option I have implemented is to add it inside: `packages/solid/store/src/store.ts` and have it exported as an EffectFunction from the `store` module.

And use it like:

```typescript
import { deep } from "solid-js/store";

createEffect(
    deep(() => store,
    v => { /* function to execute */ }
});
```

```typescript
export function deep<S, Next extends Prev, Prev = Next>(
  dep: Accessor<S>,
  fn: OnEffectFunction<S, undefined | NoInfer<Prev>, Next>,
  options?: DeepOptions
): EffectFunction<undefined | NoInfer<Next>> {
  let prevInput: S;
  return prevValue => {
    let input: S;
    let depValue = dep();
    depValue = deepTraverse(dep());
    input = depValue;

    const result = untrack(() => fn(input, prevInput, prevValue));
    prevInput = input;
    return result;
  };
}

function deepTraverse<S>(value: S, seen?: Set<unknown>): S {
  if (!(typeof value === "object")) {
    return value;
  }

  seen = seen || new Set();
  if (seen.has(value)) {
    return value;
  }
  seen.add(value);
  if (Array.isArray(value)) {
    for (let i = 0; i < value.length; i++) {
      deepTraverse(value[i], seen);
    }
  } else if (typeof value === "object" && value !== null) {
    for (const key in value) {
      deepTraverse(value[key], seen);
    }
  }
  return value;
}
```

ryansolid:
We used this approach a bit looking at an agnostic reactive core for Tanstack. But performance ended up being prohibitive moving the deep walks to reads. I had a cool idea of how to do this with a singular watcher but it required a bit more work.

My recommendation is Solid Primitives until we can introduce a native one in the store that we feel comfortable recommending.

iosamuel:
Got it! Thanks for looking into it.

I opened the PR with the effect function in `solid-primitives`: https://github.com/solidjs-community/solid-primitives/pull/350

Thanks everyone! 💙

iosamuel:
This is now living under: https://github.com/solidjs-community/solid-primitives/tree/main/packages/deep#readme

PR #1638, opened 2023-03-21
Solid v1.7

ryansolid:
This PR shows the upcoming changes for Solid 1.7. You can try it out on the`next` branch of git and `beta` tag on npm. It has numbering like v1.7.0-beta.3.

This release has a lot of TypeScript breaking changes. We've been pretty lenient on TS breaking changes in minors so we are keeping that trend, but I'm particularly concerned with this one. So I'd appreciate the ecosystem to try it out and give some feedback leading up to the release. On the positive all the TS fixes will be backwards compatible. We've just gotten more strict with the JSX mostly, so bending to that won't break any older versions of Solid.

I will publish beta versions of related packages (like vite-plugin-solid) so that you can try this. Most packages should just work with their latest versions.


juanrgm:
41 errors in [SUID ](https://github.com/swordev/suid) due to https://github.com/solidjs/solid/discussions/1554.

lxsmnsyc:
@juanrgm this is expected. solid-headless is also going to be affected by this

ryansolid:
Yeah to clarify I'm expecting a bunch of errors as this is stricter. It is more trying to determine if there is any truly awkward pieces. I don't think the errors effect external facing APIs.. just internals of libraries or applications themselves.

Issue #1642, opened 2023-03-22
InputHTMLAttributes does not have all native input attributes


jesseb34r:
### Describe the bug

The [InputHTMLAttributes interface](https://github.com/solidjs/solid/blob/main/packages/solid/h/jsx-runtime/src/jsx.d.ts#L787) does not have all of the attributes of the native `<input />` element. Specifically it doesn't handle the attributes that depend on the specified input `type`.

I initially ran into this bug when using a Kobalte text field input with `type="search"` and adding the `incremental` attribute that is specific to `<input type="search">`. It errors because the solidjs InputHTMLAttributes type does not include incremental. 

### Your Example Website or App

https://playground.solidjs.com/anonymous/67880890-0ca4-416b-a3a4-744398c71ff1

### Steps to Reproduce the Bug or Issue

add the `incremental` attribute to an `<input>` element with `type="search"` and it will type error.

### Expected behavior

I expected `<input type="search" incremental />` to be valid

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/61254589/227025289-cdd79fcf-abd5-4876-9391-ee5a759456f7.png)

### Platform

solid-js version 1.6.11

jesseb34r:
The [HTML spec](https://html.spec.whatwg.org/multipage/input.html#the-input-element) for the input element describes the expected attributes for each `type`

ryansolid:
Definitely open to PRs here. We got our types from other JSX libraries and some of those are older so we added them as we found them.

Our JSX types can be found in this package: https://github.com/ryansolid/dom-expressions/blob/main/packages/dom-expressions/src/jsx.d.ts

jesseb34r:
Yeah I linked the types file in the issue above. I will see if I can make a PR for this in the next week. It will be my first PR to a major OSS project like this so I might need some help, I'll reach out on discord if I have questions.

elite174:
@jesseb34r https://github.com/ryansolid/dom-expressions/pull/225 Could you check this?

titoBouzout:
This issue can be closed, it already has been merged in dom-expressions.

ryansolid:
Thanks.

PR #1647, opened 2023-03-25
Add dynamic typing to InputHTMLAttributes based on specified input type

jesseb34r:
## Summary

My initial motivation came because adding the `incremental` attribute to an `<input />` with `type="search"` threw a typescript error. I submitted this issue https://github.com/solidjs/solid/issues/1642#issue-1636428300 and @elite174 submitted a fix https://github.com/ryansolid/dom-expressions/pull/225#issue-1638861687. But this is only a partial fix because these attributes in addition to many of the input attributes are specific to certain input types.

I split the input attributes into three sections, `GlobalInputHTMLAttributes`, `ConditionalInputHTMLAttributeTypes`, and `NonStandardInputHTMLAttributeTypes`. Then I defined a type for each possible input type and selected the appropriate attributes as specified in the [official HTML spec](https://html.spec.whatwg.org/multipage/input.html#the-input-element:~:text=Button-,Content%20attributes,-accept). I used these types to define a dynamic `ConditionalInputHTMLAttributes` type that changes the valid attributes depending on the specified type. 

I am not sure how to best test this change. I made sure that the dynamic typing works by creating test objects in vscode, specifying a type value, and checking the resulting object type. As far as I can tell it applies all the types correctly. If there is a more thorough method for testing that you would like me to do I am happy to.


ryansolid:
Hi sorry for the confusion this file is actually a temporary build artifact and doesn't actually exist in that location permanently. It lives in DOM Expressions (https://github.com/ryansolid/dom-expressions/tree/main/packages/dom-expressions/src). I've remedied it (removing it from this repo).

PR #1609, opened 2023-03-09
Switch test runner from Jest to Vitest

niccholaspage:
## Summary
This PR switches the test runner from Jest to Vitest. By implementing this change, I was hoping to see some speed improvements in the time it takes to run tests, especially when it comes to tests running in CI. On my MacBook Pro with an M2 Max, I saw a bit of an improvement running the following command before and after this change:
```pnpm turbo run test --force```
| Run | Jest (af20f00)  | Vitest |
| - | ------------- | ------------- |
| 1 | 3.097 sec  | 2.534 sec  |
| 2 | 3.179 sec  | 2.553 sec  |
| 3 | 3.072 sec  | 2.547 sec  |
| 4 | 3.13 sec  | 2.537 sec  |
| 5 | 3.173 sec  | 2.537 sec  |
| **Average** | **3.1302 sec**  | **2.5416 sec**  |

This is one of my first PRs to an open source project, so please let me know if there is anything I can do or any changes I can make to help with this. I know it's a bit much to digest due to a lot of the changes seen here, and there are some interesting things I did that we might need to work around, like the usage of vite-plugin-solid as a dependency. I am open to making any changes as needed to help increase the viability of this PR. There are a couple of things we can tweak when it comes to running tests in CI as well, especially the threading and isolation settings in Vitest to see if it would speed up a CI run compared to my own computer locally, where it didn't seem to help.

## How did you test this change?
I tested this change by making sure all of the tests passed and that Solid still built as expected. I compared the distributable files before and after my changes and made sure that they matched exactly too, to make sure nothing weird was affected when I was doing this migration.

<details>
  <summary>SHA sums for distributable files before changes</summary>

```
d3004f5828921f4d8774799b488ae66636138319  packages/solid-element/dist/index.d.ts
6982d880c12c641061f36228e17426cd7bfdad6b  packages/solid-element/dist/index.js
122496ce27d5daac7d8a34d1ce3cb14bebe3d07c  packages/solid/dist/dev.cjs
2456670e85464922f6b3af99c613359497e3224f  packages/solid/dist/dev.js
9c936e436801b24eca0d5084f599c7fd7808e6fa  packages/solid/dist/server.cjs
82628e45b59d8cb60a7c885aadd51f9813330d52  packages/solid/dist/server.js
0f432c81bc7ce4ac620ee17227d59d6816a594cf  packages/solid/dist/solid.cjs
9fcd63bcf5accdc9e2976181fdb2268208ddfe33  packages/solid/dist/solid.js
[...]

PR #1649, opened 2023-03-25
Fix infinite loop in Suspense error handling on the server

### LINKED ISSUE #1648: Maximum call stack size exceeded in Suspense error handling on the server
### Describe the bug

If an accessed resource throws an error during SSR, Suspense will go into an infinite loop trying to handle it.
This issue only exists in versions `>=1.6.13` so most likely this is a regression caused by this commit: 60f8624db81e5491bdd3f4b69870899976212941

I think I know where the problem is, so I'll try to make a PR

image

https://stackblitz.com/edit/server-suspense-infinite-loop?file=index.ts

1. `pnpm i` to install deps
2. `pnpm dev` to run the server script
3. look at the error message

### Expected behavior

The error should be able to be handled by an error boundary

thetarnav:
## Summary

Fixes #1648 

Suspense was attaching it's `onError` not to the `o` new owner it created, but to the current `Owner`, which ends up being the parent of `o`. So when it tries to pass the error to the parent with `runWithOwner`, it's actually passing it to itself creating an infinite loop.

I see there are no server tests for solid-js at the moment. But patching solid-js locally has solved this issue for me.
I'm not sure if I should maybe try adding one using `// @jest-environment node` comment to force node env with jest?
Maybe after #1609 gets merged this could be improved.

### DIFF
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -552,7 +552,6 @@ export function SuspenseList(props: {
 
 export function Suspense(props: { fallback?: string; children: string }) {
   let done: undefined | ((html?: string, error?: any) => boolean);
-  let clean: any;
   const ctx = sharedConfig.context!;
   const id = ctx.id + ctx.count;
   const o = createOwner();
@@ -569,8 +568,8 @@ export function Suspense(props: { fallback?: string; children: string }) {
     });
   function runSuspense() {
     setHydrateContext({ ...ctx, count: 0 });
-    o && cleanNode(o);
-    return runWithOwner(o!, () => {
+    cleanNode(o);
+    return runWithOwner(o, () => {
       return createComponent(SuspenseContext.Provider, {
         value,
         get children() {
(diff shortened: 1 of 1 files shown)


packages/solid/src/server/rendering.ts
```diff
@@ -552,7 +552,6 @@ export function SuspenseList(props: {
 
 export function Suspense(props: { fallback?: string; children: string }) {
   let done: undefined | ((html?: string, error?: any) => boolean);
-  let clean: any;
```
thetarnav, line 555: `clean` wasn't being used anywhere anymore. maybe that's a mistake? not sure what it did

packages/solid/src/server/rendering.ts
```diff
@@ -569,8 +568,8 @@ export function Suspense(props: { fallback?: string; children: string }) {
     });
   function runSuspense() {
     setHydrateContext({ ...ctx, count: 0 });
-    o && cleanNode(o);
```
thetarnav, line 572: `o` is always present. before 1.6.13 `clean` was checked here

ryansolid: Are we sure we can just remove it? I hoisted it up for sure and that caused the issue, but it was always there. I guess I can retest the original React Query scenario.

thetarnav: pre `1.6.13` the current `Owner` would get assigned as `o` so it could be `null`, now we are always a new owner so the condition is pointless.

thetarnav: Also, github is showing the preview a bit weirdly. I didn't remove `cleanNode(o)`, just `o &&`

ryansolid: Oh I get it. Yeah I was looking at the wrong part. I guess the trickiest part is I change this mechanism completely in 1.7. It's going to be a bit interesting to figure out how to merge this back into next. It looks like I need to do another 1.6 release given the stuff I've broken though so I will go with this for now.

PR #1652, opened 2023-03-26
Add `loading` attribute type to `<iframe>`

kidonng:
## Summary

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#loading

N/A


ryansolid:
Hi sorry for the confusion this file is actually a temporary build artifact and doesn't actually exist in that location permanently. It lives in DOM Expressions (https://github.com/ryansolid/dom-expressions/tree/main/packages/dom-expressions/src). I've remedied it (removing it from this repo).

kidonng:
No worries! I will open a PR there instead then.

PR #1654, opened 2023-03-28
Fix external source tests

niccholaspage:
## Summary
This PR fixes the external source tests @ryansolid disabled in the next branch. This is done by only enabling external sources before the test runs, and then resetting imported modules after, bringing things back to a clean slate for the remaining tests. Also included is an override for babel-preset-solid to ensure it comes from the workspace instead of from vite-plugin-solid's dependency. I don't think this does anything specifically to fix the issue but is a good thing to have I would think, since we probably shouldn't be pulling it from npm normally?

I ran brought back the external store test file and ran ```pnpm turbo run test test-types --force``` and everything passed!

### DIFF
--- a/.changeset/tender-sheep-eat.md
+++ b/.changeset/tender-sheep-eat.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix external source tests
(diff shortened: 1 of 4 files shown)


ryansolid:
Thank you

Issue #1660, opened 2023-03-30
v1.7 does not build on skypack


xunilrj:
### Describe the bug

skypack fails to build v1.7: https://cdn.skypack.dev/solid-js/web

```
/*
 * [Package Error] "solid-js@v1.7.0" could not be built. 
 *
 *   [1/5] Verifying package is valid…
 *   [2/5] Installing dependencies from npm…
 *   [3/5] Building package using esinstall…
 *   Running esinstall...
 *   rollup-plugin-inject: failed to parse solid-js/dist/dev.cjs. Consider restricting the plugin to particular files via options.include
 *   /tmp/cdn/_VNezyagvbSfhA4EpFC9I/node_modules/solid-js/dist/dev.cjs:1794:18 Error when using sourcemap for reporting an error: Can't resolve original location of error.
 *   Failed to load node_modules/solid-js/dist/dev.cjs
 *     Duplicate export 'DEV' (Note that you need plugins to import files that are not JavaScript)
 *   Install failed.
 *   Install failed.
 *
 * How to fix:
 *   - If you believe this to be an error in Skypack, file an issue here: https://github.com/skypackjs/skypack-cdn/issues
 *   - If you believe this to be an issue in the package, share this URL with the package authors to help them debug & fix.
 *   - Use https://skypack.dev/ to find a web-friendly alternative to find another package.
 */

console.warn("[Package Error] \"solid-js@v1.7.0\" could not be built. \n[1/5] Verifying package is valid…\n[2/5] Installing dependencies from npm…\n[3/5] Building package using esinstall…\nRunning esinstall...\nrollup-plugin-inject: failed to parse solid-js/dist/dev.cjs. Consider restricting the plugin to particular files via options.include\n/tmp/cdn/_VNezyagvbSfhA4EpFC9I/node_modules/solid-js/dist/dev.cjs:1794:18 Error when using sourcemap for reporting an error: Can't resolve original location of error.\nFailed to load node_modules/solid-js/dist/dev.cjs\n  Duplicate export 'DEV' (Note that you need plugins to import files that are not JavaScript)\nInstall failed.\nInstall failed.");
throw new Error("[Package Error] \"solid-js@v1.7.0\" could not be built. ");
export default null;
```

### Your Example Website or App

 https://cdn.skypack.dev/solid-js/web

### Steps to Reproduce the Bug or Issue

1. Just navigate to https://cdn.skypack.dev/solid-js/web

### Expected behavior

v1.7 should build successfully.

[...]

PR #1664, opened 2023-03-31
make it harder for the optimizer to incorrectly export DEV twice

xunilrj:
## Summary

Try to fix https://github.com/solidjs/solid/issues/1660. I think `skypack` is complaining because `dev.cjs` after being optimized has 

```js
exports.DEV = void 0;
exports.DEV = {
  hooks: DevHooks,
  writeSignal,
  registerGraph
};
```
This change make it harder for the optimizer to export `DEV` twice.

```js
const DEV = "_SOLID_DEV_"
  ? ({ hooks: DevHooks, writeSignal, registerGraph } as {
      readonly hooks: typeof DevHooks;
      readonly writeSignal: typeof writeSignal;
      readonly registerGraph: typeof registerGraph;
    })
  : undefined;
export { DEV };
```

I can´t test if this will fix `skypack` build.

### DIFF
--- a/packages/solid/src/index.ts
+++ b/packages/solid/src/index.ts
@@ -69,19 +69,14 @@ type JSXElement = JSX.Element;
 export type { JSXElement, JSX };
 
 // dev
-import {
-  registerGraph,
-  writeSignal,
-  DevHooks
-} from "./reactive/signal.js";
-let DEV:
-  | {
+import { registerGraph, writeSignal, DevHooks } from "./reactive/signal.js";
+const DEV = "_SOLID_DEV_"
+  ? ({ hooks: DevHooks, writeSignal, registerGraph } as {
       readonly hooks: typeof DevHooks;
       readonly writeSignal: typeof writeSignal;
       readonly registerGraph: typeof registerGraph;
-    }
-  | undefined;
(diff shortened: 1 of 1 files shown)


thetarnav:
Why not use a ternary? (`"SOLID_DEV" ? { ... } : undefined`)
The `DEV` object is used in libraries as a development mode check, so it being falsy needs to cause that code path to be tree-shaken.
Also the same DEV object is exported from the /store module.

xunilrj:
Good idea. The optimizer seems happy with that.

```js
const DEV = {
  hooks: DevHooks,
  writeSignal,
  registerGraph
} ;
exports.DEV = DEV;
```

packages/solid/src/index.ts
```diff
@@ -69,19 +69,14 @@ type JSXElement = JSX.Element;
...
 // dev
-import {
-  registerGraph,
-  writeSignal,
-  DevHooks
-} from "./reactive/signal.js";
-let DEV:
-  | {
+import { registerGraph, writeSignal, DevHooks } from "./reactive/signal.js";
+const DEV = "SOLID_DEV"
```
ryansolid, line 73: I think it is "_SOLID_DEV_"

xunilrj: Fixed.

PR #1699, opened 2023-04-23
stores: array items tracking, removal, fixes #1695

### LINKED ISSUE #1695: Clearing an array with reconcile doesnt properly unset individual array items
### Describe the bug

Deleting array items with "splice" versus clearing an array with `reconcile([])` doesn't behave the same way. Currently the `reconcile([])` doesn't notify individual array items, even worse: the individual items still can be accessed ':).

https://playground.solidjs.com/anonymous/422636c2-e46c-4a65-89f9-f8b2e1f1e3f3

1. Press "Clear array" button
2. The "item" and "name" rows stay as is (i expected that they would be blank)
3. The "weird memo" row shows that the "deleted" item still exists
4. Press "Reset" button
5. Press "Delete first"
6. The result of this action is what I also expected from "Clear array"

### Expected behavior

Individual array item subscriptions should be notified when the array is cleared and the items should be deleted :). This would make it consistent with notifications on array splices.

### Screenshots or Videos

[Bildschirmaufzeichnung vom 2023-04-20, 14-12-58.webm](https://user-images.githubusercontent.com/4012401/233362735-437e2133-d614-4361-9ab8-7043d0e10be1.webm)

### Platform

- OS: Linux
- Browser: Firefox
- Version: 110.0-2

### Additional context

_No response_

mdynnl:
## Summary
Array items in stores are not removed when array `length` decreased which was initially reported in #1695 WRT `reconcile`. Actually this happens with anything that involves `length` only change. `setStore([])`, `setStore('length', v => v - 1)`, `setStore(produce(v => v.length--))`, `mutable.length = 0` etc. The test only uses the first one for `stores` but it should cover all of them hopefully.

`pnpm test`

### DIFF
--- a/packages/solid/store/src/store.ts
+++ b/packages/solid/store/src/store.ts
@@ -228,8 +228,10 @@ export function setProperty(
     node: DataNode | undefined;
   if ((node = getDataNode(nodes, property, prev))) node.$(() => value);
 
-  if (Array.isArray(state) && state.length !== len)
+  if (Array.isArray(state) && state.length !== len) {
+    for (let i = state.length; i < len; i++) (node = nodes[i]) && node.$();
     (node = getDataNode(nodes, "length", len)) && node.$(state.length);
+  }
   (node = nodes._) && node.$();
 }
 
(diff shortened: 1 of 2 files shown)


ryansolid:
Thank you. I will take a look. I was thinking that the root notification would suffice, but maybe not. Much appreciated.

ryansolid:
I see. Right it is because those other cases aren't listening to top-level so they don't trigger. Hmm.. this seems like the correct fix but it could be costly. I will merge but this needs to be benchmarked.

PR #1724, opened 2023-05-13
chore: remove ./scripts/bump.js

K-Sato1995:
## Summary

Assuming `./scripts/bump.js` is currently not being used as the project uses [changesets](https://github.com/changesets/changesets) now. 

Plz correct me if I'm wrong here.  Just checked some PRs that bump versions and assumed the script is not currently being used.

- related: #1358


ryansolid:
Thanks you are correct

PR #1730, opened 2023-05-17
Solid-Element: Add clarification on 'props' parameter in customElement function

raymondboswel:

N/A


ryansolid:
Thank you.

PR #1732, opened 2023-05-19
feature: allow configuring the initial value of from

RobinVdBroeck:
## Summary

I added an option to pass an initial value to `from`. This allows for easier integration with 3th party observables. For example, OpenLayers has their own kind of observables where they emit an event if something changed.
```typescript
const view = fromSomewhere();
const viewProperties = from<Properties>((set) => {
    const syncProperties = () => {
      const olProps = view.getProperties();
      set(olProps);
    };
    syncProperties();
    props.view.on("propertychange", syncProperties);
    return () => props.view.un("propertychange", syncProperties);
 }) as Accessor<Properties>;
```

This code has multiple problems:
1. You need to call set directly in the setup, otherwise the signal will be undefined until the first event is emitted.
2. Typescript can't possibly know that the signal will never be undefined since it does not know what happens in the setup, so you have to do a manual cast to `Accessor<Properties>`. This makes the code more brittle, considering that you might forget to call the `set()` in the setup

As such, I changed the api of from that you can define an inital value. The previous code can then be rewritten to: 
```typescript
// this will now be inferred to Accessor<Properties>
const viewProperties = from<Properties>({
  subscribe: (set) => {
    const syncProperties = () => {
      const olProps = view.getProperties();
      set(olProps);
    };
    view.on("propertychange", syncProperties);
    return () => view.un("propertychange", syncProperties);
   },
   initialValue: view.getProperties(),
});
```

We can possibly also implement this for the function syntax (the one in use on the original example). I've not done it in this PR, but if there is interest in it, I will do so.

I added some test cases

### DIFF
--- a/packages/solid/src/reactive/observable.ts
+++ b/packages/solid/src/reactive/observable.ts
@@ -87,9 +87,21 @@ export function observable<T>(input: Accessor<T>): Observable<T> {
 export function from<T>(
   producer:
     | ((setter: Setter<T | undefined>) => () => void)
-    | { subscribe: (fn: (v: T) => void) => (() => void) | { unsubscribe: () => void } }
+    | { subscribe: (fn: (v: T) => void) => (() => void) | { unsubscribe: () => void }}
+): Accessor<T | undefined>;
+export function from<T>(
+  producer:
+    | ((setter: Setter<T | undefined>) => () => void)
+    | { subscribe: (fn: (v: T) => void) => (() => void) | { unsubscribe: () => void }, initialValue: T}
+): Accessor<T>;
+
+export function from<T>(
+  producer:
+    | ((setter: Setter<T | undefined>) => () => void)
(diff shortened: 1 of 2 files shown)


ryansolid:
Is this `initialValue` on the observable itself common? Is it standard? I haven't seen it before. Otherwise wouldn't it make more sense as a second argument to the function rather than something that requires modifying the observable being passed in?

I am not super into Rx or anything but I thought the solution to these sort of things was the `startsWith` operator. While I don't see much issue with setting an initial value, it would also be natural for the produce to set a first value if it makes sense for it. But out of all options modifying the observable object to something non-standard would be the least desirable solution.

clembu:
My use case for it is external event subscription (like with Electron apps or similar, to subscribe to events from the main process), it has nothing to do with RxJs, it's just a useful function for that purpose, because it handles the unlistening part of it.

I do think the initial value would be better passed as a second argument to `from`, maybe following the same convention as all the other operators, in a `FromOpts` argument.

RobinVdBroeck:
Closing this PR, since I lost interest in solid.

PR #1740, opened 2023-05-24
feature: Add document property to owner

pedro00dk:
## Summary

I currently work on an application based on microfrontends (MFEs) architecture. We have several teams working on different parts of the application and we currently use React.
Our setup consists on having each MFE contained in its own shadow root, having its own style and dependencices.

All the MFEs also use a component library that is based on webcomponents. Since different versions of the library are going to be used by different teams, we can't use the default `window.customElements` registry to store webcomponents.
We current use this [polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) that allows us to attach custom element registries to the MFEs shadow roots. The polyfill also attaches other functions to the shadow root e.g. `createElement`, `importNode`, etc... so it can be used as if it as a `Document` instance.

We also use React to render custom elements from the shadowRoot rather than window.
This is possible because when rendering elements, react will use the `ownerDocument` of the element passed to `createRoot`, so we can easily set the element's `ownerDocument` to be the shadowRoot itself and the render phase will use the custom functions attached to the shadow root. [(this is where react gets the ownerDocument, search for getOwnerDocumentFromRootContainer) ](https://github.com/facebook/react/blob/6fc3333b685657c8e6425df9fcb4fe1ce7374d40/packages/react-dom-bindings/src/client/ReactFiberConfigDOM.js#L196)

I have been trying to use Solid as an alternative to react in our application.
One thing that is preventing me from that is the fact that when rendering templates with custom elements, solid always uses `document.importNode`, there is no way to change it to use documents other than `window.document`.

So my idea was to make it possible to change the document that is used for rendering.
The idea consists on two small changes on this project (this PR), and on dom-expressions.

In this PR I added a new `document` property to owners when solid's `createRoot` and `createComputation` are called (not sure if it is needed on computations).
When child owners are created, I copy the document ref to make access easier.
One thing that concerns me is that the signal.js file there are no mentions to any DOM apis/types. So maybe that could be done in another way. In any way, the owner.document property is never used, just set.

In the [dom-expressions PR](https://github.com/ryansolid/dom-expressions/pull/253) I make two changes to the client.js file, one in the `render` function, it now always pass a owner to `createRoot`, that owner contains the document property set to the root element's `ownerDocument`.
And in the `template` function, instead of just calling `document.importNode`, it uses `(getOwner().document || document).importNode`.

This is my first PR in solid, I did not find much information on how to create a test build for both projects at once.
Since the amount of changes is very small, made the proposed changes directly to the solid dist files in my node_modules directory to verify that it works.

All existing tests on both projects also passed, I'm not sure if a test is necessary on this PR, it is just a new optional property on Owner. Maybe on the dom-extensions PR, but again, minor changes to render and template function, which I think are being tested already.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -82,6 +82,7 @@ export interface SignalState<T> extends SourceMapValue {
 }
 
 export interface Owner {
+  root?: Owner;
   owned: Computation<any>[] | null;
   cleanups: (() => void)[] | null;
   owner: Owner | null;
@@ -145,6 +146,7 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: typeof Owner)
         ? { owned: null, cleanups: null, context: null, owner: null }
         : UNOWNED
       : {
+          root: detachedOwner?.root ?? owner?.root,
           owned: null,
           cleanups: null,
           context: null,
@@ -1396,6 +1398,7 @@ function createComputation<Next, Init = unknown>(
     sourceSlots: null,
     cleanups: null,
     value: init,
+    root: Owner?.root,
     owner: Owner,
     context: null,
     pure


ryansolid:
I have to think about this. The core library is agnostic. It isn't aware of the DOM. So patching owner doesn't feel right. But I'm not sure how else to get that in as I haven't thought about it.

pedro00dk:
Another implementation suggestion is to replace `document` in `Owner` by a reference to the root owner or any unique object/symbol that can be a index in a weakmap.
Then, on dom-expressions, a weakmap whose keys are the root owner (or symbol) and values the elements' ownerDocuments 
 setup in the render function, could be used to get the correct ownerDocument in the template function.
That makes signals.js platform agnostic again.

I can update this PR or create another one with the alternative implementation if you think it is more interesting.

ryansolid:
This is better.. only thing I want to check now is that this doesn't tie our hands on the Solid 2.0 reactivity design. I don't know if we want to look roots through like this. It is probably harmless, but I don't want to depend on something that we may change.

ryansolid:
Thanks for the PR, and sorry this sat for so long. I agree this is a real gap: scoped custom element registries / shadow-root-backed documents need Solid’s DOM creation path to use the mount root’s document-like object rather than always falling back to global `document`.

The tricky part is the layering. The original `document`-on-owner approach puts DOM-specific state into Solid’s reactive owner, which does not feel right. The later `root`/WeakMap approach is better because it keeps the document itself in DOM Expressions, but it still creates a new owner-tree contract mainly to support renderer state lookup.

That is the part I’m hesitant to merge, especially for Solid 2 where the owner model has moved further into the platform-agnostic signals layer. I don’t want DOM Expressions depending on owner shape or root propagation details that we may want to keep internal.

So I’m going to close this PR as stale in its current form, but not because the use case is invalid. I think the right follow-up is a fresh design for a renderer-scoped state channel between Solid’s `rxcore` integration and DOM Expressions, so DOM Expressions can associate a render root with its `ownerDocument` without putting DOM concepts into signals or relying on owner internals.

If we solve that, the companion DOM Expressions behavior still makes sense: template creation/cloning for custom elements should use the render root’s document-like object when available.

PR #1749, opened 2023-05-30
docs: remove an unavailable video link in `README.md`

EastSun5566:
## Summary

Hi Solidjs Team,

I just came across an unavailable video link while going through `README.md`, so I took the liberty of removing it.


ryansolid:
Thank you. There will be a replacement soon. but until then I agree.

PR #1758, opened 2023-06-04
Improve type inference of `createSelector`.

thetarnav:
The types for `createSelector` are making this primitive annoying to use if we want to use the default equality checker - omit the second param - we have to pass an explicit `<number, number>` generic to get the right `key` type. Otherwise the selector function will be `(key: unknown) => boolean`.

image

This adds overloads to the function to infer the type of the `key` from `source`. It still can be overwritten and should be comparable with current uses of the primitive.

Added some basic type tests to cover the types.

### DIFF
--- a/.changeset/five-cats-compare.md
+++ b/.changeset/five-cats-compare.md
@@ -2,4 +2,4 @@
 "solid-js": patch
 ---
 
-Add function overloads to `createSelector` to improve type inference.
+Improve type inference of `createSelector`.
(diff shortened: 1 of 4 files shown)


AlexErrant:
Note that this bumps the lockfile version, so contributors must be on PNPM [v8.6.0](https://github.com/pnpm/pnpm/releases/tag/v8.6.0) or higher.

PR #1762, opened 2023-06-05
set pnpm version

AlexErrant:
## Summary

Manually specify PNPM version so new contributors' PNPM version and lockfile don't get out of sync.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,7 +92,8 @@ This repository uses [pnpm](https://pnpm.io/) and
 [Turborepo](https://turborepo.org/).
 If you want to build Solid from scratch, use the following steps:
 
-1. `pnpm install` (install all dependencies)
-2. `pnpm run build`
+1. `corepack enable` (use the correct version of PNPM, https://nodejs.org/api/corepack.html#enabling-the-feature)
+2. `pnpm install` (install all dependencies)
+3. `pnpm build`
 
 You can then run all tests via `pnpm test`.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -92,7 +92,8 @@ This repository uses [pnpm](https://pnpm.io/) and
 [Turborepo](https://turborepo.org/).
 If you want to build Solid from scratch, use the following steps:
 
-1. `pnpm install` (install all dependencies)
-2. `pnpm run build`
+1. `corepack enable` (use the correct version of PNPM, https://nodejs.org/api/corepack.html#enabling-the-feature)
+2. `pnpm install` (install all dependencies)
+3. `pnpm build`
```
AlexErrant, line 97: Yes, corepack is experimental, but it's the best solution I know of to juggle PNPM versions.

package.json
```diff
@@ -57,5 +57,9 @@
     "overrides": {
       "babel-preset-solid": "workspace:*"
     }
-  }
+  },
+  "engines": {
+    "pnpm": ">=8.6.1"
+  },
+  "packageManager": "pnpm@8.6.1"
```
AlexErrant, line 64: https://nodejs.org/docs/latest-v18.x/api/all.html#all_packages_packagemanager

Until the above docs or [StackOverflow](https://stackoverflow.com/q/71747609) comes to a clear answer/recommendation, I lean conservative and don't think we should specify a PNPM version range.

ryansolid:
Thank you

Issue #1773, opened 2023-06-10
catchError is not propagating error to parent


Exelord:
### Describe the bug

According to the documentation `catchError` should propagate the error to the parent:

https://github.com/solidjs/solid/blob/46e5e78710cdd9f170a7afd0ddc5311676d3532a/packages/solid/src/reactive/signal.ts#L973

However during my testing I have discovered this is not the case for nested `catchError`:

```js
  catchError(() => {
    catchError(() => {
      throw "hello"
    }, (err) => {
      console.log("1:", err);
      // throws Uncaught error
      throw err;
    })
  }, (err) => {
    // never throws
    console.log("2:", err)
  })
```

### Your Example Website or App

https://playground.solidjs.com/anonymous/6ed025c5-df61-4655-b971-07ab36c76386

### Steps to Reproduce the Bug or Issue

Nest catchError in catchError:

```js
  catchError(() => {
    catchError(() => {
      throw "hello"
    }, (err) => {
      console.log("1:", err);
      // thows Uncaught error
      throw err;
    })
  }, (err) => {
    // never thows
    console.log("2:", err)
  })
```

### Expected behavior

`catchError` should propagate the error to the parent 

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

modderme123:
For comparison, this actually does work when placed outside of a createRoot: https://playground.solidjs.com/anonymous/c2dc01b8-4351-4a9c-aa41-f6d7c7e622a9

ryansolid:
Hmm.. we will definitely be changing the behavior around error propagation in 2.0 when things get lazier but this looks the be an oversight. Thank you for the PR.

PR #1774, opened 2023-06-10
Propagate errors to parents when throwing errors in nested catchError

Exelord:
## Summary

While Im not sure yet if this is really the expected behaviour (at lest that's what doc says), it was quite easy to make catchError errors bubbling through parents.

Fixes: #1773 

`pnpm test`


ryansolid:
Thank you

PR #1784, opened 2023-06-20
fix(Portal): reactive in children when pass signal directly

### LINKED ISSUE #1767: Portal's children are not reactive
### Describe the bug

The `<Portal>` component doesn't react to changes made to `props.children`.
Not sure if that was always the case, but it seems like a bug.

```tsx
<Portal><div>{reactive()}</div></Portal> // works
<Portal>{reactive()}</Portal> // doesn't work
```

https://playground.solidjs.com/anonymous/3e137d98-99eb-47db-92fa-e13ae7b4bc95

wjq990112:
## Summary

Close #1767.

https://playground.solidjs.com/anonymous/0e4e4568-58b6-450a-945d-9205f37ac50e

### DIFF
--- a/.changeset/chilled-swans-learn.md
+++ b/.changeset/chilled-swans-learn.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix(Portal): reactive in children when pass signal directly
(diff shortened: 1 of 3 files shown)


ryansolid:
The biggest danger here is timing of when things are run. I recently went through a series of fixes on Portals because of moving stuff around. I suspect I might have to tweak this solution but it is a start. Thank you.

Issue #1526, opened 2023-02-05
enhance types of mergeProps


bigmistqke:
### Describe the bug

`mergeProps` is a powerful tool, but the typing is a bit clunky.

### Your Example Website or App

https://playground.solidjs.com/anonymous/f901c3ec-82cb-4fbf-bc81-6e6583c67b64

### Steps to Reproduce the Bug or Issue

``` tsx
type Vector = [number, number];
type Props = { tuple: Vector };

const App = (props: Partial<Props>) => {
  const merged = mergeProps({ tuple: [1, 1] }, props);
  merged.tuple; // type is (number[] | Vector) & (number[] | Vector | undefined)
  // ideally types would not become less narrow by using mergeProps

  // typecasting works, but is clunky and not typesafe
  const merged2 = mergeProps({ tuple: [1, 1] as Vector }, props);
  merged2.tuple; // type is Vector
  const merged3 = mergeProps({ tuple: [1, 1] as Vector }, props) as Props;
  merged3.tuple; // type is Vector

  // no type-error
  const merged4 = mergeProps({ tuple: [1, 1, 1] }, props)
  // no type-error
  const merged5 = mergeProps({ tuple: [1, 1, 1] }, props) as Props

  return <></>;
};
```

### Expected behavior

1. ideally types would not become less narrow by using mergeProps and no additional typecasts would be needed
2. original typing of props should be respected when merging props, or there is at least a typesafe way to enforce this. 

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

otonashixav:
1. I mean this isn't really something fixable, typescript is going to infer `[1, 1]` as `number[]` and not `[number, number]`, and I don't see any way to correctly type this without `[1, 1] as const` or `[1, 1] satisfies Vector`.
2. `mergeProps` doesn't know which of the arguments passed is the original type, and we it doesn't make assumptions about what the user wants. You could make a wrapper that is meant to merge props onto default props and have the latter be the "original" type.

bigmistqke:
2. I was thinking something in the line of mergeProps taking a generic that is the 'correct' type and the result of mergeProps would be typechecked against this type. Then `mergeProps<{tuple: Tuple}>({ tuple: [1, 1, 1] }, props)` would error and you could do something like `(props: Partial<Props>) => {const merged = mergeProps<Props>({...assign all values...}, props) ...}`

[...]

PR #1794, opened 2023-06-29
Feature: Add defaultProps helper function

lunafoxfire:
## Summary

mergeProps provides the current standard way of adding default props to a component, but it has some weaknesses when it comes to typing (see #1526). defaultProps essentially adds a subset of mergeProps functionality but with more restricted typing for the specific case of providing default props to a component. It has two main advantages:
- Exact typings are preserved (e.g. `"start" | "center" | "end"` does not get transformed into `string`)
- The default props are restricted to providing keys that are already defined as Component props. This provides safety against accidentally providing defaults for props that are no longer valid.

Even though technically the functionality of defaultProps can be achieved with mergeProps, defaultProps provides a much nicer typescript experience for an extremely common use case.

- Added tests to `packages/solid/test/component.spec.ts`
- Added type tests to `packages/solid/test/component.type-tests.ts`
- Ran the full test suite with `pnpm test`

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -265,6 +265,14 @@ export function mergeProps<T extends unknown[]>(...sources: T): MergeProps<T> {
   return target as any;
 }
 
+export type DefaultProps<T, K extends keyof T> = MergeProps<[Required<Pick<T, K>>, T]>;
+export function defaultProps<T, K extends keyof T>(
+  defaults: Required<Pick<T, K>>,
+  props: T
+): DefaultProps<T, K> {
+  return mergeProps(defaults, props);
+}
+
 export type SplitProps<T, K extends (readonly (keyof T)[])[]> = [
   ...{
     [P in keyof K]: P extends `${number}`
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -267,8 +267,8 @@ export function mergeProps<T extends unknown[]>(...sources: T): MergeProps<T> {
 
 export type DefaultProps<T, K extends keyof T> = MergeProps<[Required<Pick<T, K>>, T]>;
 export function defaultProps<T, K extends keyof T>(
-  defaults: Required<Pick<T, K>>,
-  props: T
+  props: T,
+  defaults: Required<Pick<T, K>>
 ): DefaultProps<T, K> {
   return mergeProps(defaults, props);
 }
--- a/packages/solid/test/component.spec.ts
+++ b/packages/solid/test/component.spec.ts
@@ -2,6 +2,7 @@ import {
   createRoot,
   createComponent,
   mergeProps,
+  defaultProps,
   splitProps,
   createUniqueId,
   createSignal,
@@ -225,21 +226,60 @@ describe("mergeProps", () => {
   });
 });
 
-describe("Set Default Props", () => {
-  test("simple set", () => {
+describe("defaultProps", () => {
+  test("empty defaults", () => {
     let props: SimplePropTypes = {
-        get a() {
-          return "ji";
-        },
-        b: null,
-        c: "j"
+      get a() {
+        return "beep";
       },
-      defaults: SimplePropTypes = { a: "yy", b: "ggg", d: "DD" };
-    props = mergeProps(defaults, props);
-    expect(props.a).toBe("ji");
+      b: null,
+      c: "boop"
+    };
+    props = defaultProps({}, props);
+    expect(props.a).toBe("beep");
     expect(props.b).toBe(null);
-    expect(props.c).toBe("j");
-    expect(props.d).toBe("DD");
+    expect(props.c).toBe("boop");
+    expect(props.d).toBe(undefined);
+  });
+  it("overwrites only undefined values", () => {
+    let props: SimplePropTypes = {
+      get a() {
+        return "beep";
+      },
+      b: null,
+      c: "boop"
+    };
+    props = defaultProps(
+      {
+        a: "xxx",
+        b: "xxx",
+        c: "xxx",
+        d: "xxx"
+      },
+      props
+    );
+    expect(props.a).toBe("beep");
+    expect(props.b).toBe(null);
+    expect(props.c).toBe("boop");
+    expect(props.d).toBe("xxx");
+  });
+  it("allows null as a default", () => {
+    let props: SimplePropTypes = {
+      a: "defined",
+      c: null
+    };
+    props = defaultProps(
+      {
+        a: null,
+        b: null,
+        c: null
+      },
+      props
+    );
+    expect(props.a).toBe("defined");
+    expect(props.b).toBe(null);
+    expect(props.c).toBe(null);
+    expect(props.d).toBe(undefined);
   });
 });
 
--- a/packages/solid/test/component.spec.ts
+++ b/packages/solid/test/component.spec.ts
@@ -235,7 +235,7 @@ describe("defaultProps", () => {
       b: null,
       c: "boop"
     };
-    props = defaultProps({}, props);
+    props = defaultProps(props, {});
     expect(props.a).toBe("beep");
     expect(props.b).toBe(null);
     expect(props.c).toBe("boop");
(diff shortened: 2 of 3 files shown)


ryansolid:
Ok. This is interesting. I can see value in special casing it for types. I generally don't like increasing the API surface but I have a suspicion this is the 90% case so that seems like a good add. I'd love to get some more feedback on this feature, but definitely something I think we'd consider.

atk:
I believe we should try to solve this within the type system before admitting defeat and resorting to extend the API, e.g. using a switch generic like `<DEFAULTS extends boolean = false>` that needs to be set to true to toggle the "defaults" mode.

otonashixav:
I can see enough reason for this to be included, albeit maybe with some iteration of its design. As I see it, `mergeProps` is analogous to a spread operator that works with how solid implements reactivity: `mergeProps(a, b, c) <==> {...a, ...b, ...c}`, so this could try to resemble default parameters: `defaultProps({a: 1}, props) <==> ({a = 1, b, c}: Props => ...)`.

Personally I would reverse the parameters (props first, then defaults), so that the default values follow their "declaration" as they do with default parameters, and also exclude props which are required since providing a default for those props is meaningless.

lydianlights:
> Personally I would reverse the parameters (props first, then defaults), so that the default values follow their "declaration" as they do with default parameters, and also exclude props which are required since providing a default for those props is meaningless.

I agree with this. I've been using this as a helper function quite extensively in my current project and having the signature be `defaultProps(props, defaults)` makes way more sense (I actually ended up switching the order in my own code). You get the type inference immediately, which helps when you start typing out the defaults. Also it's a bit nicer formatting wise with prettier.

I implemented it this way originally to stick with the current method but I think swapping the arguments is the way to go. Here's an example of how I've been using this (with swapped arguments version):
```jsx
export type SliderProps = {
    label?: string;
    value?: number;
    onChange?: (value: number) => void;
    min?: number;
    max?: number;
    step?: number;
    origin?: "start" | "center";
    class?: string;
};
const Slider: Component<SliderProps> = (unresolvedProps) => {
    const props = defaultProps(unresolvedProps, {
        label: "",
        min: 0,
        max: 100,
        step: 1,
        origin: "start",
        class: "",
    });
    // blah...
};

**Comment by lydianlights** [comment]
> I believe we should try to solve this within the type system before admitting defeat and resorting to extend the API, e.g. using a switch generic like `<DEFAULTS extends boolean = false>` that needs to be set to true to toggle the "defaults" mode.

I see what you're saying but if I can add my 2 cents about this: I think `defaultProps` still makes a lot of sense purely semantically, even if the typing issues with `mergeProps` can be resolved. Sometimes restricting a very general function to a more narrow domain to handle a common use case makes a lot of sense. However, I also know I probably would not have made this PR if I was using pure JS without typescript.

That's just my thoughts. I definitely understand the concerns with extending the API though.

**Comment by xxxxue** [comment]
`mergeProps` doesn't work well with TypeScript
Last year I wrote a defaultProps helper function

```typescript
import { mergeProps } from "solid-js";

export function defaultProps<T, U extends any | Partial<T>>(props: T, defProps: U): T & Required<U> {
    return mergeProps(defProps, props) as any;
}
```

packages/solid/src/render/component.ts
```diff
@@ -265,6 +265,14 @@ export function mergeProps<T extends unknown[]>(...sources: T): MergeProps<T> {
   return target as any;
 }
 
+export type DefaultProps<T, K extends keyof T> = MergeProps<[Required<Pick<T, K>>, T]>;
+export function defaultProps<T, K extends keyof T>(
+  props: T,
+  defaults: Required<Pick<T, K>>
+): DefaultProps<T, K> {
+  return mergeProps(defaults, props);
+}
```
otonashixav, line 274:
```ts
type DefaultableProps<T, K extends keyof T> = {
  [P in K]-?: undefined extends T[P] ? Exclude<T[P], undefined> : never;
}
```

```diff
- Required<Pick<T, K>>
+ DefaultableProps<T, K>
```

I haven't fully tested this but this should allow only defaults that aren't already necessary in `props`.

otonashixav (changes requested):

oddcelot:
Hey there, just tried out the `defaultProps()` in this [playground](https://playground.solidjs.com/anonymous/b37df4ac-b63b-42f5-a79e-08212a89ce54) in `CounterWithOldDefaults.tsx` and there may be an issue with the signature of the return type.

```ts
type Props = {
  initialCount: number;
  firstName?: string;
  middleName?: string | undefined;
  lastName?: string;
  notDeclared?: string
};

const $someNumber = 3;

export const myDefaultProps = {
  firstName: $someNumber >= 3 ? "John" : undefined,
  middleName: "Ron",
  lastName: $someNumber > 3 ? "Doe" : undefined,
} as const satisfies Partial<Props>;

const props = defaultProps(props, myDefaultProps);
```
So given this block here 👆, this should result in something like this, no?


<table>
<thead>
    <th>expected result</th>
    <th>actual result</th>
</thead>
<tr>
<td> 

```ts
const props: {
    initialCount: number;
    firstName: string | undefined;
    middleName: string;
    lastName: string | undefined;
    notDeclared?: string | undefined;
}
```

</td>
<td> 

```ts
const props: {
    firstName: string;
    middleName: string;
    lastName: string;
    initialCount: number;
    notDeclared?: string | undefined;
}
```

</td>
</tr>
<tr>
<td>
</td>
<td>
(actual order of props from result)
</td>
</tr>
</table>

In this example `firstName` and `lastName` can be still undefined and should be returned as such.
Also `myDefaultProps` is marked as error, since it has unions with `undefined`in it (since it uses `Required<T>` )
This may be a bit of a synthetic issue, but I managed to adress this issue I think…  (_also the order of the props is still the original one_)
Maybe have a look in the `Counter.tsx` in the playground

oddcelot:
Worked the bit out with where only keys that are actually in the props should be allowed in the defaults I think, like @otonashixav mentioned.
So, yeah, better errors 😅


Check out the [new playground](https://playground.solidjs.com/anonymous/847a3e08-00f8-4b9b-9003-f937ff362ce2)
>[!NOTE]  
> A thought: is `defaultProps` a bit misleading? I mean these are not the default props but rather a merge of props _with_ defaults… hence the name I used

>[!TIP] 
> You now get proper TS suggestions with this (maybe I also missed something, but didn't work for me somehow)
>
>image

RomainLanz:
Hey all! 👋🏻 

I had exactly the same issue today when I tried to use SolidJS and do a simple `Button` component.

Using the default API with `mergeProps`  make the following code not type-safe because the result of `mergeProps` remove the proper type for the `type` attribute.

```tsx
import { mergeProps } from 'solid-js'
import type { JSX } from 'solid-js/jsx-runtime'

interface ButtonProps {
  type?: 'button' | 'submit' | 'reset'
  children: JSX.Element
}

export function Button(initialProps: ButtonProps) {
  const props = mergeProps({ type: 'button' }, initialProps)

  return <button type={props.type}>{props.children}</button>
}
```

This result in the following TypeScript error:

```
TS2322: Type string is not assignable to type 'button' | 'submit' | 'reset' | undefined
```

I am wondering if there is any chance to have this PR merged or maybe looked at again?
I believe many beginners will have those kind of issue.

ryansolid:
Thanks for the PR, and sorry this sat for so long. The use case is real: defaulting component props while preserving literal types is a common TypeScript pain point, and `mergeProps` in Solid 1 made that more awkward than it should be.

I’m going to close this PR as stale for Solid 2, though. The relevant API surface has changed: `mergeProps` is no longer the public `solid-js` helper in the same way, and the current public helper is `merge`. Its typing is also much improved compared to the old `mergeProps` behavior.

That said, I don’t think the broader idea is invalid. A focused defaults helper may still be worth considering because it can express a narrower intent than `merge`: defaults should be limited to component prop keys, and the return type should reflect which optional props are actually guaranteed after applying defaults.

If we revisit this, I’d want it to be a fresh Solid 2 design against the current `merge` API, with props-first argument order and type tests for the edge cases raised in the comments, especially defaults whose value can still be `undefined`.

So closing this as stale, but not as rejected in principle.

PR #1823, opened 2023-07-25
Update universal rendering documentation

chiefcll:
The docs for universal rendering are out of date and I used them as a starting point for my project. Stumbled across the updated way of doing it and want the docs to reflect that.


ryansolid:
Oh.. use probably should be getting generated or be part of the things we ask for. It shouldn't be a seperate import on your part. This is an oversight on my part when I added it.

chiefcll:
When I started things worked with the babel config. I started getting the use error when I started adding some solidjs primitives like keyboard.

ryansolid:
Oh it's odd, added the use method to the return value from the createRenderer call a year ago. And it is in the types too. I wouldn't expect it to be missing. 

As for the docs, probably best to keep both examples. Technically Vite plugin is just exposing the Babel plugin so I'd like to keep the babel example there too as it is more primary.

PR #1842, opened 2023-08-04
Stores: fix overwriting underscore property

morethanwords:
## Summary

Store doesn't handle `_` property well. It works fine when adding element first time, but when you try to modify the same element, property will be `undefined`.

`pnpm t`


ryansolid:
Thank you.

PR #1859, opened 2023-08-17
chore: rename property reload to load in lazy component

### LINKED ISSUE #1726: Rename preload method to load  in lazy
https://github.com/solidjs/solid/blob/1593dd65a9e08d508d860b2afa75813c6bcc0e75/packages/solid/src/render/component.ts#L316

The name `preload` implies something that happens beforehand and does not describe the action it carries out. `load` would be a more accurate name for it.

wjq990112:
## Summary

Rename `preload` to `load` in `lazy` method's return value

In this PR:

- The `preload` property from the `lazy` method's return value has been renamed to `load`.
- To ensure backward compatibility with existing project, the `preload` property has been deprecated but not removed. Developers are encouraged to use the new `load` property going forward.

It's recommended for all teams and developers to start transitioning to the use of load as preload may be removed in future releases.

Close #1726.

None.

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -342,15 +342,14 @@ export function splitProps<
 }
 
 type AsyncComponent<T extends Component<any>> = () => Promise<{ default: T }>;
-
-// lazy load a function component asynchronously
-export function lazy<T extends Component<any>>(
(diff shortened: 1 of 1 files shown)


ryansolid:
Not going to do this one.

Issue #1873, opened 2023-09-07
Portal component's wrapper div cannot be styled


mdawar:
### Describe the bug

I'm using the `Portal` component for a toast element where I noticed that the wrapper `div` element is messing up the styles of the elements.

I know that this `div` is required, but is it possible to at least add the ability to add the `class` attribute for this element for these edge cases?

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-cshflq?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the [StackBlitz](https://stackblitz.com/edit/solidjs-templates-cshflq?file=src%2FApp.tsx) demo
2. Click the button to show the toast
3. You will see that the styles are messed up
4. If you inspect the elements and manually remove the wrapping `div` the layout will be fixed

### Expected behavior

I expected to at least be able to add the `class` attribute to the `Portal`'s wrapper `div`.
I would also be great if we can pass any attributes down to the `div` element.

### Platform

- OS: Linux
- Browser: Chrome
- Version: 116

### Additional context

Sorry if this issue is too specific, feel free to close it if it's out of scope.

AFatNiBBa:
You could set the "ref" attribute on the `Portal` like so:
```tsx
<Portal ref={x => x.style.display = "contents"} />
```
This will make the `div` behave like it wasn't even there (At least visually)

kachurun:
Why do we even need this `<div>` for Portal? I made a Proxy/Bridge component to render SolidJS in Riot.js components, and it looks like this:

```
<solid component={ MyFancySolidComponent } some-prop="1" />
```

After rendering, I leave this <solid> element in the DOM for some edge cases where Riot should control it (like for the 'each' directive) and insert MyFancySolidComponent's content right before this `<solid>` element. It renders the component wrapped in a ForwardRef component, which simply runs props.ref(props.children) in an effect.

This way, I can get MyFancySolidComponent's rendered nodes, even if it's a fragment, multiple elements, or even a lazy-loaded component. Next, I insert it before `<solid>` (as this.root in the Riot component) and store refs. On each refs update from ForwardRef, I just check the old and new element references and rearrange them if needed. Removing old refs and appending new ones is enough for me, but for Portal, we could use a smarter algorithm.

So I'm pretty sure this approach will work for Portal implementation without this unnecessary `<div>`, at least as long as we're not thinking about SSR, shadow DOM, SVG, or other edge cases.

[...]

PR #1874, opened 2023-09-07
Added `class` prop for the `Portal` component

mdawar:
## Summary

This is a possible fix for #1873, it adds a `class` prop for the `Portal` component to allow styling the wrapper `div`.

I have added new tests for the `Portal` component with the `class` prop and added assertions for previous tests.

### DIFF
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -57,6 +57,7 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
   mount?: Node;
   useShadow?: T;
   isSVG?: S;
+  class?: string;
   ref?:
     | (S extends true ? SVGGElement : HTMLDivElement)
     | ((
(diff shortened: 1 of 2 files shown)


ryansolid:
I see the desire but this solution won't work well.. it will recreate the whole element on class update. We could look at spreads maybe.. But this whole area of solution I'm not a huge fan of. We should look and see if there is something else that can be done here. It might still be possible to find a way to remove this element.

mdawar:
Thank you, it would be great if we can remove this element or at least pass the props down to it.

For now as a workaround, I'm using the `ref` prop on `Portal`, I didn't know about it before because it's not documented in the [API docs](https://www.solidjs.com/docs/latest/api#portal). Shouldn't be added?

```ts
<Portal mount={container} ref={(el) => el.setAttribute('class', 'class names')}>
  //...
</Portal>
```

Issue #1875, opened 2023-09-07
mergeProps fails with SSR


juanrgm:
### Describe the bug

This code throws an exception only with SSR enabled:

```ts
import { render } from "solid-js/web";

function Div(props: any) {
  return <div {...props} />;
}

function App() {
  // `data-firstprop` and `undefined` together cause the error
  const props = () => undefined as any;
  return (
    <Div data-firstprop {...props()}>
      test
    </Div>
  );
}

render(() => <App />, document.getElementById("app")!);
```

### Your Example Website or App

https://playground.solidjs.com/anonymous/d1fa8058-4207-41c8-98fe-591be0d25909

### Steps to Reproduce the Bug or Issue

1. Execute the code with SSR enabled.

### Expected behavior

No error.

### Platform

- SolidJS: 1.7.11

ryansolid:
What's the error? The reproduction just shows code but no error message. If I can get a working reproduction, maybe Stackblitz or at least a posting of what the error message is.

juanrgm:
I tried to create an example on Stackblitz, but NodeJS 18 is not supported yet (https://github.com/stackblitz/webcontainer-core/issues/560).

You can download it (https://stackblitz.com/edit/solid-ssr-vite-c8xxyw?file=src%2FApp.tsx) and run it locally.

The error is:

> An unhandled error occured: TypeError: Cannot read properties of undefined (reading 'data-firstprop')

And the problem is [here](https://github.com/solidjs/solid/blob/f79ba4d9087d74ef7e55fd0dc336c98ce2ec8526/packages/solid/src/server/rendering.ts#L105C16-L105C16).

ryansolid:
I see the server side merge props is much "simpler" and isn't handling undefined case properly. It must try to read it through undefined..

PR #1880, opened 2023-09-11
make reactive and non reactive jsx attributes consistent in ssr when expression evaluates to falsey

samualtnorman:
## Summary
Just got a massive stack trace in a solid start project due to reactive and non reactive attributes acting differently in ssr.
```jsx
<Foo {...bar && { bar }}/>
```
is fine whether `bar` is defined or `undefined`

```jsx
<Foo {...bar() && { bar: bar() }}/>
```
throws an exception when the `bar` `Accessor` is falsey.

You'll see that the fix is very simple.

I patched `solid-js` locally in my project and am currently dogfooding the change.

### DIFF
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -101,7 +101,7 @@ export function mergeProps(...sources: any): any {
           get() {
             for (let i = sources.length - 1; i >= 0; i--) {
               let s = sources[i] || {};
-              if (typeof s === "function") s = s();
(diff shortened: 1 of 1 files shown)


ryansolid:
I see yeah it is the same as this bug: https://github.com/solidjs/solid/issues/1875

Thanks for the PR but I already went and did the same little change.

samualtnorman:
oops my bad, looking forward to the next release

PR #1881, opened 2023-09-11
Bump node version in CI to > 16

birkskyum:
Due to the [Node 16 EOL](https://nodejs.org/en/blog/announcements/nodejs16-eol):

actions/checkout: node 14 -> 20 (there was no 18 release)
pnpm/action-setup: still on node 16, but this bumps from 2.2.2 -> 2.4.0
actions/setup-node: node 16 -> 18


ryansolid:
Thank you

PR #1889, opened 2023-09-23
feat: `createServerValue`

lxsmnsyc:
## Summary

This is a PoC for exploring the new serialization method. This PR adds a new primitive called `createServerValue` which allows users to instanciate server-side values and load those exact values on hydration. This is like resumability for babies.

A great use case of this would be:

```js
// Without createServerValue, this will throw a hydration error 99.9% of the time.
const someRandomData = createServerValue(() => Math.random());

return <h1>Data: {someRandomData}</h1>;
```

By default, `createServerValue` should be able to serialize everything the `createResource` is capable of (as of course they share the same serializer), with the exception of `createServerValue` having non-blocking serialization assuming that it produces a `Promise`. This is, of course, is up for debate.

This might also be in conflict with the upcoming async primitive.


packages/solid/src/server/rendering.ts
```diff
@@ -68,6 +68,18 @@ export function createUniqueId(): string {
   return `${ctx.id}${ctx.count++}`;
 }
 
+export function createServerValue<T>(cb: () => T): T {
+  const ctx = sharedConfig.context;
+  // This might be fun specially if the value is a Promise
+  // but of course, the only way they can translate the Promise
```
apatrida, line 74: what do you mean by "translte the Promise"

lxsmnsyc: convert the Promise into something the components and the reactivity system can understand

PR #1896, opened 2023-10-02
fix: missing `has` property in `SharedConfig`

lxsmnsyc:
Very tiny change, likely to be forgotten so I just made this PR.


ryansolid:
Thanks

PR #1913, opened 2023-10-11
Add `.js` extension to import that is missing it

frank-weindel:
## Summary

Trying to build a Solid renderer using the Node16 `moduleResolution` of TypeScript and we're seeing this error:

```
node_modules/solid-js/types/render/hydration.d.ts:1:29 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

1 import { Computation } from "../reactive/signal";
                              ~~~~~~~~~~~~~~~~~~~~
```

This seems to be the only place in the `solid-js` package that doesn't add the `.js` extension to the import path.

- [x] `pnpm test`
- [x] Confirm that it fixes issue

### DIFF
--- a/.changeset/wet-forks-melt.md
+++ b/.changeset/wet-forks-melt.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Add `.js` extension to import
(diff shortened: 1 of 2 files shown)


ryansolid:
Yeah that's an odd omission. Thanks

frank-weindel:
You got the changeset in right before I could 😄

PR #1915, opened 2023-10-11
Documents `!children.length`

titoBouzout:
Spent an incredible amount of time figuring out the reasoning of this check and would like to document it.

### DIFF
--- a/packages/solid/src/server/reactive.ts
+++ b/packages/solid/src/server/reactive.ts
@@ -234,10 +234,7 @@ export function runWithOwner<T>(o: typeof Owner, fn: () => T): T | undefined {
 }
 
 function resolveChildren(children: any): unknown {
-  // `!children.length` avoids reading signals to not triggger a refresh on the parent memo.
-  // The issue manifest when `children` is an array containing more than 1 signal, because
-  // an invalidation on any, will cause invalidation on siblings, as the parent memo needs to be refreshed.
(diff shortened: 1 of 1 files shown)


deluksic:
Actually, even with your comment it took me a while. It surprised me that `function` objects have `length` and what it means. Maybe saying something like
```js
// !children.length ensures that the function is a getter (takes 0 parameters)
```

GiyoMoon:
Which isn't 100% true either. Variadic parameters or parameters with a default don't count towards this length. Its a bit of a hack to catch * some *  at least functions which aren't accessors.

```ts
const myFunction = () => {}
console.log(myFunction.length) // 0

const myFunction = (myProp) => {}
console.log(myFunction.length) // 1

const myFunction = (myProp, mySecondProp) => {}
console.log(myFunction.length) // 2

const myFunction = (myProp = 'withDefault') => {}
console.log(myFunction.length) // 0

const myFunction = (myProp, mySecondProp = 'withDefault') => {}
console.log(myFunction.length) // 1

const myFunction = (...variadicParam) => {}
console.log(myFunction.length) // 0
```

ryansolid:
Yeah I mean I'm going to merge it. It is a best try. Honestly things this gets called on tend to all be accessors but there is some world where someone is doing something odd, and this avoids something bad happening.

PR #1920, opened 2023-10-16
Add missing `indexArray` to server-side runtime.

thetarnav:
- Add missing `indexArray` to server-side runtime.
- prevent `cannot access property "length" of undefined` when passing `undefined` to `mapArray` on the server


ryansolid:
Looks good thanks.

PR #1948, opened 2023-11-07
Add types directory export for each package

thetarnav:
When selecting a `"moduleResolution"` option is tsconfig other than `"node"` you cannot import the other not-exported from main modules types, which is useful for importing a very specific internal type, eg in devtools. 🙃

Screenshot from 2023-11-07 18-35-55

But when the types directories are explicitly exported one can import a specific `.d.ts` file.

image

Another solution would be to export every internal type from main modules, which is fine too.

### DIFF
--- a/.changeset/cyan-dolls-lay.md
+++ b/.changeset/cyan-dolls-lay.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks

PR #1958, opened 2023-11-17
TypeScript: Context.Provider value should not be assignable to undefined in the default case

deluksic:
## Summary
Context providers should not accept `undefined` by default. Example of previously correct code we'd like to avoid:
```tsx
const MyContext = createContext<string>()
const someValue: string | undefined

<MyContext.Provider value={someValue}>
  ...
</MyContext>
```

Assumption is that if you **really** wanted to allow providing `undefined` you would have written:
```ts
const MyContext = createContext<string | undefined>()
```

This does not change the fact that `useContext` returns `undefined` if no default was provided:
```ts
const MyContext = createContext("hello")
const MyContextWithoutDefault = createContext<string>()
const MyContextExplicit = createContext<string | undefined>("string")

const ctx: string = useContext(MyContext)
const ctx: string | undefined = useContext(MyContextWithoutDefault)
const ctx: string | undefined = useContext(MyContextExplicit)
```

When combining Resources with Context, it is very easy to forget to test if a given resource is available before assigning it to the provider value.

I wrote an additional test to test the usage patterns.
This is type-level only change.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1130,10 +1130,10 @@ export function registerGraph(value: SourceMapValue): void {
 export type ContextProviderComponent<T> = FlowComponent<{ value: T }>;
 
 // Context API
-export interface Context<T> {
+export interface Context<T, D = T> {
   id: symbol;
   Provider: ContextProviderComponent<T>;
-  defaultValue: T;
+  defaultValue: D;
 }
 
 /**
@@ -1158,7 +1158,7 @@ export interface Context<T> {
 export function createContext<T>(
   defaultValue?: undefined,
   options?: EffectOptions
-): Context<T | undefined>;
+): Context<T, undefined>;
 export function createContext<T>(defaultValue: T, options?: EffectOptions): Context<T>;
 export function createContext<T>(
   defaultValue?: T,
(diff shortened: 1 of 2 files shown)


otonashixav:
I don't believe this requires adding a generic; regardless it will be a breaking (type) change. You'd need a second `Context` type for the defaultless overload, with `useContext` accepting either. There are a few different ways to change `useContext` for this purpose but overloading it probably makes the most sense. The second overload should accept both while the first accepts the narrower context.

deluksic:
Yes, I considered this as well, just thought this approach was a bit simpler. I'm not super worried about this detail, as long as the user-facing API is improved. Should I create another PR with your suggestion and we compare?

> regardless it will be a breaking (type) change.

Yeah, but hopefully people have not relied on this, and would actually like to see the error otherwise (given that it appears only when strict enabled). For me it was surprising that doing
```tsx
<Context.Provider value={undefined}>
```
actually reverts to the default value and doesn't literally set undefined. This could be reflected in the types, but I wonder why it is like this in the first place.

ryansolid:
Thanks for the PR, and sorry this sat for so long.

I’m going to close this as stale for Solid 2. The context model has changed there: default-less contexts now throw when read without a Provider, and `useContext(createContext<T>())` is typed as `T`, so this exact issue is no longer modeled the same way.

For Solid 1, I do agree with the underlying concern. Provider values for default-less contexts should not accidentally accept `undefined` unless the context type explicitly includes it. However, I also agree with the feedback above that the cleaner type fix would be a separate default-less context type plus `useContext` overloads, rather than adding another generic parameter to `Context`.

So I don’t think this PR is something we should merge as-is. If we address this on the 1.x line, it should be as a fresh type-only fix with that overload-based design.

PR #1964, opened 2023-11-23
fix: createSignal not found when bundled

CrabNejonas:
In some circumstances a global function export (like `createSignal`) can be transformed by rollup into a let binding, which will break the global evaluation of `transPending` since then `createSignal` will be defined after being used.
I.e. the code from `signal.ts` which looks like this:
```ts
const [transPending, setTransPending] = createSignal(false);

// ... some other code

export function createSignal(){}
```

will be translated to 

```js
let createSignal;

const [transPending, setTransPending] = createSignal(false);

export function createSignal(){}
```
which obviously won't work anymore since the let definition isn't hoisted. 

While I believe this bug should fundamentally be solved in rollup itself this is a quick, rather painless fix (it's interesting that rollup isn't smart enough to reason about in-file dependencies in this way) 

The app that this came up in is still closed-source unfortunately, and this bug - being a codegen-bug - proofed itself to be hard to reproduce, but I hope the above reasoning is easy enough to follow.

### DIFF
--- a/.changeset/nasty-coins-call.md
+++ b/.changeset/nasty-coins-call.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix: createSignal not found when bundled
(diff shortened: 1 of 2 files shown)


ryansolid:
Thank you

PR #2020, opened 2024-01-07
fix #2017 reconcile fails to replace nested array with object

### LINKED ISSUE #2017: `reconcile` fails to replace nested array with object
### Describe the bug

Basically the reverse of [this issue](https://github.com/solidjs/solid/issues/1973). Using reconcile to replace a nested value that is an array with an object still leaves it as an array.

https://playground.solidjs.com/anonymous/d43decad-3911-4beb-99e0-6a166b6655c6

```ts
  const [store, setStore] = createStore<{ value: [] | {} }>({ value: [] });
  setStore(reconcile({ value: {} }));
  expect(Array.isArray(store.value)).toBe(false);
```

### Expected behavior

Expect the array to be replaced with an object, works without reconcile.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Vivaldi
- Version: 6.5

### Additional context

I don't think the playground version of solid has been updated to use the latest version so obj -> array bug exists there, but this bug exists both on older and latest (1.8.8) version.

takeramagan:
1. fix #2017 reconcile fails to replace nested array with object
2. add tests

### DIFF
--- a/packages/solid/store/src/modifiers.ts
+++ b/packages/solid/store/src/modifiers.ts
@@ -14,9 +14,16 @@ function applyState(
   merge: boolean | undefined,
   key: string | null
 ) {
-  const previous = parent[property];
+  let previous = parent[property];
   if (target === previous) return;
   const isArray = Array.isArray(target);
+  if (Array.isArray(previous) && !isArray) {
+    parent[property] = {};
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks. I went with a smaller fix that I hope works as well (passes the tests).

PR #2026, opened 2024-01-11
feat: remove `castError`

lxsmnsyc:
# THIS IS A BREAKING CHANGE

This PR changes the behavior of Solid's error handling, mainly on the concept of "casting errors". Solid has an internal mechanism wrapping thrown values if the value isn't an `Error` instance.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -642,7 +642,9 @@ export function createResource<T, S, R>(
         }, false);
       } else completeLoad(isSuccess, v);
     }
-    return v;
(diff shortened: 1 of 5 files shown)


ryansolid:
Why? It was this way before. We intentionally went to normalizing errors. I admit it was mostly for serialization so I could see the argument for reverting it now. But on the other hand I feel as much temptation to normalize farther to prevent serialization errors for things that random libraries pass us that can't be serialized.

lxsmnsyc:
@ryansolid it's not exactly that way before. We used to check if the error is truthy, and not if the error existed.

Issue #1746, opened 2023-05-29
`createMutable` getters do not work if they're on the prototype


AFatNiBBa:
### Describe the bug

I tried to do the following:
```tsx
import { render } from "solid-js/web";
import { createEffect } from "solid-js";
import { createMutable } from "solid-js/store";

function Test() {
  const m = createMutable({
    a: 1,
    get b() { return this.a * 4; }
  });

  console.log(">>>", m.a, m.b);
  m.a++;

  createEffect(() => {
    console.log(">>>", m.a, m.b);
  });

  return (
    <button type="button" onClick={() => m.a++}>
      {m.a} - {m.b}
    </button>
  );
}

render(Test, document.getElementById("app")!);
```
This worked as espected, the "b" getter updated without problems every time "a" had changed
```tsx
import { render } from "solid-js/web";
import { createEffect } from "solid-js";
import { createMutable } from "solid-js/store";

class A {
  a = 1
  get b() { return this.a * 4; }
}

function Test() {
  const m = createMutable(new A());

  console.log(">>>", m.a, m.b);
  m.a++;

  createEffect(() => {
    console.log(">>>", m.a, m.b);
  });

  return (
    <button type="button" onClick={() => m.a++}>
      {m.a} - {m.b}
    </button>
  );
}

render(Test, document.getElementById("app")!);
```
But then I tried putting the "b" getter on a class and not only the reactivity stopped, but the getter stopped working completely giving only its first value and ignoring the changes to "a"
(This bug is perfectly reproducible on the playground but NOT only there)

### Your Example Website or App

The solid playground is ok

### Steps to Reproduce the Bug or Issue

Just run the code and click the button, you will notice something is off

### Expected behavior

The output of the first example

### Platform

- OS: Windows
- Browser: Opera
- Version: 98 I think

ryansolid:
Interesting.. technically we don't wrap classes and only wrap own properties to prevent infinite proxy recursion loops into built-ins. This is interesting though. It definitely isn't what I was thinking when adding this feature but it is probably worth looking at whether this is feasible to support without causing too many issues with edge cases.

takeramagan:
In wrap funtion, for class component, Object.getOwnPropertyDescriptors doesn't return the getter and setter. I create a MR to fix it.

PR #2027, opened 2024-01-12
fix createMutable getters do not work if they're on the prototype #1746

takeramagan:
issue related: https://github.com/solidjs/solid/issues/1746


takeramagan:
@ryansolid  Updated test, please review. if you remove the changes, then the test will fail

ryansolid:
Thank you. This is probably one of those features it would have taken me forever to get around to.

PR #2062, opened 2024-02-07
Lock on the cleanNode function

arvid220u:
## Summary

It is possible for `cleanNode` to be called recursively. This PR adds a simple `isCleaning` lock to ensure that `cleanNode` isn't called on the same node twice.

Minimal example: https://playground.solidjs.com/anonymous/9f860e26-d0a3-476d-8b20-82ebff4763d7 (when you click "hide", you get a `Cannot read properties of null` error which comes from `cleanNode` being called twice on the same node).

I believe this error is most likely to happen when your entire app isn't in Solid, but you're instead mounting several Solid roots inside an existing application. We have global state controlling which Solid roots and components to show, and a `syncState` function that reads from the global state and disposes/creates Solid roots based on that. If component `A` has a `Show` that causes component `B` to be disposed, and component `B` has an `onCleanup` that calls the global `syncState` which disposes the entire Solid root that `A` is rendered in, then `A` will error in the `cleanNode` function.

I believe that this change should solve this problem, but I would love input on whether this change is reasonable at all, or whether it may have side effects that I haven't considered.

Thank you!

Would love advice on where you think it'd be best to add a test for this — didn't do it yet, because would first love input on whether this change makes sense.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -90,6 +90,7 @@ export interface Owner {
   context: any | null;
   sourceMap?: SourceMapValue[];
   name?: string;
+  isCleaning?: boolean;
 }
 
 export interface Computation<Init, Next extends Init = Init> extends Owner {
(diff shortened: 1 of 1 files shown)


ryansolid:
I think it does. The only challenge is concurrency. That being said I don't think it changes the desire for this. Just it complicates things a bit.

ryansolid:
Thanks for the PR, and sorry this sat for so long.

I’m going to close this as stale for Solid 2. The owner/disposal system has been rewritten in `@solidjs/signals`, so this patch no longer maps cleanly to the current implementation.

For the 1.x line, I agree with the underlying concern that re-entrant cleanup can be dangerous, but I don’t think I’d merge this exact fix. A broad `isCleaning` lock can prevent the crash, but it can also silently skip cleanup work if the same owner is reached again during disposal.

If we patch this in 1.x, I’d prefer a narrower fix that snapshots/detaches the child lists before recursing, along with a minimal regression test. That avoids reading mutable `owned` / `tOwned` fields after a nested cleanup may have cleared them, without adding a general “already cleaning” escape hatch.

So closing this PR as stale, but the cleanup re-entry case is useful context if we revisit a 1.x bugfix.

PR #2066, opened 2024-02-09
Fix `DistributeOverride` losing type

BierDav:
## Summary

The problem currently is that `DistributeOverride` looses the type that is `undefinable` completely which results in bad developer experience. Additionally I have run `pnpm format` which resulted in a few files get changed from `crlf` to `lf`

```typescript
const foo: DistributeOverride<{}, { bar?: string }> = {};
console(foo.bar); // before this change `bar` was unresolved, because only `{}` was forwarded, now the types get combined and bar is resolved as `string | undefined`
```

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -139,7 +139,11 @@ const propTraps: ProxyHandler<{
   }
 };
 
-type DistributeOverride<T, F> = [T, F] extends [undefined, undefined] ? F & T : (F & T extends undefined ? never : F & T);
+type DistributeOverride<T, F> = [T, F] extends [undefined, undefined]
+  ? F & T
+  : F & T extends undefined
+  ? never
+  : F & T;
 type Override<T, U> = T extends any
   ? U extends any
     ? {
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -139,7 +139,7 @@ const propTraps: ProxyHandler<{
   }
 };
 
-type DistributeOverride<T, F> = T extends undefined ? F : T;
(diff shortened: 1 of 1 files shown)


BierDav:
I hope for a fast review, because it is literally only one line of code

ryansolid:
Unfortunately I am not confident in TS enough to do TS reviews myself so they are always by committee and take a bit longer.

ryansolid:
Ok had some feedback. 

I was informed:
>  it's a helper used by MergeProps to merge a single property of two props types, basically it converts undefined in a union of types T into F.

Do you have a concrete example of what you are trying to do and how it is failing to do that? Because my understanding this in an internal-only helper and it is working as intended for the use case. So it could be doing the wrong thing but it is hard to see without an example.

BierDav:
I have now gone through my issue again and noticed the deeper problem thus making this pr irrelevant and also wrong.
You see my problem was that
```
const foo: { bar?: {test?: string} } = {};
const baseProps = mergeProps({bar:{}}, foo)
console(baseProps.bar); // bar has now the type `{test?:string} | {}`, because `{}` would allow something like `{test:4}` where test is definitely not a string, therefore typescript cannot guarantee anything.
```
now if you want to resolve this you need to tell typescript explicitly that `{}` cannot have any properties with a cast to `{} as { [K in any]?: never }`. In my opinion this is a bit a foundational problem of typescript, because the type `{}` is not the closest fit for `{}` it actually would be `{ [K in any]?: never }`, because a object that has no properties at initialization where no properties can be defined later is a object which never can have properties.

Unfortunately typescript works that way therefore we have to always cast it like that:
```
const baseProps = mergeProps({bar:{} as { [K in any]?: never }}, foo)
```

PR #2081, opened 2024-02-22
Adjust some JSDocs

marvin-j97:
- Make casing consistent for all docs.
- Removes unnecessary prefixes in some docs like `onMount`


ryansolid:
Thank you. Looks great.

Issue #1661, opened 2023-03-30
 isWrappable(new Class()) return false


Reinhard2019:
### Describe the bug

In this issue: https://github.com/solidjs/solid/issues/19
@ryansolid Decided to only wrap plain Objects and Arrays like MobX.

But i think a class object also should be wrap.

What do you think?

### Your Example Website or App

https://stackblitz.com/edit/typescript-mkiunp?file=index.ts

### Steps to Reproduce the Bug or Issue

class Obj {}
console.log('new Obj():', isWrappable(new Obj()))

### Expected behavior

return true

### Screenshots or Videos

<img width="1471" alt="image" src="https://user-images.githubusercontent.com/13881726/228853565-e8b95703-4c75-40c8-be7d-c536aced9c61.png">

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

ryansolid:
No this is by design. Is wrappable false means that it can only wrap it atomically instead of each property. We did this to avoid wrapping built in classes like Date, Regexp, HTMLElement. And things like Set and Map which could be wrapped differently depending on intent.

Reinhard2019:
> No this is by design. Is wrappable false means that it can only wrap it atomically instead of each property. We did this to avoid wrapping built in classes like Date, Regexp, HTMLElement. And things like Set and Map which could be wrapped differently depending on intent. The original issue explains how this is a much easier identification model than trying to exempt specific classes.

But when i use createStore({ obj: new Obj() })，store.obj will not be monitored. 

Can you give me an escape window?

Reinhard2019:
@ryansolid Do you think it is possible to add a symbol to solve this problem? Like this:
![image](https://user-images.githubusercontent.com/13881726/229263613-cf00b898-19cf-4f1c-bac2-27be1b9ad2c3.png)
If you think it's ok, i will commit a pr. thx!

PR #2128, opened 2024-04-14
Allow classes to be wrapped inside a store

timbasel:
## Summary

This PR allows classes to be wrapped inside a store.
It implements the very small change outlined by @Reinhard2019 in the Issue #1661. 

Even if the issue was closed as `wontfix` I do believe that the functionality is tremendously useful.
It has no effects on existing code because it is an opt in on a class by class bases.

To make a class wrappable you just define the `$WRAP`  property (I've added a minimal helper function to do this) in the constructor of the class.
```ts
class Test {
  constructor() {
    makeWrappable(this);
  }
}
const [store, setStore] = createStore(new Test())
```
You can even make just individual instances wrap, by calling `makeWrappable` just on the created instance.

### DIFF
--- a/.changeset/empty-cars-jog.md
+++ b/.changeset/empty-cars-jog.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Reordering setter overloads
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -179,12 +179,12 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: typeof Owner)
 export type Accessor<T> = () => T;
 
 export type Setter<in out T> = {
-  <U extends T>(value: Exclude<U, Function> | ((prev: T) => U)): U;
   <U extends T>(
     ...args: undefined extends T ? [] : [value: Exclude<U, Function> | ((prev: T) => U)]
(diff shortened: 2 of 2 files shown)


ryansolid:
Directionally I'm not opposed. I've considered adding symbol or having helpers to do the wrapping. Technically you can add the PROXY symbol to your class and define your own wrapper today.

My thinking though is like MobX there is perhaps more involved than just auto wrapping every property. I think we should consider something like MobX `makeObservable` if we go this way. And I think that requires a bit more thought than just enabling it.

timbasel:
@ryansolid This PR has been stale for a long time. Is there anything more I can do.

PS: I have just thought about the ability to wrap only some fields would already work by just passing only the desired fields into `makeWrappable` (example below). If this is desired it would probably be good to allow `makeWrappable` to take multiple arguments to prevent multiple calls for multiple fields, but that change is trivial.

```ts
class Test {
  field1: string;
  field2: string;

  constructor() {
    makeWrappable(this.field1);
  }
}
```

timbasel:
@ryansolid This change is definitely not necessary. Coming back to my project I have just figured out you can just call `createStore` in the constructor of the class to create instances that are wrapped with the `$PROXY`.

```ts
class Test {
  constructor() {
    createStore(this)
  }
}

const [instance, setInstance] = createStore({ nested: new Test() });
```

PR #2143, opened 2024-04-23
Point API references to new docs entries

atilafassina:
## Summary

The documentation entries at `solidjs.com/docs` are not receiving updates anymore.
So even with the new docs app being still in beta, I believe it's best to already point the JSDoc `@description` suggestions to `docs.solidjs.com`


packages/solid/store/src/store.ts
```diff
@@ -493,7 +493,7 @@ export interface SetStoreFunction<T> {
 /**
  * Creates a reactive store that can be read through a proxy object and written with a setter function
  *
- * @description https://www.solidjs.com/docs/latest/api#createstore
+ * @description https://docs.solidjs.com/reference/store-utilities/create-store
```
atilafassina, line 496: this entry depends on:

https://github.com/solidjs/solid-docs-next/pull/694

packages/solid/src/reactive/signal.ts
```diff
@@ -1750,7 +1750,7 @@ type TODO = any;
  *
  * * If the error is thrown again inside the error handler, it will trigger the next available parent handler
  *
- * @description https://www.solidjs.com/docs/latest/api#onerror
+ * @description https://www.solidjs.com/docs/latest/api#onerror | https://docs.solidjs.com/reference/reactive-utilities/catch-error
```
danieljcafonso, line 1753: Just out of curiosity, Is there a reason why we kept the previous one here?

atilafassina: as far as I can see, `onError` has not been taken to the new docs because the api is deprecated. So I left the old link and added the link to the recommended alternative

ryansolid:
Looks great thanks.

PR #2226, opened 2024-07-18
optimize useContext call

mdynnl:
## Summary
No benchmarks are ran as this should not need one. Feel free to do though 😂.

`pnpm test`

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1185,8 +1185,9 @@ export function createContext<T>(
  * @description https://docs.solidjs.com/reference/component-apis/use-context
  */
 export function useContext<T>(context: Context<T>): T {
-  return Owner && Owner.context && Owner.context[context.id] !== undefined
-    ? Owner.context[context.id]
+  let value: undefined | T;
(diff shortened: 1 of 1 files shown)


ryansolid:
Nice that saves 8 bytes and is probably a tiny bit faster

packages/solid/src/reactive/signal.ts
```diff
@@ -1185,8 +1185,9 @@ export function createContext<T>(
  * @description https://docs.solidjs.com/reference/component-apis/use-context
  */
 export function useContext<T>(context: Context<T>): T {
-  return Owner && Owner.context && Owner.context[context.id] !== undefined
-    ? Owner.context[context.id]
+  let value: undefined | T;
+  return Owner && Owner.context && (value = Owner.context[context.id]) !== undefined
```
maciek50322, line 1189: how about
```tsx
Owner?.context?.[context.id] ?? context.defaultValue
```

ryansolid: I've been avoiding optional chaining as we don't transform JS in the repo and some platforms down't support optional chaining. We've had issue trying to introduce it in the past.

Issue #1895, opened 2023-10-02
Uncaught TypeError: evalConditions is not a function or its return value is not iterable


mathieuprog:
### Describe the bug

I started to have this cryptic error message when starting my app:

> Uncaught TypeError: evalConditions is not a function or its return value is not iterable
    at Switch.createMemo.name [as fn] (dev.js:1517:33)
    at runComputation (dev.js:691:22)
    at updateComputation (dev.js:674:3)
    at Object.readSignal (dev.js:607:67)
    at resolveChildren (dev.js:983:82)
    at resolveChildren (dev.js:983:66)
    at createMemo.name [as fn] (dev.js:573:33)
    at runComputation (dev.js:691:22)
    at updateComputation (dev.js:674:3)
    at createMemo (dev.js:251:10)
    
There is no indication of the problem in my code or the line that causes this error.

### Platform

- OS: Windows
- Browser: Chrome 117
- SolidJS: 1.7.12

Update:

The cause seems to be this line:
```jsx
<Match when={!currentUser.latest && currentUser.loading}>
```
If I remove the condition `!currentUser.latest` (or `!currentUser()`), the error disappears. Basically what I am trying to match on is: "the current user is loading for the first time (i.e. has not been previously fetched)"

So there are 2 issues:
1. cryptic error message (had to remove one by one every `Match` in different `Switch` elements to see which caused the error);
2. I don't see why I can't check if the resource value is undefined or has been previously fetched in a `Match`.

ryansolid:
Yeah this seems like an odd Transition or Suspense issue. Without a reproduction it might be harder to determine the cause exactly. `latest` does bypass Suspense which makes me think so. The thing is the error is coming form switch match which always returns an array... but initial value doesn't have one, which could be where it is choking. It should run it once but if there is something odd going on we might have our issue.

mathieuprog:
I switched to using `Show` to match on this condition. I tried to reproduce using the playground but can't, the real-world case also involves routing and other features - it is hard to replicate the case in the playground.

gidra5:
I stumbled upon the same error message today, and figured out minimal reproduction. Basically whenever condition computation throws switch will fail with that error. For me exact issue was accessing field on undefined, which was obscured by this error. When I removed switches as was suggested, I was able to see it.
Take a look at playground below. Although error is from minified code, which obscures it, I verified locally that its the same error.

[...]

PR #2227, opened 2024-07-18
fix catchError propagation

mdynnl:
## Summary
Opening this up just for the idea. I initially used a separate `const ErrorHandled = Symbol()` but changed to `ERROR`.

fixes https://github.com/solidjs/solid/issues/1895

IMO, `catchError` should work like one we have in the language. It should stop further execution inside the boundary. Everything outside should work as usual.
https://playground.solidjs.com/anonymous/a42cb5b3-37b9-4469-9b48-a5fc0d50937e

Added a narrowed down test for `catchError` though I'm not sure if there's any other edge cases.
`pnpm test` seems to pass including the one added.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1380,7 +1380,8 @@ function runComputation(node: Computation<any>, value: any, time: number) {
     }
     // won't be picked up until next update
     node.updatedAt = time + 1;
-    return handleError(err);
+    handleError(err);
+    throw ERROR;
   } finally {
     Listener = listener;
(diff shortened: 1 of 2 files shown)


ryansolid:
I'm going to link a conversation I had with Jin which lead to the last change in error behavior: https://dev.to/ninjin/comment/206d2

I'm not saying this doesn't uphold this but I want to review it in this context. I was thinking that moving towards lazy computation would mean we wouldn't want to error out everything but just things that depended on the broken stuff.. That being said we aren't there in current Solid so we have to make a decision.

PR #2237, opened 2024-07-25
chore: remove vitest globals types from base tsconfig

Gpx:
## Summary

The `tsconfig.json` file includes types for `vitest/globals`. Those types are already included in `tsconfig.test.json` which makes them available in test files. If we include the types in `tsconfig.json` we make them available in _all files_.

This is, for example, what happens in `packages/solid/src/index.ts`:

Screenshot 2024-07-25 at 16 48 47

After this change a TypeScript error is shown instead:

Screenshot 2024-07-25 at 16 49 04

I run `pnpm run build && pnpm run test && pnpm run types`.
This is my first PR to this project so please let me know if I missed something.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,9 +6,7 @@
     "newLine": "LF",
     "moduleResolution": "node",
     "strict": true,
-    "jsx": "preserve",
-    // Allows TS to see the Vitest globals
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

PR #2267, opened 2024-09-01
make a named export for solid-js/html for autocompletion in IDEs

trusktr:
## Summary

Clearly-named exports appear as autocomplete results in IDEs like VS Code. This allows someone to, for example, write the following JS code with `|` representing the cursor location:

```js
function Comp() {
  return html|
}
```

and IDEs that use TypeScript Language Server (f.e. VS Code) will propose to automatically add `import {html} from 'solid-js/html'`, converting the code to the following when the user hits the Tab key (default in most IDEs) to accept the auto-completion:

```js
import {html} from 'solid-js/html'

function Comp() {
  return html|
}
```

### DIFF
--- a/.changeset/grumpy-jars-invite.md
+++ b/.changeset/grumpy-jars-invite.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+expose `html` with a named export from `solid-js/html`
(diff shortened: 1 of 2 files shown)


trusktr:
Actually I just tested, and autocompletion will work as long as the app already imports the default `html` somewhere first. So this change is not needed for that.

This does make it easier to write things like `export * from `solid-js/html`, so I'd say keep it, causes no harm.

ryansolid:
I'm glad this wasn't a blocker. It seems reasonable to consider for Solid 2.0. But given Solid 1 might not see another minor. I'm ok just to leave things as they are.

Issue #1984, opened 2023-12-09
Importing solid-js/web (as ESM, from Node.js) throws a SyntaxError


vinsonchuong:
### Describe the bug

When I import `solid-js/html` as an ES module from Node.js, I get the following error:

```
SyntaxError: The requested module 'solid-js/web' does not provide an export named 'Aliases'
```

### Your Example Website or App

n/a

### Steps to Reproduce the Bug or Issue

For example, in the Node.js REPL:

```
>> node
Welcome to Node.js v21.1.0.
Type ".help" for more information.
> await import('solid-js/html')
file:///home/vinsonchuong/projects/test-tube/node_modules/solid-js/html/dist/html.js:15
  Aliases,
  ^^^^^^^
Uncaught:
SyntaxError: The requested module 'solid-js/web' does not provide an export named 'Aliases'
```

### Expected behavior

I expect the `import` to succeed and not throw an error.

### Platform

- OS: Linux
- Browser: Running in Node.js v21.1.0
- Version: 1.8.7

### Additional context

I'm looking to write some custom tooling for writing automated tests from Node.js and was surprised to see this error.

After some searching, I found this issue: https://github.com/solidjs/solid/issues/1483, which clued me into the fact that there are separate builds for the browser and for Node.js.

It looks like because the `html` tag function returns a DOM `Node`, `solid-js/html` is a browser-only package. I am, however, able to proceed if I configure Node.js to prefer browser packages:

```
>> node --conditions browser
Welcome to Node.js v21.4.0.
Type ".help" for more information.
> await import('solid-js/html')
[Module: null prototype] { default: [Function: html] }
```

Because `solid-js/html` imports `solid-js/web`, there's no way to specify that the browser packages should be used without passing in arguments to Node.js. To make this easier, could `solid-js/html` directly import the browser build of `solid-js/web` instead? There are probably trade-offs to doing that, but I'm hoping to have easy ways to run Solid from a Node.js + JSDOM environment.

ryansolid:
So far we've pushed people to use "browser" export conditions for this. Like in `solid-jest` when doing testing we use the browser condition. We could make it use the browser version. Although there are 2 versions of Solid as well, and it is less clear what version you want. Like testing I want the browser reactive version, but for doing some sort of SSR you probably want the inert server version. To be fair right now because they all connect to the same condition it isn't that configurable. 

[...]

PR #2269, opened 2024-09-01
update dom-expressions, solid-js/web, solid-js/html, solid-js/store to make the exports isomorphic

trusktr:
Depends on:

- [x] https://github.com/ryansolid/dom-expressions/pull/345 before this can be merged

## Summary

This PR will make exports between client and server match ("isomorphic"), to stop people from running into build errors.

We will update the dom-expressions version in this PR once this dom-expressions PR is merged:

- https://github.com/ryansolid/dom-expressions/pull/345

The dom-expressions PR adds a version of client-side APIs to server.js that throw when they are used on the server side. This bring the exports of server.js closer to parity with client.js so that we can avoid issues like:

- https://github.com/solidjs/solid/issues/1984
- https://github.com/solidjs/solid-start/issues/1618
- https://github.com/solidjs/solid-start/issues/1614

Furthermore we update `solid-js/store` to also have matching exports between server and client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "babel-plugin-jsx-dom-expressions": "^0.38.5",
     "coveralls": "^3.1.1",
     "csstype": "^3.1.0",
-    "dom-expressions": "0.38.5",
+    "dom-expressions": "https://gitpkg.vercel.app/trusktr/dom-expressions/packages/dom-expressions?update-exports",
     "hyper-dom-expressions": "0.38.5",
     "jsdom": "^21.1.1",
     "lit-dom-expressions": "0.38.5",
--- a/packages/solid/store/src/server.ts
+++ b/packages/solid/store/src/server.ts
@@ -1,5 +1,19 @@
 import type { SetStoreFunction, Store } from "./store.js";
 
+export type {
+  ArrayFilterFn,
+  DeepMutable,
+  DeepReadonly,
+  NotWrappable,
+  Part,
+  SetStoreFunction,
+  SolidStore,
+  Store,
+  StoreNode,
+  StorePathRange,
+  StoreSetter
+} from "./store.js";
+
 export const $RAW = Symbol("state-raw");
 
 export function isWrappable(obj: any) {
@@ -103,6 +117,10 @@ export function createMutable<T>(state: T | Store<T>): T {
   return state as T;
 }
 
+export function modifyMutable<T>(state: T, modifier: (state: T) => T) {
+  notSup();
+}
+
 type ReconcileOptions = {
   key?: string | null;
   merge?: boolean;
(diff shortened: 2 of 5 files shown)


packages/solid/web/server/index.ts
```diff
@@ -12,12 +12,15 @@ export {
...
 } from "solid-js";
 
 export const isServer: boolean = true;
 export const isDev: boolean = false;
 
+// Types for these come from dom-expressions/src/server.d.ts
+// These override the functions from dom-expressions that throw on the serverside.
 export function render() {}
 export function hydrate() {}
 export function insert() {}
```
trusktr, line 26: These functions override the error-throwing functions that are now exported from dom-expressions, making this particular set of functions be no-ops instead of throwing errors.

All the other newly-added functions in dom-expressions continue to throw errors if they are not overriden here, which is non-breaking. 

Removing the set of functions here could be breaking if anyone is already depending on them being no-ops, but the new functions added to dom-expressions/src/server.js never existed there, so the introduction of those new error-throwing functions cannot be breaking because no one used them before.

If desired, those additional functions in dom-expressions can be overriden here to make them no-ops instead of error-throwing.

package.json
```diff
@@ -37,7 +37,7 @@
     "babel-plugin-jsx-dom-expressions": "^0.38.5",
     "coveralls": "^3.1.1",
     "csstype": "^3.1.0",
-    "dom-expressions": "0.38.5",
+    "dom-expressions": "https://gitpkg.vercel.app/trusktr/dom-expressions/packages/dom-expressions?update-exports",
```
trusktr, line 40: This is a quick way to test a package from any git repo, including sub-folders.

See https://gitpkg.vercel.app

trusktr: All tests are passing. Anything I may not have thought of that needs testing?

packages/solid/store/src/server.ts
```diff
@@ -103,6 +117,10 @@ export function createMutable<T>(state: T | Store<T>): T {
   return state as T;
 }
 
+export function modifyMutable<T>(state: T, modifier: (state: T) => T) {
+  notSup();
+}
```
trusktr, line 122: Should this throw? Or should it be a no-op? I think that a no-op could be misleading, f.e. someone would think it worked, then log values only to see they didn't change.

packages/solid/web/server/index.ts
ryansolid: This seems fine to me and honestly I think I'm ok even making them throw if called. I don't see any reason why these should ever be called on the server with a server build.

packages/solid/store/src/server.ts
ryansolid: To be fair a lot of these methods still work on the server but just don't do anything. Like modifyMutable being missing is an oversight and should just function the same as in the client.

PR #2280, opened 2024-09-12
adds `reconcileWithKeys`

titoBouzout:
## Summary

A suggestion to add `reconcileWithKeys` that @katywings and I have requested in the past. In some situations you want to use reconcile with objects that have keys differently named to `id`. 

```js
       const [target, setStore] = createStore({
        c: {
          a: [
            { ida: 1, name: "1" },
            { ida: 0, name: "0" }
          ],
          b: [
            { idb: 1, name: "1" },
            { idb: 0, name: "0" }
          ]
        }
      });

      const ref = target.c.a[1];

      const source = {
        c: {
          a: [{ ida: 0, name: "0 modified" }],
          b: [{ idb: 0, name: "0 modified" }]
        }
      };

      /*
      setStore("c", reconcile(source.c));
      expect(target.c.a[0]).not.toBe(ref);
      */

      setStore(
        "c",
        reconcileWithKeys(source.c, {
          keys: {
            a: { _key: "ida" },
            b: { _key: "idb" }
          }
        })
      );
      expect(target.c.a[0]).toBe(ref);

      expect(target).toEqual({
        c: {
          a: [{ ida: 0, name: "0 modified" }],
          b: [{ idb: 0, name: "0 modified" }]
        }
      });
```

Hehe, Im running into an issue when trying to write tests for it, when I run `npm run test` I get this error:

```
solid-js:test:  FAIL  store/test/modifiers.spec.ts > reconcileWithKeys
solid-js:test: TypeError: Cannot assign to read only property 'Symbol(solid-proxy)' of object '#<Object>'
solid-js:test: ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```

It works in the playground, so I'm lost at to whats going on, any ideas? 🙏🏼 
https://playground.solidjs.com/anonymous/9167e413-e0b2-45bf-942a-3ca2153c0416

## References

https://discord.com/channels/722131463138705510/780502110772658196/971305532034011156
https://discord.com/channels/722131463138705510/780502110772658196/1278247899616247808

### DIFF
--- a/packages/solid/store/src/modifiers.ts
+++ b/packages/solid/store/src/modifiers.ts
@@ -6,13 +6,20 @@ export type ReconcileOptions = {
   key?: string | null;
   merge?: boolean;
 };
+export type ReconcileWithKeysOptions = {
+  keys?: object | null;
+  merge?: boolean;
+};
+const emptyKeys = /* #__PURE__*/ Object.create(null);
 
 function applyState(
   target: any,
   parent: any,
   property: PropertyKey,
   merge: boolean | undefined,
-  key: string | null
+  key: string | null,
+  keys: object | null,
+  useKeys: boolean | null
 ) {
   const previous = parent[property];
   if (target === previous) return;
@@ -43,7 +50,7 @@ function applyState(
           (key && previous[start] && target[start] && previous[start][key] === target[start][key]));
         start++
       ) {
-        applyState(target[start], previous, start, merge, key);
+        applyState(target[start], previous, start, merge, key, keys, useKeys);
       }
 
       const temp = new Array(target.length),
@@ -65,7 +72,7 @@ function applyState(
         for (j = start; j <= newEnd; j++) setProperty(previous, j, target[j]);
         for (; j < target.length; j++) {
           setProperty(previous, j, temp[j]);
-          applyState(target[j], previous, j, merge, key);
+          applyState(target[j], previous, j, merge, key, keys, useKeys);
         }
         if (previous.length > target.length) setProperty(previous, "length", target.length);
         return;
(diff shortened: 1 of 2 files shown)


titoBouzout:
Ah, I see what's going on, it's failing a test (as expected), but then its throwing an unrelated error, that put me off, and I wasn't understanding. I think I can continue writing tests..

ryansolid:
It's a cool Idea. There have been a few design variations on this. Not completely sure which makes the most sense. Mostly wonder if the changes to Stores in 2.0 will have bigger bearing on what the design ultimately is.

ryansolid:
Thanks for the PR. This is a real use case, and it’s one that came up a few times around `reconcile`: different nested collections often use different identity fields, so a single default key like `id` is not always enough.

I’m going to close this as stale for Solid 2, though. Stores changed significantly in 2.0, and keyed reconciliation is now centered around the current `reconcile(value, key)` / projection-store APIs rather than adding a separate `reconcileWithKeys` helper.

I also don’t think this exact API is the shape I’d want to carry forward. The nested `keys` object with `_key` sentinels is powerful, but it adds a second reconciliation API and leaves some design questions around path semantics, swapped array behavior, and how this should compose with the newer projection store model.

If we revisit this, I’d rather do it as a fresh Solid 2 design. The current `key` function may already cover many cases, since users can choose identity from the item shape itself. For more ergonomic object-based key maps, I’d first want to explore whether they can compile to a normal custom key function outside the core reconciliation path, so the common `reconcile` implementation stays small.

Issue #1973, opened 2023-12-05
`reconcile` problem when array should replace an object


jcmonnin:
### Describe the bug

I have a data structure coming as JSON from the server that is put in a store. When a new state from server comes, reconcile is used to update the store in order to keep fine grained updates.

When there is an existing object in store, which should be an array after store update, reconcile shows strange behavior.

### Your Example Website or App

https://playground.solidjs.com/anonymous/ab7db9d8-47b6-4fd7-8f0b-fda541030b92

### Steps to Reproduce the Bug or Issue

The example below shows the strange behavior of `reconcile`
```
const [store, setStore] = createStore({"a":{"b":1}});
setStore(reconcile({"a":[{"b": 1}]}));
console.log("with reconcile", JSON.stringify(store));
```
outputs:
```
with reconcile {"a":{"0":{"b":1},"b":1}}
```

### Expected behavior

When not using `reconcile` the store has the expected state:
```
const [store, setStore] = createStore({"a":{"b":1}});
setStore({"a":[{"b": 1}]});
console.log("without reconcile", JSON.stringify(store));
```
outputs:
```
without reconcile {"a":[{"b":1}]}
```

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

raviqqe:
The fix seems to be deployed to the npm registry. But I still see the problem. https://github.com/raviqqe/til/blob/9b08c2db269100e6eac19dd7c10ef73e4788cc86/solid-array-store/src/store.test.ts#L32

PR #2281, opened 2024-09-13
adds test cases for swapping objects/arrays on stores

titoBouzout:
## Summary
Was reading the `isArray` condition on `applyState` ..  investigating the issues here found this comment https://github.com/solidjs/solid/issues/1973#issuecomment-2152496877  
This other post may be relevant https://github.com/solidjs/solid/issues/2017#issuecomment-1881586257 I haven't considered you could patch an array at an index with an object describing the patch.

Admittedly, needing to swap object types seems weird, don't remember having run into this myself. In any case this PR adds two failing test cases for it. I tried to give it a shot but couldn't pass the test, so Im leaving this here.

### DIFF
--- a/packages/solid/store/test/modifiers.spec.ts
+++ b/packages/solid/store/test/modifiers.spec.ts
@@ -156,6 +156,32 @@ describe("setState with reconcile", () => {
     setStore(reconcile({ value: { q: "aa" } }));
     expect(store.value).toEqual({ q: "aa" });
   });
+  test("reconciles an object with an array", () => {
+    const [store, setStore] = createStore<{ value: {} | [] }>({
+      value: { foo: "bar" }
+    });
+
+    const value = [0, 1, 2];
+    setStore("value", reconcile(value));
+
+    expect(Array.isArray(store.value)).toBe(true);
+    expect(store).toEqual({
+      value: [0, 1, 2]
+    });
+  });
+
+  test("reconciles an array with an object", () => {
+    const [store, setStore] = createStore<{ value: {} | [] }>({
+      value: [0, 1, 2]
+    });
+
+    const value = { foo: "bar" };
+    setStore("value", reconcile(value));
+    expect(Array.isArray(store.value)).toBe(false);
+    expect(store).toEqual({
+      value: { foo: "bar" }
+    });
+  });
 });
 
 describe("setState with produce", () => {


ryansolid:
Yeah I admit I don't know I considered ever supporting this. Because we can't change the nature of the topmost proxy. We connect it to an object or array and we need to keep that reference.

titoBouzout:
The test seems to be intended to change the property `.value`, seems doable if the proxy is on the parent. I am not sure if on the implementation it is considered as topmost proxy. I get the idea that some stuff it's tracked by returning new proxies to the data, but how all plays together is complicated to think about.

I'm not suggesting, but theoretically I suppose you can proxy a `listener` object, hold a reference to it, and then swap the underlying data by changing its prototype. Something like this  https://playground.solidjs.com/anonymous/e2ab459b-7c0c-4583-9c72-7ca431cf100c (cant find a way to fool `Array.isArray` so probably has downsides)

ryansolid:
Yeah I read this wrong. This should be workable.

titoBouzout:
ok, I understood how this works, simplifying It's something on the lines of:

```js
// store.value === {}

setStore('value', reconcile([1, 2]))

function setStore(key, reconcile){
  // reconcile receives `store.value` which is `{}`
  reconcile(store[key]) 
}

function reconcile(newValue) {
  // newValue is `[1, 2]`
  return state => { 
    // state is `{}`
    state = newValue
  }
}
```

I presume we can error out when `state` and `newValue` are both object, but one it's an Array and the other isn't.

titoBouzout:
There's this to consider https://github.com/solidjs/solid/issues/2017#issuecomment-1881586257
Basically reconcile can patch an array at an index with an object that describes the patch
```js
setStore("value", {0: something });
```

Kind of an awkward api if you ask me

ryansolid:
Yeah I just don't know if it can be "solved".

Typical `setStore` an object is a patch instruction since we can't change the identity of the proxy itself top-level. If objects were not to merge we couldn't accept top level objects. In our case it happened that it started that way (based on React's setState) and continued out.

`reconcile` is a different beast. It is supposed to transform what is there to be exactly what you pass in. But we can't change the identity of the underlying proxy. We can change nested properties but we can change an array into an object or vice versa, otherwise we'd need to set a new proxy on all parents, which we can't get to. 

To me these are both isolated considerations. There could be a future where we only mutably change stores.. like `produce` by default and there again you still won't be able to change the nature of top-level proxy even without merging. But `reconcile`still will have this limitation. As long as you would want to be able to reconcile top-level (I believe we need to) then we don't have a way to change the identity of the top-most thing passed in.

titoBouzout:
> Typical setStore an object is a patch instruction since we can't change the identity of the proxy itself top-level. 

I understand the identity is to be preserved, but in my opinion this edge case is about losing that identity. It would be on the same lines as patching an array that has some objects with an empty array, the identity of the objects will vanish because the array It's now empty, so it's expected to lose information. From `[{u:1}, {u:2}, {u:3}]` to `[]` there's no identity to preserve for the `ù` objects.

I'm considering this edge case could happen not intentionally by the developer but when patching objects from API results (idk I am trying to imagine a situation). The issue I have with this, is that if stuff changes from object to array, then the API of these objects change (imagine trying to `.push` into an object) it will error out unexpectedly.

ryansolid:
I guess the only potential bug here from my perspective is that an object can be used to set a previously undefined key.

```js
const [store, setStore] = createStore({ value: null });

// should throw?
setStore("value", { something: "hey" })
```
Now the patch becomes the object. In a sense an object passed into `setStore` should be seen as an `assign` operation. So you can't assign `something` to null.

This however would conceptually be completely fine:
```js
const [store, setStore] = createStore({ value: null });

setStore({ value: { something: "hey" } })
```
Here you are assigning a new value to `value`.

I'm not going to say this is the most intuitive API but it is what the DSL is. 

From that perspective there is no problem with:
```js
const [store, setStore] = createStore({ value: [] });

// update index 0
setStore("value", { 0: "hey" })
```
Which is completely different then:
```js
const [store, setStore] = createStore({ value: [] });

// replace value with an object
setStore("value", { something: "hey" })
```

titoBouzout:
Thanks, that's very insightful. I've been thinking of this from the perspective of the user, as in `<>{store.value}</>`. While I wouldn't have expected the first example to throw I would welcome throwing on the cases that it can't handle.  So maybe from my suggestion it can throw when trying to set an array when it's an object, but accept an object when its an array (as its a describing patch) [unless I am misunderstanding]

ryansolid:
Thanks for digging into this. I’m going to close this as stale for Solid 2.

The distinction we worked through in the thread is now reflected in the current store behavior/tests. Reconcile can replace a nested value’s shape when it is applied at the parent level, for example reconciling `{ value: [...] }` into the store. In that case the parent can update the `value` reference.

What it can’t do is change the fundamental object-vs-array nature of the proxy that is passed in as the top-level state, e.g. `setStore("value", reconcile([...]))` when `value` is currently an object. At that point `reconcile` only has the existing `value` proxy; it cannot replace the parent’s reference from inside that proxy.

So I don’t think these failing tests describe behavior we can support in that form. The remaining improvement here is probably documentation around parent-level replacement vs path-level reconciliation, rather than changing the store semantics.

Issue #2302, opened 2024-09-22
Multiple dynamic imports always load the first component


coralet:
### Describe the bug

In Solid.js 1.8.19 and above, SolidStart applications have an issue where dynamic imports render only the first component multiple times when lazy loading different components. This issue does not occur in Solid.js 1.8.18, where rendering works as expected.

### Your Example Website or App

https://stackblitz.com/edit/solidstart-dynamic-import

### Steps to Reproduce the Bug or Issue

The following code outputs `AA` in Solid.js 1.8.19 while `AB` in 1.8.18.

```typescript
import { lazy, Suspense } from "solid-js";

const A = lazy(() => import("./components/a.tsx"));
const B = lazy(() => import("./components/b.tsx"));

export default () => (
  <Suspense>
    <A />
    <B />
  </Suspense>
);
```

### Expected behavior

Expect to output `AB` in Solid.js 1.8.19 and above.

### Platform

- OS: Windows
- Browser: Chrome
- Version: 129.0

PR #2303, opened 2024-09-23
use correct hydration id in server lazy

mdynnl:
fixes https://github.com/solidjs/solid/issues/2302

with the new hydration key generation introduced in https://github.com/solidjs/solid/commit/424a31a31a956ad84f7e49c6b654e290ebea4fac, it's not necessary to cut the `-` at the end anymore.


ryansolid:
Oh nice. I knew there was still some edge case missing. I wonder if this is the one that has been getting us and hadn't been narrowed down. Thank you.

PR #2318, opened 2024-10-01
Fix setter type compatibility with kobalte select and add tests

Huliiiiii:
## Summary
fix https://discord.com/channels/722131463138705510/1063803756388548709/1289935849932456036


ryansolid:
Thanks

PR #2333, opened 2024-10-08
Reordering setter overloads

Huliiiiii:
Made changes based on discord discussions.

Compared to the type before #2297, the current type only modifies the first overload to widen the type of the parameter to allow auto-completion for literal unions, and it shouldn't break anything now.

### DIFF
--- a/.changeset/empty-cars-jog.md
+++ b/.changeset/empty-cars-jog.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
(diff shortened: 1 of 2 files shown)


ryansolid:
I've merged a few of these now.. Are we good after this one?

Huliiiiii:
Yes.

ryansolid:
Thank you for all your work on this.

PR #2340, opened 2024-10-17
upgrade Show component - added prop checkObjectValues

madaxen86:
First of all - this PR is open for discussion

Before adding changes, documentation, etc. I just wanted to get some feedback if this has a chance to be merged.

-- no breaking change -- 

Show will just behave as before when `checkObjectValues` is not set / true.
Motivation: e.g.[From Discord]( https://discord.com/channels/722131463138705510/1295944738163392534/1295944738163392534)

Setting it to true will allow narrow the type of all values to NonNullable.
Imagine this case where you have a signal and store that relies in user input without initial values and you need to pass them to a child component which has strict non-null types.
With current Show typescript will complain that first and last might be undefined. So you either have to use the bang operator or use createMemo to narrow like 

```ts
const narrow = createMemo(()=> {
  if (selected() && store.data?.first && store.data?.second) return {selected:selected(),first:store.data?.first, second:store.data?.second}
  return false
}
```
And pass it to Show which now correctly narrows to non-null types.

With the added prop you can just:
```ts
const [selected,setSelected] = createSignal<Date|null>(null);
const [store,setStore] = createStore<{data?:{first?:Date,last?:Date}}>({})

return <Show when={{selected:selected(),first:store.data?.first, second:store.data?.second}} checkObjectValues>
// children will only render if all values are truthy 
// item is now narrowed to {selected:Date, first:Date,second:Date}
   {item => <StrictChild {...item} />}
</Show>

```

### DIFF
--- a/packages/solid/src/render/flow.ts
+++ b/packages/solid/src/render/flow.ts
@@ -78,35 +78,95 @@ export function Index<T extends readonly any[], U extends JSX.Element>(props: {
       )) as unknown as JSX.Element;
 }
 
+// Helper to prettify types in errors
+type Prettify<T> = {
+  [K in keyof T]: T[K];
+} & {};
+
+// Helper to make all values non-nullable
+type NonNullableValues<T> = {
+  [K in keyof T]: NonNullable<T[K]>;
+};
+
+// Strict object type to exclude arrays, maps, sets, and functions
+type StrictKeyValueObject = {
+  [key: string]: string | number | boolean | object | null | undefined;
+};
+
+// ConditionalRecord ensures that when `checkObjectValues` is true, `T` must be a StrictKeyValueObject (not an array or other types)
+type NarrowedRecord<
+  T,
+  CheckObjectValues extends boolean | undefined
+> = CheckObjectValues extends true
+  ? T extends StrictKeyValueObject
+    ? NonNullableValues<T> // Ensure all values are NonNullable
+    : never // Throw an error if not a strict object
+  : T;
+
+// Enforce when to be a StrictKeyValueObject if checkObjectValues is true
+type EnforceStrictKeyValueObject<
+  T,
+  CheckObjectValues extends boolean | undefined
+> = CheckObjectValues extends true
+  ? T extends StrictKeyValueObject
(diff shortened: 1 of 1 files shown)


madaxen86:
@davedbase you're also quite active on discord. What are your thoughts on this?

davedbase:
> @davedbase you're also quite active on discord. What are your thoughts on this?

That would be quite handy tbh.

madaxen86:
So,...
I dug deeper and tested a lot, realizing that the with the first implementation, granular updates ware lost. To fix that I implemented a `store` in the case the checkObjectValues prop is set to true, if false the store won't be created.
So we get fine grained updates for the cost of having `createStore` and `reconcile` as dependencies.
But we get closer to type narrowing like in `React`.
Few questions:

1. Is it worth to proceed and invest more time, or does this not have a chance to be merged. Then I'll remove the PR. @ryansolid and @davedbase 
2. With latest commit it is still possible to get the callback as an accessor or 'keyed'. But there's really no point of accessing a store. So we could always just return the store. But this would then make the `keyed` prop obsolete when `checkObjectValues` is set, because it doesn't change anything. So we have 3 options
2a. Keep the current implementation
2b. always return the store
2c. remove the `checkObjectValues` prop and add it as an option of keyed `keyed:boolean | "checkObjectValues"`

3. if 1. true :-P I would add documentation and also implement the same options for Switch/Match

ryansolid:
This feels Solid Primitives area. As you have noted the implementation becomes more complicated as you try to do these checks. Generally I try avoid any base implementation that gets us into Stores because not every platform supports Proxies. It is unlikely a change like this happens in 1.x time period but I'm definitely interested in this area to understand the problems this solves and understand all the scenarios as we approach control flow improvements for 2.0.

That being said it seems to have value today which is why I point to Solid Primitives:  https://github.com/solidjs-community/solid-primitives

madaxen86:
Just pasting some comments from the SolidJack 2024 channel about this topic:

[Discord](https://discord.com/channels/722131463138705510/1054806092825239683/1308790650027638804)

PR #2343, opened 2024-10-20
fix class/prototype getters not being reactive

timbasel:
This fixes class getter methods not being reactive.
It is the same fix that was already applied for `createMutable` in #2027.

### DIFF
--- a/.changeset/young-mirrors-begin.md
+++ b/.changeset/young-mirrors-begin.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix class/prototype getters not getting wrapped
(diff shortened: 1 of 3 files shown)


timbasel:
@ryansolid It's been a while, but is there anything that can be done to get this fix implemented so I don't have to run a custom version of solid for some my projects?

ryansolid:
Sorry it has taken me a while to get back to Solid 1.0 stuff especially wanted to make sure I would be fine with these changes in 2.0 (didn't have this concern for createMutable).

ryansolid:
Thank you.

PR #2353, opened 2024-10-31
Improve resolving arguments in createResource

thetarnav:
Check `typeof` the second argument instead of using the `arguments` variable.
This fixes case where some arguments were passed explicitely as `undefined` instead of being omitted.
Which helps with wrapping the resource in own function.

```ts
createRecource(fetcher, undefined) // used to cause TypeError: fetcher is not a function
```


ryansolid:
Thanks.. Looks good.

PR #2363, opened 2024-11-13
Add `StoreBundle<T>` type

mustevenplay:
## Summary

Here I added `StoreBundle<T>`; a useful type extracted directly from the `createStore` function's return type:
`type StoreBundle<T> = [get: Store<T>, set: SetStoreFunction<T>]`

This type is identical to the return type of the `createStore` function, so nothing really should change, besides the fact that it's definitely more portable, more compact, and depending on who we ask, more readable.
All tests have been successfully ran, though.

### DIFF
--- a/.changeset/lovely-mayflies-sniff.md
+++ b/.changeset/lovely-mayflies-sniff.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Add `StoreReturn<T>` type
--- a/packages/solid/store/src/index.ts
+++ b/packages/solid/store/src/index.ts
@@ -8,6 +8,7 @@ export type {
   SetStoreFunction,
   SolidStore,
   Store,
+  StoreBundle,
   StoreNode,
   StorePathRange,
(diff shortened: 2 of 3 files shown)


mustevenplay:
The current return type of `createSignal` is `Signal<T>`, which is alias to `[get: Accessor<T>, set: Setter<T>]`. In my opinion, the chosen names `Accessor` and `Setter` might be kind of ambiguous, to which I suggest we might want to try coming up with a convention that could unify return types of functions like `createSignal` or `createStore` with a naming such as "`FooBundle`" (like in this case, `StoreBundle`). Keeping the current naming could be a good idea too, but I could as well suggest to give it a specific name for each return type.
These conventions applied on the current `Signal<T>` type would look like this:
```ts
type SignalBundle<T> = [get: SignalAccessor<T>, set: SignalSetter<T>]
```

ryansolid:
Thanks. I'm going with StoreReturn. But I've merged it as the base.

PR #2382, opened 2024-12-16
Update typescript to 5.7

thetarnav:
Since version 5.7 typescript doesn't like comparing strings for truthiness

image

And inlining `"_SOLID_DEV_" as any as boolean` doesn't seem to do the trick.

So I've added a constant and use it in place of strings which silences the errors.
```ts
const IS_DEV = "_SOLID_DEV_" as string | boolean
```

The output looks almost identical to using strings directly - most code paths behind a dev check get removed in prod entries. Some uncleaned if statements remain, but as they did with strings before.
```js
const IS_DEV = false;
// ...
if (IS_DEV); // the block gets removed but if remains
else {
	// ...
}
```

The variable is not exported outside of the module, just internally between files to not redeclare it in every file.

Other option would be to use `@ts-expect-error` everywhere but that just seem a bit noisy.

---

Also had to add some additional assertions `!` to variable declarations for refs, as they are "being used before assigned" from ts point of view.

### DIFF
--- a/.changeset/many-crabs-reply.md
+++ b/.changeset/many-crabs-reply.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Update typescript to 5.7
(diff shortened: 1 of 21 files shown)


ryansolid:
Thanks

PR #2389, opened 2024-12-27
dev: Use `devComponent` in `lazy`

thetarnav:
Because `lazy` doesn't call the component function using JSX or `createComponent` it doesn't show up in the devtools properly.
This replaces the function call with `devComponent` if in dev mode.

Although it may be a bit odd, as `devComponent` is an internal-only api, and `lazy` should probably be fully implementable in userland, right?
Maybe there is a way to just use `createComponent` there, as the logic is very similar.
```ts
return createMemo(() => (Comp = comp()) ? createComponent(Comp, props) : "")
```


ryansolid:
My only concern with this is does this impact hydration.. like component counting mismatch. I haven't been able to verify but when we choose to create component vs not is usually very intentional for matching hydration IDs.

ryansolid:
Thanks for the PR. The devtools issue is real: `lazy` currently bypasses the normal component creation path, so lazy-loaded components do not get the same dev metadata as regular JSX-created components.

I’m hesitant to merge this as-is for 1.x, though. `lazy` has always been a sensitive path for hydration, and switching from a direct component call to `devComponent` changes owner creation in dev mode. Hydration IDs/counting are very sensitive to where owners are created, so I’d want targeted SSR/hydration tests before making that change.

For Solid 2 this area has changed enough that this PR is stale it should be easy to implement because components no longer count towards ID. For 1.x, I think the issue is worth solving, but the fix needs to either prove hydration parity or use a metadata approach that does not alter the lazy component owner structure.

PR #2401, opened 2025-01-08
Export ClassList type

marvin-j97:


ryansolid:
Is this the right place to export it. I wonder if it should be part of the JSX types? I'm suggesting that is the right answer I just really don't know what would be more expected.

ryansolid:
Thinking this through. This type should probably be exposed from the JSX types in dom-expressions, since classList is part of the shared JSX attribute contract rather than a solid-js/web runtime API. In Solid it would ideally be consumed as JSX.ClassList via the generated JSX types, with classList?: ClassList inside JSX.CustomAttributes<T>, instead of adding a separate top-level export from solid-js/web.

PR #2404, opened 2025-01-13
Fix pnpm corepack version

DaniGuardiola:
When trying to clone and set up the solid repo (on `next` branch), I realized the corepack version of pnpm (`packageManager` field in `package.json`) was set to v8. When running `pnpm i`, it seems to ignore the lockfile (says it's broken), probably because the lockfile version (9) is incompatible with pnpm v8.

To remediate this, I upgraded the corepack version to latest (v10) and re-ran `pnpm i`, which seems to add some minor changes to the lockfile itself.

Feel free to replicate this process to verify the contents of this PR. From the `next` branch:

```sh
corepack use pnpm # to upgrade pnpm
pnpm i # to update lockfile
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,5 @@
   "engines": {
     "pnpm": ">=8.6.1"
   },
-  "packageManager": "pnpm@8.6.1"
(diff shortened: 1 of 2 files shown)


ryansolid:
The next branch is a complete WIP.. completely broken at this point. I'm closing this for now..

PR #2411, opened 2025-01-24
Show double evaluation

### LINKED ISSUE #2406: The condition of <Show> runs another time unnecessarily when using the function form
### Describe the bug

It looks like the `when` condition of the `<Show>` component is run twice unnecessarily at initialisation when a function as passed as `children`. It is first run when creating the `condition` memo, but then it is run again to be passed as the parameter to the `children` func, instead of passing the `condition` directly to the `children` func.

https://stackblitz.com/edit/solidjs-templates-blqgxpec?file=src%2FApp.tsx

1. Use a `<Show>` component (not `keyed`), with function passed as children.

The `when` condition is immediately evaluated twice.

### Expected behavior

The `when` expression is evaluated once, and then reevaluated whenever there's an update notifying the value might have changed.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Chrome
- Version: 1.7.6

### Additional context

The problem is especially big if the condition creates some JSX, e.g. when JSX is passed through props. Normally this JSX is not considered to change, even if some signals are updated, but rather it updates its contents reactively, but here two instances of the JSX are created, and the first one is not cleaned up when the second one is created, but instead they coexist (although only one of them is mounted to the document). See the provided stackblitz for an example.

The `Show` component has this code:
```ts
(child as any)(
  keyed ? (c as T) : () => {
    if (!untrack(condition)) throw narrowedError("Show");
    return props.when;
  }
)
```
It is surprising to me that `props.when` is passed to `child`, instead of passing there the already memoised `condition` created before. This would fix the problem described here, but I'm not able to guess if it creates any other problems.

TPReal:
## Summary

This change removes unnecessary evaluations of the `when` conditions in the `<Show>` and `<Match>` components. Both these components used to evaluate the condition twice immediately after creation if the condition was true, children was specified as a function and keyed was not specified.

This was especially problematic when the condition produced `JSX.Element` which was then displayed by the children function, as two copies of the elements were instantiated.

This change also removes unnecessary evaluations of `<Match>` conditions in a sutiation where the condition did not change, but only a later `<Match>`'s condition changed.

Added unit tests demonstrating the described problems, and made `pnpm test` pass.

fixes #2406

### DIFF
--- a/.changeset/hot-jeans-cheer.md
+++ b/.changeset/hot-jeans-cheer.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Removed unnecessary evaluations of <Show> and <Match> conditions.
--- a/packages/solid/src/render/flow.ts
+++ b/packages/solid/src/render/flow.ts
@@ -106,16 +106,23 @@ export function Show<T>(props: {
   children: JSX.Element | ((item: NonNullable<T> | Accessor<NonNullable<T>>) => JSX.Element);
 }): JSX.Element {
   const keyed = props.keyed;
-  const condition = createMemo<T | undefined | null | boolean>(
+  const conditionValue = createMemo<T | undefined | null | boolean>(
     () => props.when,
     undefined,
-    IS_DEV
-      ? {
-          equals: (a, b) => (keyed ? a === b : !a === !b),
-          name: "condition"
-        }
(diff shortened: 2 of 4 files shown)


ryansolid:
Thank you

PR #2422, opened 2025-02-06
Add `createDynamic` utility.

DaniGuardiola:
## Summary

`Dynamic` is very helpful for polymorphism use-cases. However, in libraries, there's a potential optimization to be leveraged by not performing a prop split, and leaving that detail to the library author.

For this purpose, I've extracted most of the logic in `Dynamic` into a `createDynamic` utility (now used by `Dynamic` under the hood.

In my case (Ariakit Solid), due to how things are laid out, I can effectively fulfill this API without splitting, which wouldn't be doable with `Dynamic` alone.

TODO

### DIFF
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -115,6 +115,42 @@ export type DynamicProps<T extends ValidComponent, P = ComponentProps<T>> = {
 };
 
 /**
+ * Renders an arbitrary component or element with the given props
+ *
+ * This is a lower level version of the `Dynamic` component, useful for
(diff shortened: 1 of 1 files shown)


ryansolid:
This probably should be a minor release but I don't want to jam you up. Hopefully by the time your library releases most people will be past 1.9.5. We can consider this experimental/undocumented behavior until 1.10.x.

DaniGuardiola:
Oof I accidentally missed a feature I also needed here to make the props reactive. I'll submit another PR. My bad!

PR #2429, opened 2025-02-19
add optional initalValue argument to `from` helper

chris-kruining:
I just discovered the from helper and immediately went and used it to change my own helper I had made to create a clock and then discovered that the `Accessor` that it returned was always `T | undefined`. I then went to the source code of `from` and found out that the underlying signal was initialized with undefined. 

So this PR aims to add an optional `initalValue` argument so that the return value can be `Accessor<T>` if you would prefer that

### DIFF
--- a/.changeset/dull-bikes-reply.md
+++ b/.changeset/dull-bikes-reply.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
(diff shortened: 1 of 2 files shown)


ryansolid:
This looks good. I don't know why I didn't think of doing this before. Thank you.

ryansolid:
Thank you

Issue #2436, opened 2025-02-28
Hydration bug in v1.9.5


brenelz:
### Describe the bug

I was working on some SSR stuff for Tanstack Start for Solid and noticed our hydration logic worked for solidjs version 1.9.4 but didn't work in 1.9.5. This only affected the prod build and worked fine in dev. 

Either our logic is relying on something that changed or there is a bug here.

### Your Example Website or App

https://github.com/TanStack/router/tree/5f1d044f969327945fbbdaffceabe7da7e8036f4/e2e/solid-start/server-functions

### Steps to Reproduce the Bug or Issue

- git clone
- pnpm i
- pnpm build:all
- cd e2e/solid-start/server-functions/
- downgrade solid in package.json
- pnpm build
- pnpm start
- go to http://localhost:3000/consistent
- test clicking on the button

### Expected behavior

Before clicking the button:

<img width="566" alt="Image" src="https://github.com/user-attachments/assets/8eecb22b-f0ec-4f35-a940-ca1783907711" />

After clicking the button:

<img width="898" alt="Image" src="https://github.com/user-attachments/assets/9a56de21-73b7-4693-aea3-87fa14908ff8" />

### Platform

- OS: macOs
- Browser: Chrome

birkskyum:
@brenelz , after the git clone, checkout [this PR](https://github.com/TanStack/router/pull/3522)

ryansolid:
Oh so there isn't any error? As much as it isn't working.. Yeah this should be interesting because most of the changes in 1.9.5 are types, escaping logic, and code refactoring. It is odd that anything would be impacted. This might take some doing to track down.

birkskyum:
No error

birkskyum:
I've tested more versions:

1.9.2 breaks
1.9.3 breaks
1.9.4 works 🎉
1.9.5 breaks

birkskyum:
@ryansolid was anything introduced in 1.9.4 that was reverted later?

brenelz:
So it looks like this was just a solid js version mismatch between packages. Closing this issue

PR #2439, opened 2025-03-01
ci(enable lockfile) bump pnpm and node for development

birkskyum:
This updates the pnpm version to one compatible with the existing lockfile, and hte same as solid-start uses (faster ci, because it can install from lockfile)
This updates node to v22 in ci, which is the current LTS

**we need this for reproduction of builds**
Without this PR each CI release is made without a lockfile, making ti impossible for us to go back in time and make a build of say version 1.9.4 exactly like it was at the time of release, because we don't have proper information about which nested deps versions were used. This makes e.g. debugging of #2436 a lot harder, because a local build at that commit isn't behaving like the published 1.9.4 distribution.

(this is also a problem on the `next` branch, where similar changes should be applied)


ryansolid:
Thanks

PR #2466, opened 2025-04-08
dev: add warning when context is undefined

Tommypop2:
## Summary

Using `useContext` outside of a provider can cause unforeseen and cryptic errors. This PR adds a warning (in dev) when the result of `useContext` is falsy.

See https://discord.com/channels/722131463138705510/1358899310116213008

Typescript does protect against using an undefined context somewhat, but I thought a warning could be nice, especially for new users.

Created a project and consumed a context without being in its provider:
image

Warning is printed to the console.

After the project has been built, the warning is not logged.

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1201,9 +1201,11 @@ export function createContext<T>(
  */
 export function useContext<T>(context: Context<T>): T {
   let value: undefined | T;
-  return Owner && Owner.context && (value = Owner.context[context.id]) !== undefined
-    ? value
-    : context.defaultValue;
+	let ctx = Owner && Owner.context && (value = Owner.context[context.id]) !== undefined
+	? value
+	: context.defaultValue;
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks for the PR. I’m going to close this as stale for Solid 2.

The underlying issue is real: reading a default-less context outside its Provider can lead to confusing errors later. In Solid 2, though, default-less contexts now throw immediately when read without a Provider, and `useContext(createContext<T>())` is typed as `T`, so this is handled more directly.

I also don’t think this exact warning is safe to add. Context values can legitimately be falsy — `false`, `0`, `""`, or even `null` depending on the context type — so warning based on a falsy result would produce false positives.

For the 1.x line, I think the better fix is a type-level one: distinguish default-less contexts so providers don’t accidentally accept `undefined` unless the context type explicitly includes it. That addresses the common mistake without warning on valid context values. As outlined as my response to https://github.com/solidjs/solid/pull/1958.

PR #2474, opened 2025-05-06
Fix seroval 1.3.0 compat

lxsmnsyc:
> [!CAUTION]
> I haven't updated the lockfile because Solid is currently locked to PNPM 9, I'm using PNPM 10.

This PR fixes a potential incompatibility with Seroval 1.3.0 (latest release) which reworks the serialized status/value of a serialized Promise. The change was imminent because it made Seroval incompatible for use in the React ecosystem (due to React's internal suspense mechanism which relies coincidentally with the `status` property)

### DIFF
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -237,7 +237,7 @@
   ],
   "dependencies": {
     "csstype": "^3.1.0",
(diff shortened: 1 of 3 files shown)


elite174:
Would be nice to fix package versions

PR #2476, opened 2025-05-09
fix: catch synchronous error

juanrgm:
## Summary

Synchronous (non-promise) errors cannot be caught by `ErrorBoundary`.

https://playground.solidjs.com/anonymous/98e1258d-67c8-4c53-b9cb-4bfa029d451e

```ts
import { render } from "solid-js/web";
import { ErrorBoundary, createResource } from "solid-js";

function Counter() {
  const [res] = createResource(() => {
    if (1) throw new Error(); // this error is not captured by ErrorBoundary
    return 1;
  });

  return (
    <ErrorBoundary fallback={(error) => <>{error.toString()}</>}>
      {res()}
    </ErrorBoundary>
  );
}

render(() => <Counter />, document.getElementById("app")!);
```

Using an async function as fetcher it works:

https://playground.solidjs.com/anonymous/0e5e5348-57ef-41b3-aebf-66f8567122c8

An unit test was added to the commit.

### DIFF
--- a/.changeset/chilled-buttons-hammer.md
+++ b/.changeset/chilled-buttons-hammer.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Catch synchronous errors in `createResource`.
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks

PR #2497, opened 2025-05-27
Fix resource instances always getting cached on SSR

XiNiHa:
## Summary

Fixes `createResource()` calls on SSR always caching the initially created resource instance, potentially resulting in stale closure issues on edge cases.

Before this PR, the following code didn't fetch and render data on SSR. Now it does.

```jsx
import { createSignal, createComputed, createResource, lazy } from "solid-js";
const Profile = lazy(() => import("./Profile"));

export default () => {
  const [userId, setUserId] = createSignal();

  const [user] = createResource(() => {
    console.log("LOAD USER");
    return new Promise(res => {
      setTimeout(() => res({ id: "1", firstName: "Jon", lastName: "Snow" }), 400);
    });
  });
  createComputed(() => {
    const u = user();
    if (!u) return;
    setUserId(u.id);
  });
  const [info] = createResource(
    userId,
    () => {
      // simulate cascading data loading
      console.log("LOAD INFO");
      return new Promise(res => {
        setTimeout(
          () =>
            res(["Something Interesting", "Something else you might care about", "Or maybe not"]),
          400
        );
      });
    },
    { initialValue: [] }
  );

  return <Profile user={user()} info={info()} />;
};
```

### DIFF
--- a/.changeset/young-donuts-smell.md
+++ b/.changeset/young-donuts-smell.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix resource instances always getting cached on SSR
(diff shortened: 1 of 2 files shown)


ryansolid:
I see it only now caches it if it is actually fetching. Ok. This is s tricky area in general and one of the reasons I want to get rid of `createComputed` in the future but I think this should be ok.

PR #2502, opened 2025-06-12
improve `<Portal/>` container, add 2 new properties

ngdangtu-vn:
I don't know what is "three mandatory fields" since I only saw 2 headings. But I filled all of them.

## Summary

```git
improve `<Portal/>` container, add 2 new properties

- `asElement` allows to change Portal container element type
- `class` allows to set class for Portal container
```

Bascially, I don't like a bare `<div/>` wrap children component so I would like to suggest make Portal `container` be more customisable with tagName and class. We can find a way to support dataset later if it requires.

I've update test file and run test script from `package.json`s:

1. `/` → `pnpm test`
2. `cd ./packages/solid` → `pnpm test`

Both of them return all pass results so I assume there is nothing would break the framework. As for testing on real field, I'm not sure how to do that yet.

### DIFF
--- a/packages/solid/web/src/index.ts
+++ b/packages/solid/web/src/index.ts
@@ -63,6 +63,8 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
         el: (T extends true ? { readonly shadowRoot: ShadowRoot } : {}) &
           (S extends true ? SVGGElement : HTMLDivElement)
       ) => void);
+  asElement?: keyof HTMLElementTagNameMap;
+  class?: string[];
   children: JSX.Element;
 }) {
   const { useShadow } = props,
@@ -84,11 +86,13 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
         createRoot(dispose => insert(el, () => (!clean() ? content!() : dispose()), null));
         onCleanup(cleanup);
       } else {
-        const container = createElement(props.isSVG ? "g" : "div", props.isSVG),
-          renderRoot =
-            useShadow && container.attachShadow
-              ? container.attachShadow({ mode: "open" })
-              : container;
+        const container = createElement(props.isSVG ? "g" : props.asElement || "div", props.isSVG);
+        props.class && container.classList.add(...props.class.flatMap(str => str.split(" ")));
+
+        const renderRoot =
+          useShadow && container.attachShadow
+            ? container.attachShadow({ mode: "open" })
+            : container;
 
         Object.defineProperty(container, "_$host", {
           get() {
--- a/packages/solid/web/test/portal.spec.tsx
+++ b/packages/solid/web/test/portal.spec.tsx
@@ -25,6 +25,56 @@ describe("Testing a simple Portal", () => {
   });
 });
 
+describe("Testing an Portal custom container", () => {
+  let div = document.createElement("div"),
+    disposer: () => void;
+  const testMount = document.createElement("div");
+  const Component = () => (
+    <Portal mount={testMount} asElement="section">
+      Hi
+    </Portal>
+  );
+
+  test("Create portal control flow", () => {
+    disposer = render(Component, div);
+    expect(div.innerHTML).toBe("");
+    const container = testMount.firstChild as HTMLElement & { _$host: HTMLElement };
+    expect(container.tagName).toBe("SECTION");
+    expect(container.innerHTML).toBe("Hi");
+    expect(container._$host).toBe(div);
+  });
+
+  test("dispose", () => {
+    disposer();
+    expect(div.innerHTML).toBe("");
+  });
+});
+
+describe('Testing Portal "class" property', () => {
+  let div = document.createElement("div"),
+    disposer: () => void;
+  const testMount = document.createElement("div");
+  const Component = () => (
+    <Portal mount={testMount} class={["flex flex-col", "text-gold bg-red"]}>
+      ★
+    </Portal>
+  );
+
+  test("Create portal control flow", () => {
+    disposer = render(Component, div);
+    expect(div.innerHTML).toBe("");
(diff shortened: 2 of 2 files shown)


mizulu:
I wonder, if defaulting the container with this inline style will solve 
most issues that will require a custom element 

it is possible to hack it as shown here
https://github.com/solidjs/solid/issues/1873#issuecomment-1875092507

if Portal has intentions to support fragments in the future ( no wrapper )
then these might not be necessary.  ( as might have been implied by https://github.com/solidjs/solid/issues/1873#issuecomment-2623761894) 

it is nice however to have control of the container element however and its styling. 
 with the current implementation of Portal. without resorting to the implementation details.

ngdangtu-vn:
I'm not sure I would want a hacky solution and it is not friendly with accessibility since `div` and `article` can be meant different thing. For now, without this pr, we can use your suggestion set `aria-*=` and `role=` on it. But I prefer to have native html tag more.

Fragment support is nice though.

ngdangtu-vn:
Yeah, when I made this, I didn't think that far. But you are right. It should support all html attributes. We can do that with `setAtrribute`. But how should handle it in the Portal component property? Put all of them in one propety like `containerAttributes` or give them first level property like `class`?

@mizulu what's your thought?

ryansolid:
The truth is we are redesigning Portal's in 2.0. I figured out a way to use proxies to proxy DOM elements that will allow us to get rid of the intermediate DOM element. So while I know it is a bit clunky right now I'm less incentivized to change the API to encourage patterns that will be going away anyway. For now messing with the ref is probably the way to go.

Perfumere:
If you need `<Fragment>` functionality, please consider not forcing the use of DOM element wrapping.

Perfumere:
I have a temp custom component: `Teleport`，no `div` wrapper:

## use case

```tsx
// 1. default to document.body
<Teleport when={mounted()}></Template>

// 2. cowork with Transition
const [show, setShow] = createSignal(false);
const [mounted, setMounted] = createSignal(false);
<Teleport when={mounted()}>
        <Transition name="fade-in" appear onAfterExit={() => setMounted(false)}>
        { show() && <div>Toast...</div> }
        </Transition>
<Teleport>

// 3. 动态修改mount，move props.children from  `old mount` to `new mount`
const [el, setEl] = createSignal(document.body);
setTimeout();
<Teleport mount={el}>
  <For each={[0, 1,2,3]}>{i => i}</For>
  Anytype...
</Teleport>
```

## Source Code

```tsx
import {
  createEffect,
  createRenderEffect,
  createRoot,
  getOwner,
  runWithOwner,
} from 'solid-js';
import { insert } from 'solid-js/web';
import type { JSX, Accessor } from 'solid-js';

type TeleportProps = {
  mount?: HTMLElement | null;
  when?: boolean | Accessor<boolean>;
  children: JSX.Element;
};

export function Teleport(props: TeleportProps) {
  const owner = getOwner();
  let parentEl: HTMLElement | null = null;
  let disposeCurrent: (() => void) | null = null;
  let baselineTail: ChildNode | null = null;
  let managedFirst: ChildNode | null = null;
  let managedLast: ChildNode | null = null;

  const disposeRoot = () => {
    if (disposeCurrent) {
      disposeCurrent();
      disposeCurrent = null;
    }
  };

  const updateManagedBounds = () => {
    if (!parentEl) return;
    const first = baselineTail ? baselineTail.nextSibling : parentEl.firstChild;
    const last = parentEl.lastChild;
    if (!first || !last || first === baselineTail) {
      managedFirst = null;
      managedLast = null;
      return;
    }
    managedFirst = first;
    managedLast = last;
  };

  const mountInto = (target: HTMLElement) => {
    parentEl = target;
    baselineTail = parentEl.lastChild;
    runWithOwner(owner, () => {
      createRoot(dispose => {
        disposeCurrent = dispose;
        createRenderEffect(() => {
          insert(parentEl as Node, () => props.children, null);
          updateManagedBounds();
        });
        return undefined as unknown as void;
      });
    });
  };

  const moveManagedTo = (nextParent: HTMLElement) => {
    if (!parentEl || parentEl === nextParent) return;
    const newBaseline = nextParent.lastChild;
    if (!managedFirst || !managedLast) {
      parentEl = nextParent;
      baselineTail = newBaseline;
      updateManagedBounds();
      return;
    }
    let node: ChildNode | null = managedFirst;
    const lastToMove: ChildNode = managedLast;
    while (node) {
      const next = node === lastToMove ? null : node.nextSibling;
      nextParent.appendChild(node);
      if (node === lastToMove) break;
      node = next;
    }
    parentEl = nextParent;
    baselineTail = newBaseline;
    // remark start & end
    managedFirst = newBaseline ? (newBaseline.nextSibling as ChildNode | null) : (nextParent.firstChild as ChildNode | null);
    managedLast = lastToMove;
  };

  const cleanupManaged = () => {
    if (parentEl && managedFirst && managedLast) {
      let node: ChildNode | null = managedFirst;
      const lastToRemove: ChildNode = managedLast;
      while (node) {
        const next = node === lastToRemove ? null : node.nextSibling;
        parentEl.removeChild(node);
        if (node === lastToRemove) break;
        node = next;
      }
    }
    managedFirst = null;
    managedLast = null;
    baselineTail = null;
    parentEl = null;
  };

  createEffect(() => {
    const shouldRender = props.when ?? true;
    const target = (props.mount || document.body) as HTMLElement;

    if (!shouldRender) {
      disposeRoot();
      cleanupManaged();
      return;
    }

    if (parentEl && parentEl !== target) {
      moveManagedTo(target);
    } else if (!parentEl) {
      mountInto(target);
    } else if (!disposeCurrent) {
      mountInto(target);
    }
  });

  onCleanup(() => {
    disposeRoot();
    cleanupManaged();
  });

  return null;
}
```

ryansolid:
Yeah given teh 2.0 direction. It is unlikely this gets merged. Thanks for the work here.

PR #2505, opened 2025-06-16
export deep util

brenelz:
Export the deep helper from @solidjs/signals


packages/solid/src/index.ts
```diff
@@ -28,7 +28,8 @@ export {
   runWithOwner,
   tryCatch,
   untrack,
-  unwrap
+  unwrap,
+  deep,
```
brenelz, line 32: I was also looking to export the `path` helper but I don't think that's implemented yet

PR #2588, opened 2026-02-14
fix: prevent createDeferred from keeping Node.js process alive

### LINKED ISSUE #2570: `createDeferred` hangs the process in nodejs
### Describe the bug

using `createDeferred` within node, while using `node -C browser file.js`, causes the process thread to hang and never exit.

Add this to a file:

```js
import { createSignal, createEffect, createRoot, createDeferred } from "solid-js";

function createCounter(initial = 0) {
  const [count, setCount] = createSignal(initial);
  const def = createDeferred(() => (count() * 2), { timeoutMs: 2000 })

  createEffect(() => {
    console.log(`Count changed, new value: ${count()}`);
  });

  function increment() {
    setCount((c) => c + 1);
  }

  function reset() {
    setCount(initial);
  }

  return { count, increment, reset, def };
}

createRoot((dispose) => {
  const { count, increment, reset, def } = createCounter(10);

  console.log(count())
  console.log(def())
  increment();
  console.log(count())
  console.log(def())
  reset();
  console.log(count())
  console.log(def())

  dispose();
});
```

Run it with `node -C browser ./file.js`, it will print out correctly but the process will not end properly.

### Platform

    OS: macOS
    Browser: nodejs
    Version: ^1.9.10

veeceey:
Fixes #2570

When using `createDeferred` in Node.js (via `node -C browser`), the process hangs after all work is done because the internal `MessageChannel` keeps the event loop alive. The scheduler creates a `MessageChannel` for task scheduling, and in Node.js, active `MessagePort` listeners prevent the process from exiting.

The fix calls `unref()` on both ports right after creating the channel. This tells Node.js not to keep the process alive solely because of these ports. In browsers `unref` doesn't exist, so the call is guarded with a typeof check -- no behavior change there.

Tested by running the full solid-js test suite (453 tests passing).

### DIFF
--- a/.changeset/cyan-comics-worry.md
+++ b/.changeset/cyan-comics-worry.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix: prevent createDeferred from keeping Node.js process alive


veeceey:
just wanted to check back - the createDeferred timer keeping Node.js alive is a real issue for SSR/testing. anything needed from my end?

PR #2660, opened 2026-04-02
fix(signal): reduce duplicate observer registrations in repeated signal reads

oanlit:
## **Summary**

When a signal is read repeatedly within the same computation (e.g. inside loops), 
the current implementation registers the same observer multiple times.

This leads to unnecessary growth of internal tracking structures, including:
- `observers`
- `observerSlots`
- `sources`
- `sourceSlots`

This issue is especially noticeable in patterns such as:

- Tight loops:
```ts
createEffect(() => {
  for (let i = 0; i < 1000; i++) a();
});
```

- Iteration over store arrays:

```ts
createEffect(() => {
  for (const item of arrayStore) {
    _ = item;
  }
});
```

In these cases, repeated signal reads (e.g. `length`, indexed access, or direct signal calls)
cause the same computation to be registered multiple times.

---

This PR introduces a lightweight deduplication mechanism by tracking the last registered observer per signal.

* Adds `lastObserver` to `SignalState`
* Only registers the current computation if it differs from `lastObserver`
* Resets `lastObserver` during cleanup

This significantly reduces redundant registrations in common scenarios while keeping the cost of deduplication O(1) and preserving existing behavior.

---

## **How did I test this change?**

To enable verification of internal observer behavior in tests, this PR slightly adjusts the return value of `createSignal`:

```ts
const result = [readSignal.bind(s), setter] as Signal<T | undefined>;
if (IS_DEV) (result as any).state = s;
return result;
```

2. Added a test case that repeatedly reads a signal inside a loop:

```ts
const a = createSignal(0);
const dispose = createRoot(dispose => {
  createEffect(() => {
    for (let i = 0; i < 1000; i++) {
      a[0]();
    }
  });
  return dispose;
});
expect((a as any).state.observers.length).toBe(1);
expect((a as any).state.observerSlots.length).toBe(1);
expect((a as any).state.observers[0].sources.length).toBe(1);
expect((a as any).state.observers[0].sourceSlots.length).toBe(1);
dispose();
```

Before this change:

* `observers.length === 1000`

After this change:

* `observers.length === 1`

before_change

2. Verified similar behavior with store array iteration:

```ts
const [arrayStore] = createStore([1, 2, 3, 4, 5]);

createEffect(() => {
  for (const item of arrayStore) {
    _ = item;
  }
});
```

Confirmed that repeated reads (e.g. `length` and indexed access) no longer
cause excessive observer registrations.

3. Ran full test suite:

```bash
pnpm build
pnpm test
```

All tests pass.

4. Manually inspected that:

* `observerSlots` and `sourceSlots` no longer grow excessively
* No regression in reactivity behavior

after_change

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -86,6 +86,7 @@ export interface SignalState<T> extends SourceMapValue {
   value: T;
   observers: Computation<any>[] | null;
   observerSlots: number[] | null;
+  lastObserver: Computation<any> | null;
   tValue?: T;
   comparator?: (prev: T, next: T) => boolean;
   // development-only
@@ -157,17 +158,17 @@ export function createRoot<T>(fn: RootFunction<T>, detachedOwner?: typeof Owner)
         ? { owned: null, cleanups: null, context: null, owner: null }
         : UNOWNED
       : {
-          owned: null,
-          cleanups: null,
-          context: current ? current.context : null,
-          owner: current
-        },
+        owned: null,
+        cleanups: null,
+        context: current ? current.context : null,
+        owner: current
+      },
     updateFn = unowned
       ? IS_DEV
         ? () =>
-            fn(() => {
-              throw new Error("Dispose method must be an explicit argument to createRoot function");
-            })
+          fn(() => {
+            throw new Error("Dispose method must be an explicit argument to createRoot function");
+          })
         : fn
       : () => fn(() => untrack(() => cleanNode(root)));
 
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks for digging into this. The repeated registration issue is real, but I wanted to avoid adding extra signal state or exposing internal SignalState through the createSignal return tuple for testing.

I landed a smaller version in fae8fbe3 that uses the existing observer slot structure instead: when the current listener is already the last observer for a source, readSignal skips adding another source/observer pair. This keeps the common repeated-read case from growing the tracking arrays while avoiding new cleanup state.

I also added behavior coverage for repeated reads, disposal cleanup, and conditional dependency cleanup/re-subscription. Thanks again for identifying the issue and providing the repro direction.

PR #2683, opened 2026-04-23
fix(store): prevent prototype pollution via setStore paths

thomasbuilds:
Fixes a prototype-pollution vulnerability in `solid-js/store`. Any code path that forwards untrusted input into `setStore(...)`, a common pattern for query params, form fields, or JSON bodies, could mutate `Object.prototype` and affect every object in the runtime.

Reproduction (against `main`)

```js
import { createStore } from "solid-js/store";
const [state, setState] = createStore({ a: 1 });

setState("__proto__", "polluted", true);
console.log(({}).polluted);        // true  ❌ global pollution

setState("__proto__", { x: true });
console.log(({}).x);               // true  ❌ global pollution

setState("constructor", "prototype", "y", true);
console.log(({}).y);               // true  ❌ global pollution

// Realistic attacker payload:
const evil = JSON.parse('{"__proto__":{"z":true}}');
setState(evil);                    // ❌ redefines the store's prototype

### DIFF
--- a/.changeset/orange-ghosts-walk.md
+++ b/.changeset/orange-ghosts-walk.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix(store): prevent prototype pollution via setStore paths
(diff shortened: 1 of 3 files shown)


ryansolid:
thank you

PR #2696, opened 2026-05-08
Fix createResource initial value inference

### LINKED ISSUE #2676: createResource doesn't infer type from initialValue
Consider the code below. The resource type should be inferred to `number | null`, but instead, the compiler infers the type from the fetcher alone, to `number`, and reports a type error on the initialValue field. The behavior I expect would align `createResource` with `useFetch` in Nuxt.

```ts
const [resource] = createResource(
  async () => {
    await new Promise((resolve) => setTimeout(resolve, 1000));
    return 42;
  },
  {
    initialValue: null,
  },
);
```

In most cases, the initial value that can be generated or otherwise known client-side, is not trivially the same type as the response from the server. In cases where the whole space of the response type is in use, like boolean or numbers, there might not be any natural value that the initialValue can take up anyway inside the server response type. It is also not acceptable to give up inferred-but-strict typing for something as simple as initialValue. Explicit typing can get very bothersome very fast if complex union and intersection types is returned from the fetcher.

```
"dependencies": {
  // 4 omitted
},
"devDependencies": {
  // 4 omitted
  "solid-js": "1.9.12",
  "vite": "8.0.8",
  "vite-plugin-solid": "2.11.12"
}
```

brenelz:
Used AI to change the types for `createResource` to close #2676 

Not 100% sure how well the tests cover type regressions but the new test passes

### DIFF
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -577,19 +577,19 @@ function isPromise(v: any): v is Promise<any> {
  *
  * @description https://docs.solidjs.com/reference/basic-reactivity/create-resource
  */
-export function createResource<T, R = unknown>(
-  fetcher: ResourceFetcher<true, T, R>,
-  options: InitializedResourceOptions<NoInfer<T>, true>
-): InitializedResourceReturn<T, R>;
+export function createResource<T, R = unknown, I = T>(
+  fetcher: (k: true, info: ResourceFetcherInfo<T | I, R>) => T | Promise<T>,
+  options: ResourceOptions<T | I, true> & { initialValue: I }
+): InitializedResourceReturn<T | I, R>;
 export function createResource<T, R = unknown>(
   fetcher: ResourceFetcher<true, T, R>,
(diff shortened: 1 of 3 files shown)


packages/solid/src/reactive/signal.ts
```diff
@@ -577,19 +577,19 @@ function isPromise(v: any): v is Promise<any> {
  *
  * @description https://docs.solidjs.com/reference/basic-reactivity/create-resource
  */
-export function createResource<T, R = unknown>(
-  fetcher: ResourceFetcher<true, T, R>,
-  options: InitializedResourceOptions<NoInfer<T>, true>
```
le0-0, line 582: Could the problem be resolved by just removing the NoInfer of the T in the InitializedResourceOptions here? In that way, what goes on inside there can contribute to the type T. This would be cleaner than the parallel I mechanism if it works.

ryansolid:
Thanks. I also tried the NoInfer approach but it had holes. Atleast what I tried.

PR #2704, opened 2026-05-15
Update Solid 2.0 documentation for initial values

titoBouzout:
## Summary

 
Add docs about removal of `@once` and provides a migration path with alternative solutions. 

Notes: needs from https://github.com/ryansolid/dom-expressions/pull/523 to be merged


ryansolid:
I agree with removing @once as a thing in docs. I've updated the migration docs myself a bit more subtly. This was a fringe feature anyway and I almost don't want to give it too much weight.

Issue #2131, opened 2024-04-16
[Astro] `createResource` results get mixed up when rendering server only components with Suspense


PeterDraex:
### Describe the bug

`resource.data()` is returning data from another resource, in another component

### Your Example Website or App

https://stackblitz.com/edit/github-pqy8j7-y8z78u?file=src%2Fcomponents%2FPage.tsx

### Steps to Reproduce the Bug or Issue

1. Open Page.tsx
2. See that `pageQuery` resolves to `'pageQuery API response'` in it's fetcher function, but when you check `pageQuery.data()`, it contains `"Component A API response"`

### Expected behavior

This text is displayed
> pageQuery.data: "pageQuery API response"

### Platform

Windows, Chrome 123

### Additional context

@birkskyum:
> It appear to happen because the data from the CompA resource isn't used anywhere, and then it ends up being picked up inside Page instead of the resource defined later there...

ryansolid:
The setup for this is really odd. In that Astro seems to SSR Solid twice at least in dev and that is causing a whole bunch of chaos (not to mention sort of terrible for performance). I'm going to try to comment out the Astro code and see if there is any issue on Solid's side.

PeterDraex:
> Astro seems to SSR Solid twice

Unfortunately, that's by design according to https://github.com/withastro/astro/issues/8368
> Unfortunately this is because we can't detect if a given function is a Solid component until we run it. Our React and Preact integrations do the same thing.

ryansolid:
Ok positive is I don't think this is the cause even if it causes some weirdness in terms of multiple copies of resources. I will continue to investigate.

ryansolid:
Ok I see the problem I just need to decide where the fix goes. There is an optimization it looks like someone put in to Solid-Astro integration to generate the non-hydration markup when you don't add client directives to your Solid components. NoHydration doesn't increment component IDs and parallel Suspense is resetting at the same id making those resources have identical IDs.

I admit I wasn't expecting this optimization. But I think this could happen in Solid's own Island implementation so I guess I need to make sure Suspense generates its own unique IDs. It will add depth but I'm not sure how else to avoid it.

PeterDraex:
Hi, can I ask if this will be fixed in foreseeable future? Or is there any other way to avoid sending serialized data to the browser?

ryansolid:
[...]

PR #2707, opened 2026-05-19
Isolate context id space inside createComponent under NoHydration

### LINKED ISSUE #2546: `renderToStringAsync` with `NoHydration` and nested resources + `Suspense` boundaries triggers resource collision
### Describe the bug

Suppose you have the following `Post` component, which uses a resource to load a user:

```jsx
async function loadPost(id) {
	await new Promise((resolve) => setTimeout(resolve, 10));
	return {
		id,
		userId: 'alice',
		content: 'Hello world',
	};
}

export function Post(props) {
	const [post] = createResource(
		() => props.id,
		(postId) => loadPost(postId),
	);

	return (
		<div>
			<Suspense>
				<User id={post()?.userId} />
			</Suspense>
			<span>Post: {post()?.content}</span>
		</div>
	);
}
```

The User component in turn loads data via a resource and renders that:

```jsx
async function loadUser(id) {
	await new Promise((resolve) => setTimeout(resolve, 10));
	return {
		id,
		name: `User ${id}`,
	};
}

export function User(props) {
	const [user] = createResource(
		() => props.id,
		(id) => loadUser(id),
	);

	return <span>User: {user()?.name}</span>;
}
```

If you render with `renderToStringAsync`, this will wait for all the resources / Suspense boundaries to resolve. But if you render this under `NoHydrate` like so:

```
await renderToStringAsync(() => (
	<NoHydration>
		<Post id="post123" />
	</NoHydration>
));
```

This can result in the nested `User` resource returning the return value from the parent `Post` resource. So `user()` returns the same value as `post()` does.

This only happens with `NoHydration`. Removing it seems to resolve the issue.

https://github.com/fongandrew/solidjs-resource-repro

1. Nest components with `createResource` calls and `Suspense` boundaries
2. Render with `renderToStringAsync` and `<NoHydrate>`

### Expected behavior

I expect the accessor returned by each `createResource` to always return the value returned by the fetcher it was created with (or undefined if still loading). And I expect it to work with `NoHydrate` (the use case here is that we're just rendering a pure HTML page that will never be hydrated but still relies on a waterfall of async data loading).

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Node
- Version: 23.5.0

### Additional context

_No response_

LeSingh1:
Closes #2546.

Per the existing maintainer analysis in the issue thread (#2546 / #2131): under `<NoHydration>`, IDs aren't managed correctly, so two `createResource` calls in nested components can collide on the same key in `context.resources`.

Walking through the repro (https://github.com/fongandrew/solidjs-resource-repro):

```
<NoHydration>
  <Post id="post123">         // outer createComponent
    // const [post] = createResource(...)  -> getNextContextId
    <Suspense>
      <User id={post()?.userId}>   // inner createComponent
        // const [user] = createResource(...) -> getNextContextId
```

`createComponent` was:

```ts
export function createComponent<T>(Comp: (props: T) => JSX.Element, props: T): JSX.Element {
  if (sharedConfig.context && !sharedConfig.context.noHydrate) {
    const c = sharedConfig.context;
    setHydrateContext(nextHydrateContext());
    const r = Comp(props || ({} as T));
    setHydrateContext(c);
    return r;
  }
  return Comp(props || ({} as T));
}
```

Under noHydrate the function returned the child directly without pushing a new `HydrationContext`. So:

1. Post runs with the parent context. `post` resource gets `id = "parent-0"`, `count` becomes 1.
2. `<Suspense>` reads the current id (`"parent-1"`) and then runs its body with `setHydrateContext({ ...ctx, count: 0 })` — count is reset for the suspense boundary.
3. User runs with that same parent context (no isolation under noHydrate). `user` resource gets `id = "parent-0"` again because count was reset by Suspense — collision with `post.resources["parent-0"]`.

The fix is to always isolate the id space per component, even when noHydrate is set:

```ts
export function createComponent<T>(Comp: (props: T) => JSX.Element, props: T): JSX.Element {
  if (sharedConfig.context) {
    const c = sharedConfig.context;
    setHydrateContext(nextHydrateContext());
    const r = Comp(props || ({} as T));
    setHydrateContext(c);
    return r;
  }
  return Comp(props || ({} as T));
}
```

`nextHydrateContext` already spreads `...sharedConfig.context`, so the new context inherits `noHydrate: true` and downstream hydration markers continue to be skipped. The only behavior change is that the inner component's resource ids are derived from the outer component's freshly-allocated id, which keeps them unique across Suspense-driven count resets.

Verified: `pnpm test` in packages/solid → 469 / 469 pass (26 files).

No regression test added because the existing test surface for renderToStringAsync + NoHydration lives in `solid-ssr` / `dom-expressions` rather than this package's vitest suite. Happy to add one if you'd like — would need a few lines of harness to wire up `dom-expressions/src/server.js` here.

### DIFF
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -146,7 +146,15 @@ export function createUniqueId(): string {
 }
 
 export function createComponent<T>(Comp: (props: T) => JSX.Element, props: T): JSX.Element {
-  if (sharedConfig.context && !sharedConfig.context.noHydrate) {
+  if (sharedConfig.context) {
+    // Even under noHydrate we still need to isolate the id space for each
+    // component, so that `createResource` calls in sibling/nested components
+    // do not collide on the same context id when a parent `Suspense` resets
+    // `count` back to 0. Without this, a non-hydrating render of
+    // `<Post><Suspense><User/></Suspense></Post>` (each with a resource)
+    // produces two resources sharing the same id key in
+    // `context.resources` and the inner resource reads back the outer one's
+    // value. See #2546.
     const c = sharedConfig.context;
     setHydrateContext(nextHydrateContext());
     const r = Comp(props || ({} as T));


ryansolid:
Yeah as I commented in the original issue the challenge here is downstream libraries in the ecosystem like SolidStart literally using the equivalent of Dummy components to match id gen between client/server where the mount and ssr render points are different. I can't make this change without potentially breaking libraries that depend on this behavior.  Which is very awkward at this point. At minimum we'd need to makie it a minor release and then be like for this version of Start you need atleast this version of Solid. But more than likely this is just something that gets fixed in Solid 2.0 (and it already is).

ryansolid:
Won't fix in 1.x. Isolating id generation under NoHydration would change SSR ids for libraries that rely on the current behavior — SolidStart dummy components exist specifically to match client/server id gen across different mount points. This is already fixed in 2.0. Closing the related issues as not planned for 1.x.

PR #2710, opened 2026-05-21
fix: keep onCleanup order consistent between dev and prod

### LINKED ISSUE #1561: inconsistent owner in Component function (product vs development build)
### Describe the bug

A component function owner is different between a dev and prod build

```jsx
function MyComponent(){
     getOwner()   // <--  this is different between prod/dev builds.
                  //      in prod the owner is of the parent, in dev it is a new owner scope.
     return <></>
}
```

this inconsistency causes unexpected behavior, for example when using solid's `onCleanup` primitive 

[dev build playground](https://playground.solidjs.com/anonymous/42ddfed8-dade-4486-bae6-9b2e20647286)
image

[production build playground](https://solid-playground.netlify.app/?version=1.6.2#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfr2ZCouOAGU0Ac2hqsvWhABhNTgoCHpCKVl5RRU6DW09CysrGXoIZlw0f14AgAs0NREACgBKfiteP0Dg0PCi0t4AXgA+cohKyuZ-OLgAOjVaDyKVZnzClRLLdqlJisE4XCZ2lSSISWSIVPTM7IBBQkIGkDn-IJCww4aWto6HbvU+gaGVKAOJqcrJWYg5oUXGdoAHjyBREvH0zSm6x+6W6fGAZFytEMvjIC1cSMMAF0mg4nC53F4oGoirhGPQ4N80bgACocOAMXD1MrXakY5FFGTEtElb5WISicTMpqtQHNECIjllABk0t4gP2EQhkkBEN8IlozHo8AguF6HgWAFFgjrcAAhfAASWKLzeYBKAEJvmBJFigA)

image

note: production build is working as designed, it is dev that alters the expectations 

.

see playground links  for the difference between dev and prod

### Expected behavior

it is expected that dev and prod will work similarly, in regard to owner scopes.

### Screenshots or Videos

_No response_

### Platform

.

### Additional context

_No response_

ATOM00blue:
## Summary
In dev builds, components are wrapped in an extra owner scope by `devComponent` so devtools can observe component boundaries. This wrapper does not exist in production, which caused `onCleanup` callbacks registered directly in a component body to fire in a different order on disposal than in production.

This registers such cleanups on the same owner they would be in production by skipping past any dev component wrapper(s). The owner returned by `getOwner()` inside a component is intentionally left unchanged, so devtools continues to see the component owner.

## Test plan
- New tests in `dev.spec.ts` assert dev `onCleanup` order matches prod for sibling, nested, and re-created component cases
- Existing solid package test suite shows no new failures

Closes #1561

### DIFF
--- a/.changeset/dev-owner-cleanup-order.md
+++ b/.changeset/dev-owner-cleanup-order.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Keep `onCleanup` ordering consistent between development and production builds. In dev, components are wrapped in an extra owner scope for devtools; cleanups registered directly in a component body now register on the same owner they would have in production, so disposal order matches across build modes.
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1005,10 +1005,17 @@ export function onMount(fn: () => void) {
  * @description https://docs.solidjs.com/reference/lifecycle/on-cleanup
  */
 export function onCleanup<T extends () => any>(fn: T): T {
-  if (Owner === null)
+  let o = Owner;
+  // In dev, components are wrapped in an extra owner scope so that devtools can
+  // observe them (see `devComponent`). That wrapper does not exist in production,
+  // which means cleanups registered directly in a component body would otherwise
+  // be ordered differently on disposal. Skip past any dev component wrapper(s) so
+  // the cleanup is registered on the same owner it would be in production. See #1561.
+  if (IS_DEV) while (o && (o as DevComponent<any>).component) o = o.owner;
+  if (o === null)
     IS_DEV && console.warn("cleanups created outside a `createRoot` or `render` will never be run");
-  else if (Owner.cleanups === null) Owner.cleanups = [fn];
-  else Owner.cleanups.push(fn);
+  else if (o.cleanups === null) o.cleanups = [fn];
+  else o.cleanups.push(fn);
   return fn;
 }
 
--- a/packages/solid/test/dev.spec.ts
+++ b/packages/solid/test/dev.spec.ts
@@ -7,7 +7,9 @@ import {
   createComputed,
   DEV,
   createContext,
-  createComponent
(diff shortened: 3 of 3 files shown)


ryansolid:
Thanks for this — the underlying dev/prod discrepancy still reproduces on current `next`, so the problem is not obsolete. However, this implementation can no longer merge after the Solid 2 signals-core rewrite:

- `packages/solid/src/reactive/signal.ts` was deleted; `onCleanup` now delegates to `@solidjs/signals`.
- The owner fields this patch uses (`component`, `owner`, `cleanups`) are now `_component`, `_parent`, and `_disposal`, with ownership/disposal living in the signals package.
- Current same-owner cleanup order is FIFO, not the old LIFO behavior pinned here. Current reproduction is production-equivalent `[before, child, after]` vs dev wrapper `[child, before, after]`; the PR expects `[after, child, before]`.
- The tests also need the 2.0 two-phase `createEffect(compute, effect)` API and explicit `flush()` where updates are involved.

The remaining bug is narrower in 2.0: effect-returned cleanup (the idiomatic path) already has parity; raw component-body `onCleanup` still lands on the dev component wrapper and runs out of order.

Please rebase and rework around the current architecture rather than resolving the old-file conflict mechanically. A Solid-specific `_component` check should not be pushed down into `@solidjs/signals` (that would invert the dependency boundary). The likely design space is either a wrapper at the `solid-js` component layer or a generic lifecycle-transparency mechanism in signals, with new behavioral tests asserting current FIFO semantics in actual dev and production-equivalent paths. If the fix touches `packages/signals/src`, the changeset must include `@solidjs/signals: patch` (and `solid-js` only if its source also changes).

Leaving this open because the issue is real; it just needs a 2.0-native implementation.

PR #2731, opened 2026-05-30
[next] fix: snapshot array length override fallback

mathieuprog:
Use nullish coalescing so snapshot preserves an overridden array length of 0 instead of falling back to the source length

### DIFF
--- a/packages/solid-signals/src/store/utils.ts
+++ b/packages/solid-signals/src/store/utils.ts
@@ -13,7 +13,6 @@ import {
   storeLookup,
   trackSelf,
   wrap,
-  type Store,
   type StoreNode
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks @mathieuprog. I pulled this fix into `next` with an added regression test and changeset, and kept your contribution credited via Co-authored-by.

Merged locally as baa47d23 (`fix: preserve zero snapshot array length overrides`).

ryansolid:
Landed on  in baa47d23 with test coverage and changeset. Thanks again!

ryansolid:
Correction: landed on the next branch in baa47d23 with test coverage and a changeset.

PR #2742, opened 2026-06-04
Fix hydrated innerHTML with attribute spreads

### LINKED ISSUE #2737: Spread on element with innerHTML clears content during reactive updates after SSR hydration
### Describe the bug

  When using the spread operator on an element that also has innerHTML defined a reactive update to the spread
  object causes the element's content to be cleared if the app was hydrated from SSR. This does not happen when SSR is disabled or
  when attributes are bound individually.

```tsx
function App() {
  const [ref, setRef] = createSignal<HTMLDivElement>();

  // Reactive object for spread
  const data = createMemo(() => {
    return {
      "data-whatever": count() === 1 ? "yes" : "no",
    };
  });

  onMount(() => {
    console.log(ref()?.textContent); // Hello world! (Correct)
    setCount(1);
    setTimeout(() => {
      console.log(ref()?.textContent); // "" (BUG: Should be "Hello world!")
      console.log(ref()?.getAttribute("data-whatever")); // yes (Correct)
      console.log(ref()?.getAttribute("class")); // flex (correct)
      console.log(ref()?.getAttribute("style")); // color: red (Correct)
    }, 500);
  });

  return (
    <div
      ref={setRef}
      {...data()}
      class="flex"
      style="color: red"
      innerHTML="Hello world!"
    ></div>
  );
}

```

https://stackblitz.com/github/youyoumu/solid-spread-hydration-bug

https://github.com/youyoumu/solid-spread-hydration-bug

1. Clone the repo and install the dependency
2. Run the dev server
3. Open the dev server in browser and see the dom and the console.log ouput
4. change `ssr` to `false` on `app.config.ts` to confirm the bug only happen with ssr

### Expected behavior

The innerHTML should remain intact when the spread object is updated, as long as the spread object does not contain a children or innerHTML key.

## Actual Behavior
The element's content is cleared immediately upon the first reactive update of the spread object after hydration.

### Screenshots or Videos

Picture of chrome dev tools showing div with `data-whatever="yes"` having no children/innerHTML
Image

Picture of console output
Image

### Platform

- OS: Linux
- Browser: Chrome
- Version: 147.0.7727.116 (Official Build) (64-bit)

### Additional context

- The bug only occurs when ssr: true. With ssr: false, it works as expected.
- If the attributes are bound individually (e.g., `data-whatever={count() === 1 ? "yes" : "no"}`) instead of using a spread, the bug does not occur.

### SolidJS + Vite
```json
  "dependencies": {
    "@floating-ui/dom": "^1.7.6",
    "solid-js": "^1.9.13"
  },
```

### SolidStart 1
```json
  "dependencies": {
    "@solidjs/start": "1.3.2",
    "solid-js": "^1.9.5",
    "vinxi": "^0.5.7"
  },
```

### SolidStart 2
```json
  "dependencies": {
    "@solidjs/start": "2.0.0-alpha.2",
    "solid-js": "^1.9.5",
    "vite": "^7.0.0",
    "@solidjs/vite-plugin-nitro-2": "^0.1.0"
  },
```

puneetdixit200:
Fixes #2737.

This prevents an attribute-only spread from installing a child reconciliation effect when the spread props do not include child-bearing props. During hydration, that effect could clear existing innerHTML content on the first reactive spread update.

Added a regression test that hydrates server-rendered innerHTML, updates a spread-only data attribute, and verifies the content remains intact.

Verification:
- pnpm --dir packages/solid test web/test/element.spec.tsx
- git diff --check

Note: `pnpm --dir packages/solid exec tsc --project web/tsconfig.build.json` was attempted in a fresh checkout, but it requires generated local package links/types first and failed before checking this change.

### DIFF
--- a/packages/solid/web/src/client.ts
+++ b/packages/solid/web/src/client.ts
@@ -1,1 +1,18 @@
 export * from "dom-expressions/src/client.js";
+import { spread as baseSpread } from "dom-expressions/src/client.js";
+
+const childPropNames = ["children", "innerHTML", "textContent", "innerText"];
+
+export function spread(
+  node: Element,
+  props: Record<string, unknown> = {},
+  isSVG?: boolean,
+  skipChildren?: boolean
+) {
+  return baseSpread(
+    node,
+    props,
+    isSVG,
+    skipChildren || !childPropNames.some(prop => prop in props)
+  );
+}
--- a/packages/solid/web/test/element.spec.tsx
+++ b/packages/solid/web/test/element.spec.tsx
@@ -3,7 +3,15 @@
  * @vitest-environment jsdom
  */
 import { describe, expect, test } from "vitest";
-import { createRoot, createSignal, createUniqueId, JSX, children } from "../../src/index.js";
+import {
+  createMemo,
+  createRoot,
+  createSignal,
+  createUniqueId,
+  JSX,
+  children
+} from "../../src/index.js";
+import { hydrate } from "../src/index.js";
 
 declare module "solid-js/jsx-runtime" {
   namespace JSX {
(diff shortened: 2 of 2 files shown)


ryansolid:
Thanks for the repro and test — this is a real hydration bug, but the fix belongs upstream in **dom-expressions**, not in Solid.

PR #2756, opened 2026-06-12
perf: optimize signals hot paths

brenelz:
## Summary

Four targeted optimizations to `@solidjs/signals` hot paths, driven by the CodSpeed flamegraphs from [run 6a2a5421](https://app.codspeed.io/solidjs/solid/runs/6a2a54212e49bc3a0c2d0f55) plus local `--cpu-prof` profiles of the same workloads.

### 1. O(1) dependency revalidation (`graph.ts`, `types.ts`, `core.ts`)

When a computation re-reads a dependency it already saw earlier in the same recompute pass non-consecutively (e.g. `doc.meta.author` then `doc.meta.timestamp` — `doc.meta` read twice), `link()` fell back to `isValidLink`, which scans the dep list **from the head** to check membership. With large dep lists this is O(n²): it was **51% of total time** in the `reconcile: deep tree, all ~12k paths subscribed` bench.

Links now carry a `_gen` stamp of the subscriber's `_depGen` recompute counter, set on creation and on in-order reuse. The membership question "was this dep already revalidated this pass?" becomes one integer compare. Equivalence: a link is inside the `[head.._depsTail]` prefix iff it was created or (B)-reused during the current pass, which is exactly what the stamp records.

### 2. Skip redundant subscriber walks on batched writes (`core.ts`, `heap.ts`, `effect.ts`)

Writing the same signal N times in one batch walked the full subscriber list N times — after the first walk, every subscriber is already queued, so the remaining walks are provable no-ops (the `updateSignals:update1to1000` flamegraph is ~95% `insertSubs` + `insertIntoHeap`).

`setSignal` now skips the walk while a global `notifyEpoch` is unchanged. The epoch advances whenever a queued notification is *consumed* — any heap removal (`deleteFromHeap`, which all recompute paths go through) or a tracked-effect run draining `_modified` — and a signal's cached epoch is invalidated when it gains a new subscriber (`link()`) and on optimistic writes, which mutate subscriber lane state and always walk.

### 3. Reconcile allocation reduction (`reconcile.ts`)

- `getAllKeys` allocated `[...keys, ...nextKeys]` + a `Set` + `Array.from` for every tracked node in the tree; it now returns the existing keys array when key sets match (the common shape), and merges without the spread copies otherwise.
- `unwrap` did `$TARGET` symbol lookups on primitive leaves, forcing boxing; primitives now return immediately.

### 4. Small allocation cleanups (`store.ts`, `utils.ts`, `core.ts`)

- `getKeys` only wraps `Object.keys` in an `untrack` closure when the source is itself a wrapped store (plain objects can't trigger traps).
- `snapshotImpl`'s no-override walk reads each property once via the descriptor instead of three times.
- Effects reuse one bound runner instead of allocating `_runEffect.bind(null, el)` per update (`runEffect` already no-ops on stale `_modified`, so re-enqueueing the same function is safe).

## Results (local, dev build, same workloads as the CodSpeed suite)

| Benchmark | Before | After | Δ |
|---|---|---|---|
| reconcile: all ~12k paths subscribed | 27.5 ms | 5.6 ms | **4.9x** |
| reconcile: single `deep()` effect | 13.4 ms | 5.6 ms | **2.4x** |
| updateSignals:update1to1000 | 1.44 ms | 0.06 ms | **24x** |
| reconcile: 10 sparse paths | 0.43 ms | 0.38 ms | ~12% |

Expected CodSpeed movement: `update1to1000` (183.8 ms), `reconcile all paths` (84.2 ms), and `reconcile deep()` (76.6 ms) should drop substantially. The DOM-lane benches won't move much — their flamegraphs are 75–95% jsdom internals.

## Trade-offs

- +1 field per link (`_gen`), +2 fields per node (`_depGen`, `_notifyEpoch`): ~8–16 bytes each, may show as a sub-percent tick on the creation benches.
- `notifyEpoch` is a monotonically increasing double — no wraparound concern in practice.

## Testing

- 749 solid-signals + 293 solid-web + 415 solid tests pass
- `tsc -p tsconfig.build.json` clean
- Verified the epoch-skip invalidation paths: heap pops, tracked-effect runs, sync pull-based recomputes, `suppressComputedRecompute`, dispose, and mid-batch new subscribers

### DIFF
--- a/.changeset/perf-signals-hot-paths.md
+++ b/.changeset/perf-signals-hot-paths.md
@@ -1,0 +1,18 @@
+---
+"@solidjs/signals": patch
+---
+
+perf: optimize reactive hot paths
+
+- O(1) dependency revalidation: replace the `isValidLink` dep-list scan with a
+  per-recompute generation stamp on links, eliminating O(n²) behavior when a
+  computation re-reads a dependency it already saw during the same pass
+- Skip redundant subscriber walks when a signal is written multiple times in
+  the same batch (epoch-invalidated by heap consumption, tracked-effect runs,
+  or new subscribers)
+- Reduce reconcile allocations: reuse the existing key array when key sets
+  match in `getAllKeys`, and skip symbol lookups on primitive leaves in
+  `unwrap`
+- Avoid the `untrack` closure in `getKeys` for plain (non-proxy) sources
+- Specialize `snapshotImpl`'s no-override walk to read each property once
+- Cache one bound effect runner per effect instead of allocating per update
--- a/packages/solid-signals/src/core/core.ts
+++ b/packages/solid-signals/src/core/core.ts
@@ -39,7 +39,8 @@ import {
   insertIntoHeap,
   insertIntoHeapHeight,
   markHeap,
-  markNode
+  markNode,
+  notifyEpoch
 } from "./heap.js";
 import {
   findLane,
@@ -179,6 +180,7 @@ export function recompute(el: Computed<any>, create: boolean = false): void {
   const oldcontext = context;
   context = el;
   el._depsTail = null;
+  el._depGen++;
   el._flags = REACTIVE_RECOMPUTING_DEPS;
   el._time = clock;
   let value = el._pendingValue === NOT_PENDING ? el._value : el._pendingValue;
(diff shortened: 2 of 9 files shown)


brenelz:
Used the new Fable 5 model to test so see if it had any perf suggestions and it came up with these.

I more or less wanted to see how codspeed would show up for these perf improvements.

Feel free to take any of them or discard them

ryansolid:
Great find on `isValidLink` — the head-scan on non-consecutive dep re-reads was exactly the right diagnosis, and it's the dominant cost in store-heavy recomputes (your 51% flamegraph number reproduced on our end). We've landed the gen-stamp revalidation plus the reconcile/store allocation trims (`getAllKeys` fast path, `unwrap` primitive early-return, `getKeys` untrack skip, cached bound runner) on `next`, re-implemented against the current tree since the branch predates a couple of `link()`/`setSignal` rewrites — with your numbers verified: ~7x on deep-tree reconcile with all paths subscribed, ~5x on a `deep()` effect.

The one piece we deliberately didn't take is the `notifyEpoch` skip-walk. It's a global invalidation scheme where correctness depends on catching every path that consumes a queued notification, and a missed one means silently stale UI — and the scheduler internals it has to track (optimistic lanes, transition holds, gated subscribers) have been substantially rebuilt since June. The bench it targets (writing one signal 1000× in a batch) is also not a shape real apps hit. If that path ever shows up in real workloads we'd want to re-derive the invalidation set against the current scheduler as its own change. The `snapshotImpl` specialized walk landed alongside the #2850 fix (it was waiting on that ruling, since it touches the same function).

Closing since the taken parts are in — thanks, this was a genuinely valuable profile-driven find.

PR #2782, opened 2026-06-13
fix(signals): defer onSettled fallback cleanup to owner disposal (closes #2766)

### LINKED ISSUE #2766: 2.0.0-beta.14: Cleanup returned from nested `onSettled` runs immediately
### Describe the bug

In Solid 2.0.0-beta.14, a composable helper that registers `onSettled()` can behave incorrectly when the helper is called from inside another `onSettled()` callback.

The inner `onSettled()` callback runs, but its returned cleanup function is called immediately during the same flush instead of being deferred until the surrounding owner is disposed. This causes setup/teardown helpers to immediately tear themselves down.

https://stackblitz.com/edit/solidjs-templates-ayujgi7k?file=src%2FApp.tsx

1. Open the StackBlitz repro.
2. Wait for the app to finish its initial render/settle.
3. Observe the rendered status text.

### Expected behavior

The helper's setup should remain active until the component/root is disposed.

The page should render:

```text
subscribed
```

The cleanup returned by the inner `onSettled()` should run on owner disposal, not immediately after setup.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.14

### Additional context

This can happen through normal composition, not only by intentionally nesting lifecycle calls. For example, a helper may use `onSettled()` internally to install and clean up a subscription, while a parent calls that helper from an outer `onSettled()` because it wants to wait until the component has reached a stable settled state.

The issue appears related to `packages/solid-signals/src/signals.ts`. When `onSettled()` is called while the current owner has children forbidden, it falls back to the ownerless queue branch. That branch invokes the returned cleanup immediately after the callback runs.

tsushanth:
## Summary

Closes #2766.

`onSettled` falls back to the global queue when the current owner is null or has `CONFIG_CHILDREN_FORBIDDEN`. The most common trigger for that fallback is `onSettled()` called from inside another `onSettled()` callback (the outer callback runs as a tracked effect, which sets `CONFIG_CHILDREN_FORBIDDEN`). The previous fallback shape:

```ts
globalQueue.enqueue(EFFECT_USER, () => {
  const cleanup = callback();
  …
  cleanup?.();        // ❌ runs the cleanup eagerly in the same flush
});
```

invoked the returned cleanup eagerly inside the same flush, immediately tearing down helpers whose intent was "install on settle, tear down on dispose." Composable helpers using `onSettled` internally — subscription setup, listener install, focus management — set up and ripped themselves down in the same tick.

Capture the calling owner and, when the callback returns a function, register it via `cleanup()` under that owner with `runWithOwner`. Cleanup now fires when the captured owner is disposed, matching the contract of the non-forbidden path. Orphaned helpers (no owner at all) drop the cleanup rather than run it immediately — there is no point at which firing it would be correct.

## Test plan

Added `defers cleanup from nested onSettled to owner disposal (#2766)` to `tests/onSettled.test.ts` mirroring the reporter's scenario:

```ts
function useSubscription() {
  onSettled(() => {
    log.push("subscribed");
    return () => log.push("unsubscribed");
  });
}
const dispose = createRoot(stop => {
  onSettled(() => { useSubscription(); });
  return stop;
});
flush();
expect(log).toEqual(["subscribed"]);
dispose();
expect(log).toEqual(["subscribed", "unsubscribed"]);
```

```
pnpm --filter @solidjs/signals build
cd packages/solid-signals && npx vitest run tests/onSettled.test.ts
```

Result: 13/13 pass with the fix; reverting `signals.ts` to upstream and re-running shows the new case fails (`expected ["subscribed","unsubscribed"] to equal ["subscribed"]`) — confirming the test isolates the regression.

### DIFF
--- a/packages/solid-signals/src/signals.ts
+++ b/packages/solid-signals/src/signals.ts
@@ -745,15 +745,33 @@ export function createOptimistic<T>(
  */
 export function onSettled(callback: () => void | (() => void)): void {
   const owner = getOwner();
-  owner && !(owner._config & CONFIG_CHILDREN_FORBIDDEN)
-    ? createTrackedEffect(() => untrack(callback), __DEV__ ? { name: "onSettled" } : undefined)
-    : globalQueue.enqueue(EFFECT_USER, () => {
-        const cleanup = callback();
-        if (__DEV__ && cleanup !== undefined && typeof cleanup !== "function") {
-          throw new Error(
-            "onSettled callback returned an invalid cleanup value. Return a cleanup function or undefined."
-          );
-        }
-        cleanup?.();
-      });
+  if (owner && !(owner._config & CONFIG_CHILDREN_FORBIDDEN)) {
+    createTrackedEffect(() => untrack(callback), __DEV__ ? { name: "onSettled" } : undefined);
+    return;
+  }
+  // Fallback path: either no owner, or the current owner is a children-
+  // forbidden scope (the most common case is `onSettled()` called from
+  // inside another `onSettled()` callback, which itself runs as a
+  // tracked effect). The original implementation invoked the returned
+  // cleanup eagerly inside the same flush, which immediately tore down
+  // setup helpers (#2766). Defer cleanup via the captured owner's
+  // disposal chain so it fires on real disposal, not on settle.
+  const capturedOwner = owner;
+  globalQueue.enqueue(EFFECT_USER, () => {
+    const result = callback();
+    if (__DEV__ && result !== undefined && typeof result !== "function") {
+      throw new Error(
+        "onSettled callback returned an invalid cleanup value. Return a cleanup function or undefined."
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks for taking this on. We resolved #2766 with a different contract than this PR takes — see https://github.com/solidjs/solid/issues/2766#issuecomment-4860524647 for the full reasoning. In short: when `onSettled` runs in an unowned scope there's no lifecycle for a returned cleanup to bind to, and tying it to the surrounding owner's disposal manufactures a relationship the primitive never established. Instead, returning a cleanup from an unowned `onSettled` is now a dev-mode error (dropped in production), and the owned form — calling the setup helper directly from the component body — already covers the reporter's use case. Closing in favor of that approach.

PR #2783, opened 2026-06-13
fix(signals): surface sync rejection from thenables (closes #2764)

### LINKED ISSUE #2764: 2.0.0-beta.14: Synchronously rejecting thenable keeps Loading pending and loses the error
### Describe the bug

In Solid 2.0.0-beta.14, a memo that returns a Promise-like thenable can get stuck in the pending path if the thenable calls its rejection handler synchronously during `.then(...)`.

The same boundary setup handles normal async rejections, but the synchronous rejection never reaches `<Errored>`. The UI stays on the `<Loading>` fallback forever and the original error is lost.

https://stackblitz.com/edit/solidjs-templates-t7nncvm9?file=src%2FApp.tsx

1. Open the StackBlitz reproduction
2. Observe that the app renders `loading...` and never changes to the error fallback.

### Expected behavior

The synchronous rejection should be handled the same way as an asynchronous rejection from a Promise or thenable.

The pending state should settle, and `<Errored>` should render the error fallback, for example:

```text
error: Error: sync-reject
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.14

### Additional context

This appears to come from the thenable branch in `packages/solid-signals/src/core/async.ts`. The rejection callback only handled errors after the `.then(...)` call had returned, so a synchronously rejecting thenable could throw the pending `NotReadyError` path without recording or propagating the actual rejection.

The practical shape is a library/cache thenable that already knows it failed and calls `reject(...)` during the synchronous `.then(...)` call.

tsushanth:
Closes #2764.

When a thenable invokes its rejection handler synchronously during `.then(...)`, the error was being silently dropped. The handler's `if (!isSync) handleError(e)` branch skipped the call during the sync window, and the subsequent `if (!resolved)` branch unconditionally threw `NotReadyError` — so the boundary observed pending forever and never reached `Errored`.

The fix mirrors the success handler's sync/async split for rejections: capture the error inside the sync window, then after `.then` returns, surface it through `handleError` and rethrow so the calling computation propagates it to the nearest error boundary.

## Test

Added a regression test in `tests/syncThenable.test.ts` that wraps a synchronously-rejecting thenable in a `createMemo` under `createErrorBoundary` and asserts the boundary receives the error.

Without the fix: test times out / boundary stays pending.
With the fix: boundary catches the error.

### DIFF
--- a/packages/solid-signals/src/core/async.ts
+++ b/packages/solid-signals/src/core/async.ts
@@ -233,6 +233,8 @@ export function handleAsync<T>(
 
   if (isThenable) {
     let resolved = false,
+      rejected = false,
+      syncError: any,
       isSync = true;
     (result as PromiseLike<T>).then(
       v => {
@@ -242,11 +244,21 @@ export function handleAsync<T>(
         } else asyncWrite(v);
       },
       e => {
-        if (!isSync) handleError(e);
+        if (isSync) {
+          // Library/cache thenables can invoke the rejection handler
+          // synchronously during `.then(…)`. Capturing the error here and
+          // surfacing it after `.then` returns is what lets `<Errored>` see
+          // it instead of being stuck on the pending path forever (#2764).
+          syncError = e;
+          rejected = true;
+        } else handleError(e);
       }
     );
     isSync = false;
-    if (!resolved) {
+    if (rejected) {
+      handleError(syncError);
+      throw syncError;
+    } else if (!resolved) {
       globalQueue.initTransition(resolveTransition(el as any));
       throw new NotReadyError(context!);
     }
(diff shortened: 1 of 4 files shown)


ryansolid:
Thanks for this — the sync-rejection diagnosis is exactly right. A thenable that rejects synchronously during `.then()` was having its error dropped (the reject handler only acted in the async case), so the node stayed pending and `<Errored>` never fired. We've landed an equivalent fix on `next`: the thenable branch now captures a synchronous rejection alongside the sync-resolve path and settles it, with a regression test.

One thing we can't take from this PR is the bundled `onSettled` change. We resolved #2766 with a different contract — returning a cleanup from an `onSettled` that runs in an unowned scope is a dev-mode error (dropped in production), rather than deferring the cleanup to the surrounding owner's disposal — so the two approaches conflict. Since the thenable fix is already in, I'll close this; the credit for the #2764 diagnosis is yours.

PR #2785, opened 2026-06-13
fix(signals): await non-Promise thenables yielded from action (closes #2765)

### LINKED ISSUE #2765: 2.0.0-beta.14: `action()` resumes with raw Promise-like thenable instead of awaiting it
### Describe the bug

In Solid 2.0.0-beta.14, `action()` only awaits yielded native `Promise` instances. If an action generator yields a valid Promise-like thenable that is not `instanceof Promise`, the generator resumes immediately with the raw thenable object instead of waiting for its settled value.

This affects valid promise-like values that work with await but are not instance of Promise.

https://stackblitz.com/edit/solidjs-templates-mqmnn6u9?file=src%2FApp.tsx

1. Open the StackBlitz repro.
2. Click `run action`.
3. Observe the rendered result.
4. See `got: [object Object]`

### Expected behavior

`action()` should treat yielded Promise-like thenables the same way `await` does: wait for the thenable to settle, then resume the generator with the fulfilled value.

The page should render:

```text
got: 42
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.14

### Additional context

The issue appears to come from `packages/solid-signals/src/core/action.ts`, where yielded values were checked with `instanceof Promise`. That misses valid Promise-like values whose shape is `{ then(...) { ... } }`.

The check should be based on thenability, for example whether the value is non-null and has a callable `then` property.

tsushanth:
Closes #2765.

\`action.ts\` was checking yielded values with \`instanceof Promise\`. That skips anything whose \`then\` is a method but whose prototype isn't \`Promise.prototype\` — userland thenables, cache library handles (the \`syncThenable\` shape used elsewhere in this same package), and engines that pass cross-realm promises across worker boundaries. Every such yield landed in the synchronous \`run(r)\` branch, so the generator resumed immediately with the raw thenable object — \`got: [object Object]\` in the StackBlitz repro.

The fix replaces both \`instanceof Promise\` checks with a small \`isThenable\` predicate (\`value != null && (object|function) && typeof value.then === "function"\`), the same shape \`await\` itself uses, and casts the awaited path to \`PromiseLike<...>\`. Native Promises still satisfy the predicate, so the existing test cases for \`yield Promise.resolve(42)\` etc. keep their behavior — the change only widens what counts as awaitable.

## Test

Added \`should await a non-Promise thenable and resume with its fulfilled value (#2765)\` in \`packages/solid-signals/tests/action.test.ts\`:

- yields \`{ then(onFulfilled) { queueMicrotask(() => onFulfilled(42)); } }\`
- asserts the generator does NOT resume synchronously with the raw object (pre-fix bug)
- asserts the resumed value is \`42\` after the microtask drains

Without the patch the new test fails at \`expect(receivedValue).toBe(42)\` (receives the thenable object); with the patch it passes. The full \`pnpm --filter @solidjs/signals test tests/action.test.ts\` suite (57 tests) stays green.

### DIFF
--- a/packages/solid-signals/src/core/action.ts
+++ b/packages/solid-signals/src/core/action.ts
@@ -15,6 +15,14 @@ function restoreTransition<T>(transition: Transition, fn: () => T): T {
   return result;
 }
 
+function isThenable(value: unknown): value is PromiseLike<unknown> {
+  return (
+    value != null &&
+    (typeof value === "object" || typeof value === "function") &&
+    typeof (value as { then?: unknown }).then === "function"
+  );
+}
+
 /**
  * Wraps a generator function so each invocation runs as a single transaction
  * (a "transition") that batches every signal/store write between yields. The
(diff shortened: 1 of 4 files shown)


ryansolid:
Good catch — `action()` only awaited native `Promise` instances, so a yielded thenable that wasn't `instanceof Promise` resumed the generator with the raw object instead of its settled value. We've landed an equivalent fix on `next`: yield handling now uses a thenability check (`typeof value.then === "function"`) at both sites, matching how `await` treats thenables, with a regression test.

The `map.ts`/repeat change bundled here is already on `next` (the `_offset` reset from the earlier repeat fix), so there's nothing to take there. Closing since the action fix is in — thanks for the thenable fix.

PR #2786, opened 2026-06-14
Fix createProjection seed type inference

brenelz:
## Summary
- type createProjection seeds like createStore's projection seed
- keep the solid-js hydration wrapper signature aligned with @solidjs/signals
- add a patch changeset

## Testing
- pnpm --filter @solidjs/signals types
- pnpm --filter solid-js types


brenelz:
Yeah doesn't look like this is a complete fix. Its break's some tests. Ill try to make it more robust

brenelz:
I think this is a better fix now

PR #2788, opened 2026-06-18
fix: handle non-object entries in keyed reconcile

### LINKED ISSUE #2772: 2.0.0-beta.14: Keyed `reconcile` crashes on nullable or primitive array entries
### Describe the bug

In Solid 2.0.0-beta.14, keyed `reconcile(..., "id")` can crash when an array contains entries that are not keyed objects.

Two related cases:

1. An array contains `null` entries between keyed objects.
2. An existing object entry is replaced by a primitive value.

Both are valid array shapes for API data or mixed collections, and `reconcile` should preserve/replace those entries rather than assuming every item can be keyed as an object.

https://stackblitz.com/edit/solidjs-templates-gqv5mur8?file=src%2FApp.tsx

```tsx
import { createSignal, createStore, reconcile } from 'solid-js';

function NullEntryRepro() {
  const [list, setList] = createStore<any[]>([{ id: 1 }, null, { id: 2 }]);
  const [result, setResult] = createSignal('not run');

  return (
    <section>
      <h2>null entry</h2>

      <button
        onClick={() => {
          try {
            setList(
              reconcile([{ id: 1 }, null, { id: 2, value: 'updated' }], 'id')
            );
            setResult(JSON.stringify(list));
          } catch (error) {
            setResult(String(error));
          }
        }}
      >
        reconcile null entry
      </button>

      <p>{result()}</p>
    </section>
  );
}

function PrimitiveEntryRepro() {
  const [list, setList] = createStore<any[]>([{ id: 1, value: 'object' }]);
  const [result, setResult] = createSignal('not run');

  return (
    <section>
      <h2>object to primitive</h2>

      <button
        onClick={() => {
          try {
            setList(reconcile([5], 'id'));
            setResult(JSON.stringify(list));
          } catch (error) {
            setResult(String(error));
          }
        }}
      >
        reconcile object to primitive
      </button>

      <p>{result()}</p>
    </section>
  );
}

  return (
    <main>
      <NullEntryRepro />
      <PrimitiveEntryRepro />
    </main>
  );
}

```

### Repro 1: `null` entry

1. Open the repro.
2. Click `reconcile null entry`.
3. Observe the rendered result.

### Repro 2: object entry replaced by primitive

1. Open or refresh the repro.
2. Click `reconcile object to primitive`.
3. Observe the rendered result.

### Expected behavior

The `null` entry should be preserved and the keyed object after it should update:

```json
[{"id":1},null,{"id":2,"value":"updated"}]
```

### Repro 2

The object entry should be replaced by the primitive:

```json
[5]
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: current stable

### Additional context

These appear related to `packages/solid-signals/src/store/reconcile.ts` in the keyed array reconciliation paths. The implementation appears to assume keyed object entries in places where arrays may contain `null`, `undefined`, or primitive values.

cyphercodes:
## Summary

- Fix keyed `reconcile` in `@solidjs/signals` so array entries that are `null`, `undefined`, or primitives are treated as direct values instead of being wrapped as store objects.
- Preserve keyed object reconciliation for wrappable entries while replacing non-object entries safely.
- Add regression coverage for nullable keyed arrays and object-to-primitive array replacement.

Closes #2772

- `pnpm --dir packages/solid-signals exec vitest run tests/store/reconcile.test.ts`
- `pnpm exec prettier --check packages/solid-signals/src/store/reconcile.ts packages/solid-signals/tests/store/reconcile.test.ts`
- `git diff --check`
- `pnpm test`

### DIFF
--- a/packages/solid-signals/src/store/reconcile.ts
+++ b/packages/solid-signals/src/store/reconcile.ts
@@ -31,6 +31,34 @@ function getAllKeys(value, override, next) {
   return Array.from(new Set([...keys, ...nextKeys]));
 }
 
+function wrapValue(value: any, target: any) {
+  return isWrappable(value) ? wrap(value, target) : value;
+}
+
+function keyedMatch(a: any, b: any, keyFn: (item: NonNullable<any>) => any) {
+  return a === b || (isWrappable(a) && isWrappable(b) && keyFn(a) === keyFn(b));
+}
+
+function itemKey(item: any, keyFn: (item: NonNullable<any>) => any) {
+  return isWrappable(item) ? keyFn(item) : item;
+}
+
+function applyArrayItem(
+  next: any,
+  previous: any,
+  target: any,
+  node: any,
+  keyFn: (item: NonNullable<any>) => any
+) {
+  if (isWrappable(next) && isWrappable(previous)) {
+    const wrapped = wrap(previous, target);
+    node && setSignal(node, wrapped);
+    applyState(next, wrapped, keyFn);
+  } else {
+    node && setSignal(node, wrapValue(next, target));
+  }
+}
+
 // Dispatcher: every applyState call (including recursion) checks for the
 // presence of override / optimistic-override slots once and routes to the
 // appropriate body. The fast body never calls `getOverrideValue` and never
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks for catching this and for the PR, @cyphercodes! 🙏

We've landed a fix for #2772 in `bc92d002` on `next`. It takes a slightly broader approach than this PR: rather than guarding a single site, it guards every keyed-diff touch point (match / key / wrap) in both the fast and slow `applyState` paths, so `null`/`undefined`/primitive entries are safely preserved or replaced wholesale in any position (leading/trailing scans and the map/LIS reorder), including object→primitive and back.

We also added a UIBench-`tree` store-`reconcile` Tier-1 bench (`reconcile-tree.bench.ts`) and confirmed the guards add no measurable cost to the keyed reorder path.

Closing since the issue is now fixed on `next`, but thank you again — the report and repro were spot on.

PR #2810, opened 2026-07-01
fix(repo): patch npm-run-all to detect cjs npm exec paths

better-salmon:
## Summary

`pnpm` managed by `vite-plus` causes this error:

```
ERROR: spawn /Users/name/.vite-plus/package_manager/pnpm/11.1.1/pnpm/bin/pnpm.cjs EACCES
```

In this particular setup `pnpm 11` exposes `npm_execpath` as `pnpm.cjs`. But `npm-run-all` only treated `.js/.mjs` paths as Node shims. It tried to execute `pnpm.cjs` directly and hit `EACCES`, so the patch treats `.cjs` as a JS shim and runs it through `node`.

Updating `pnpm` to 11.9.0 did not help.

`pnpm build` now works. The patch is minimal and sufficient.

### DIFF
--- a/patches/npm-run-all@4.1.5.patch
+++ b/patches/npm-run-all@4.1.5.patch
@@ -1,0 +1,13 @@
+diff --git a/lib/run-task.js b/lib/run-task.js
+index 8cabc03c7ef64f5ca57378009618222fc97ac9b7..53166e6a81a1fb5dfcf0c489ef5815f864f80a13 100644
+--- a/lib/run-task.js
++++ b/lib/run-task.js
+@@ -155,7 +155,7 @@ module.exports = function runTask(task, options) {
+ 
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks for the report and the patch — the diagnosis is right. Rather than carrying a patch for `npm-run-all` (unmaintained since 2018), we've switched to `npm-run-all2` (`d6bc696c`), the maintained fork, which has the same `.cjs` handling built in along with other pnpm fixes. Same CLI, no script changes. Your setup should work on `next` now. Closing in favor of that, but appreciate you tracking it down.

better-salmon:
Yeah, this is better! I was a little afraid to update a dependency as my first contribution, so I tried to keep the patch minimal and easy to reason about.

Issue #2046, opened 2024-01-23
Stale read from <Show> under a transition


deluksic:
### Describe the bug

I'm encountering `Stale read from <Show>` while a transition is happening. I'm not sure exactly what the problem is, but I created a simple reproduction.

### Your Example Website or App

https://playground.solidjs.com/anonymous/bdbea778-bc1d-4508-963f-6683c38de3d5

### Steps to Reproduce the Bug or Issue

1. Click the counter and move the mouse at the same time

### Expected behavior

I expect to see the tooltip instead (some text and coordinates).

### Platform

- SolidJS 1.8

ryansolid:
Yeah.. I see it.. interestingly only when the browser console is open but that suggests there is a timing issue here. Or more likely the internals of Show that do the check aren't concurrent safe. Thanks for reporting.

deluksic:
Can similar issues happen even if we don't use the callback form? I use this form a lot:
```tsx
<Show when={condition()}>
  <Component prop={condition()!.something} />
</Show>
```

In this case, I assumed that `<Component>` would be unmounted before it had a chance to read from `condition()` and failing the `.something`. Reading the implementation of `<Show>`, I don't see much difference between callback vs non-cb forms, except for the arrow fn wrap and throw. I will try to create a test when I find time.

ryansolid:
You won't get the error because we don't do the checks. The reason we do is because someone could try to access the Signal inside like a timeout where it was no longer defined with a value which would mean TS lied. https://github.com/solidjs/solid/blob/main/packages/solid/src/render/flow.ts#L130C36-L130C45

Hmm.. interestingly it doesn't look like it is what expected. Reproducing this one will probably require me to run it in replay.

mdynnl:
more controllable repro https://playground.solidjs.com/anonymous/e5a35370-6712-46cb-af2b-78d7eca848c0

deluksic:
> more controllable repro https://playground.solidjs.com/anonymous/e5a35370-6712-46cb-af2b-78d7eca848c0

Can you please add steps for reproduction? I'm not sure what am I looking at. 😅

emansom:
This seems to happen when events trigger DOM changes too fast, and the internal state in Solid does not match the real DOM anymore.

Workaround that worked for me, was limiting the amount of changes that could possibly happen at a single moment by disabling buttons that triggered a lot of DOM changes for a few seconds.

[...]

PR #2812, opened 2026-07-01
fix(solid): exclude keys shared across splitProps groups on the proxy path

spokodev:
### Problem

`splitProps` proxy path duplicates a key that is shared across two groups, disagreeing with its own plain-object path and with the documented contract that a property belongs to a single group.

### Repro

```js
// plain object source
const [g1, g2, rest] = splitProps({ a: 1, b: 2, c: 3 }, ["a", "b"], ["b", "c"]);
// g2.b === undefined, "b" in g2 === false, Object.keys(g2) === ["c"], { ...g2 } === { c: 3 }   (correct)

// createStore proxy source (same as component props at runtime)
const [state] = createStore({ a: 1, b: 2, c: 3 });
const [p1, p2, prest] = splitProps(state, ["a", "b"], ["b", "c"]);
// ACTUAL:   p2.b === 2, "b" in p2 === true, Object.keys(p2) === ["b", "c"], { ...p2 } === { b: 2, c: 3 }
// EXPECTED: p2.b === undefined, "b" in p2 === false, Object.keys(p2) === ["c"], { ...p2 } === { c: 3 }
```

The key `b` is consumed by the first group but still leaks into the second group through get, has, enumeration, and spread.

### Root cause

`splitProps` has two implementations. The plain-object path walks each property once and assigns it to the first group that lists it (first-match `break`), so keys are exclusive. The proxy path instead builds one `Proxy` per group whose `get`/`has`/`keys` each test `k.includes(property)` against that group's own key list, with no cross-group exclusion. A key present in two groups is therefore reported by both proxied groups. Component props and stores are proxies at runtime, so the proxy path is the one that actually runs for components.

### Fix

In the proxy branch, track keys claimed by earlier groups and let each group own only its not-yet-claimed keys. `get`, `has`, and `keys` all read from that owned set, bringing the proxy path to parity with the plain-object first-match semantics. The `rest` proxy is unchanged, since it already excludes `keys.flat()`.

### Authority

Solid docs state `splitProps` partitions props and each property belongs to one group, and the plain-object implementation in the same function already enforces this with a first-match `break`.

### Tests

Added a case to `packages/solid/test/component.spec.ts` asserting first-group-only ownership for both a plain object and a `createStore` proxy across get, has, keys, and spread. It fails on the current proxy path (`expected 2 to be undefined`) and passes with the fix.

`packages/solid` suite: 482 passing plus the new test. The only failure is the unrelated pre-existing `web/test/transition.spec.tsx` "Transition memo stale read (#2046)" flake, which is nondeterministic and passes when run in isolation on the clean tree.

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -293,17 +293,20 @@ export function splitProps<
 
   if (SUPPORTS_PROXY && $PROXY in props) {
     const blocked = len > 1 ? keys.flat() : keys[0];
+    const claimed = new Set<PropertyKey>();
     const res = keys.map(k => {
+      // a key belongs to the first group that lists it (matches non-proxy path)
+      const owned = k.filter(property => !claimed.has(property) && (claimed.add(property), true));
       return new Proxy(
         {
           get(property) {
-            return k.includes(property) ? props[property as any] : undefined;
+            return owned.includes(property) ? props[property as any] : undefined;
           },
           has(property) {
-            return k.includes(property) && property in props;
+            return owned.includes(property) && property in props;
           },
           keys() {
-            return k.filter(property => property in props);
+            return owned.filter(property => property in props);
           }
         },
         propTraps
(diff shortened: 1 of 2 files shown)


ryansolid:
Landed on main in 59649bc7. The proxy path now tracks keys claimed by earlier groups so get/has/keys/spread match the plain-object first-match contract. Thanks.

Issue #2813, opened 2026-07-02
2.0.0-beta.15: A throwing effect cleanup permanently wedges its effect (stale cleanup re-invoked on every run)


yumemi-thomas:
### Describe the bug

If a cleanup function returned by a `createEffect` callback throws, the effect is permanently broken:

1. The effect body never runs again for any subsequent update.
2. Instead, the **same dead cleanup is re-invoked** (and re-throws) on every later run — the stale cleanup reference is never cleared.
3. An `<ErrorBoundary>` wrapping the effect cannot catch the cleanup error — it escapes as uncaught even though effect-body errors thrown one line later are caught fine.
4. Dev-only side effect: the strict-read guard armed for the effect run is never restored, so after the first throw every legal untracked signal read anywhere in the app (e.g. `n()` inside a click handler) falsely warns `[STRICT_READ_UNTRACKED] Reactive value read directly in an effect callback will not update.` for the rest of the session.

Related to but distinct from #2761 / #2762: a throwing cleanup can also trigger the scheduler stall those issues describe, but this wedge is a separate per-effect defect that survives fixing them. The repro is arranged so neither scheduler bug masks it (see Additional context).

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-uphtinhn?file=src%2FApp.tsx

```tsx
import { createSignal, createEffect } from 'solid-js';

export default function App() {
  const [n, setN] = createSignal(0);
  const [b, setB] = createSignal(0);

  // The wedge target. Deliberately arranged so neither scheduler bug can
  // mask it: this is the ONLY user effect subscribed to `n` (nothing in its
  // batch for #2762 to abandon), and it writes no signals (nothing to strand
  // the scheduler, so no #2761 stall).
  createEffect(n, (v) => {
    console.log('body ran with n =', v);
    return () => {
      console.log('cleanup ran (registered by n =', v, ')');
      if (v === 0) throw new Error('cleanup boom');
    };
  });

  // Control effect on a DIFFERENT signal, driven by its own button so it is
  // never in the same batch as the thrower. It keeps working throughout,
  // proving the app is healthy and the wedge is per-effect.
  createEffect(b, (v) => {
    console.log('control effect saw b =', v);
  });

  return (
    <div>
      <p>n = {n()}</p>
      <p>b = {b()}</p>
      <button onClick={() => setN(n() + 1)}>bump n (watch the console)</button>
      <button onClick={() => setB(b() + 1)}>
        bump b (control — always works)
[...]

Issue #2762, opened 2026-06-13
2.0.0-beta.14: An uncaught throwing effect abandons later queued siblings


yumemi-thomas:
### Describe the bug

If a user effect throws during a flush, later effects already queued in the same batch are permanently skipped. The thrown error escapes, which is expected enough, but the remaining queued callbacks are lost because the queue is detached before iteration.

One important wrinkle: this is order-dependent. If the throwing effect is queued before a sibling effect, the sibling never observes the update. If the sibling is queued first, it runs before the throw.

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-thqrgu5x?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

1. Open the StackBlitz repro and the browser DevTools console.
2. On mount, the console shows sibling effect saw: 0 (expected).
3. Click `set n = 1`.
4. The console shows an uncaught Error: boom from the first effect, and the sibling effect never logs 1.

### Expected behavior

A throwing effect may propagate its error, but it should not permanently discard unrelated queued effects from the same flush. After the error is caught, either the remaining queued work should already have run, or it should still be available to drain.

### Platform

- OS: macOs
- Browser: chrome
- Version: 2.0.0-beta.14

ryansolid:
Same resolution as #2761 (90238e79): an error escaping every boundary now halts the reactive system entirely, by design. Sibling effects behind the thrower intentionally don't run — after an unhandled error the flush's writes are partially applied, so running the rest of the queue would execute effects against state we can't vouch for. If sibling work must survive a failure, the failing effect needs an error boundary; errors caught by boundaries don't halt anything.

PR #2814, opened 2026-07-02
fix(signals): contain throwing effects and detach cleanups (closes #2761, #2762, #2813)

### LINKED ISSUE #2761: 2.0.0-beta.14: One throwing effect can kill all reactivity
### Describe the bug

If a `createEffect` callback writes a signal and then throws during the same flush, the scheduler gets stuck and no further updates run until you manually call flush(). A throw alone does not cause this particular stall.

https://stackblitz.com/edit/solidjs-templates-vmutpnzg?file=src%2FApp.tsx

1. Click “1. break the scheduler (throws once)”.
 The console shows an uncaught error (boom).

2 - Click “2. increment b” several times. Actual: b = … never updates again — the UI is frozen.

### Expected behavior

After the one-time error from step 1, step 2 should keep updating the display (b = 101, b = 102, …). Reactivity should recover automatically; only the throwing effect should fail, not the whole app.

### Screenshots or Videos

_No response_

### Platform

- OS: macOs
- Browser: Chrome
- Version: 2.0.0-beta-14

### Root cause

The stall requires **write + throw in the same flush**, not a throw alone:

1. `setA(1)` sets `scheduled = true` and queues a microtask to `flush()`.
2. During that flush, before effects run, `scheduled` is reset to `false` when the dirty queue is empty (`scheduler.ts:380`).
3. The effect calls `setB(100)` → `schedule()` sets `scheduled = true` again, but **does not queue a new microtask** because `globalQueue._running` is already `true` (`scheduler.ts:221–225`).
4. The effect then throws; the flush aborts with **`scheduled` still `true` and no pending microtask**.
5. Every later write hits `if (scheduled) return` in `schedule()` — reactivity is dead until something calls `flush()` manually.

A throw **without** an in-flush signal write does not leave `scheduled` stuck this way. The app can still update on the next user action.

### Suggested fix direction

Ensure the effect-running phase clears or re-arms `scheduled` on abort (e.g. `try/finally` around effect execution, or re-queue a microtask when a flush exits early with pending work).

yumemi-thomas:
## Summary

Fixes #2761, #2762 and #2813 — three ways a single throwing effect (or cleanup) can break an app on 2.0.0-beta.15.

The first two share a root cause: when an effect throws and nothing handles the error, the exception unwinds `GlobalQueue.flush` mid-queue. The queue array is already detached at that point, so sibling effects queued behind the thrower are lost (#2762). And if the effect wrote a signal earlier in the same flush, `scheduled` stays true with no microtask armed, so every later write short-circuits in `schedule()` and the app is dead until something calls `flush()` manually (#2761).

Rather than patching the individual unwind paths, this removes the mechanism: unhandled effect errors no longer throw through the scheduler at all. They go through a small funnel (`reportUncaughtError`) that collects them during a drain and rethrows the first one from `flush()` once the drain completes (extra ones are `console.error`'d so they don't disappear). Outside a drain it throws in place, so creation-time behavior is unchanged. I also split "work is pending" from "a drain microtask is armed" (new `drainArmed` flag), so even if a drain dies for some unforeseen reason, the next write can always arm a fresh one instead of trusting a flag that lied.

#2813 is a separate bug the scheduler stall was hiding: `runEffect` invokes the previous cleanup before clearing `node._cleanup` and before the `try`. A throwing cleanup is therefore never detached (every later run re-invokes the dead cleanup and the effect body never runs again), never reaches the `catch` (error boundaries can't see cleanup errors), and never reaches the `finally` (the dev strict-read guard leaks, and every untracked read in the app starts warning `STRICT_READ_UNTRACKED`). The fix is to detach the cleanup to a local before calling it, inside the `try`. Same reorder in `trackedEffect` and the error handler's cleanup callback.

The observable changes, in one list:

- sibling effects queued after a throwing effect now run (#2762)
- a signal write followed by a throw in the same flush no longer permanently stalls scheduling (#2761)
- unhandled effect errors surface from `flush()` after the drain completes, instead of aborting it mid-queue
- a throwing cleanup costs one round instead of permanently disabling its effect (#2813)
- error boundaries now catch cleanup errors (previously impossible — the throw bypassed `runEffect`'s catch)
- the dev-only cascade of `STRICT_READ_UNTRACKED` false positives after a cleanup throw is gone

Two commits, one per root cause, each green on its own. One existing test changed (`onSettled` reentrant-flush): it pinned the old abort behavior where the nested callback and a pending write were silently dropped; the guardrail error still throws, but queued work now drains.

<details>
<summary>Why that test's expectations changed (the only existing-behavior change in the suite)</summary>

The test's purpose — the reentrant-flush guardrail throws — is untouched, as are the two writes after the `flush()` call (the throw still aborts the user callback at that statement, so `values.at(-1)` is `5`, not `15`). What the old secondary assertions pinned was the #2762 unwind itself, observed from an `onSettled` callback: the nested `onSettled`, registered *before* the violation, was silently discarded (`log = ["outer"]`), and the effect stayed permanently stale against a write that had already committed — `count()` read `5` while the effect had seen `0` and would never be told otherwise (`values.at(-1) = 0`). That's an internally inconsistent reactive graph frozen into a test.

This change is forced by any fix to #2762, not by this particular design: even the most minimal per-callback containment in `runQueue` produces the new values. Keeping the old expectations would require special-casing the guardrail error to still unwind the drain — i.e. deliberately preserving one live instance of the bug.

</details>

Two judgment calls reviewers may want to weigh in on: when several effects throw in one drain, the first is thrown and the rest are `console.error`'d (an `AggregateError` or an `onUncaughtError` hook would be alternatives, but both change public API), and errors now surface at the end of the drain rather than mid-flush (no existing test cared, but it is observable).

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. All new tests were written first and confirmed failing on the unfixed code, and the bugs and fixes were both verified against the published beta.15 npm package, not just this repo.

- 10 new unit tests across `flush` / `strictRead` / `createTrackedEffect` / `createErrorBoundary`, each verified red on the pre-fix code before applying the fix. The strict-read one is telling: reverting the fix fails it plus 11 unrelated tests in the same file, because the leaked flag poisons everything that runs after it.
- `pnpm build`, then `turbo run test test-types --force` (uncached): all 19 tasks green — signals 782, solid-js 416, web 409. Commit 1 was also tested in isolation (778/778).
- Node scripts against the built dist using only microtask scheduling (no manual `flush()`), replicating the StackBlitz repros from the three issues: the error surfaces once, siblings survive, later writes keep flushing, the wedged effect recovers, and no `STRICT_READ_UNTRACKED` warnings appear. The same scripts run against the published `@solidjs/signals@2.0.0-beta.15` package reproduce all three failures — the fixes target what is actually shipped.

### DIFF
--- a/.changeset/fix-effect-cleanup-detach.md
+++ b/.changeset/fix-effect-cleanup-detach.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Fix a throwing effect cleanup permanently breaking its effect (#2813). The previous cleanup is detached before it runs, so a cleanup that throws is never re-invoked on later runs and the effect recovers on the next update. Cleanup errors now take the standard effect error path: error boundaries can catch them, and the dev-mode strict-read guard is restored instead of leaking `STRICT_READ_UNTRACKED` false positives app-wide.
--- a/.changeset/fix-scheduler-throwing-effect-stall.md
+++ b/.changeset/fix-scheduler-throwing-effect-stall.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Fix one throwing effect killing all reactivity (#2761, #2762). An unhandled effect error no longer unwinds the flush: sibling effects queued behind it still run, and a write during the failed flush can no longer permanently stall scheduling. Errors are collected and rethrown from `flush()` once the drain completes; when several effects throw in one drain, the first is rethrown and the rest are reported via `console.error`.
--- a/packages/solid-signals/src/core/effect.ts
+++ b/packages/solid-signals/src/core/effect.ts
@@ -90,8 +90,9 @@ function notifyEffectStatus(this: Effect<any>, status?: number, error?: any): vo
       try {
         return this._errorFn
           ? this._errorFn(err, () => {
-              this._cleanup?.();
+              const prevCleanup = this._cleanup;
               this._cleanup = undefined;
+              prevCleanup?.();
             })
           : console.error(err);
       } catch (e) {
@@ -126,9 +127,11 @@ function runEffect(node: Effect<any>): void {
   if (__DEV__) {
     prevStrictRead = setStrictRead("an effect callback");
   }
-  node._cleanup?.();
+  // Detach first so a throwing cleanup cannot wedge future runs.
+  const prevCleanup = node._cleanup;
   node._cleanup = undefined;
   try {
+    prevCleanup?.();
     const nextCleanup = node._effectFn(node._value, node._prevValue);
     if (__DEV__ && nextCleanup !== undefined && typeof nextCleanup !== "function") {
       throw new Error(
@@ -179,8 +182,10 @@ export function trackedEffect(fn: () => void | (() => void), options?: NodeOptio
 
   const node = computed<void>(
     () => {
-      node._cleanup?.();
+      // Detach first so a throwing cleanup cannot wedge future runs.
+      const prevCleanup = node._cleanup;
       node._cleanup = undefined;
+      prevCleanup?.();
       const cleanup = staleValues(fn);
       if (__DEV__ && cleanup !== undefined && typeof cleanup !== "function") {
         throw new Error(
--- a/packages/solid-signals/src/core/effect.ts
+++ b/packages/solid-signals/src/core/effect.ts
@@ -22,6 +22,7 @@ import { cleanup } from "./owner.js";
 import {
   _hitUnhandledAsync,
   GlobalQueue,
+  reportUncaughtError,
   resetUnhandledAsync,
   setTrackedQueueCallback
 } from "./scheduler.js";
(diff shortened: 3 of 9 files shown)


ryansolid:
Thanks for this — the diagnosis of all three bugs is excellent, and the #2813 portion (detaching cleanups before invocation, inside the try) is essentially what we landed in 90238e79.

We're declining the containment half, though, on philosophy rather than execution. Collecting unhandled errors and draining the remaining queue keeps the app running after its state is already undefined — the flush that threw had writes half-applied, so the drained siblings run against state no one can reason about, and the judgment calls you flagged (first-error-wins, end-of-drain surfacing) are symptoms of that. We went the other way: an error that escapes every boundary now permanently halts the scheduler, loudly. Unhandled errors are a crash, not a condition to route around; recovery belongs to error boundaries, which do contain errors without halting anything.

Appreciate the rigor here — red-first tests and verifying against the published package is exactly how we'd want it done.

PR #2819, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This PR adds a CVE Lite dependency audit workflow to help catch vulnerable dependencies before they land in main.

CVE Lite CLI is an OWASP Lab Project that scans lockfiles locally without installing packages. It reads the pnpm lockfile directly, queries the OSV vulnerability database, and classifies findings as direct or transitive so you know exactly what you control. A scan of the current main branch found 34 findings total - 3 critical, 16 high, 14 medium, and 1 low.

The workflow runs on every push to main, every pull request targeting main, and on a weekly schedule every Monday morning. When vulnerabilities at high severity or above are found, the job fails so they do not go unnoticed in CI. Results are also exported as a SARIF file and uploaded to GitHub Code Scanning, which surfaces findings directly on the Security tab with file and line context.

All Actions are pinned to immutable commit SHAs rather than mutable version tags, so the supply chain for the workflow itself is locked down.

A companion PR with direct dependency upgrades is coming separately once this is reviewed.

Full documentation and the OWASP project page are at https://owasp.org/cve-lite-cli

### DIFF
--- a/.github/workflows/cve-lite.yml
+++ b/.github/workflows/cve-lite.yml
@@ -1,0 +1,36 @@
+name: CVE Lite dependency audit
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'
+
+permissions:
+  contents: read
+  security-events: write
+
+jobs:
+  scan:
+    name: Scan dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Scan for vulnerabilities
+        uses: OWASP/cve-lite-cli@2eed959b8641042472d2810444393b88d5454e62 # v1
+        with:
+          fail-on: high
+          sarif: 'true'
+
+      - name: Upload SARIF to Code Scanning
+        if: always() && hashFiles('*.sarif') != ''
+        uses: github/codeql-action/upload-sarif@411bbbe57033eedfc1a82d68c01345aa96c737d7 # v4
+        with:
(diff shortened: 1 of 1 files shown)


sonukapoor:
One quick note: CVE Lite CLI v1.26.0 shipped a scheduled fix mode that can automate what the companion fix PR does manually. Adding `fix: true` and `create-pr: true` to this workflow would have it open a single batched PR with the next round of upgrades automatically - OSV-confirmed versions and advisory IDs included. Here's an example of what that looks like: https://github.com/sonukapoor/cve-lite-test/pull/2

Want me to update this PR to include that? It does require enabling "Allow GitHub Actions to create and approve pull requests" in your repo settings first.

sonukapoor:
@ryansolid With 2.0 RC out, wanted to gently check back in on this one. It wires a lightweight dependency audit into CI so vulnerable deps get surfaced on PRs before a release, and it's config-driven so you can set the severity threshold or turn off the fail gate to fit your flow. I just bumped the pinned action to our latest release so it's current and ready to go. Totally understand if it's not a priority with everything else on your plate right now.

ryansolid:
Landed on main in c257b9ee without `fail-on: high`, so leftover transitive findings do not block PRs. SARIF upload is in place. Thanks for the workflow.

sonukapoor:
Thank you so much @ryansolid

sonukapoor:
Hey Ryan - just wanted to let you know that the CVE Lite CLI GitHub Action had a bug where not setting `fail-on` still caused the step to exit 1 on critical findings (the CLI's internal default was applying even though the action documented `""` as "no threshold"). That's why your workflow has been failing on every run since day one.

Fixed in v1.31.0 (#1048). Since you're pinned to `fd481fff` (v1.29.0), you'll need to bump the SHA to pick it up. The floating `@v1` tag is already updated if you want to switch to that instead.

Issue #2801, opened 2026-06-26
2.0.0-beta.15 Many hydration bugs


dangkyokhoang:
### Describe the bug

Bug 1: Plain simple hydration issue. Try clicking Refreshing several times.
Bug 2: JSX syntax `{data().value && <h4>{data().value}</h4>}` causes bug in `createMemo` order. `keyFn` memo returns an object instead of the path string. See console logs.
Bug 3: `Loading > Errored > Nested` async memo causes the page to load forever. `SSR stream transform exceeded maximum lifetime (120000ms), forcing cleanup`
Bug 4: `createEffect` on async memo causes the page to load forever.
Bug 5: Reusing promise stored in global cache doesn't work.
Bug 6: Rendering `{data()}` where data() is an object causes infinite loop. Does not loop but logs errors in Vite Nitro SSR reproduction.

### Your Example Website or App

Tanstack SSR
https://stackblitz.com/edit/solidjs-templates-jskxvhkj?file=src%2Froutes%2Fhydration-bug-1.tsx

Vite Nitro SSR
https://stackblitz.com/edit/nitrojs-nitro-6gibxz9z?file=src%2Froutes%2Findex.tsx

### Steps to Reproduce the Bug or Issue

Go the the link above.

### Expected behavior

No bugs.

### Platform

Doesn't matter.

brenelz:
It would be helpful to see if you could replicate these outside of Tanstack Start if possible. That way we know whether the issue is in solid core or tanstack

brenelz:
# SolidJS 2.0.0-beta.15 — Hydration Bug Findings

Reproduced and traced all six bugs on `2.0.0-beta.15` (SSR + TanStack Start).
Method: reproduce locally, drive SSR with `curl` and hydration with Playwright/Chrome,
then trace into the framework source.

## Summary

| # | Verdict | Where | Cause (simple) | Fix / workaround |
|---|---------|-------|----------------|------------------|
| 1 | Real bug | `@solidjs/web` (hydration) | Tracked node drifts from the real DOM node | needs runtime fix |
| 2 | Real bug | `babel-plugin-jsx-dom-expressions` | Server and client number hydration keys differently | use `<Show>` instead of `&&` |
| 3 | Real bug | `solid-js` SSR | Error boundary re-throws "pending" and loops forever | runtime patch |
| 4 | Real bug | `solid-js` SSR | Effect re-throws "pending" and loops forever | runtime patch / `ssr:false` |
[...]

PR #2827, opened 2026-07-04
fix(server): reset sync memo child state on re-pull so hydration keys stay aligned

brenelz:
Fixes bug 2 of #2801 (`{data().value && <h4>{data().value}</h4>}` before a `<For>` breaking hydration).

## Root cause

Hydration keys are owner-path child-id slots. On the client, every recompute resets the owner's child state (dispose + `_childCount = 0`). The server's lean sync memo — what compiler-emitted `_$memo()` expression wrappers use — re-runs its compute on every pull after a `NotReadyError`, but did so **without resetting child state**, so each failed pull permanently leaked the child-id slots it consumed.

The `&&` shape is uniquely exposed because it compiles to an inner condition memo created *inside* the outer compute body. With the condition reading a pending async memo in a streamed `<Loading>`, the server needed three pulls (eager create → pending, discovery read → pending, resolved → success), leaking two slots. The successful pull emitted `<h4 _hk=10003>` while the client's single successful compute claimed `10001` — nobody matched, so the h4 went unclaimed (duplicated in prod; dev warns "unclaimed server-rendered node"). Sync conditions pull once and are unaffected, and `<Show>` avoids it via its own already-aligned owner scaffolding — matching the symptoms reported in the issue.

## Fix

`createSyncMemo.pull()` now disposes children and resets `_childCount` before re-running the compute, mirroring the client's owner reset on recompute, so every pull emits the same ids. First pulls skip the dispose branch (`_firstChild`/`_disposal` are null), so the hot path pays one null check. This follows the same server-matches-client slot-alignment pattern as 4cab2480 and 1122d741, and also covers the equivalent latent leak when a `mapArray`/`repeat` item throws `NotReadyError` mid-loop.

## Tests

- Server streaming test asserting the emitted keys for the async `&&`-before-`<For>` shape (`<h4 _hk=10001>`, previously `10003`).
- jsdom round-trip test hydrating captured stream output for both the async (streamed fragment) and sync condition shapes, asserting full claim with no unclaimed-node/mismatch warnings and in-place claiming of server nodes.

Suites: solid-web server 116/116, solid-web client 299/299, solid-web hydration 23/23*, packages/solid 417/417.

*Known follow-up, not addressed here: the non-sync server `createMemo.update()` retry path has the same theoretical leak but needs a children-only reset (a naive `disposeOwner` would fire the memo's own cleanup and kill the retry); no failing case observed. Bug 1 of #2801 (ghost node after `refresh()`) is a separate dom-expressions `insert()` issue and will be a PR there.

### DIFF
--- a/.changeset/fix-ssr-sync-memo-child-id-leak.md
+++ b/.changeset/fix-ssr-sync-memo-child-id-leak.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix hydration key drift when a compiler-emitted expression memo reads a pending async source during streamed SSR (#2801). The server's lean sync memo re-runs its compute on every pull after a `NotReadyError`, but did so without resetting the owner's child state — each failed pull leaked the child-id slots it consumed (e.g. the inner condition memo of `{data().value && <h4>...</h4>}`), so hydration keys produced by the eventual successful pull drifted ahead of the client's single successful compute and the affected nodes went unclaimed (duplicated in prod, "unclaimed server-rendered node" warning in dev). The sync memo now disposes children and resets `_childCount` before each re-pull, mirroring how the client resets an owner on recompute, so every pull emits the same ids.
--- a/packages/solid-web/test/hydration/async-condition-keys.spec.tsx
+++ b/packages/solid-web/test/hydration/async-condition-keys.spec.tsx
@@ -1,0 +1,141 @@
+/**
+ * @jsxImportSource @solidjs/web
+ * @vitest-environment jsdom
+ *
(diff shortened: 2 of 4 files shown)


ryansolid:
Landed on `next` as 14589077 with you as commit author — the runtime fix, server stream test, and changeset are verbatim. The jsdom spec was adapted into the compiled hydration parity harness (which regenerates server artifacts from the live compiler) instead of hand-captured stream HTML: the async-cond-before-for scenario flipped from known-failure to passing, plus a sync control scenario. Thanks for the thorough root-cause work — the within-hole slot-leak diagnosis was exactly right.

PR #2834, opened 2026-07-06
fix(web): remove Portal's end marker on cleanup

### LINKED ISSUE #2833: 2.0.0-beta.15: `Portal` leaks one empty text node in the mount target per instance
### Describe the bug

Every `<Portal>` instance leaves one empty text node behind in its mount target after cleanup: the disposer removes the nodes in `[startMarker, endMarker)` but never removes `endMarker` itself.

1. Mount + dispose a Portal once → 1 stray empty text node stays in the mount element.
2. Toggle a Portal inside `<Show>` N times (the classic modal pattern, mounting into `document.body`) → N empty text nodes accumulate, unbounded.

Related to (but distinct from) #2757, which was about the Portal's *content* accumulating on reactive swaps and is fixed in beta.15 — this is *marker* accumulation on unmount. It is invisible to `innerHTML`-based checks (empty text nodes don't serialize), but shows up in `childNodes.length`, breaks `:empty` CSS selectors on the mount target, and is a real unbounded DOM leak for long-lived apps that open/close modals.

Regression from 1.x (verified on 1.9.14): the mount target returns to 0 child nodes after every close and after final disposal.

https://stackblitz.com/edit/solidjs-templates-kwwlzkir?file=src%2FApp.tsx

```tsx
import { createSignal, Show } from "solid-js";
import { Portal } from "@solidjs/web";

const target =
  document.getElementById("portal-target") ??
  document.body.appendChild(document.createElement("div"));
target.id = "portal-target";
target.replaceChildren();

function dump() {
  return [...target.childNodes]
    .map(node => `${node.nodeName}(${JSON.stringify(node.textContent)})`)
    .join(", ");
}

  const [open, setOpen] = createSignal(false);
  const [count, setCount] = createSignal(target.childNodes.length);
  const [nodes, setNodes] = createSignal(dump() || "(empty)");

  function openThenClose() {
    setOpen(true);
    // open and close in separate ticks so each phase actually renders,
    // then sample the mount target after the close has been applied
    setTimeout(() => {
      setOpen(false);
      setTimeout(() => {
        setCount(target.childNodes.length);
        setNodes(dump() || "(empty)");
      }, 0);
    }, 0);
  }

  return (
    <div style={{ "font-family": "system-ui", padding: "16px" }}>
      <button onClick={openThenClose}>open then close Portal</button>
      <p>
        target.childNodes.length: <b>{count()}</b>
      </p>
      <pre>{nodes()}</pre>

      <Show when={open()}>
        <Portal mount={target}>
          <div>modal content</div>
        </Portal>
      </Show>
    </div>
  );
}
```

1. Click **open then close Portal** repeatedly.
2. Watch the reported `childNodes` count of the mount target.

Actual — each open/close cycle permanently adds one empty text node:

```text
after 1 click: target.childNodes.length === 1
after 2 clicks: target.childNodes.length === 2
after 3 clicks: target.childNodes.length === 3
```

The displayed dump shows the leftovers are empty text nodes:

```text
#text(""), #text(""), #text("")
```

### Expected behavior

Disposing a Portal removes everything it put into the mount target, including its own markers — the count must return to 0 after every close and stay flat:

```text
modal closed — mount childNodes: 0   (after every cycle)
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.15 (verified at `next` @ `bad66625`)

### Additional context

Root cause in `packages/solid-web/src/index.ts:237-245` (Portal's render-effect cleanup): the loop removes nodes strictly *before* `endMarker` and then stops, so `endMarker` — appended to the mount by this same effect at `index.ts:226` (`m.appendChild(endMarker)`) — survives every cleanup:

```ts
return () => {
  dispose!();
  let c: Node | null = startMarker;
  while (c && c !== endMarker) {   // ← half-open range: endMarker itself is never removed
    const n: Node | null = c.nextSibling;
    m.removeChild(c);
    c = n;
  }
};
```

(Switching `props.mount` does *not* leak into the previous mount — the next effect run's `m.appendChild(endMarker)` moves the surviving marker into the new mount, so the stranding only materializes on final disposal, i.e. every time the Portal unmounts.)

Suggested fix direction: make the removal inclusive of the marker the effect itself appended — e.g. loop `while (c) { … }` over `[startMarker, endMarker]` ending after removing `endMarker`, or simply `m.removeChild(endMarker)` after the loop.

**Regression from 1.x.** Verified against solid-js **1.9.14**: toggling `<Show><Portal mount={target}>...` on and off five times leaves the mount target completely empty (`target.childNodes.length === 0`). In 2.0 each mount/unmount cycle strands one empty text node in the target.

yumemi-thomas:
Closes #2833

## Summary

Every `Portal` unmount strands one empty text node in its mount target. Toggling a Portal inside `<Show>` — the standard modal open/close pattern, typically mounting into `document.body` — accumulates one node per cycle, unbounded. The leak is invisible to `innerHTML`-based checks (empty text nodes don't serialize) but shows up in `childNodes.length` and breaks `:empty` CSS selectors on the mount target. Regression from 1.x, where the mount returns to 0 child nodes after every close.

Repro: https://stackblitz.com/edit/solidjs-templates-kwwlzkir?file=src%2FApp.tsx

## Root cause

Portal's render-effect cleanup removes the half-open range `[startMarker, endMarker)` and stops — but `endMarker` was appended to the mount by that same effect run (`m.appendChild(endMarker)`) and nothing else ever removes it:

```ts
let c: Node | null = startMarker;
while (c && c !== endMarker) {   // ← endMarker itself survives every cleanup
  const n: Node | null = c.nextSibling;
  m.removeChild(c);
  c = n;
}
```

Switching `props.mount` does *not* leak into the previous mount — the next run's `appendChild(endMarker)` moves the surviving marker into the new mount — so the stranding materializes exactly once per unmount.

## Fix

Make the removal range inclusive: the loop now removes through `endMarker` and stops there. Four lines, no behavior change beyond the mount target being left truly empty.

## Solid 1.x note

Verified on `solid-js@1.9.14` with the identical toggle sequence: the mount target holds 0 child nodes after every close and after final disposal. Clean regression, no caveats.

## Tests

Three red-first tests in `test/portal.spec.tsx` ("Testing Portal marker cleanup"), all asserting on `childNodes.length` since `innerHTML` checks stay green even with stranded markers — which is exactly why the existing disposal tests never caught this:

- disposing a Portal leaves the mount at 0 child nodes
- 5× `<Show>` toggle cycles accumulate nothing
- switching mounts leaves no markers behind in either mount

All three confirmed failing against the unfixed build. Full solid-web client suite, the separately-configured hydration suite, and the server suite all pass at their existing baselines.

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. All new tests were written first and confirmed failing on the unfixed code, and the leak and the fix were verified in both jsdom and the [StackBlitz repro app](https://stackblitz.com/edit/solidjs-templates-kt1gg3uf?file=src%2FApp.tsx).

### DIFF
--- a/.changeset/portal-marker-cleanup.md
+++ b/.changeset/portal-marker-cleanup.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/web": patch
+---
+
+Fix `Portal` stranding one empty text node in its mount target per unmount: the cleanup removed the nodes in `[startMarker, endMarker)` but never `endMarker` itself, which the same effect run had appended. Toggling a Portal (the modal open/close pattern) accumulated one node per cycle, unbounded — invisible to `innerHTML` checks but breaking `:empty` selectors and `childNodes` counts on the mount target. The removal range is now inclusive of `endMarker`.
--- a/packages/solid-web/src/index.ts
+++ b/packages/solid-web/src/index.ts
@@ -236,10 +236,13 @@ export function Portal<T extends boolean = false, S extends boolean = false>(pro
       );
       return () => {
         dispose!();
+        // remove [startMarker, endMarker] inclusive — endMarker was appended
+        // by this same effect run and must not survive the cleanup
         let c: Node | null = startMarker;
-        while (c && c !== endMarker) {
+        while (c) {
           const n: Node | null = c.nextSibling;
           m.removeChild(c);
+          if (c === endMarker) break;
           c = n;
         }
       };
--- a/packages/solid-web/test/portal.spec.tsx
+++ b/packages/solid-web/test/portal.spec.tsx
@@ -457,6 +457,80 @@ describe("Testing Portal insert effect ownership (#2758)", () => {
   });
 });
 
+describe("Testing Portal marker cleanup", () => {
+  // empty text markers don't serialize, so these assert on childNodes.length —
+  // innerHTML checks stay green even when markers are stranded
+  test("disposing a Portal removes its markers from the mount", () => {
+    const div = document.createElement("div");
+    const mount = document.createElement("div");
+
+    const disposer = render(
+      () => (
+        <Portal mount={mount}>
+          <span>content</span>
+        </Portal>
+      ),
+      div
+    );
+    expect(mount.querySelector("span")).toBeTruthy();
+
+    disposer();
+    expect(mount.childNodes.length).toBe(0);
+  });
+
+  test("toggling a Portal does not accumulate markers in the mount", () => {
+    const div = document.createElement("div");
+    const mount = document.createElement("div");
+    const [open, setOpen] = createSignal(false);
+
+    const disposer = render(
+      () => (
+        <Show when={open()}>
+          <Portal mount={mount}>
+            <span>modal</span>
+          </Portal>
+        </Show>
+      ),
+      div
+    );
+
+    for (let i = 0; i < 5; i++) {
+      setOpen(true);
+      flush();
+      expect(mount.querySelector("span")).toBeTruthy();
+      setOpen(false);
+      flush();
+      expect(mount.childNodes.length).toBe(0);
+    }
+    disposer();
+  });
+
+  test("switching mounts leaves no markers behind in either mount", () => {
+    const div = document.createElement("div");
+    const mountA = document.createElement("div");
+    const mountB = document.createElement("div");
+    const [mount, setMount] = createSignal(mountA);
+
+    const disposer = render(
+      () => (
+        <Portal mount={mount()}>
+          <span>content</span>
+        </Portal>
+      ),
+      div
+    );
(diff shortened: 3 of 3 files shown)


ryansolid:
Merged — textbook fix. The half-open removal range stranding `endMarker` is exactly right, and asserting on `childNodes.length` instead of `innerHTML` is the reason this ever gets caught; the existing disposal tests were green through the leak. Verified the toggle repro and ran the full monorepo suite with the change in place.

Cherry-picked onto `next` with your authorship preserved (the branch is maintained locally ahead of the release push, so GitHub won't mark this as merged automatically — it will appear on `next` shortly). Ships in the next beta. Thanks!

PR #2847, opened 2026-07-07
fix(signals): settle async-generator actions on uncaught errors instead of freezing

### LINKED ISSUE #2841: 2.0.0-beta.15: uncaught error in an async-generator `action()` freezes the JS thread
## Describe the bug

When an `action()` is given an **async generator** (`async function*`) and an uncaught error occurs inside it, the returned promise never settles and the whole JS thread freezes in an infinite microtask loop (frozen tab in the browser, hung process in Node). The same code written as a sync `function*` rejects correctly.

Since the transition's iterator is never removed from `_actions`, the transition also never completes.

## Reproduction

```ts
import { action } from "@solidjs/signals";

const act = action(async function* () {
  throw new Error("boom"); // any uncaught throw: failed fetch, TypeError, etc.
});

act(); // never settles; event loop is starved — a setTimeout after this never fires
```

Real-world shape:

```ts
const save = action(async function* (text: string) {
  const res = await fetch("/api/todos", { method: "POST", body: text });
  if (!res.ok) throw new Error("save failed"); // ← freezes the tab
  return yield res.json();
});
```

brenelz:
## Summary

Fixes #2841 — an uncaught error inside an async-generator `action()` froze the JS thread: the returned promise never settled, the event loop was starved by an infinite microtask loop, and the transition never completed because the iterator was never removed from `_actions`.

## Root cause

When an async generator's body throws, `it.next()` returns a **rejected promise** with the generator already in the completed state. The action runner responded to that rejection by calling `it.throw(e)` — but `throw()` on a completed async generator just returns another rejected promise, whose rejection handler called `it.throw(e)` again, forever. Each iteration is a microtask, so the loop starves the event loop (frozen tab / hung process). Sync generators were unaffected because their `it.next()` throws synchronously, which was already handled.

## Fix

A rejected iterator-result promise is terminal: async generators run user `try`/`catch`/`finally` internally before the rejection ever reaches the runner. So the rejection handler now settles the action via `done(undefined, e)` — mirroring how the fulfilled path handles `done: true`. The returned promise rejects, the iterator is removed from the transition's `_actions`, and the transition can complete.

`try/catch` around `yield`/`await` inside async generators still works (handled by the generator machinery itself), and the sync-generator `it.throw` path for yielded promise rejections is unchanged.

## Tests

Four regression tests in `action.test.ts`:
- immediate throw in an async generator rejects and does not starve the event loop (the issue's repro — this test hangs without the fix)
- throw after an `await`/`yield` rejects
- `try/catch` inside an async generator around an awaited rejection recovers and continues
- a failing async-generator action is removed from the transition so a concurrent action still commits

Full solid-signals suite passes (818 tests, 49 files).

### DIFF
--- a/packages/solid-signals/src/core/action.ts
+++ b/packages/solid-signals/src/core/action.ts
@@ -76,8 +76,10 @@ export function action<Args extends any[], Y, R>(
         } catch (e) {
           return done(undefined, e);
         }
-        if (isThenable(r))
-          return void r.then(run, e => restoreTransition(ctx, () => step(e, true)));
+        // A rejected iterator result (async generators) means the error already
+        // escaped the generator body — it is completed, and throwing back in
+        // would just reject again forever. Settle instead.
+        if (isThenable(r)) return void r.then(run, e => done(undefined, e));
         run(r);
       };
 
--- a/packages/solid-signals/tests/action.test.ts
+++ b/packages/solid-signals/tests/action.test.ts
@@ -886,6 +886,73 @@ describe("action", () => {
       await expect(myAction()).rejects.toThrow("error after yield");
     });
 
+    it("should reject the promise when an async generator throws immediately", async () => {
+      const myAction = action(async function* () {
+        throw new Error("boom");
+      });
+
+      await expect(myAction()).rejects.toThrow("boom");
+
+      // The event loop must not be starved — a macrotask still runs
+      await new Promise<void>(r => setTimeout(r, 0));
+    });
+
+    it("should reject the promise when an async generator throws after an await", async () => {
+      const myAction = action(async function* () {
+        await Promise.resolve();
+        yield undefined;
+        throw new Error("async error after yield");
+      });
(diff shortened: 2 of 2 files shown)


ryansolid:
Diagnosis and fix are both exactly right — a rejected iterator-result promise from an async generator is terminal (the body's error already escaped, user `try`/`catch` already ran inside the generator machinery), so throwing back in could only ever produce another rejected promise, and each round trip being a microtask is what starved the loop. Settling via `done` mirrors the synchronous catch path and correctly removes the iterator from the transition's `_actions`.

We verified the freeze on the current tree (the repro genuinely hangs a vitest worker), confirmed no conflict with the unhandled-error-halts philosophy (this is the action promise contract, which sync generators already honored — not an effect-queue error), and landed your diff as-is with a changeset, credited as co-author: `0e811993`. All four regression tests included. Ships in the next beta — thanks!

PR #2848, opened 2026-07-07
fix(store): snapshot() and deep() preserve array length when trailing indices were deleted (closes #2846)

### LINKED ISSUE #2846: 2.0.0-beta.15: `snapshot()` and `deep()` shrink the array length when a trailing index was deleted
### Describe the bug

`delete arr[i]` on an array store leaves a hole without changing `length` — plain JavaScript semantics, and the store proxy agrees (`store.length` stays `3`, `store[2]` reads `undefined`). But `snapshot()` of that store returns an array whose `length` has silently shrunk: the trailing hole is dropped entirely. Any consumer that serializes, structured-clones, or diffs the snapshot sees an array that disagrees with the store it came from.

The affected surface is wider than one call (probed each variant individually):

- **`deep()` is equally affected** — it shares `snapshotImpl`, so the tracked deep-copy loses the length the same way.
- **Any nesting depth** — `snapshot(store).list` of a store holding the array as a property shrinks identically (the recursion reaches the same branch).
- **Multiple trailing holes** shrink by the full run (`["a","b",<hole>,<hole>]` → length 2).

And the boundary of the bug, for contrast: a **middle** hole is preserved correctly (`["a",null,"c"]`, length 3), writing any index past the hole restores the length, `length` extension (`arr.length = 5`) copies correctly — and every **proxy-side** path agrees with plain JS (`store.length === 3`, `JSON.stringify(store)`, `[...store]`, `store.map`, `Array.from` all yield `["a","b",null]`). The store disagrees only with its own materialized copies.

This is a gap in `snapshotImpl` only: the sibling code path `unwrapStoreValue` (used by the set trap) handles the same case correctly by restoring the length after skipping deleted slots. Not covered by the truncation fixes for #2768 (5894f2a9 / bc92d002), which are about `length` writes through the set trap, not about `snapshot`'s read path.

https://stackblitz.com/edit/solidjs-templates-dvf3msqv?file=src%2FApp.tsx

```tsx
import { createSignal, createStore, deep, flush, Show, snapshot } from "solid-js";

type Verdict = { ok: boolean; actual: string };

  const [store, setStore] = createStore(["a", "b", "c"]);
  const [verdict, setVerdict] = createSignal<Verdict>();

  function deleteLastAndCopy() {
    setStore(draft => {
      delete draft[2];
    });
    flush();

    const snap = snapshot(store);
    const deepCopy = deep(store);
    setVerdict({
      ok: snap.length === store.length && deepCopy.length === store.length,
      actual: [
        `store (proxy):  length ${store.length}  ${JSON.stringify([...store])}`,
        `snapshot():     length ${snap.length}  ${JSON.stringify(snap)}`,
        `deep():         length ${deepCopy.length}  ${JSON.stringify(deepCopy)}`
      ].join("\n")
    });
  }

  return (
      <h2>snapshot() and deep() lose trailing holes</h2>
      <p>
        items: {JSON.stringify([...store])} (length: {store.length})
      </p>
      <button onClick={deleteLastAndCopy}>delete last item and copy</button>
      <Show when={verdict()}>
        {v => (
          <section
            style={{
              padding: "12px",
              "margin-top": "12px",
              color: "white",
              background: v().ok ? "#137333" : "#c5221f"
            }}
          >
            <b>{v().ok ? "PASS - bug is fixed" : "FAIL - bug reproduced"}</b>
            <pre>{v().actual}</pre>
          </section>
        )}
      </Show>
    </main>
  );
}
```

1. Click **delete last item and copy**. The rendered store shows the deletion behaving like plain JS — the hole appears (rendered as `null`) and the length stays `3`:

```text
items: ["a","b",null] (length: 3)
```

2. Actual — the verdict banner shows both copy APIs disagreeing with the store they were taken from, while the proxy stays correct:

```text
FAIL - bug reproduced
store (proxy):  length 3  ["a","b",null]
snapshot():     length 2  ["a","b"]
deep():         length 2  ["a","b"]
```

### Expected behavior

Both copies preserve the store's length, keeping the trailing hole like plain JS (`["a", "b", <hole>].length === 3`):

```text
PASS - bug is fixed
store (proxy):  length 3  ["a","b",null]
snapshot():     length 3  ["a","b",null]
deep():         length 3  ["a","b",null]
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.15 (all variants re-verified at `next` @ `e2ebc11c`)

### Additional context

Root cause: `snapshotImpl`'s array branch in `packages/solid-signals/src/store/utils.ts:48-58` skips `$DELETED` slots but never restores the result's length afterwards:

```ts
if (isArray) {
  const len = override?.length ?? item.length;
  for (let i = 0; i < len; i++) {
    v = override && i in override ? override[i] : item[i];
    if (v === $DELETED) continue;      // ← skipped, so result never reaches index i
    ...
    result[i] = unwrapped;
  }
}
```

When the deleted index is trailing, the fresh `result` array simply ends before it, so `result.length` comes out shorter than the store's. `unwrapStoreValue` in `store.ts:164-184` handles the identical situation correctly:

```ts
if (isArray) result.length = override.length ?? source.length;   // store.ts:183
```

Suggested fix direction: after the array loop in `snapshotImpl`, apply the same restoration (`if (result) result.length = len;`), so deleted slots stay holes instead of truncating the copy. One change covers every affected variant above, since `snapshot()` and `deep()` share the implementation and nesting recurses through it.

**Regression from 1.x.** Verified against solid-js **1.9.14** using the 1.x equivalents (`setState(2, undefined)` deletes the property in 1.x; `unwrap` is 1.x's snapshot): after deleting the trailing index of `[1, 2, 3]`, both `state.length` and `unwrap(state).length` remain `3` (`2 in state` is `false` — a hole, matching plain-JS `delete` semantics). 2.0's `snapshot()` shrinks the length instead.
Working solid 1.x example: https://playground.solidjs.com/anonymous/32bb125d-9292-4723-bd28-4e992662ee71

yumemi-thomas:
Closes #2846

## Summary

`delete arr[i]` on an array store leaves a hole without changing `length` — plain JavaScript semantics, and every proxy-side path agrees (`store.length`, `JSON.stringify(store)`, `[...store]`, `store.map`, `Array.from` all yield `["a","b",null]`). But both copy APIs return an array whose length has silently shrunk: `snapshot(store)` and `deep(store)` drop trailing holes entirely, at any nesting depth, and multi-hole runs shrink by the full run. The store disagrees only with its own materialized copies — any consumer that serializes, diffs, or `reconcile()`s the copy sees silently lost slots.

https://stackblitz.com/edit/solidjs-templates-dvf3msqv?file=src%2FApp.tsx

## Root cause

The array branch of `snapshotImpl` (`packages/solid-signals/src/store/utils.ts`) skips `$DELETED` slots but never restores the result's length afterwards, so a copy ends at the last written index:

```ts
for (let i = 0; i < len; i++) {
  v = override && i in override ? override[i] : item[i];
  if (v === $DELETED) continue;   // ← skipped; result never reaches index i
  ...
  result[i] = unwrapped;
}
```

The sibling path `unwrapStoreValue` (used by the set trap) handles the identical situation correctly by restoring the length after its loop (`store.ts`: `result.length = override.length ?? source.length`). This is the one divergent site.

## Fix

One line, mirroring `unwrapStoreValue`: after the array loop, `if (result) result.length = len;`. Deleted slots stay true holes (`i in copy === false`, serialized as `null`), matching plain-JS `delete` and the proxy's own view. By construction it can't overreach: the loop only writes indices `< len` so the restoration never truncates written data, the `if (result)` guard leaves the identity/no-copy path untouched, and `len` already honors `override.length` so length-truncation writes keep working. One change covers every affected variant, since `snapshot()` and `deep()` share the implementation and nesting recurses through it.

## Tests

Four tests added to the existing `snapshot` describe in `tests/store/utilities.test.ts`, three confirmed failing red-first on the unfixed source:

- trailing hole: copy keeps `length 3`, the hole stays a *hole* (`2 in copy === false`), serializes as `["a","b",null]`
- `deep()` with the same shape
- nested array + multi-hole run (`{ list: ["a","b",<hole>,<hole>] }` keeps `length 4`)
- controls (passing before and after): middle holes preserved, `length`-truncation writes still truncate

Edge shapes probed additionally: deleting the only element (`[null]`, length 1, index still a hole), `length`-extension with a delete in the extended region, mixed middle+trailing holes. Full solid-signals suite: the diff versus pristine is exactly the intended flips; solid and @solidjs/web rebuilt and re-run at their baselines; test files pass the tests-inclusive typecheck.

## Solid 1.x note

Regression from 1.x, verified on 1.9.14 with the 1.x equivalents (`setState(2, undefined)` to delete; `unwrap` as the copy): both `state.length` and `unwrap(state).length` remain `3` after deleting the trailing index — a hole, matching plain-JS `delete` semantics. 2.0's copies shrink instead.

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. The tests were written first and confirmed failing on the unfixed code, and the affected/unaffected variant boundary was established by probing each shape individually.

### DIFF
--- a/.changeset/snapshot-trailing-holes.md
+++ b/.changeset/snapshot-trailing-holes.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Fix `snapshot()` and `deep()` silently shrinking an array's length when trailing indices were deleted. `delete arr[i]` on a store leaves a hole without changing `length` — plain JS semantics, and every proxy-side read agrees — but the copy loop skipped `$DELETED` slots without restoring the result's length afterwards, so trailing holes truncated the copy (at any nesting depth, for either API). The array branch of `snapshotImpl` now restores the length after the loop, mirroring `unwrapStoreValue`, so deleted slots stay holes (`i in copy === false`, serialized as `null`).
--- a/packages/solid-signals/src/store/utils.ts
+++ b/packages/solid-signals/src/store/utils.ts
@@ -56,6 +56,8 @@ function snapshotImpl<T>(
         result[i] = unwrapped;
       }
     }
+    // Deleted trailing slots are skipped above, so restore length to preserve holes.
+    if (result) result.length = len;
   } else {
     const keys = getKeys(item, override);
     for (let i = 0, l = keys.length; i < l; i++) {
--- a/packages/solid-signals/tests/store/utilities.test.ts
+++ b/packages/solid-signals/tests/store/utilities.test.ts
@@ -651,6 +651,66 @@ describe("deep", () => {
 });
 
 describe("snapshot", () => {
+  test("preserves array length when a trailing index was deleted", () => {
+    const [state, setState] = createStore<(string | undefined)[]>(["a", "b", "c"]);
+    setState(s => {
+      delete s[2];
+    });
+    flush();
+    expect(state.length).toBe(3); // proxy: plain-JS delete semantics
+
+    const copy = snapshot(state);
+    expect(copy.length).toBe(3); // the copy must agree with the store
+    expect(2 in copy).toBe(false); // and keep the hole a hole, not undefined
+    expect(JSON.stringify(copy)).toBe('["a","b",null]');
+  });
+
+  test("deep() preserves array length for trailing holes too", () => {
+    const [state, setState] = createStore<(string | undefined)[]>(["a", "b", "c"]);
+    setState(s => {
+      delete s[2];
+    });
+    flush();
+    createRoot(dispose => {
+      const copy = deep(state);
+      expect(copy.length).toBe(3);
+      expect(2 in copy).toBe(false);
+      dispose();
+    });
+  });
+
+  test("preserves length for nested arrays and multi-hole runs", () => {
+    const [state, setState] = createStore<{ list: (string | undefined)[] }>({
+      list: ["a", "b", "c", "d"]
+    });
+    setState(s => {
+      delete s.list[2];
+      delete s.list[3];
+    });
+    flush();
+    const copy = snapshot(state);
+    expect(copy.list.length).toBe(4);
+    expect(JSON.stringify(copy.list)).toBe('["a","b",null,null]');
+  });
+
+  test("middle holes and length truncation keep working (controls)", () => {
+    const [mid, setMid] = createStore<(string | undefined)[]>(["a", "b", "c"]);
+    setMid(s => {
+      delete s[1];
+    });
+    flush();
+    expect(JSON.stringify(snapshot(mid))).toBe('["a",null,"c"]');
+    expect(snapshot(mid).length).toBe(3);
+
+    const [cut, setCut] = createStore<string[]>(["a", "b", "c"]);
+    setCut(s => {
+      s.length = 2;
+    });
+    flush();
+    expect(snapshot(cut).length).toBe(2);
+    expect(JSON.stringify(snapshot(cut))).toBe('["a","b"]');
+  });
+
   test("returns same object if unchanged", () => {
     const ref = { a: 1, b: 2 };
     const [state, setState] = createStore(ref);


ryansolid:
Correct on all counts — `unwrapStoreValue` already restored the length in the identical situation, so `snapshotImpl` skipping `$DELETED` without the restoration was a plain omission, and the one-line fix covers every variant (both APIs share the impl, nesting recurses through it). The 1.x regression check was appreciated context.

Your branch predates the #2850/#2756 `snapshotImpl` rework (merged overlay + the specialized no-overlay walk), so the diff didn't apply — the same one-liner is now placed in the reworked array branch, with your four tests taken verbatim (verified red-first: 3 fail unfixed, the controls pass). Landed as `faf78ebd` with co-author credit. Ships in the next beta — thanks!

PR #2849, opened 2026-07-07
fix(signals): settle resolve() when the async source rejects

### LINKED ISSUE #2842: 2.0.0-beta.15: `resolve()` never settles (and leaks its root) when the async source rejects
## Describe the bug

`resolve(fn)` is documented to await a reactive expression and return its first fully-settled value as a `Promise`. When the underlying async source **resolves**, this works. When it **rejects** after having been pending, the promise returned by `resolve()` never settles — it neither resolves nor rejects — and the internal `createRoot` is never disposed, leaking the root and everything under it.

Typical real-world hit: SSR or a route preloader doing `await resolve(...)` on data that 404s — the request hangs instead of surfacing the error.

## Reproduction

```ts
import { createMemo, createRoot, resolve } from "@solidjs/signals";

const delay = (ms: number) => new Promise(r => setTimeout(r, ms));

await createRoot(async () => {
  const m = createMemo(async () => {
    await delay(10);
    throw new Error("boom");
  });

  await resolve(() => m()); // hangs forever — never resolves, never rejects
});
```

brenelz:
## Summary

Fixes #2842 — `resolve(fn)` hung forever (never resolved, never rejected) and leaked its internal `createRoot` whenever the awaited async source **rejected** after having been pending. Resolution worked; rejection did not.

## Root cause

`resolve()` was built on a bare `computed` inside a root, and computeds are pull-based. The two settlement paths are asymmetric:

- When a pending source **resolves**, `settlePendingSource()` explicitly re-enqueues blocked dependents, so the computed re-ran, called `res(fn())`, and disposed the root.
- When a pending source **rejects**, `handleAsync`'s error path only propagates `STATUS_ERROR` through `notifyStatus()`, which *marks* subscribers errored for their next pull but never re-runs them. Nothing ever reads `resolve()`'s computed (it has no subscribers), so neither `res` nor `rej` fired and `dispose()` never ran.

## Fix

Rebuild `resolve()` on a **user effect** — the primitive that owns an error-notification channel (`_notifyStatus` → `_errorFn`) and is therefore actively told when a pending source rejects:

- effect callback → `res(value)` + dispose (same observable timing as before)
- error callback → `rej(err)` + dispose, unwrapping the internal `StatusError` wrapper so the promise rejects with the user's original error, matching what `createErrorBoundary` exposes

Also documents the rejection behavior in the JSDoc.

## Tests

New `tests/resolve.test.ts` pins the settlement contract:

- rejects with the original error when the async source rejects after pending *(hung before this fix)*
- disposes the internal root on rejection *(hung before this fix)*
- still resolves with the settled value on success
- rejects on a synchronous throw

Full `solid-signals` suite (814 tests) and `solid` suite (418 tests) pass.

### DIFF
--- a/packages/solid-signals/src/signals.ts
+++ b/packages/solid-signals/src/signals.ts
@@ -23,6 +23,7 @@ import {
   untrack
 } from "./core/index.js";
 import { emitDiagnostic, registerGraph } from "./core/dev.js";
+import { StatusError } from "./core/error.js";
 import { globalQueue } from "./core/scheduler.js";
 
 /**
@@ -588,7 +589,9 @@ export function createReaction(
  * Awaits a reactive expression and returns its first fully-settled value as a
  * `Promise`. Pending async reads (`createMemo` returning a promise, etc.) are
  * waited on; once the expression returns synchronously without `NotReadyError`
- * the promise resolves with that value.
+ * the promise resolves with that value. If the expression settles with an
+ * error instead — including an async source that rejects — the promise
+ * rejects with it.
  *
  * Must be called *outside* a tracking scope — it doesn't subscribe, it just
  * resolves the current value once.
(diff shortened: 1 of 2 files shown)


ryansolid:
Root cause is exactly right, and the asymmetry framing is the valuable part: `settlePendingSource` re-enqueues blocked dependents on *resolution*, but a *rejection* only propagates `STATUS_ERROR` through `notifyStatus` — which marks a pull-based computed errored for a pull that, with no subscribers, never comes. `resolve()`'s bare computed was exactly that shape, so the promise hung and the root leaked.

Rebuilding on a user effect is the right fix — the effect's error-notification channel (`_notifyStatus` → `_errorFn`) is the primitive that's actively told about rejections, and unwrapping the `StatusError` so the caller gets their original error matches what error boundaries expose. Verified the hang on the current tree, confirmed the full monorepo suite passes with your diff, and landed it as-is with a changeset, credited as co-author: `c4ba5268`. Ships in the next beta — thanks for another precise one.

Issue #2769, opened 2026-06-13
2.0.0-beta.14: # Store helpers inconsistently handle symbol-keyed properties


yumemi-thomas:
### Describe the bug

In Solid 2.0.0-beta.14, store APIs inconsistently handle symbol-keyed own properties.

The original symptom was that writing a symbol-keyed property on an array store throws. The issue appears broader than that: some store helper/replacement paths silently drop symbol keys, and `merge()` / `omit()` can read symbol-keyed properties but omit them from own-key enumeration.

This makes symbol-keyed metadata unreliable across store update and props-helper paths.

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-mpwaeaj7?file=src%2FApp.tsx

```tsx
import { createSignal, createStore, flush, merge, omit, storePath } from "solid-js";

const meta = Symbol("meta");

export default function App() {
  const [result, setResult] = createSignal("click run");

  return (
    <main>
      <button
        onClick={() => {
          const lines: string[] = [];

          // Original symptom: array symbol writes should not be treated as index writes.
          const [list, setList] = createStore<any>([1, 2, 3]);
          try {
            setList((list: any) => {
              list[meta] = 42;
            });
            flush();
            lines.push(`array symbol write: ${String(list[meta])}`);
          } catch (error) {
            lines.push(`array symbol write threw: ${String(error)}`);
          }

          // Control: direct object draft write works.
          const [direct, setDirect] = createStore<Record<PropertyKey, unknown>>({});
          setDirect(s => {
            s[meta] = 42;
          });
          flush();
          lines.push(`direct object draft write: ${String(direct[meta])}`);

          // Bug: returned replacement drops symbol keys.
          const [returned, setReturned] = createStore<Record<PropertyKey, unknown>>({});
          setReturned(() => ({ [meta]: 42 }));
          flush();
          lines.push(`return replacement: ${String(returned[meta])}`);

          // Bug: storePath root merge drops symbol keys.
          const [pathStore, setPathStore] = createStore<Record<PropertyKey, unknown>>({});
          setPathStore(storePath({ [meta]: 42 }));
          flush();
          lines.push(`storePath root merge: ${String(pathStore[meta])}`);

          // Bug: merge/omit can read symbol values, but own-key enumeration drops them.
          const [source] = createStore<Record<PropertyKey, unknown>>({
            visible: true,
            [meta]: 42
          });

          const merged = merge({ fallback: true }, source);
          const omitted = omit(source, "visible");

[...]

PR #2852, opened 2026-07-07
fix(store): reconcile() notifies symbol-keyed properties (closes #2851)

### LINKED ISSUE #2851: 2.0.0-beta.15: `reconcile()` never notifies symbol-keyed nodes — the #2769 fix covers the set trap but not reconcile's diff loops
### Describe the bug

`reconcile()` never notifies nodes for symbol-keyed store properties:

1. An effect (or rendered binding) tracking `state[SYM]` never re-runs when `reconcile()` changes that property.
2. Because the stale node then exists, even plain untracked reads of `state[SYM]` keep returning the old value — the new value written by `reconcile` is unreachable. (Without any prior tracked read there is no node, and reads fall through to the swapped value — which is why the bug hides in non-rendering tests.)

This is a remaining variant of #2769. The committed fix (bc92d002) covered symbol keys in the **set trap** (array metadata writes), `storeSetter`, `storePath`, and `merge`/`omit` enumeration — all via the new `ownEnumerableKeys` helper. `reconcile()`'s diff loops still enumerate with `Object.keys`, which drops symbol keys, so they were not reached by that fix. String-keyed properties update fine through the identical code path (verified as a control).

https://stackblitz.com/edit/solidjs-templates-krnhumcg?file=src%2FApp.tsx

```tsx
import { createStore, reconcile } from "solid-js";

const META = Symbol("meta");

  const [store, setStore] = createStore<any>({ id: 1, label: "old", [META]: "old" });

  return (
      <h2>reconcile() skips symbol-keyed bindings</h2>
      <p>
        string key <code>label</code>: <b>{store.label}</b> (control)
      </p>
      <p>
        symbol key <code>[META]</code>: <b>{store[META]}</b>
      </p>
      <button onClick={() => setStore(reconcile({ id: 1, label: "new", [META]: "new" }, "id"))}>
        reconcile fresh data
      </button>
    </main>
  );
}
```

1. Initial render shows both rows at `old`.
2. Click **reconcile fresh data** — one `reconcile()` carrying the new value for both keys. Actual — the string row updates, the symbol row does not:

```text
string key label: new     (control)
symbol key [META]: old
```

The symbol binding never re-ran, and even a plain `store[META]` read still returns `"old"` — the value `reconcile` wrote is unreachable behind the stale node. The string row, updated by the same call, is the built-in control: the only difference is the key type.

### Expected behavior

Symbol-keyed properties reconcile like string-keyed ones — both rows update:

```text
string key label: new     (control)
symbol key [META]: new
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.15 (re-verified at `next` @ `e2ebc11c`)

### Additional context

Root cause: every key enumeration in `reconcile`'s object diff drops symbols (`packages/solid-signals/src/store/reconcile.ts`):

```ts
// reconcile.ts:28-32 — used when the store's $TRACK node is tracked
function getAllKeys(value, override, next) {
  const keys = getKeys(value, override) as string[];   // getKeys → Object.keys
  const nextKeys = Object.keys(next);
  return Array.from(new Set([...keys, ...nextKeys]));
}

// reconcile.ts:187 (applyStateFast) / :337 (applyStateSlow) — untracked path
const keys = tracked ? getAllKeys(previous, undefined, next) : Object.keys(nodes);
```

Symbol-keyed nodes in the `nodes` record are invisible to `Object.keys(nodes)`, and symbol keys of `next` are invisible to `getAllKeys`, so `setSignal` is never called for them; the `STORE_HAS` loops (reconcile.ts:209/:359) have the same gap. Meanwhile `applyState*` has already swapped `STORE_VALUE` to `next`, so the stale node shadows the new value for direct reads too.

**Exists in 1.x too — in a worse form.** Verified against solid-js **1.9.14**: 1.x `reconcile` doesn't even *merge* symbol-keyed values (`state[sym]` stays at the old value after `setState(reconcile({ [sym]: 2, a: 2 }))`; string keys update fine), so no notification question even arises. 2.0 half-fixed this — the value now merges — but subscribers of the symbol-keyed node are still never notified. Not a regression, but 2.0 already handles symbols in its other store paths (`getAllKeys`, set trap after #2769), so finishing the job in reconcile's diff loops is consistent.

yumemi-thomas:
Closes #2851

## Summary

`reconcile()` never notifies subscribers of **symbol-keyed** store properties. An effect or binding tracking `state[SYM]` doesn't re-run when `reconcile()` changes that property, and because the stale node then shadows the value, even a plain `state[SYM]` read keeps returning the old value — the reconciled value is unreachable. String-keyed properties on the same store update fine through the identical path, so the only difference is the key type.

https://stackblitz.com/edit/solidjs-templates-krnhumcg?file=src%2FApp.tsx

This is the remaining variant of #2769: that fix (`bc92d002`) covered symbol keys in the set trap, `storeSetter`, `storePath`, and `merge`/`omit` via the `ownEnumerableKeys` helper — but `reconcile()`'s own diff loops still enumerated with `Object.keys`, which drops symbols.

## Root cause

Every key enumeration in reconcile's diff dropped symbols:

- `getAllKeys` (the `$TRACK`-tracked path) built from `getKeys` (string-only) + `Object.keys(next)`.
- The untracked value loops and the `STORE_HAS` (`in`-dependency) loops iterated `Object.keys(nodes)`.

So `setSignal` was never called for symbol-keyed nodes, while `applyState*` had already swapped `STORE_VALUE` to the new value — leaving the stale node shadowing it for direct reads too.

## Fix

Enumerate symbol keys everywhere reconcile enumerates string keys, while keeping the common symbol-free path on the exact `Object.keys` fast path:

- **`getAllKeys`** builds string keys via `Object.keys` (unchanged) and appends *enumerable* symbol keys via `getOwnPropertySymbols` — symbol-free objects pay only an empty call, no per-key predicate.
- **`nodeKeys`** (new) enumerates a node record's keys: `Object.keys` always, plus symbol nodes only for records flagged in a `symbolKeyedRecords` WeakSet. `getNode` marks a record when it creates a user (non-`$TRACK`) symbol node, and its `unobserved` cleanup drops the mark once the last such node goes — so the flag means exactly "this record currently holds a user symbol node", and the reconcile hot path stays on `Object.keys` gated by a free `WeakSet.has`.
- `$TRACK` (the internal self-node, the only internal symbol a node record can hold — the get/has traps early-return the others) is filtered out of symbol enumeration; callers handle it separately.

Semantics are identical to the old code for string keys, and match `ownEnumerableKeys` (enumerable strings + enumerable symbols) for the extended set; key order is irrelevant since each key is diffed independently.

## Tests

Seven tests in `tests/store/reconcile.test.ts` (`reconcile with symbol-keyed properties`), the symbol cases confirmed failing red-first on the unfixed source:

- effect tracking a symbol-keyed property is notified; the value is reachable, not shadowed
- string control updates through the identical path
- string and symbol keys on one store both update
- a symbol key removed by reconcile notifies `undefined`
- a symbol `in` check updates when reconcile adds the key
- a nested symbol-keyed value reconciles
- perf invariant: the `symbolKeyedRecords` mark is set while tracked and cleared once unobserved (guards the fast-path optimization against a monotonic leak)

Full solid-signals suite: the diff versus pristine is exactly the intended flips, nothing else; solid and @solidjs/web rebuilt against the fixed package and re-run at their baselines; source and test files pass the tests-inclusive typecheck.

## Solid 1.x note

Exists in 1.x too, in a worse form: 1.x `reconcile` doesn't even *merge* symbol-keyed values (the value stays stale; no notification question arises). 2.0 half-fixed this — the value merges — but subscribers of the symbol-keyed node are still never notified. Not a regression, but 2.0 already handles symbols in its other store paths (set trap, `getAllKeys`, `merge`/`omit` after #2769), so finishing the job in reconcile's diff loops is consistent.

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. All new tests were written first and confirmed failing on the unfixed code

### DIFF
--- a/.changeset/reconcile-symbol-keyed-nodes.md
+++ b/.changeset/reconcile-symbol-keyed-nodes.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Fix `reconcile()` updates for symbol-keyed store properties so tracked reads and `in` checks are notified like string-keyed properties.
--- a/packages/solid-signals/src/store/reconcile.ts
+++ b/packages/solid-signals/src/store/reconcile.ts
@@ -1,9 +1,9 @@
 import { setSignal } from "../core/index.js";
 import {
+  $DELETED,
   $PROXY,
   $TARGET,
   $TRACK,
-  getKeys,
   isWrappable,
   STORE_HAS,
   STORE_LOOKUP,
@@ -13,22 +13,56 @@ import {
   STORE_VALUE,
   notifySelf,
   storeLookup,
+  symbolKeyedRecords,
   wrap
 } from "./store.js";
+
+function nodeKeys(nodes: Record<PropertyKey, any>): PropertyKey[] {
+  const keys: PropertyKey[] = Object.keys(nodes);
+  // Keep the common string-key path cheap; only symbol-tracked records pay for
+  // symbol enumeration. `$TRACK` is handled separately by callers.
+  if (symbolKeyedRecords.has(nodes)) {
+    const syms = Object.getOwnPropertySymbols(nodes);
+    for (let i = 0, len = syms.length; i < len; i++) {
+      if (syms[i] !== $TRACK) keys.push(syms[i]);
+    }
+  }
+  return keys;
+}
 
 function unwrap(value: any) {
   return value?.[$TARGET]?.[STORE_VALUE] ?? value;
 }
 
-function getOverrideValue(value: any, override: any, key: string, optOverride?: any) {
+function getOverrideValue(value: any, override: any, key: PropertyKey, optOverride?: any) {
   if (optOverride && key in optOverride) return optOverride[key];
   return override && key in override ? override[key] : value[key];
 }
 
+// Append a value's *enumerable* own symbol keys. Symbol-free objects (the
+// common case) pay only an empty `getOwnPropertySymbols` call — no per-key
+// predicate — so the string fast path stays on `Object.keys`.
+function addEnumSymbols(o: any, keys: Set<PropertyKey>) {
+  const syms = Object.getOwnPropertySymbols(o);
+  for (let i = 0, len = syms.length; i < len; i++) {
+    if (Object.prototype.propertyIsEnumerable.call(o, syms[i])) keys.add(syms[i]);
+  }
+}
+
 function getAllKeys(value, override, next) {
-  const keys = getKeys(value, override) as string[];
+  // Reconcile must diff enumerable symbol keys the same way it diffs strings,
+  // but keep the string keys on the `Object.keys` fast path — symbols are
+  // appended only when the object actually has them.
+  const keys = new Set<PropertyKey>(Object.keys(value));
+  addEnumSymbols(value, keys);
+  if (override) {
+    for (const key of Reflect.ownKeys(override))
+      override[key] === $DELETED ? keys.delete(key) : keys.add(key);
+  }
   const nextKeys = Object.keys(next);
-  return Array.from(new Set([...keys, ...nextKeys]));
+  for (let i = 0, len = nextKeys.length; i < len; i++) keys.add(nextKeys[i]);
+  addEnumSymbols(next, keys);
+  return Array.from(keys);
 }
 
 // Array entries can be `null`/`undefined`/primitives, not just keyed objects.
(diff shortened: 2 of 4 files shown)


ryansolid:
Right diagnosis and the right performance-conscious shape — the `symbolKeyedRecords` WeakSet mark maintained at node creation/cleanup is exactly how to keep the symbol-free hot path on `Object.keys` without a per-key predicate, and filtering `$TRACK` (the only internal symbol a node record can hold) at the enumeration site is correct.

Your branch predates the #2756 `getAllKeys` rewrite (the identical-key-sets fast path landed there), so the diff no longer applied — we re-implemented it against the current tree preserving your design: symbol-free diffs stay byte-for-byte on the old path including the fast path, the symbol-aware branch re-applies override `$DELETED` after base symbols so deletes win, and we added one guard you couldn't have known to need — store-in-store values enumerate their symbols untracked, mirroring `getKeys`. All seven of your tests are included verbatim (symbol cases re-verified red-first), including the perf-invariant pin on the WeakSet mark lifecycle. Landed as `b9f27372` with co-author credit. Ships in the next beta — thanks!

Issue #2767, opened 2026-06-13
2.0.0-beta.14: `<Repeat>` window bookkeeping disposes wrong rows and can crash after `count=0`


yumemi-thomas:
### Describe the bug

`<Repeat>` appears to lose track of its window owner/index bookkeeping when `from` changes.

There are two related symptoms:

1. When a fixed-size window advances, `<Repeat>` disposes the wrong row owners.
2. When `count` goes to `0`, then `from` changes, then `count` becomes nonzero again, `<Repeat>` can crash.

Both cases point to the same underlying invariant: row owners are stored relative to the current window, but some update paths appear to use stale or absolute indexes.

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-ua7m7sct?file=src%2FApp.tsx

### Repro 1: advancing `from` disposes the wrong rows

Minimal app:

```tsx
import { createSignal, onCleanup, Repeat } from "solid-js";

function Row(props: { index: number }) {
  onCleanup(() => {
    console.log("cleanup row", props.index);
  });

  return <p>row {props.index}</p>;
}

export default function App() {
  const [from, setFrom] = createSignal(1);

  return (
    <>
      <Repeat count={3} from={from()}>
        {index => <Row index={index} />}
      </Repeat>

      <button onClick={() => setFrom(3)}>slide window from 1-3 to 3-5</button>
    </>
  );
}
```

Steps:

1. Open the repro.
2. Open the browser console.
3. Click `slide window from 1-3 to 3-5`.
4. Observe the cleanup logs.

### Repro 2: `count -> 0 -> nonzero` with changed `from` crashes

Minimal app:

```tsx
import { createSignal, Repeat } from "solid-js";

export default function App() {
  const [from, setFrom] = createSignal(2);
  const [count, setCount] = createSignal(2);

  return (
    <>
      <Repeat count={count()} from={from()}>
        {index => <p>row {index}</p>}
      </Repeat>

      <button onClick={() => setCount(0)}>1. clear rows</button>
      <button
        onClick={() => {
          setFrom(0);
          setCount(1);
        }}
      >
        2. show row 0
      </button>
    </>
  );
}
```

Steps:

1. Open the repro.
2. Click `1. clear rows`.
3. Click `2. show row 0`.
4. Observe the crash.

### Repro 1

The visible window changes from:

```text
row 1
row 2
row 3
```

to:

```text
row 3
row 4
row 5
```

[...]

PR #2854, opened 2026-07-07
fix(signals): repeat() handles disjoint window jumps (closes #2853)

### LINKED ISSUE #2853: 2.0.0-beta.15: `<Repeat>` leaks live rows and can crash on disjoint window jumps
### Describe the bug

`repeat()` / `<Repeat>` with a reactive `from` handles small window slides, but not jumps where the old and new windows do not overlap.

Think of `count={3}` as a moving window:

```text
from=0   -> rows 0,1,2
from=10  -> rows 10,11,12
```

On a correct implementation, clicking **jump +10** from the initial state would leave exactly three live row scopes:

```text
from=10, count=3
created=6  cleaned=3  LIVE SCOPES=3
calls: cleanup(0) cleanup(1) cleanup(2) map(10) map(11) map(12)
```

What actually happens is that `<Repeat>` creates every gap row too:

```text
from=10, count=3
created=13  cleaned=3  LIVE SCOPES=10
calls: cleanup(0) cleanup(1) cleanup(2) map(3) map(4) map(5) map(6) map(7) map(8) map(9) map(10) map(11) map(12)
```

Rows `3..9` never appear in the DOM, but their owners/effects stay alive because they are stored outside the normal window bookkeeping. Repeating the jump leaks more rows. Jumping backward across a disjoint gap can also throw before the window is replaced.

So the bug has three visible forms:

1. **Initial nonzero `from` maps too much.** `from=10,count=3` maps `0..12` instead of `10..12`.
2. **Forward disjoint jumps leak invisible row scopes.** `0..2 -> 10..12` maps and leaks `3..9`.
3. **Backward disjoint jumps can crash.** `20..22 -> 0..2` throws while trying to dispose a row at a negative local index.

This is a normal virtualized-list shape: dragging the scrollbar or jumping to an index moves `from` by more than `count` all the time.

Related to but distinct from #2767 (closed, fixed by #2784 / d8921ac1): that fix covered the empty-window reset and *overlapping* backward slides, and those cases still work correctly. The disjoint geometries (`prevTo < from` and `to <= offset`) were never handled.

The invariant is simple: after any update, `created - cleaned` should equal `count`, and the map function should only run for indices in `[from, from + count)`. Disjoint jumps violate both sides of that contract.

https://stackblitz.com/edit/solidjs-templates-de3jtq5s?file=src%2FApp.tsx

The rendered list always *looks* correct — the leaked rows never enter the DOM, only their owners/effects stay alive — so the panel shows `created`/`cleaned`/`live` counts and the exact `map`/`cleanup` calls each action makes. (It's written with raw DOM writes because the backward jump crashes and freezes reactivity — a reactive panel couldn't render the final state.) Validated against 2.0.0-beta.15.

```tsx
import { createSignal, onCleanup, Repeat, flush } from 'solid-js';

  const [from, setFrom] = createSignal(0);
  let created = 0;
  let cleaned = 0;
  let calls: string[] = [];
  let panel!: HTMLPreElement;

  // Raw DOM write: the backward jump crashes and freezes reactivity, so the
  // panel must not depend on a reactive read to show the final leaked state.
  const report = (action: string, note = '') => {
    panel.textContent =
      `${action}
` +
      `from=${from()}, count=3 — the DOM list shows only these 3 rows
` +
      `created=${created}  cleaned=${cleaned}  LIVE SCOPES=${
        created - cleaned
      } (should be 3)
` +
      (note ? note + '' : '') +
      `calls: ${calls.join(' ') || '(none)'}`;
  };

  const go = (to: number, label: string) => {
    calls = [];
    try {
      setFrom(to);
      flush();
      report(label);
    } catch (e) {
      report(label, `CRASH: ${e} — UI frozen, buttons dead`);
    }
  };

  return (
    <main style={{ 'font-family': 'system-ui', padding: '16px' }}>
      <h2>&lt;Repeat&gt; leaks live row scopes on disjoint window jumps</h2>
      <button onClick={() => go(from() + 1, 'step +1')}>step +1</button>{' '}
      <button onClick={() => go(from() + 10, 'jump +10')}>jump +10</button>{' '}
      <button onClick={() => go(0, 'jump to 0 (disjoint backward)')}>
        jump to 0
      </button>
      <ul>
        <Repeat count={3} from={from()}>
          {(i: number) => {
            created++;
            calls.push(`map(${i})`);
            onCleanup(() => {
              cleaned++;
              calls.push(`cleanup(${i})`);
            });
            return <li>row {i}</li>;
          }}
        </Repeat>
      </ul>
      {/* the pre's ref callback fires the initial report once it mounts */}
      <pre ref={(el) => ((panel = el), report('mount'))} />
    </main>
  );
}

```

1. On mount the panel reads `created=3 cleaned=0 LIVE SCOPES=3` and the list shows rows 0–2. Correct.
2. Click **"jump +10"**. This changes the requested window from `0..2` to `10..12`.

Expected panel:

```text
jump +10
from=10, count=3 — the DOM list shows only these 3 rows
created=6  cleaned=3  LIVE SCOPES=3 (should be 3)
calls: cleanup(0) cleanup(1) cleanup(2) map(10) map(11) map(12)
```

Actual panel:

```text
jump +10
from=10, count=3 — the DOM list shows only these 3 rows
created=13  cleaned=3  LIVE SCOPES=10 (should be 3)
calls: cleanup(0) cleanup(1) cleanup(2) map(3) map(4) map(5) map(6) map(7) map(8) map(9) map(10) map(11) map(12)
```

Rows `3..9` were created, never cleaned, and stay live — invisible in the DOM. Click **"jump +10"** again and `LIVE SCOPES` grows to `17`.

3. Click **"jump to 0"**. This is a backward disjoint jump.

Expected panel:

```text
jump to 0 (disjoint backward)
from=0, count=3 — the DOM list shows only these 3 rows
created=9  cleaned=6  LIVE SCOPES=3 (should be 3)
calls: cleanup(10) cleanup(11) cleanup(12) map(0) map(1) map(2)
```

Actual panel:

```text
CRASH: Error: Cannot read properties of undefined (reading 'dispose') — UI frozen, buttons dead
```

The list stays frozen on the old rows and every button is dead from then on.

4. Control: reload and use **"step +1"** instead. That move overlaps the old window, so everything stays correct and `LIVE SCOPES` remains `3`.

5. Bonus: change the initial signal to `createSignal(10)` and reload — the mount panel reads `created=13` for a 3-row window (the whole 0..12 prefix is mapped).

### Expected behavior

Rows are only created inside the requested window, every row that leaves the window is cleaned up, and backward jumps work:

```text
after any jump:  created - cleaned == 3, list shows the 3 requested rows
jump to 0:       no crash — rows 0–2 render, LIVE SCOPES=3
```

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 2.0.0-beta.15 (re-verified at `next` @ `61722cbe`, which includes the #2784 fix; all three symptoms and the exact counts still reproduce)

### Exact map/cleanup call sequences

Instrumenting the map function with `console.log` shows precisely which rows each jump creates and disposes:

```text
mount:        map(0) map(1) map(2)                                    ✓
jump to 10:   cleanup(0..2), then map(3) map(4) ... map(12)           ← 10 created for a 3-row window;
                                                                        rows 3..9 never get a cleanup
jump to 20:   cleanup(10..12), then map(13) ... map(22)               ← 7 more leaked (13..19);
                                                                        23 mapped, 6 cleaned = 17 live
jump to 0:    nothing — no map, no cleanup, StatusError (see above)
```

Quick matrix:

```text
action                        expected calls                   actual calls / result
initial from=10,count=3       map(10) map(11) map(12)          map(0) ... map(12)
0..2 -> 10..12                cleanup(0..2), map(10..12)       cleanup(0..2), map(3..12), leaks 3..9
10..12 -> 0..2                cleanup(10..12), map(0..2)       throws before replacement
overlap, 0..2 -> 1..3         cleanup(0), map(3)               works
```

### Root cause

The row-creation loop in `updateRepeat` (`packages/solid-signals/src/map.ts:377`) assumes the old and new windows overlap:

```ts
for (let i = prevTo; i < to; i++) {
  this._mappings[i - from] = runWithOwner<MappedItem>(
    (this._nodes[i - from] = createOwner()),  // i - from is NEGATIVE for every i < from
    () => this._map(i)
  );
}
```

When `prevTo < from` (first render with nonzero `from`, where `prevTo` is 0, or any forward jump larger than the window), the loop starts at `prevTo` and walks the entire gap. Every `i < from` writes its owner at a **negative** `_nodes` index — invisible to all the dispose/splice bookkeeping, which only operates on local indices `0..len-1`. That is symptoms 1 and 2.

Symptom 3 is the mirror image in the end-clear loop (`map.ts:350`):

```ts
for (let i = to; i < prevTo; i++) this._nodes[i - this._offset].dispose();
```

On a disjoint backward jump (`to <= this._offset`), `i - this._offset` goes negative, `this._nodes[-7]` is `undefined`, and `.dispose()` throws — aborting the update before anything is created or cleaned. The `_offset > from` shift branch just below (`map.ts:360-375`) has the same overlap assumption: on a disjoint backward move it would overwrite live owners in `_nodes` without disposing them.

### Why #2784 / d8921ac1 don't cover this

That fix resets `_offset` when the window goes *empty* and corrects which rows get disposed on overlapping slides from a nonzero offset. All the loops above still model window movement as "shift + fill the difference", which is only valid when `old ∩ new ≠ ∅`. Disjoint jumps never take a sane path.

### Suggested fix direction

Detect disjoint windows up front — `from >= prevTo || to <= this._offset` — and treat them as a full window replacement: dispose all current rows (local indices `0..len-1`), reset `_nodes`/`_mappings`, and create exactly the rows in `[from, to)`. That condition is the exact complement of "the windows overlap" (`from < prevTo && to > offset`), so partial-overlap slides still take the existing preserving path and keep shared rows' identity; only truly disjoint moves are replaced wholesale (which is unavoidable work — no shared rows to preserve). It also subsumes the nonzero-first-render case, where `prevTo` is 0.

Working patch with tests in hand (fix PR to follow): the ~15-line guard above fixes all three symptoms with no regression to the overlapping paths, and the common overlapping-slide path benchmarks unchanged (the two added comparisons are free). No loop-clamping is needed — the guard's classification is exact, so the negative-index accesses become unreachable rather than merely clamped.

**Not applicable — new 2.0 API.** `repeat()`/`<Repeat>` has no Solid 1.x counterpart (the closest 1.x pattern, `<Index>` over a sliced range array, has no windowing bookkeeping and no equivalent failure mode).

yumemi-thomas:
Closes #2853

## Summary

`repeat()` / `<Repeat>` with a reactive `from` only handled window moves that **overlap** the previous window. Any disjoint jump — the everyday shape of a virtualized list, where dragging the scrollbar or jumping to an index moves `from` by more than `count` — broke three ways:

1. **Initial render with a nonzero `from` maps the whole prefix** — `from: 10, count: 3` creates rows `0..12` (13 rows for a 3-row window).
2. **A forward jump larger than the window leaks every gap row** — jumping `0 → 10` maps `3..12` and stores the gap rows (`3..9`) at *negative* `_nodes` indices, so nothing ever disposes them; their owners, effects and `onCleanup`s stay live until the whole list is torn down. Each disjoint jump adds more (`0 → 10 → 20` ⇒ 17 live scopes for a 3-row window).
3. **A backward disjoint jump crashes** — `20 → 0` throws `Cannot read properties of undefined (reading 'dispose')`, the update aborts, and the list freezes on the old rows.

The rendered list always *looks* right (leaked rows never enter the DOM), so the leak is invisible without watching live scope counts. Repro: https://stackblitz.com/edit/solidjs-templates-de3jtq5s?file=src%2FApp.tsx

Distinct from #2767 (fixed by #2784 / `d8921ac1`): that covered the empty-window reset and *overlapping* backward slides, which still work. The disjoint geometries were never handled.

## Root cause

`updateRepeat` (`packages/solid-signals/src/map.ts`) models every window move as "clear the end, shift the front, fill the difference" — valid only when `old ∩ new ≠ ∅`. The fill loop `for (i = prevTo; i < to; i++) this._nodes[i - from] = …` writes negative local indices for every `i < from` (symptoms 1–2), and the end-clear loop `this._nodes[i - this._offset].dispose()` reads `this._nodes[negative]` (`undefined`) on a backward disjoint jump (symptom 3).

## Fix

Detect the disjoint case up front and replace the window wholesale:

```ts
if (from >= prevTo || to <= this._offset) {
  for (let i = 0; i < this._len; i++) this._nodes[i].dispose();
  this._nodes = new Array(newLen);
  this._mappings = new Array(newLen);
  for (let i = 0; i < newLen; i++)
    this._mappings[i] = runWithOwner(this._nodes[i] = createOwner(), () => this._map(from + i));
  this._offset = from;
  this._len = newLen;
  return;
}
```

That condition is the exact boolean complement of "the windows overlap" (`from < prevTo && to > offset`), so:

- **partial-overlap slides still take the existing preserving path** and keep shared rows' identity/state — verified: `[5,8) → [3,6)` creates only rows `3,4` and preserves row `5`;
- **only truly disjoint moves are replaced wholesale**, which is unavoidable work (no shared rows to preserve);
- it also subsumes the nonzero-first-render case (`prevTo` is 0 on first render).

The existing overlapping-slide branches (the #2784 fix) are untouched, and no loop-clamping is added — the guard makes the negative-index accesses unreachable rather than merely clamped.

## Tests

`tests/repeat.test.ts` gains a "disjoint window jumps" block, all confirmed red-first on the unfixed source:

- a forward jump larger than the window creates only the window's rows (`live === 3`, not 10)
- repeated disjoint jumps do not accumulate live scopes (`live === 3`, not 17)
- a backward disjoint jump maps the new window without crashing
- initial render with a nonzero `from` maps only the window (3 created, not 13)

## Solid 1.x note

Not applicable — `repeat()` / `<Repeat>` is a new 2.0 API with no 1.x counterpart (the closest 1.x pattern, `<Index>` over a sliced range, has no windowing bookkeeping and no equivalent failure mode).

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. All new tests were written first and confirmed failing on the unfixed code

### DIFF
--- a/.changeset/repeat-disjoint-window.md
+++ b/.changeset/repeat-disjoint-window.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Fix `repeat()` / `<Repeat>` leaking live row scopes and crashing on disjoint window jumps. When a reactive `from` moves the window to indices that don't overlap the previous window, the shift-and-fill update walked `_nodes` at negative local indices: a forward jump larger than the window created every gap row and left them alive (owners, effects, and `onCleanup`s never disposed), and a backward disjoint jump threw `Cannot read properties of undefined (reading 'dispose')` and froze the list. The first render with a nonzero `from` mapped the whole `0..from+count` prefix for the same reason. Disjoint windows are now detected and replaced wholesale; overlapping slides (the #2784 fix) are unchanged.
--- a/packages/solid-signals/src/map.ts
+++ b/packages/solid-signals/src/map.ts
@@ -346,6 +346,20 @@ function updateRepeat<MappedItem>(this: RepeatData<MappedItem>): any[] {
     // remove fallback
     if (this._len === 0 && this._nodes[0]) this._nodes[0].dispose();
 
+    // Disjoint windows have no rows to retain; replace them wholesale.
+    if (from >= prevTo || to <= this._offset) {
+      for (let i = 0; i < this._len; i++) this._nodes[i].dispose();
+      this._nodes = new Array(newLen);
+      this._mappings = new Array(newLen);
+      for (let i = 0; i < newLen; i++)
+        this._mappings[i] = runWithOwner<MappedItem>((this._nodes[i] = createOwner()), () =>
+          this._map(from + i)
+        );
+      this._offset = from;
+      this._len = newLen;
+      return;
+    }
+
     // clear the end
     for (let i = to; i < prevTo; i++) this._nodes[i - this._offset].dispose();
 
--- a/packages/solid-signals/tests/repeat.test.ts
+++ b/packages/solid-signals/tests/repeat.test.ts
@@ -346,3 +346,76 @@ it("slides the window backwards: disposes departing rows and retains the overlap
   // only the departing rows were disposed
   expect(disposed.sort()).toEqual([4, 5]);
 });
+
+describe("disjoint window jumps", () => {
+  function windowed() {
+    const [from, setFrom] = createSignal(0);
+    let created = 0;
+    let cleaned = 0;
+    let view!: () => number[];
+    createRoot(() => {
+      view = repeat(
+        () => 3,
+        i => {
+          created++;
+          onCleanup(() => cleaned++);
+          return i;
+        },
+        { from }
+      );
+    });
+    flush();
+    return {
+      view,
+      jump: (n: number) => {
+        setFrom(n);
+        flush();
+      },
+      live: () => created - cleaned
+    };
+  }
+
+  it("a forward jump larger than the window creates only the window's rows", () => {
+    const w = windowed();
+    expect(w.view()).toEqual([0, 1, 2]);
+    expect(w.live()).toBe(3);
+    w.jump(10);
+    expect(w.view()).toEqual([10, 11, 12]);
+    expect(w.live()).toBe(3);
+  });
+
+  it("repeated disjoint jumps do not accumulate live scopes", () => {
+    const w = windowed();
+    w.jump(10);
+    w.jump(20);
+    expect(w.view()).toEqual([20, 21, 22]);
+    expect(w.live()).toBe(3);
+  });
+
+  it("a backward disjoint jump maps the new window without crashing", () => {
+    const w = windowed();
+    w.jump(20);
+    expect(() => w.jump(0)).not.toThrow();
+    expect(w.view()).toEqual([0, 1, 2]);
+    expect(w.live()).toBe(3);
+  });
+
+  it("initial render with a nonzero `from` maps only the window", () => {
+    const [from] = createSignal(10);
+    let created = 0;
+    let view!: () => number[];
+    createRoot(() => {
+      view = repeat(
+        () => 3,
+        i => {
+          created++;
+          return i;
+        },
+        { from }
+      );
+    });
+    flush();
+    expect(view()).toEqual([10, 11, 12]);
+    expect(created).toBe(3);
+  });
+});


ryansolid:
Excellent analysis — the "shift + fill is only valid when old ∩ new ≠ ∅" framing pins all three symptoms to one modeling assumption, and the guard being the exact boolean complement of the overlap condition is what makes this safe: partial-overlap slides can't accidentally take the replacement path, and disjoint moves have no shared rows to preserve anyway, so wholesale replacement is the minimum possible work. Also appreciated the red-first tests — we re-verified all four fail on the unfixed source before landing.

Checked the interaction with the fallback-dispose line above the guard (fallback is removed before the wholesale path runs, so no double-dispose) and the adjacent-window boundary (`from === prevTo` — half-open windows share nothing, replacement is correct). Landed as-is, changeset included, credited as co-author: `4e81e9c5`. Ships in the next beta — thanks!

PR #2863, opened 2026-07-09
fix: handle runtime edge case regressions

brenelz:
Used the new GPT 5.6 Sol model to hunt bugs and this what it came up with. 

Feel free to take as many or as little as you want

## Summary

- reject actions correctly when generators throw falsy values and settle consumers of asynchronously empty iterables
- preserve native array index semantics and keyed store identity with leading null entries
- unlink disposed SSR owners before pooling so reused owners cannot be disposed through stale parent links
- add regression coverage for each corrected path

## Testing

- `pnpm test` in `packages/solid-signals` (900 tests)
- `pnpm test` in `packages/solid` (432 tests)
- `pnpm types` in `packages/solid-signals`
- `pnpm types` in `packages/solid`
- `pnpm test-types` in `packages/solid`

### DIFF
--- a/packages/solid-signals/src/core/action.ts
+++ b/packages/solid-signals/src/core/action.ts
@@ -89,13 +89,13 @@ export function action<Args extends any[], Y, R>(
       let ctx = activeTransition!;
       ctx._actions.push(it);
 
-      const done = (v?: R, e?: any) => {
+      const done = (v?: R, e?: any, failed = false) => {
         ctx = currentTransition(ctx);
         const i = ctx._actions.indexOf(it);
         if (i >= 0) ctx._actions.splice(i, 1);
         setActiveTransition(ctx);
         schedule();
-        e ? reject(e) : resolve(v!);
+        failed ? reject(e) : resolve(v!);
       };
 
       const step = (v?: any, err?: boolean): void => {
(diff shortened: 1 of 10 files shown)


ryansolid:
Thanks @brenelz — took you up on the à la carte offer. Here's the accounting:

**Landed (committed to `next`, authored to you):**
- **`action.ts` falsy generator throws** → bcb0ca6e. Confirmed red-first: `throw undefined` inside an action generator resolved the promise instead of rejecting. The explicit `failed` flag is the right shape — same bug class as #2866.
- **`store.ts` strict array-index detection** → fda28a90. Verified the repro: `store["01"] = x` on an empty array grew `length` to 2 via `parseInt`. Your canonical-index check (`Number.isInteger` + bounds + `String(index) === property`) matches the spec definition and subsumes the old symbol/`length` guards.
- **`server/signals.ts` owner pool unlink** → 25a5685b. This was the most serious of the four — a real use-after-recycle: the pool cleared `_parent`/`_nextSibling` on the pooled node but the parent's child chain still referenced it, so disposing the old parent could walk into whatever tree the recycled owner ended up in and dispose live owners. Your test proved it red-first. I ported it by hand (the diff had drifted against recent hydration changes) and verified subtree disposal stays O(1) per unlink and boundary retries (`self=false`) are unaffected.

**Superseded:**
- The `async.ts` empty-iterable settle landed via #2870, which also covers the sync-rejection case (#2868, #2869).

**Declined:**
- The `reconcile.ts` `hasKeyedItem` change. The leading-null inconsistency is real (mid-array nulls already survive keyed reconcile per the #2772 tests), but scanning `next` for any keyed item makes the keyed/positional decision O(n) for the common `reconcile(primitiveArray)` case where it's O(1) today, and it changes behavior for arrays mixing keyless objects among keyed ones (their undefined keys would collide in the index map). The failure mode it fixes is also mild — positional fallback still reconciles correct values, it only loses referential identity. Not worth the cost for `[null, ...items]`-shaped data; happy to revisit if it bites someone in practice.

Closing since everything here is either landed or intentionally passed on. Appreciate the hunting — 3 of 4 were genuine bugs with clean fixes.

Issue #2871, opened 2026-07-11
2.0.0-beta.17: streaming SSR/client asymmetry — a nested `<Loading>` joins the ancestor `<Reveal>` group and delays its direct slots


yumemi-thomas:
### Describe the bug

In Solid 2.0.0-beta.17 streaming SSR, a `<Loading>` boundary nested **inside** another `<Loading>` boundary's content is registered into the ancestor `<Reveal>` group as if it were a direct slot:

1. The repro has two direct slots under `<Reveal order="together">`: product details resolve at ~10ms and reviews at ~40ms. The product slot also contains a nested recommendations boundary that resolves at ~80ms and has its own fallback.
2. Under the client's direct-slot semantics, those two slots reveal together at ~40ms. The server instead generates a three-key group containing the two direct slots **and the nested boundary**, so both ready slots remain hidden until recommendations resolve at ~80ms.
3. The client behaves differently: a loading boundary clears the reveal context for its children, so the same nested boundary does not participate in the ancestor's group.

**Design question:** should a plain `<Loading>` nested inside a direct `<Loading>` child of `<Reveal>` participate in that ancestor's reveal group, or should the outer `<Loading>` isolate its subtree, as the client implementation currently does? SSR currently enrolls the nested boundary while the client does not, creating a server/client semantic asymmetry. This report uses the client/direct-slot behavior as the expected result, but the intended contract should be confirmed.

The same script first runs an otherwise-identical control without the nested recommendations boundary. That control produces the correct two-key group and reveals at ~40ms. Adding only the nested boundary changes the group to three keys and delays the activation until ~80ms.

This models a common product page: the main details and reviews reveal together, while a slower recommendations panel uses its own loading fallback. A slow descendant should not hold the entire page behind the outer fallbacks.

Related to but distinct from #2776 (rejected async boundary inside `<Reveal>` stalls sibling reveal during SSR streaming — closed, fixed in beta.17 by 4608539f): that issue was about error handling within a group; this one is about *group membership* — non-direct descendants being enrolled at all.

### Your Example Website or App

- [Server reproduction](https://stackblitz.com/edit/solidjs-templates-pabkbshg?file=src%2Frepro.tsx): the nested `<Loading>` joins the ancestor `<Reveal>` group.
- [Client comparison](https://stackblitz.com/edit/shzx7jmx?file=src%2FApp.tsx): the client isolates the nested `<Loading>` from the ancestor `<Reveal>` group.

[...]

PR #2874, opened 2026-07-11
fix(server): sever RevealGroupContext in createErrorBoundary to match client behavior

### LINKED ISSUE #2872: 2.0.0-beta.17: streaming SSR — an `<Errored>`-wrapped `<Loading>` joins the ancestor `<Reveal>` group on the server, but not on the client
### Describe the bug

Put an `<Errored>` between a `<Reveal>` and one of its `<Loading>` slots:

```tsx
<Reveal order="together">
  <Loading fallback={<b>loading order…</b>}>
    <OrderSummary />       {/* resolves at ~300ms */}
  </Loading>
  <Errored fallback={<i>payment panel unavailable</i>}>
    <Loading fallback={<b>loading payment options…</b>}>
      <PaymentOptions />   {/* flaky third-party, resolves at ~2000ms */}
    </Loading>
  </Errored>
</Reveal>
```

The client and the server disagree about whether the wrapped `<Loading>` is part of the reveal group:

- **Client: it is not.** Every boundary — `Loading` *and* `Errored` — clears the reveal context for its children (`createCollectionBoundary` in `packages/solid-signals/src/boundaries.ts`). The group contains only the order-summary slot: it reveals at ~300ms, the panel independently at ~2000ms.
- **Server: it is.** `createErrorBoundary` (`packages/solid/src/server/signals.ts`) does not clear `RevealGroupContext`, so the wrapped `<Loading>` registers into the group. The stream emits a two-key `$dfj` only at ~2000ms — the order summary is held on its skeleton the whole time.

Same tree, different reveal timing depending on how the user arrives: hard refresh holds everything to ~2s, client-side navigation shows the summary at ~300ms.

Sibling of #2871: the same client/server severing asymmetry, but through `<Errored>` instead of `<Loading>`. Fixing #2871 (`createLoadingBoundary`) does not cover this case.

**TanStack Start × Solid 2.0 StackBlitz:** <https://stackblitz.com/edit/1snzezwb?file=src%2Froutes%2Findex.tsx>

The checkout page above in a real `@tanstack/solid-start@2.0.0-beta.24` app (`solid-js@2.0.0-beta.17`). `npm start` runs automatically: it streams a control route and the repro route, prints a PASS/FAIL verdict in the **terminal**, and leaves the dev server running so both behaviors are visible in the preview.

1. Open the StackBlitz and wait for the terminal verdict:

```text
CONTROL /control (no Errored)
  groups: [["…10","…11"]] — two direct slots coordinating is correct
  PASS

REPRO / (Errored-wrapped panel)
  groups: [["…10","…1100"]]; activation at ~2112ms; panel content at ~2112ms
    — wrapped panel was enrolled into the group
  FAIL
```

2. `/control` (no `Errored`) coordinates its two direct slots at ~2s — correct.
3. `/` enrolls the `Errored`-wrapped panel as a group member and holds activation to ~2s. Client semantics require a one-key group at ~300ms with the panel activating independently.
4. In the preview: SPA-navigate `/control` → `/` — the order summary reveals at ~300ms, the panel trails independently. Hard-refresh `/` — everything is held on skeletons to ~2s.

### Expected behavior

The server matches the client: a `<Loading>` separated from `<Reveal>` by an `<Errored>` does not register into the reveal group. The order summary reveals at ~300ms and the panel independently at ~2000ms in both rendering modes.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome (StackBlitz preview); server render under Node 20
- Version: 2.0.0-beta.17 (`next` @ `a51cac19`)

### Additional context

**The client behavior is deliberate, not incidental.** Two client tests pin it, both specifically about error handling in reveal order:

- `createRevealOrder › advances reveal order when first slot errors inside error boundary`
- `createRevealOrder › nested composite recovers from error without breaking outer progression`

They encode this design: an error-boundary-wrapped slot is intentionally **outside** the group, so that if it errors, its error fallback can appear without stalling the group's progression. (Verified: restricting the client severing to loading boundaries only breaks exactly these two tests and nothing else.) It is the client's structural answer to the "erroring member stalls the group" problem — the same problem that was patched separately on the server for direct slots in #2776.

**Suggested fix:** port the severing to the server — clear `RevealGroupContext` on the scope that `createErrorBoundary`'s children render under, the same pattern as the #2871 fix direction for loading boundaries. Two practical notes:

1. **Module layout:** `RevealGroupContext` is defined in `server/hydration.ts`, which imports *from* `server/signals.ts` where `createErrorBoundary` lives. The context (or a severing helper) has to move to a shared module first.
2. **Sequencing:** this should land **after** the #2871 runtime fix — once the wrapped boundary activates independently, one that settles while the group is still held needs the deferred `$df` activation queue or its content is silently dropped.

Since the resulting rule — *wrapping a slot in `<Errored>` opts it out of reveal coordination* — is intended but currently undocumented, a sentence in the `<Reveal>` docs would prevent surprises.

**Not applicable — new 2.0 API.** `<Reveal>`/reveal groups have no Solid 1.x counterpart.

AmariahAK:
## Description

```markdown
## Summary

The server-side `createErrorBoundary` did not sever `RevealGroupContext` for its
children, unlike the client-side `createCollectionBoundary` which clears the
reveal controller context for all boundary types (both Loading and Errored).

This caused an `<Errored>`-wrapped `<Loading>` to incorrectly register with an
ancestor `<Reveal>` group on the server, holding the entire group to the slowest
slot on hard refresh while SPA navigation revealed independently.

**Fix:** Moved `RevealGroupContext` from `server/hydration.ts` to
`server/signals.ts` (resolving a circular dependency), then added
`setContext(RevealGroupContext, null, owner)` in `createErrorBoundary` — matching
the client's pattern at `boundaries.ts:386`.

Closes #2872.

```bash
# Server-side reveal tests (27 passed, including the new test)
pnpm build && cd packages/solid && npx vitest run test/server/reveal-ssr.spec.ts

# Client-side reveal order tests (29 passed — no regressions)
cd packages/solid-signals && npx vitest run tests/createRevealOrder.test.ts

# Full server test suite (173 passed)
cd packages/solid && npx vitest run test/server/
```

New test added: "Errored-wrapped Loading does not join ancestor Reveal group" in
`reveal-ssr.spec.ts` — verifies that only the direct Loading registers in the
group and the Errored-wrapped one does not.

### DIFF
--- a/packages/solid/src/server/hydration.ts
+++ b/packages/solid/src/server/hydration.ts
@@ -8,7 +8,8 @@ import {
   ErrorContext,
   getContext,
   setContext,
-  runWithBoundaryErrorContext
+  runWithBoundaryErrorContext,
+  RevealGroupContext
 } from "./signals.js";
 import { sharedConfig, NoHydrateContext } from "./shared.js";
 import type { SSRTemplateObject, HydrationContext } from "./shared.js";
@@ -20,33 +21,8 @@ export type { HydrationContext, SSRTemplateObject } from "./shared.js";
 
 // --- Reveal SSR coordination ---
 
-export type ServerRevealGroup = {
-  id: string;
-  /**
-   * Register a child fragment (Loading) or composite child (inner Reveal).
-   * Returns `collapseFallback` (hide fallback visually, used for collapsed-sequential
-   * tail) and `held` (stash `revealFragments` swaps until the parent releases us).
-   * `held` only applies when the caller is a nested Reveal — Loadings ignore it.
-   */
-  register(
-    key: string,
-    options?: { onActivate?: () => void }
-  ): { collapseFallback: boolean; held: boolean };
-  /** Called by a child when its subtree is fully resolved. */
-  onResolved(key: string): void;
-  /**
-   * Called by a nested Reveal when it becomes "minimally resolved" under its own
-   * order (together: fully resolved; sequential: first registered fragment resolved;
-   * natural: any fragment resolved). Loadings don't fire this — their `onResolved`
-   * implies minimal readiness at the same time.
-   */
-  onMinimallyResolved?(key: string): void;
-};
-
-export const RevealGroupContext: Context<ServerRevealGroup | null> = {
-  id: Symbol("RevealGroupContext"),
-  defaultValue: null
-};
+export type { ServerRevealGroup } from "./signals.js";
+export { RevealGroupContext } from "./signals.js";
 
 /**
  * Handles errors during SSR rendering.
--- a/packages/solid/src/server/signals.ts
+++ b/packages/solid/src/server/signals.ts
@@ -1608,6 +1608,21 @@ const ErrorContext: Context<((err: any) => void) | null> = {
 };
 
 export { ErrorContext };
+
+export type ServerRevealGroup = {
+  id: string;
+  register(
+    key: string,
+    options?: { onActivate?: () => void }
+  ): { collapseFallback: boolean; held: boolean };
+  onResolved(key: string): void;
+  onMinimallyResolved?(key: string): void;
+};
+
+export const RevealGroupContext: Context<ServerRevealGroup | null> = {
+  id: Symbol("RevealGroupContext"),
+  defaultValue: null
+};
 export function runWithBoundaryErrorContext<T>(
   owner: Owner,
   render: () => T,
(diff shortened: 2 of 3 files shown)


ryansolid:
Thanks for this — the direction was right, and this landed as part of 1561c7ed on `next`, which severs `RevealGroupContext` in both `createErrorBoundary` and `ssrLoadingBoundary` (covering #2871 as well). It depends on the deferred-activation queue added to @dom-expressions/runtime 0.50.0-next.20, so the swaps a severed boundary streams while its enclosing slot is still held are queued client-side instead of dropped. Closing as superseded.

PR #2906, opened 2026-07-17
ci probe: CodSpeed zero-diff baseline check (do not merge)

ryansolid:
Empty commit off `next`. #2905 shows two −6.8% regressions that are byte-identical across three different HEAD runs and carry CodSpeed's *different-runtime-environments* warning; a hot-path fix moved the numbers by exactly zero. If this zero-diff PR reproduces the same regressions, the base measurement is environmentally skewed and #2905's flags are phantom. Will close after the check reports either way.


ryansolid:
Probe result: clean (121/121 untouched) — baseline is healthy, #2905's regressions are real code effects. Proceeding with single-patch attribution probes.

PR #2908, opened 2026-07-17
ci probe: tier-2 config-brand in isolation (do not merge)

ryansolid:
Single-patch attribution probe for #2905's CodSpeed regressions. Zero-diff probe (#2906) came back clean, so the regressions are real; this measures the `config-brand` consolidation alone. Will close after CodSpeed reports.

### DIFF
--- a/packages/solid-signals/src/core/async.ts
+++ b/packages/solid-signals/src/core/async.ts
@@ -1,4 +1,5 @@
 import {
+  CONFIG_OPTIMISTIC,
   CONFIG_SYNC,
   EFFECT_TRACKED,
   EFFECT_USER,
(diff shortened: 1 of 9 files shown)


ryansolid:
Probe served its purpose: **guilty** — config-brand alone reproduces merge −7% and carries sub-threshold −3..4% on update1to1/propagation (dashboard), making it the driver of both #2905 regressions. The change is load-neutral by inspection and byte-flat in an isolated merge harness; the diffuse cross-bench signature points at shared-process JIT/IC state redistribution from the hasOverride shape change. Pulled from #2905; to be redesigned shape-stably (dedicated boolean in the node creation literal rather than a `_config` RMW).

PR #2909, opened 2026-07-17
ci probe: tier-2 ambient-txn in isolation (do not merge)

ryansolid:
Single-patch attribution probe for #2905's CodSpeed regressions. Zero-diff probe (#2906) came back clean, so the regressions are real; this measures the `ambient-txn` consolidation alone. Will close after CodSpeed reports.


ryansolid:
Probe served its purpose: this patch measured CLEAN in isolation (CodSpeed gate passed). Findings folded into #2905.

Issue #2919, opened 2026-07-19
Streamed Loading boundary duplicates content after Solid Refresh HMR


brenelz:
## Describe the bug

With Solid 2 whole-document streamed SSR, a `<Loading>` boundary can append duplicate resolved content after a Solid Refresh HMR update instead of replacing the previous content.

The page initially renders `Hello World2`. After reloading once and making repeated text-only edits inside the boundary, the DOM can end up containing output such as:

```text
Hello World2Hello World2333
```

The duplication only appears after a full page reload followed by HMR. A cold first navigation does not reliably reproduce it, which suggests the cached async client entry allows hydration to begin against the streamed fallback before the server reveal arrives.

This looks related to #2801 bug 1, but that regression coverage exercises a reactive refresh after streamed hydration rather than a component replacement through Solid Refresh HMR.

## Your Example Website or App

https://github.com/brenelz/solid-2/tree/b8c94efa

Relevant files:

- [`src/App.tsx`](https://github.com/brenelz/solid-2/blob/b8c94efa/src/App.tsx)
- [`src/entry-server.tsx`](https://github.com/brenelz/solid-2/blob/b8c94efa/src/entry-server.tsx)
- [`src/entry-client.tsx`](https://github.com/brenelz/solid-2/blob/b8c94efa/src/entry-client.tsx)
- [`src/HtmlDocument.tsx`](https://github.com/brenelz/solid-2/blob/b8c94efa/src/HtmlDocument.tsx)

Versions are pinned in the reproduction:

- `solid-js@2.0.0-beta.20`
- `@solidjs/web@2.0.0-beta.20`
- `vite-plugin-solid@3.0.0-next.13`
- `vite@8.0.10`
- `nitro@3.0.260610-beta`

## Steps to Reproduce the Bug or Issue

1. Clone the reproduction and check out `b8c94efa`.
2. Run `pnpm install`.
3. Run `pnpm dev`.
4. Open `http://localhost:3000` in Chrome and wait for `Hello World2` to resolve.
5. Reload the page once and wait for `Hello World2` again.
6. In `src/App.tsx`, change the boundary child from `{asyncValue()}` to `{asyncValue()}111` and save.
7. Wait for the HMR-triggered async operation to settle.
8. Change `111` to `222`, save, and wait for it to settle.
9. Repeat with `333` if needed.
10. Observe that resolved values can accumulate instead of replacing the previous DOM.

[...]

PR #2921, opened 2026-07-19
fix(solid): avoid hydrating refresh registration

### LINKED ISSUE #2920: Lazy component hydration crashes in Solid Refresh module registration
## Describe the bug

Hydrating a server-rendered `lazy()` component crashes in development when Solid Refresh wraps the component exported by the dynamically imported module.

The lazy module is evaluated while hydration is active. Its generated top-level `$$component(...)` registration calls `createSignal` in `solid-js/dist/refresh.dev.js:90`. The signal enters `hydratedCreateSignal`, but module evaluation has no reactive owner, so `peekNextChildId()` eventually reads `_config` from `null`.

Solid catches the failed hydration module preload and falls back to client rendering.

## Reproduction

https://github.com/brenelz/solid-2/tree/9d295727496d194b53964fb1b8d8a86a46c895b4

The reproduction currently includes the `/* @refresh reload */` workaround. To reproduce:

1. Clone the repository and check out `9d295727496d194b53964fb1b8d8a86a46c895b4`.
2. Remove `/* @refresh reload */` from `src/routes/Index.tsx`.
3. Run `pnpm install`.
4. Run `pnpm dev`.
5. Open `http://localhost:3000/`.
6. Observe the hydration failure in the browser console.

The relevant route declaration is:

```tsx
const Index = lazy(() => import('./routes/Index.tsx'))

<Route path="/" component={Index} />
```

The development transform evaluates this at module scope in the lazy chunk:

```js
const Index = _$$component(_REGISTRY, "Index", function Index() {
  // ...
})
```

## Actual behavior

```text
Hydration module preload failed, falling back to client render: TypeError: Cannot read properties of null (reading '_config')
    at childId (dev.js:1354:18)
    at peekNextChildId (dev.js:1386:10)
    at hydrateSignalFromAsyncIterable (dev.js:289:22)
    at hydratedCreateSignal (dev.js:422:20)
    at createSignal (dev.js:612:53)
    at $$component (refresh.dev.js:90:27)
    at Index.tsx:4:16
```

`refresh.dev.js:90` is the Refresh registration signal:

```js
const [comp, setComp] = createSignal(() => current);
```

## Expected behavior

A server-rendered lazy component should hydrate in development with Solid Refresh enabled. Refresh's module-level bookkeeping signals should not consume hydration IDs or require a hydration owner.

## Workarounds

Either workaround prevents the crash:

```tsx
/* @refresh reload */
```

on every lazy route module, or globally:

```ts
solid({
  ssr: true,
  refresh: { disabled: true },
})
```

Switching `vite-plugin-solid` to `compiler: 'babel'` does not help because the Refresh transform still emits the same `$$component(...)` wrapper.

## Versions

- `solid-js@2.0.0-beta.20`
- `@solidjs/web@2.0.0-beta.20`
- `vite-plugin-solid@3.0.0-next.13`
- `@solidjs/router@0.17.0-next.5`
- `vite@8.0.10`
- `nitro@3.0.260610-beta`
- macOS, Chrome

## Additional context

This is related to hydration plus Solid Refresh, but distinct from #2919, which reports duplicate streamed boundary content after an HMR replacement. This issue occurs on the initial hydration as soon as a Refresh-wrapped lazy module is imported.

brenelz:
## Summary
- store Solid Refresh component pointers in a plain signal instead of a function-valued writable memo
- keep module-level refresh registration out of hydration serialization and child ID allocation
- cover ownerless registration and hydration ID preservation

## Tests
- `pnpm test` in `packages/solid` (488 passed)
- `pnpm test-types` in `packages/solid`
- Prettier check and `git diff --check`

Fixes #2920

### DIFF
--- a/.changeset/fix-refresh-lazy-hydration.md
+++ b/.changeset/fix-refresh-lazy-hydration.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Fix hydration crashing when a lazy component module is evaluated with Solid Refresh enabled (#2920). Refresh component registrations now use plain signals so module-level bookkeeping does not require a reactive owner or consume hydration child IDs.
--- a/packages/solid/src/refresh/index.ts
+++ b/packages/solid/src/refresh/index.ts
@@ -176,16 +176,15 @@ export function $$component<P extends Record<string, any>>(
 ): (props: P) => SolidElement {
   if (!IS_DEV) return component;
   let current = component;
-  // A writable memo: `createSignal(fn)` in 2.0 computes its value from `fn`
-  // on first read and accepts updater writes. `current` mirrors the latest
-  // component so a recompute (which has no dependencies) stays consistent.
-  const [comp, setComp] = createSignal<(props: P) => SolidElement>(() => current);
+  // Keep the component inside a plain signal so module-level registration is
+  // never treated as a hydration-aware computation.
+  const [comp, setComp] = createSignal({ current });
   const update = (action: () => (props: P) => SolidElement): void => {
     current = action();
-    setComp(() => current);
+    setComp({ current });
   };
   const instances: LiveInstances = { count: 0 };
-  const proxy = createProxy(comp, id, instances, options.location);
+  const proxy = createProxy(() => comp().current, id, instances, options.location);
   registry.components.set(id, {
     id,
     component,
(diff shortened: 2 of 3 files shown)


ryansolid:
Thanks @brenelz! A fix for #2920 was already in flight locally, but your plain-signal approach is cleaner — no writable memo, nothing hydration-aware to work around — so I adopted it. It just landed on `next` with your co-author credit: b1b2f82fa7406274d03d0d438d68e167e15ea500. Your "does not consume a hydration child id" test came along too. The related hydration-timing fix for #2919 landed alongside it. Both will be in the next beta.

PR #2923, opened 2026-07-20
refactor(signals): affects() dedicated channel + delete stashedOptimisticReads (arch audit tier-1)

ryansolid:
The two Tier-1 architecture consolidations that survived design-spike measurement (spikes 1 and 3 were declined — pull-verdicts is spec-pinned by the #2831/lane-held wake schedule and measured size-neutral; store node-authority costs +585 min). Each of these retires a *mechanism family*, not a line.

| | Δ full (min/gz) | Δ floor (min/gz) |
|---|---|---|
| Combined | **−775 / −304** | **−388 / −169** |

## affects() off the async status rails

A mark becomes a refcount on the marked node(s) plus a pull-computed coverage walk at verdict time; Loading interaction survives as a boundary-only *visual* notification. Because marks never touch `_statusFlags`/`_error`/`_pendingSources`/`_asyncReporters`, the five carve-out families that existed only to un-teach the async rails everything a mark does — value-transparency, transaction-inertness, error-inferiority, settle-release, quiet-reask immunity — are **deleted as unrepresentable** rather than maintained. The per-read affects collection and per-recompute mark bookkeeping also leave the core `read`/`recompute` hot paths (a small simplification of the paths Sierpinski exercises).

Designer rulings encoded as the channel's structural contract: marks are invisible to all completion/settlement accounting *by construction* — a mark can never block or legitimize ending its own transaction (the action holding the window is the real work); ambient marks (no transaction) are verdict-only and release at flush end; a mark serializes only inside a resumable transaction. Hydration capture now takes marked-covered nodes' real committed values (they carry no status), consistent with value-transparency.

## delete `stashedOptimisticReads`

It forced a committed-view effect re-run when a transaction backgrounded — masking an active override from stale tracked reads (against A17's "do not mask the override from any read path") and un-rendering co-written optimistic flags mid-window — while engaging **zero times across all prior tests**. The queue stash and `_gatedSubs` replay are retained: they cover transaction-held reveal and silent-revert re-ask respectively, which are irreducibly distinct reveal *actions*; a spec comment records why they cannot physically merge (the design-spike negative result).

## Tests

1077 pass (1073 prior, unchanged, + 4 new pins):
- **S3** — the backgrounded-transaction reveal shape: co-written optimistic flag stays rendered, user effects fire once.
- **S4** — a transaction-held effect must not leak during an interleaved ambient flush. This was a **pre-existing coverage hole** (the suite never interleaved ambient work with a stash); the queue stash is load-bearing and was previously untested.
- **isolation** — deterministic recompute-count guards for the propagation/Sierpinski surgical-update property (the isolation CodSpeed's `propagation:*` benches measure by time; this asserts it by count, noise-free).

CodSpeed is the arbiter on hot-path shape — the coverage walk is O(upstream reachability) per marked-node verdict (was O(pendingSources)); flagged for the perf run. Every prior spike round taught that local wall-clock can't resolve these deltas, so the benches decide.

### DIFF
--- a/.changeset/arch-tier1-affects-reveal.md
+++ b/.changeset/arch-tier1-affects-reveal.md
@@ -1,0 +1,11 @@
+---
+"@solidjs/signals": patch
+---
+
+Two architecture consolidations from the growth audit's Tier-1 spikes, each retiring a mechanism family rather than shrinking one.
+
+`affects()` marks move off the async status rails onto a dedicated verdict channel. A mark is now only a refcount on the marked node(s); derived coverage is pull-computed at verdict time by a dependency-graph walk, and Loading interaction survives as a boundary-only visual notification. Because marks never touch `_statusFlags`/`_error`/`_pendingSources`/`_asyncReporters`, the five carve-out families that previously un-taught the async rails everything a mark does — value-transparency, transaction-inertness, error-inferiority, settle-release, quiet-reask immunity — are deleted as unrepresentable rather than maintained. The per-read affects collection and per-recompute mark bookkeeping also leave the core read/recompute hot paths. Marks stay invisible to all completion and settlement accounting by construction (a mark can never block or end its own transaction), ambient marks are verdict-only and release at flush end, and mark state serializes only inside a resumable transaction.
+
+`stashedOptimisticReads` is deleted. It forced a committed-view effect re-run when a transaction backgrounded — masking an active override from stale tracked reads (against A17) and un-rendering co-written optimistic flags mid-window (against the affordance idiom) — while engaging zero times across the whole suite. The queue stash and `_gatedSubs` replay are retained (they cover transaction-held reveal and silent-revert re-ask, which are irreducibly distinct); a spec comment records why they cannot merge.
+
+Full bundle −775 min / −304 gzip bytes; core floor −388 / −169. All prior tests pass unchanged plus new contract pins: the backgrounded-transaction reveal shape (S3), the interleaved-ambient-flush queue-stash guard (S4, a pre-existing coverage hole), and a deterministic recompute-count isolation guard for the propagation/Sierpinski surgical-update property. No public API change.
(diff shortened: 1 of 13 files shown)


ryansolid:
## CodSpeed: net strongly positive; the one regression is a shared-VM artifact

**⚡ 3 improved · ❌ 1 regressed · ✅ 117 untouched**

| Benchmark | BASE → HEAD | |
|---|---|---|
| `merge-merge-static(both 5,1,3,2)` | 165.2 → 73.3 µs | **×2.3** |
| `reconcile: deep tree, 10 of ~12k paths` | 6.4 → 3.3 ms | **+92%** |
| `updateSignals:update1to1` | 54.4 → 51.5 ms | **+5.6%** |
| `merge-mixed(0,100,3,2)` | 266.5 → 369.6 µs | −27.9% |

### Attribution (two isolation probe PRs, now closed)

- **stashedOptimisticReads deletion alone** → clean, gate passed, 1 improvement.
- **affects-channel alone** → reproduces the merge pair exactly: `merge-mixed −27.9%` **and** `merge-static ×2.3`. They are one coupled effect of simplifying `read()`'s shape.

### Why merge-mixed is a measurement artifact, not a cost

`merge-mixed` builds objects whose properties alternate between a plain value descriptor (`{value: 1}`) and a **plain-JS getter** (`{ get() { return 1 } }` — no signal, no Solid primitive). It exercises `merge`'s property-descriptor resolution over getter/value polymorphism; **it never calls `read()`, `link()`, or any code this diff touches.** So the −28% cannot be a per-operation cost from the change — it is V8 retiering the shared `merge` bench file's megamorphic descriptor-dispatch site because `read()` (called by *other* benches in the same process) changed shape. `merge-static ×2.3` is the coupled other half of the same recompilation; the two always move together and neither reflects a real cost in the merged code.

Acknowledging `merge-mixed` on that basis. The real ledger — a 2.3× and a 92% improvement on paths the diff *does* touch, against zero reachable regressions — plus the architecture win (five affects carve-out families deleted as unrepresentable; the affects fix-cluster retired) is the reason to land.

PR #2924, opened 2026-07-20
ci probe: tier1-spike2 isolation (do not merge)

ryansolid:
Attribution probe for #2923's CodSpeed result (merge-mixed −28%, alongside merge-static ×2.2 / reconcile +92% / update1to1 +5.5% improvements). Isolates one spike. Closes after CodSpeed reports.

### DIFF
--- a/packages/solid-signals/src/affects.ts
+++ b/packages/solid-signals/src/affects.ts
@@ -1,10 +1,4 @@
-import {
-  addPendingSource,
-  forEachDependent,
-  notifyStatus,
-  setPendingError,
-  settlePendingSource
-} from "./core/async.js";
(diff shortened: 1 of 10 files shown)


ryansolid:
Probe result: reproduces #2923's merge deltas exactly in isolation — merge-mixed −27.9% AND merge-static ×2.3, the coupled pair. Confirms both come solely from the affects-channel read()-shape change. Since merge-mixed is a plain-JS getter/value descriptor bench (getter body `return 1`, no Solid primitive) that never calls read(), the −28% is shared-VM JIT redistribution, not a per-op cost. Findings folded into #2923.

PR #2931, opened 2026-07-23
Store perf: structural hot-path pass, shallow stores + markRaw, effect-run trims

ryansolid:
Four commits from the store/reconcile performance effort, each independently measured on a signals-only dbmon-shaped micro-bench (1000 rows × 13 bindings, fresh keyed payload per tick) and validated on the octane benchmark harness.

## Commits

1. **Store structural hot-path pass** — `storeLookup` maps raw → target (reconcile recurses target-to-target, no wrap/`$PROXY`/`$TARGET` round-trips; lookup miss ⇒ never-observed subtree skipped, making the diff O(observed)); trap fast path returns node values without re-wrapping (dev assertion enforces the wrap-at-write invariant); keyed arrays skip identity-equal slots and exit fully-matched passes before staging; allocation-free node iteration; effect values direct-commit on plain sync flushes. **Deep-store dbmon tick 15.1 → 10.9ms** (now faster than solid-js 1.9 on every op).

2. **Shallow stores + markRaw** — `createStore(value, { shallow: true })`: root keys fully reactive, values raw records replaced by reference; positional `reconcile`; write-scope reads serve raws (replacement-only contract); optimistic staging/revert through the existing override layers with the base untouched. `markRaw(value)`: sticky raw registry consulted only on wrap-creation/ingest — a value serves raw in every store (single identity). Plumbed through `createProjection`/`createOptimisticStore`. **Micro tick 5.95 → 2.61ms (reconcile half 14×); octane dbmon 2.6× → 1.74× (1.36× with `textContent` bindings), all keyed lifecycle gates passing, ahead of React on every op.**

3. **Effect-run trims** — three no-op gates (clearStatus on status-free nodes, insertSubs on subscriber-less nodes, syncCompanions on companion-less writes) + `readNodeFast` (read()'s plain-signal fast path hoisted over the trap's node read). Individually A/B-measured; readNodeFast is a consistent ~2–3% on store-read-heavy flushes.

4. **Shallow hardening** (adversarial audit follow-up) — raw-marked values are leaves in every reconcile recursion path (three verified blockers shared that root cause); setter-staged overrides fold into the shallow diff; `markRaw` exposed from solid-js (client + server stub, export parity green); plain-form `createStore`/`createOptimisticStore` accept options; new CodSpeed bench `reconcile-dbmon.bench.ts` covering deep and shallow reconcile lanes.

## Numbers

- Micro-bench (reactive layer only): deep store 7.14 → 5.98ms/tick, shallow 2.51ms, plain-signal ceiling 1.15ms.
- Octane harness dbmon (all gates): 2.64× → 1.37× octane deep / **1.36× with shallow + textContent** — ahead of React on every op.
- Full 15-suite octane table geomean: 1.110× → **0.863×**.
- Size: +2.8KB gz total on the prod store payload (structural pass ± 0, shallow + markRaw + hardening +2.8; shallow tree-shakes out of apps that don't use it — the octane bundle-size fixture got smaller, 14.7 → 14.1KB gz).
- Tests: 1,118 signals / 492 solid / 544 web green; 18-test shallow suite covering the audit's blocker scenarios.

Purpose of this PR per @ryansolid: let CodSpeed sweep for regressions before merge.


ryansolid:
CI note: the initial red was inherited from `next` — `8a4b42dd` broke `types:web-frames` on fresh installs (nested `frames/package.json` names itself `@solidjs/web/frames`, so NodeNext self-reference doesn't cover the `@solidjs/web/server-functions/client` import, and fresh installs lack the `link`-script symlink dev checkouts have). `c8422492` maps the specifier via tsconfig `paths` — reproduced and verified on a fresh clone (12/12 turbo tasks). **`next` needs the same one-liner** or its own CI stays red independent of this PR.

ryansolid:
Second CI item: the `signals: + createStore` size budget tripped — **+728 B brotli over the 11.6 KB limit**. Cause: the `options.shallow` branch keeps `wrapShallow`/`applyStateShallow` in the plain-createStore graph under tree-shaking, plus reconcile's raw-leaf handling. `b15fb562` bumps that one scenario to 12.5 KB with reasoning inline — **explicitly your trade-off to adjudicate**: accept the cost, demand deeper golf, or ask for a separate tree-shakeable `createShallowStore` entry instead of the option. All other scenarios pass with room (core floor 6.77/7.1, isPending 8.39/8.75, app floor 9.43/10, CSR 11.55/12).

PR #2946, opened 2026-07-28
fix(signals): guard reconcile's array paths against array/object shape changes

brenelz:
## The bug

`reconcile()`'s object diff refuses to recurse into a pair whose container kinds disagree — `Array.isArray(previous) !== Array.isArray(next)` replaces the slot instead of merging into it. The **array** paths reach the same recursion through `keyedMatch` and positional pairing, and never applied that rule: two keyless wrappables "match" because `keyFn` returns `undefined` for both, regardless of whether they're arrays or objects.

So an incoming array gets merged into a proxy whose target is an object. `Array.isArray` on a Proxy inspects the *target*, which is fixed at wrap time, so the slot is wrong forever after:

```js
const [state, setState] = createStore({ list: [{ x: 1 }] });
setState(reconcile({ list: [[10, 20]] }, "id"));

Array.isArray(state.list[0]);       // false
JSON.stringify(state.list[0]);      // {"0":10,"1":20}
```

Spread, `.map`, `<For each>` and `JSON.stringify` all see an object. The reverse direction is worse — an array-shaped target receiving an object reads `length` off that object and the store presents `[]`, silently dropping the data.

## The fix

Five call sites shared the gap: the keyed prefix loop and the positional loop in both `applyStateFast` and `applyStateSlow`, plus `applyArrayItem` (which covers the keyed diff's trailing and moved slots). They now share a `recursablePair()` helper that folds the existing `markRaw` check together with the container-kind check, so a kind change replaces the slot by reference — exactly what the object diff and `descendInto` already do.

Net effect on the hot paths is neutral: `recursablePair` subsumes checks those sites already performed inline.

## Tests

Four cases added to `tests/store/reconcile.test.ts`, covering both directions, the keyed-array trailing slot (`applyArrayItem`), and positional `key: null` merging. All four fail on `next` and pass with the fix.

Found by differential fuzzing (random trees reconciled against a plain-object oracle, checking both direct reads and effect-observed values). Full suite green: 19/19 turbo tasks, 1150 `@solidjs/signals` tests.

### DIFF
--- a/.changeset/reconcile-array-object-shape-guard.md
+++ b/.changeset/reconcile-array-object-shape-guard.md
@@ -1,0 +1,18 @@
+---
+"@solidjs/signals": patch
+---
+
+Stop `reconcile` from merging an array into an object slot (and vice versa) inside arrays
+
+The object diff has always refused to recurse into a pair whose kinds disagree — `Array.isArray(previous) !== Array.isArray(next)` replaces the slot instead of merging. The array paths reach the same recursion through `keyedMatch` and positional pairing, and neither applied that rule: two keyless wrappables "match" because `keyFn` returns `undefined` for both, regardless of whether they are arrays or objects.
+
+The result was a slot whose proxy is permanently the wrong kind. `Array.isArray` on a store proxy inspects the proxy's target, which is fixed at wrap time, so after
+
+```js
+const [state, setState] = createStore({ list: [{ x: 1 }] });
+setState(reconcile({ list: [[10, 20]] }, "id"));
+```
+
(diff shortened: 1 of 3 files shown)


ryansolid:
Merged — verified the diagnosis independently: `descendInto` (reconcile.ts:275) has always applied the container-kind rule, and `keyedMatch` really does let two keyless wrappables pair across kinds. Also confirmed both rewritten call sites are reached only with both sides wrappable (the `keyedMatch` loop guarantees it), so folding the checks into `recursablePair` is behaviour-preserving apart from the intended kind guard. All four new tests fail on `next` and pass with the fix; full suite and CodSpeed green.

PR #2947, opened 2026-07-28
fix(signals): preserve sibling queues during traversal

brenelz:
## The bug

`Queue.run()` traverses child queues with a monotonically increasing array index. A boundary effect can dispose its owner while its queue is running, which removes that queue from `_children`. The next sibling then shifts into the current index and is skipped, leaving its already-queued effects unexecuted with no later flush scheduled.

## The fix

Advance the traversal index only while the child that just ran still occupies that slot. If it removed itself or an earlier sibling, process the child that shifted into the current index before advancing.

A regression test creates two loading-boundary queues, disposes the first from its effect, and verifies that the second effect still runs in the same flush.

## Tests

- `vitest run`: 78 files, 1,147 tests passed
- `tsc -p tsconfig.build.json --noEmit`
- Prettier check for the changed source and test
- Changeset status validates an `@solidjs/signals` patch release

### DIFF
--- a/.changeset/fix-scheduler-child-queue-removal.md
+++ b/.changeset/fix-scheduler-child-queue-removal.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Keep traversing sibling effect queues when a child queue removes itself or an earlier sibling during a flush. Previously, the shifted sibling could be skipped and leave its queued effects unexecuted.
--- a/packages/solid-signals/src/core/scheduler.ts
+++ b/packages/solid-signals/src/core/scheduler.ts
@@ -295,7 +295,12 @@ export class Queue implements IQueue {
       this._queues[type - 1] = [];
       runQueue(effects, type);
     }
-    for (let i = 0; i < this._children.length; i++) (this._children[i] as any).run?.(type);
+    for (let i = 0; i < this._children.length; ) {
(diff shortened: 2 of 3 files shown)


ryansolid:
Merged — thanks, the diagnosis was right and the regression test is a good one.

One thing worth flagging for the record: the fix as written recovers the cursor for **one** removal, but a single `dispose()` can remove **several** sibling queues at once — a root owning more than one boundary removes them all from the same parent's `_children`. The list then shifts by more than a slot and the cursor lands past an unprocessed later sibling, which is the same symptom the PR fixes:

```
[Q_a1, Q_a2, Q_b]   // Q_a1 and Q_a2 belong to root A, Q_b to root B
                    // an effect in Q_a2 disposes root A -> [Q_b]
                    // cursor was at 1; _children[1] is now undefined, so it
                    // doesn't advance -> length is 1 -> Q_b never runs
```

I've landed a follow-up on `next` (10173930) that generalises it: each child is stamped with the current pass before it runs, so the traversal is idempotent and a shift can be recovered by rescanning from the front — every child still runs exactly once, and children appended mid-pass still run in that pass. Your test plus a three-queue variant both cover it now.

Costs ~30 B brotli in the `+createStore` size scenario, which is a fair trade for closing the case properly.

PR #2955, opened 2026-07-28
fix: handle rejected module promises in lazy()

### LINKED ISSUE #1648: [Bug?]: syntax errors in ./app.tsx causes uncaught 500 error

### LINKED ISSUE #2260: Syntax error in a lazy route module hangs the SSR stream and shows no overlay
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

A syntax error in a file route, or in anything a route imports, makes the page go blank in dev with no error overlay and no console output. The HTTP request never completes: the document shell is flushed and the connection then stays open indefinitely, so the browser considers the page "loaded" and Vite never shows its overlay.

Reproduced on `apps/fixtures/basic` at 2.0.0-rc.5 by breaking a JSX closing tag in `src/components/Counter.tsx`:

```
$ curl -s --max-time 10 http://localhost:4321/ -o /tmp/out.html -w "%{http_code} %{size_download}\n"
200 17154
exit=28          # timed out, stream never terminated
```

Cause is upstream, not in Start: solidjs/solid#2945. Solid's server-side `lazy()` attaches only a success handler to the module promise, so a rejected dynamic import leaves the Suspense boundary's `_loading` flag set forever, `notifySuspense` never fires, and `renderToStream` never calls `end()`. Every file route in Start goes through `lazy()` (`packages/start/src/router.tsx`), which is why any route with a syntax error hits it.

Note this is the **lazy** sibling of #1648, which covers the eager `app.tsx` path. Those behave differently: an error in `app.tsx` is an eager import that fails at handler level and does produce an error page, whereas a lazy route module wedges the stream with no output at all.

### Expected behavior 🤔

The error reaches the nearest `ErrorBoundary`, the dev overlay shows the syntax error and source frame, and the response completes.

### Steps to reproduce 🕹

1. `pnpm create solid` (or use `apps/fixtures/basic`), start the dev server.
2. Introduce a syntax error in any route under `src/routes/`, or in a component the route imports.
3. Load the route.
4. The page is blank, no overlay appears, and the request never finishes.

### Context 🔦

Start can work around this locally: its server-only `withAssets` wrapper in `packages/start/src/shared/lazy.ts` already wraps the route import for asset collection, so it can catch the rejection there and resolve with a component that throws during render. The throw then reaches the nearest `ErrorBoundary` and the dev overlay displays it. Once solidjs/solid#2945 lands, that catch becomes redundant and can be dropped.

Filing this to track the upstream dependency. Three related gaps a local workaround does **not** address:

1. **Client hydration path.** Solid's browser `lazy` uses the same success-only `.then` in its hydration branch (`dist/solid.js:1442`); on rejection `sharedConfig.count` is never decremented, so a chunk that 404s during initial hydration can still stall. Not reproduced, noted in solidjs/solid#2945. SPA mode is unaffected, since it goes through `createResource`.
2. **No auto-reload after the file is fixed.** With the page already open, correcting the syntax error does not trigger a reload; a manual refresh is needed.
3. **Wrong error on HMR without a reload.** The toolbar reports `TypeError: Failed to fetch dynamically imported module` rather than the underlying syntax error. Reloading surfaces the real one.

### Your environment 🌎

- `@solidjs/start` 2.0.0-rc.5
- `solid-js` 1.9.14
- `vite` 8.1.5
- Node v26.3.0

### LINKED ISSUE #2945: [solid v1] renderToStream() never closes when a lazy() component's module promise rejects
### Describe the bug

On the server, `lazy()` attaches only a success handler to the module promise. If the promise rejects (a chunk that fails to load, or in dev a module with a syntax error), the Suspense boundary's `_loading` flag is never cleared, `notifySuspense` never fires, the registered fragment never resolves, and `renderToStream` never calls `end()`. The response hangs open indefinitely.

The error also never reaches an enclosing `ErrorBoundary`, so there is nothing to render and nothing logged. For a framework doing streaming SSR this shows up as a blank page on a request that never finishes.

Standalone script, no bundler needed.

`package.json`:

```json
{ "name": "solid-lazy-repro", "type": "module", "dependencies": { "solid-js": "1.9.14" } }
```

`repro.mjs`:

```js
import { createComponent, ErrorBoundary, lazy, Suspense } from "solid-js";
import { renderToStream } from "solid-js/web";

// A real SSR server survives unhandled rejections; without this, Node exits
// before the hang is observable (see Additional Information).
process.on("unhandledRejection", () => {});

const Broken = lazy(() => Promise.reject(new Error("boom")));

renderToStream(() =>
  createComponent(ErrorBoundary, {
    fallback: e => `error boundary: ${e.message}`,
    get children() {
      return createComponent(Suspense, {
        fallback: "loading",
        get children() {
          return createComponent(Broken, {});
        },
      });
    },
  }),
).pipe({
  write: () => {},
  end: () => {
    console.log("stream ended");
    process.exit(0);
  },
});

setTimeout(() => {
  console.log("5s elapsed, end() was never called");
  process.exit(1);
}, 5000);
```

Then `npm install && node repro.mjs`.

### Expected behavior

The rejection reaches the `ErrorBoundary`, which renders its fallback, and the stream closes.

### Actual behavior

Prints `5s elapsed, end() was never called`. `end()` is never called and the boundary fallback never renders.

### Additional Information

The cause is in `dist/server.js`, `function lazy` (line 628 in 1.9.14):

```js
p = fn();
p.then(mod => p.resolved = mod.default);   // no rejection handler
```

`wrap` then sets `track = { _loading: true }` and registers it on the `SuspenseContext`. `track._loading` is only cleared inside the fulfillment callback at line 655, so on rejection `suspenseComplete()` stays false forever and the fragment registered by `registerFragment` never resolves.

Two related notes:

- Without the `unhandledRejection` handler in the repro, Node exits on the un-caught `p.then()` at `load` (`server.js:632`) before the hang is visible. So the same defect is either a hang or a hard crash depending on whether the host installs a handler.
- The browser build has the same success-only pattern in its hydration branch (`dist/solid.js:1442`): on rejection `sharedConfig.count` is never decremented. I have not reproduced that one, but it looks like the same class of bug. The non-hydrating path is fine, since it goes through `createResource`.

`main` still has this: [`packages/solid/src/server/rendering.ts`](https://github.com/solidjs/solid/blob/main/packages/solid/src/server/rendering.ts).

Found via SolidStart, where every file route is a `lazy()`, so any route with a syntax error blanks the page in dev with no overlay.

### Related

#2780 reported the same class of defect against the 2.0 beta implementation (`p.v` / `NotReadyError`) and was fixed in 2.0.0-beta.15. The 1.x implementation on `main` (`p.resolved`) still has the un-caught `.then`, and that is the line SolidStart 2 depends on (`solid-js ^1.9.14`), so the bug is still reachable in shipping apps.

Causing downstream problems
- https://github.com/solidjs/solid-start/issues/1648
- https://github.com/solidjs/solid-start/issues/2260

### Solid Version

1.9.14

### Node Version

v26.3.0

birkskyum:
On the server `lazy()` attached only a success handler to the module promise. When the promise rejected, the Suspense boundary's `_loading` flag was never cleared, so `notifySuspense` never fired, the registered fragment never resolved and `renderToStream` never called `end()`: the response hung open and the error never reached an enclosing `ErrorBoundary`. Without a host `unhandledRejection` handler the un-caught `.then` in `load` crashed the process instead.

The rejection is now recorded next to the resolved module, the boundary is notified so the fragment resolves, and the error is thrown where the component renders so `ErrorBoundary` can catch it.

The client's hydration branch had the same success-only handler, leaving `sharedConfig.count` pinned and swallowing the error. It now decrements the count and surfaces the rejection to the nearest `ErrorBoundary`.

SSR had no test harness, so this adds a second vitest project that resolves `solid-js` to the server entries.

- Closes #2945
- Closes https://github.com/solidjs/solid-start/issues/1648
- Closes https://github.com/solidjs/solid-start/issues/2260

### DIFF
--- a/.changeset/fix-lazy-rejected-module.md
+++ b/.changeset/fix-lazy-rejected-module.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Handle rejected module promises in `lazy()`. On the server the promise only had a success handler, so a chunk that failed to load left the Suspense boundary's `_loading` flag set forever: `renderToStream` never called `end()`, `renderToStringAsync` never resolved, and the error never reached an enclosing `ErrorBoundary`. The rejection is now tracked alongside the resolved module, the boundary is notified, and the error is thrown where the component renders so `ErrorBoundary` can catch it. The client's hydration branch had the same success-only handler, leaving `sharedConfig.count` pinned and swallowing the error; it now decrements the count and surfaces the rejection to the nearest `ErrorBoundary`.
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -365,12 +365,25 @@ export function lazy<T extends Component<any>>(
       const [s, set] = createSignal<T>();
       sharedConfig.count || (sharedConfig.count = 0);
       sharedConfig.count++;
-      (p || (p = fn())).then(mod => {
-        !sharedConfig.done && setHydrateContext(ctx);
-        sharedConfig.count!--;
-        set(() => mod.default);
-        setHydrateContext();
-      });
+      (p || (p = fn())).then(
+        mod => {
+          !sharedConfig.done && setHydrateContext(ctx);
+          sharedConfig.count!--;
+          set(() => mod.default);
+          setHydrateContext();
+        },
+        err => {
+          !sharedConfig.done && setHydrateContext(ctx);
+          sharedConfig.count!--;
+          set(
+            () =>
+              (() => {
+                throw err;
+              }) as unknown as T
+          );
+          setHydrateContext();
+        }
+      );
       comp = s;
     } else if (!comp) {
       const [s] = createResource<T>(() => (p || (p = fn())).then(mod => mod.default));
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -556,11 +556,14 @@ export function createResource<T, S>(
 export function lazy<T extends Component<any>>(
   fn: () => Promise<{ default: T }>
 ): T & { preload: () => Promise<{ default: T }> } {
-  let p: Promise<{ default: T }> & { resolved?: T };
+  let p: Promise<{ default: T }> & { resolved?: T; error?: any };
   let load = (id?: string) => {
     if (!p) {
       p = fn();
-      p.then(mod => (p.resolved = mod.default));
+      p.then(
+        mod => (p.resolved = mod.default),
+        err => (p.error = castError(err))
+      );
       if (id) sharedConfig.context!.lazy[id] = p;
     }
     return p;
@@ -574,18 +577,26 @@ export function lazy<T extends Component<any>>(
     if (ref) p = ref;
     else load(id);
     if (p.resolved) return p.resolved(props);
+    if (p.error) throw p.error;
     const ctx = useContext(SuspenseContext);
-    const track = { _loading: true, error: undefined };
+    const track = { _loading: true, error: undefined as any };
     if (ctx) {
       ctx.resources.set(id, track);
       contexts.add(ctx);
     }
     if (sharedConfig.context!.async) {
       sharedConfig.context!.block(
-        p.then(() => {
-          track._loading = false;
-          notifySuspense(contexts);
-        })
+        p.then(
+          () => {
+            track._loading = false;
+            notifySuspense(contexts);
+          },
+          err => {
+            track._loading = false;
+            track.error = castError(err);
+            notifySuspense(contexts);
+          }
+        )
       );
     }
     return "";
--- a/packages/solid/test/server/lazy.spec.ts
+++ b/packages/solid/test/server/lazy.spec.ts
@@ -1,0 +1,63 @@
+import { describe, expect, test } from "vitest";
+import { createComponent, ErrorBoundary, lazy, Suspense } from "../../src/server/index.js";
+import { renderToStream, renderToStringAsync } from "../../web/server/index.js";
+
+type Text = () => string;
+
+function renderBroken(render: (code: () => any) => any) {
+  const Broken = lazy<Text>(() => Promise.reject(new Error("boom")));
+  return render(() =>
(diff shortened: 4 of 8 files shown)


ryansolid:
Landed on main in 77003419, combined with #3002. Rejected lazy() modules now clear Suspense `_loading` (stream ends, ErrorBoundary sees the error) and drop the cached promise so a later request or preload can retry. Thanks.

PR #2961, opened 2026-07-31
fix(frames): hand slot claims over from the enclosing hydration registry

brenelz:
### Problem

Every page load of an app with client components inside server components warns, even though hydration is correct. The `examples/notes` dev server reports all nine of its client slot roots:

```
Hydration completed with 9 unclaimed server-rendered node(s):
  <form _hk="sc-84ece717-0-appView-search-0" class="search" role="search">…
  <a _hk="sc-84ece717-0-appView-editButton-0" href="/new" class="edit-button…
  <div _hk="sc-a112e1b2-0-noteListView-item#0-1" style="color:black" class="sidebar-note-list-item…
  …
```

The listed nodes are exactly the client slot roots — `SearchField`, `EditButton`, each `SidebarNoteContent` and its `Show` fallback, and the route outlet. They are claimed in place and fully live (the expand toggle, the search field and the route outlet all work); only the accounting is wrong.

### Cause

`hydrate()` builds the page-wide registry with `gatherHydratable`, which is a blanket `querySelectorAll("*[_hk]")` over the container — it does not skip frame regions, so every server component's client slot roots land in the root registry.

Adoption never claims from that registry. `claimRender` gathers the slot range's own `_hk` nodes into a **scoped** registry, swaps it in for the synchronous render, and restores the previous one after. `getNextElement` deletes the key it claims from whichever registry is current, so the claim consumes the scoped copy and the root registry's copy survives to the end of hydration — where `verifyHydration` reports it as unclaimed server markup.

### Fix

When the claim scope takes ownership of a range, drop its keys from the enclosing registry. Only the scoped pass ever claims those nodes, so nothing else loses a claim.

### Test plan

- `examples/notes` under `pnpm dev`: warning gone, sidebar expand/collapse, search and navigation still work.

### DIFF
--- a/packages/solid-web/frames/src/client.ts
+++ b/packages/solid-web/frames/src/client.ts
@@ -145,6 +145,12 @@ function claimRender(prefix: string, existing: Node[], render: () => any) {
   const prevRegistry = sc.registry;
   const prevHydrating = sc.hydrating;
   const prevClaimRoots = sc.claimRoots;
+  // The enclosing pass gathered these same nodes: gatherHydratable sweeps the
+  // whole document for `_hk`, frame regions included, so every slot root ends
+  // up in the root registry too. Only this scoped registry ever claims them,
+  // so hand ownership over — otherwise the root's completion check reports
+  // each claimed slot node as unclaimed server markup.
+  if (prevRegistry) for (const key of registry.keys()) prevRegistry.delete(key);
   sc.registry = registry;
   sc.hydrating = true;
   // The range may be DETACHED right now (an async slot fill renders before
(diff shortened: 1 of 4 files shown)


packages/solid-web/frames/src/client.ts
```diff
@@ -145,6 +145,12 @@ function claimRender(prefix: string, existing: Node[], render: () => any) {
   const prevRegistry = sc.registry;
   const prevHydrating = sc.hydrating;
   const prevClaimRoots = sc.claimRoots;
+  // The enclosing pass gathered these same nodes: gatherHydratable sweeps the
+  // whole document for `_hk`, frame regions included, so every slot root ends
+  // up in the root registry too. Only this scoped registry ever claims them,
+  // so hand ownership over — otherwise the root's completion check reports
+  // each claimed slot node as unclaimed server markup.
+  if (prevRegistry) for (const key of registry.keys()) prevRegistry.delete(key);
```
brenelz, line 153: maybe this is a bad fix

Issue #2981, opened 2026-08-08
2.0.0-beta.31 |  `createProjection` `ssrSource:client` leaks initial value  on initial render.


mizulu:
### Describe the bug

a `createProjection`  or `createStore` with ssrSource `client` 
will render the initial value client side before the store  or projection is resolved. 

if the ssrSource is `server` or `hydration` the  client will be pending  where the store is read
with `client` the store can be read with the initially seeded value. 

### Your Example Website or App

https://s.olid.uk/id/zP2S3c1ERKWcLOtGC1po7g

### Steps to Reproduce the Bug or Issue

```tsx
  const v = createProjection(async () => {
    await new Promise(r => setTimeout(r, 3000))

    return { a: 1 }
  }, { a: "Initial Value" },
    { ssrSource: ssrSource }
  )
```

```tsx
  return (

      <Loading fallback="Loading..">
        {v.a}
      </Loading>
)
```

first client will render "Initial Value" once async is resolved it will become "1"

### Expected behavior

it might be expected that the read for v.a will also be pending. 
and wont leak initially seeded value of store. 

which is mostly not readble as part of the reactive system without raising NotReady

### Platform

.

### Additional context

additionally when reading v.a in outside the owned scoped
like in a setTimeout callback. 

the read for v.a is not throwing in that scope, however if we set ssrSource 
to server, the timeout callback that reads v.a will propagate the NotReady exception 
which it does not do for ssrSource client when evaluate on the client. 

this is just one more thing to possibly confirm to see if there are some inconsistent path
if it is reproducible 

```tsx
import { render, isServer } from '@solidjs/web';
import { createSignal, createProjection, createStore, Loading } from 'solid-js';

export default function App() {
  const [count, setCount] = createSignal(0);
  const ssrSource: "client" | "server" | "hybrid" = "client"


  // Store Case ( same )
  // ==========
  // const [v, vs] = createStore(async () => {
  //   await new Promise(r=> setTimeout(r,3000))
  //   return { a: 1 }
  // }, { a: "Initial Value" },
  //   { ssrSource: ssrSource }
  // )
  // setTimeout(() => {
  //   console.log(isServer)
  //   console.log(v.a)
  // })

  // Projection Case: 
  // ===================

  const v = createProjection(async () => {
    await new Promise(r => setTimeout(r, 3000))

    return { a: 1 }
  }, { a: "Initial Value" },
    { ssrSource: ssrSource }
  )
  setTimeout(() => {
[...]

Issue #2897, opened 2026-07-16
2.0.0-beta.19 | a pending optimistic store, "initial value" is readable in untracked part of the component, but throw NotReady in render part.


mizulu:
### Describe the bug

```
function Comp(){
  let [s,ss] = createOptimisticStore(async ()=>{
      await wait(1000)
      return {}
  },{a:1})

  console.log("It is ", s.a)// s should throw as it is not ready. 

  // this should suspend because of no <Loading>
  return <>{s.a}</>
}

```

console output 
```
⚠
[STRICT_READ_UNTRACKED] Reactive value read directly in <App> will not update. Move it into a tracking scope (JSX, a memo, or an effect's compute function).
It is 
1
⚠
[ASYNC_OUTSIDE_LOADING_BOUNDARY] An async value was read outside a Loading boundary. The root mount will be deferred until all pending async settles.
```

### Your Example Website or App

https://s.olid.uk/id/H3xbHt16ShekQkSSOMk2kg

### Steps to Reproduce the Bug or Issue

playground

### Expected behavior

it is expected that a not ready store, throws not ready error 
even when not used in the render part. 

### Screenshots or Videos

<img width="1906" height="604" alt="Image" src="https://github.com/user-attachments/assets/d52e0501-f8ee-454e-b479-85dc29f9c008" />

### Platform

.

### Additional context

[source for llm](https://source.s.olid.uk/id/H3xbHt16ShekQkSSOMk2kg)

ryansolid:
Fixed on `next` in 2dc2c8fb. You found a real gap in the safeguard, though the ruled behavior is slightly different from the expectation in the report:

**What was wrong:** untracked store reads skip reactive node creation, so they never reached the `PENDING_ASYNC_UNTRACKED_READ` check that pending memos hit. A component-body read of an in-flight derived store silently returned the seed value (`1`) with only the generic strict-read warning. The store proxy's strict-read branch now consults the firewall's status first and throws the same dev error a pending memo read does:

```
[PENDING_ASYNC_UNTRACKED_READ] Reading a pending async value directly in <App>.
Async values must be read within a tracking scope (JSX, a memo, or an effect's compute function).
```

[...]

PR #2986, opened 2026-08-10
fix(solid): ssrSource "client" projections/stores don't leak their seed before first resolution (#2981)

SnowingFox:
## What

`createProjection`/`createStore(fn)` with `ssrSource: "client"` leaked their seeded draft value on the client's initial render — `<Loading>` showed the seed ("Initial Value") instead of the fallback until the async source resolved.

## Why

A `"client"` source is gated behind a `hydrated` signal during hydration so the derive runs only after hydration completes. The gate previously returned `undefined`, which settled the projection against its seed (the #2897 guard cleared `STATUS_UNINITIALIZED`), leaving the seed readable. `ssrSource: "server"`/`"hybrid"` sources stay pending (NotReady) during this window, so "client" was inconsistent.

## Fix

The gate now throws `NotReady` while hydration is incomplete, keeping the projection in its pending/uninitialized state. Reads stay pending (the `Loading` fallback shows) and the seed is never observable before first resolution — matching the #2897 rule that a derived store's seed is a draft, never an observable value. The fix is in `hydrateStoreLikeFn`, so it applies symmetrically to `createProjection`, `createStore(fn)`, and `createOptimisticStore(fn)`.

## Tests

Updated every `ssrSource "client"` hydration test that asserted the seed was readable during hydration (two in the "ssrSource client modes" block and two in the "post-hydration transition" block, covering both `createProjection` and `createStore(fn)`), and added an async projection test reproducing the exact issue scenario.

### DIFF
--- a/.changeset/fix-ssr-source-client-seed-leak.md
+++ b/.changeset/fix-ssr-source-client-seed-leak.md
@@ -1,0 +1,7 @@
+---
+"solid-js": patch
+---
+
+Fix `createProjection`/`createStore(fn)` with `ssrSource: "client"` leaking their seeded draft value on the client's initial render (#2981).
+
+A `"client"` source is gated behind a `hydrated` signal during hydration so the derive runs only after hydration completes. The gate previously returned `undefined`, which settled the projection against its seed — leaving the seed readable ("Initial Value" rendered before the async source resolved) instead of staying pending. The gate now throws `NotReady` while hydration is incomplete, so reads stay pending (the `Loading` fallback shows) exactly like `ssrSource: "server"`/`"hybrid"` sources, and the seed is never observable before first resolution (#2897).
--- a/packages/solid/src/client/hydration.ts
+++ b/packages/solid/src/client/hydration.ts
@@ -6,6 +6,7 @@ import {
   runWithOwner,
   onCleanup,
   isDisposed,
+  NotReadyError,
   getNextChildId,
   peekNextChildId,
   createRevealOrder as coreRevealOrder,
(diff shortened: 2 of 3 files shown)


ryansolid:
Thanks for digging into this — the diagnosis of *where* the seed becomes observable is exactly right, and the #2897 framing ("a seed is a draft, never a value") is the correct instinct. But we can't take the fix, because it breaks the hydration contract that `ssrSource: "client"` currently depends on.

The server never runs a client-sourced store — it renders the **seed** into the HTML (`createProjection` on the server returns the seeded store untouched for `ssrSource: "client"`). So during hydration the client *must* produce the seed to claim that markup. With this change, the store is pending through hydration, the `Loading` boundary shows its fallback where the server rendered seeded content, and you get a hydration mismatch plus a content flash on every client-sourced store. The tests this PR updates were the ones pinning that parity contract. The signal/memo family behaves the same way (the gate returns `prev` rather than throwing; the server keeps client-sourced memos settled at `undefined`), so the pre-PR store behavior is also the internally consistent one.

The real defect is that `ssrSource: "client"` promotes the seed *implicitly* — and that's being fixed at the design level rather than the gate: an upcoming loading-value feature introduces an explicit "commit #0" (`loadingValue` / `seedLoadingValue`), and with it `ssrSource: "client"` will **require** an explicit loading value (type-enforced + dev error). Seeds then stay drafts everywhere, `<Loading>` never lies, and the server still always has real markup to flush — no implicit promotion, no parity break.

Closing in favor of that direction, with appreciation — this PR is what forced the inconsistency to be resolved properly. (See #2981 for the full ruling.)

SnowingFox:
Thanks for the thorough write-up, Ryan — the hydration-contract point makes sense, and I agree the explicit `loadingValue`/`seedLoadingValue` direction is the right fix. Glad the PR helped surface the implicit-seed inconsistency. I'll leave it closed and watch for that feature.

PR #2991, opened 2026-08-11
fix: preserve errors while loading retries are parked

### LINKED ISSUE #2989: Parked retry on an errored loadingValue node replaces the settled error with a NotReadyError
**Versions:** `solid-js` 2.0.0-beta.32, repo `next` at 766ea303.

## Repro

```ts
const dFirst = deferred<string>();
const dDep = deferred<number>();
let user!: () => string;
let setS!: (v: number) => void;
createRoot(() => {
  const [s, set] = createSignal(0);
  setS = set;
  const dep = createMemo(() => dDep.promise);
  user = createMemo<string>(
    () => (s() === 0 ? (dFirst.promise as any) : `dep:${dep()}`),
    { loadingValue: "placeholder" }
  );
  createEffect(() => user(), { effect: () => {}, error: () => {} });
});
flush();

// 1. First flight rejects: the error is the settled answer.
dFirst.reject(new Error("boom"));
await tick();
flush();
expect(() => user()).toThrow("boom"); // passes — settled at error

// 2. Retry via a real input change; the retry reads an unready sync source and parks.
setS(1);
flush();

// 3. Reads should still throw the settled error…
try { user(); } catch (e) { thrown = e; }
expect(thrown instanceof NotReadyError).toBe(false); // FAILS — it IS a NotReadyError
expect((thrown as Error).message).toBe("boom");      // FAILS — original error lost
```

## Expected

The inline contract in `recompute`'s catch (the `parkLoadingWindow` branch, `packages/solid-signals/src/core/core.ts`) states it directly: *"If the node is currently errored the error stays the answer until this retry can actually run."* And the loading window's design invariant is that a park has **no read-visible pending status**. So while the retry is parked, reads should keep throwing the original `Error("boom")`.

## Actual

Reads throw a `NotReadyError`. Untracked readers get a pending-class error with the real error's identity lost; tracked readers and boundaries treat a node that is settled-at-error as pending.

## Mechanism

`parkLoadingWindow` (`packages/solid-signals/src/core/async.ts:79`) calls `setPendingError`, which overwrites `el._error` with the `NotReadyError` — while `_statusFlags` still carries `STATUS_ERROR` (the park branch in `recompute`'s catch never touches status flags). The `STATUS_ERROR` read branch (`core.ts` ~900, `throw owner._error`) then throws the wrong object.

## Severity bounds (both pinned by passing tests in the same file)

- The parked-after-error window stays verdict-quiet: `isPending(user)` is `false` throughout, per the window contract.
- No wedge: when the unready source settles, the retry runs and the answer lands (`dep:7`).

So the damage is misreported error identity plus a pending-class throw leaking from a window that is contractually read-invisible — not a stall.

## Repro in tree

`packages/solid-signals/tests/__hunt15-error-then-park.spec.ts` — the main test fails; a sanity control (same shape, non-errored) and the two severity pins pass.

chyroc:
## Summary

Fixes #2989.

When a `loadingValue` memo is settled at an error and a reactive retry parks on
an unready dependency, `parkLoadingWindow` currently replaces the memo's
read-visible settled error with the dependency's `NotReadyError`. This change
keeps the settled error while `STATUS_ERROR` is active. The existing
`_pendingSources` and `_blocked` bookkeeping still schedules the retry when the
dependency settles.

The regression test covers all parts of the window contract:

- the settled error remains the same object while the retry is parked;
- `isPending` stays false during the parked window;
- settling the dependency retries the memo;
- the retry commits the final value successfully.

- `pnpm --filter @solidjs/signals exec vitest run tests/loading-value.test.ts`
  (28 tests passed)
  (86 test files, 1231 tests passed)
- `pnpm --filter @solidjs/signals types`
- `pnpm exec prettier --check .changeset/preserve-parked-loading-error.md packages/solid-signals/src/core/async.ts packages/solid-signals/tests/loading-value.test.ts`

The repository-wide `pnpm build` and `pnpm test` currently stop in the unchanged
`@solidjs/web#types` task because the locked
`@dom-expressions/runtime@0.50.0-next.40` package does not contain
`src/serializer-decode.d.ts`, which that package's script tries to copy. The
`@solidjs/signals` build, type check, and complete test suite pass independently.

### DIFF
--- a/.changeset/preserve-parked-loading-error.md
+++ b/.changeset/preserve-parked-loading-error.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Preserve a loading-value memo's settled error while a retry is parked on an unready dependency. The parked retry remains verdict-quiet and resumes when its dependency settles instead of exposing a `NotReadyError` in place of the original error.
--- a/packages/solid-signals/src/core/async.ts
+++ b/packages/solid-signals/src/core/async.ts
@@ -79,7 +79,10 @@ function retryReaches(el: Computed<any>, source: any): boolean {
 export function parkLoadingWindow(el: Computed<any>, e: NotReadyError): void {
   el._blocked = true;
   if (e.source) addPendingSource(el, e.source as Computed<any>);
-  setPendingError(el, e.source as Computed<any>, e);
+  // An errored node keeps its settled error as the read-visible answer while
+  // the retry is parked. `_pendingSources` and `_blocked` are sufficient to
+  // schedule it again when the dependency settles.
+  if (!(el._statusFlags & STATUS_ERROR)) setPendingError(el, e.source as Computed<any>, e);
 }
 
 export function setPendingError(el: Computed<any>, source?: Computed<any>, error?: any): void {
--- a/packages/solid-signals/tests/loading-value.test.ts
+++ b/packages/solid-signals/tests/loading-value.test.ts
@@ -311,6 +311,56 @@ describe("createMemo with loadingValue", () => {
     expect(isPending(user)).toBe(false);
   });
 
+  it("keeps the settled error while a retry is parked on an unready dependency", async () => {
+    const first = deferred<string>();
+    const dependency = deferred<number>();
+    const originalError = new Error("boom");
+    let setRetry!: (retry: boolean) => void;
+    let user!: () => string;
+
+    createRoot(() => {
+      const [retry, set] = createSignal(false);
+      setRetry = set;
+      const dep = createMemo(() => dependency.promise);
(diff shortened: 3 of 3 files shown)


ryansolid:
Thanks for this — the diagnosis was exactly right, and your test contract (error identity through the parked window, quiet `isPending`, retry on dependency settle) matches what we want pinned. A fix for the full report landed on `next` in ba7560ff before this could be reviewed; it covers the park path the same way yours does and additionally guards the source-settle path, which could reintroduce the same clobber when the unready dependency settles. Closing as superseded — sorry for the overlap, and the effort is appreciated (the regression coverage that landed asserts the same invariants your tests did).

PR #2992, opened 2026-08-11
fix: re-subscribe to ordinary external source after a transition

### LINKED ISSUE #2953: `enableExternalSource`: a computation created during a transition loses its external subscription after one update
When `enableExternalSource` is enabled, a computation created while a transition is pending registers its external dependencies on the transition-scoped source rather than the ordinary one. That source is then disposed by `triggerInTransition`'s `.then()` callback, which is the same code path that delivers an update, and nothing ever re-registers the computation on its ordinary source.

So a computation created while in a transition receives exactly one external update, then stops responding to the external system permanently. In an app it shows up as a subtree that renders correctly, updates once, and is stale from then on until something else forces it to re-run.

I believe the root cause can be found in the `createComputation` function:

https://github.com/solidjs/solid/blob/3be495cec52bf78d7cc61f054af00320ecf4058c/packages/solid/src/reactive/signal.ts#L1475-L1497

An external source records the computation's dependencies when its `track()` runs. A computation created during a transition only ever has `inTransition.track()` called, so ordinary exists with zero recorded dependencies.

SnowingFox:
## Problem

A computation created while a `startTransition` is pending never receives more than one external update when `enableExternalSource` is enabled. In an app this shows up as a subtree that renders correctly, updates once, then stays stale until something else forces it to re-run.

## Root cause

When `ExternalSourceConfig` is set, `createComputation` (in `packages/solid/src/reactive/signal.ts`) wraps the computation's fn so that a run while a transition is pending tracks the computation on a transition-scoped external source (`inTransition`) instead of the ordinary one. A computation created during a transition only ever has `inTransition.track()` called, so its ordinary source records zero dependencies. After the transition completes, `triggerInTransition`'s `.then()` disposes the transition-scoped source and nothing ever re-registers the computation on the ordinary source — the external system can no longer invalidate it.

## Fix

Track whether the computation has ever run through the ordinary external source. When the transition-scoped source is disposed, if it never did, fire the computation's trigger once so it re-runs through the ordinary source and re-subscribes. If the computation was disposed in the meantime the trigger is a no-op, because the computation is no longer an observer of the internal `track` signal.

## Test

Adds a regression test to `packages/solid/test/external-source.spec.ts` that creates a memo inside a transition, applies two external updates, and asserts both are delivered. On `main` the second update is lost and the test fails.

Fixes #2953

### DIFF
--- a/.changeset/external-source-transition-subscription.md
+++ b/.changeset/external-source-transition-subscription.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+fix: re-establish the ordinary external source subscription after a transition
--- a/packages/solid/src/reactive/signal.ts
+++ b/packages/solid/src/reactive/signal.ts
@@ -1486,11 +1486,20 @@ function createComputation<Next, Init = unknown>(
     const ordinary = ExternalSourceConfig.factory(sourceFn, trigger);
     onCleanup(() => ordinary.dispose());
     let inTransition: ExternalSource | undefined;
+    let trackedOrdinary = false;
     const triggerInTransition: () => void = () =>
       startTransition(trigger).then(() => {
         if (inTransition) {
           inTransition.dispose();
           inTransition = undefined;
+          // A computation created while a transition was running only ever
+          // tracked the transition-scoped source, so its ordinary source has no
+          // recorded dependencies. Now that the transition is over, re-trigger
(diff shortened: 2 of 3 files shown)


ryansolid:
Landed on main in c6aa6728. After the transition-scoped source is disposed, a computation that never ran through the ordinary source is triggered once so it re-subscribes. Thanks.

PR #3002, opened 2026-08-16
fix(core): clear cached promise in lazy() on rejection (#2999)

### LINKED ISSUE #2999: [2.0] lazy() caches rejected module promises: retry is impossible on the client and one failed import poisons SSR for the process lifetime
**Affected version:** `solid-js@2.0.0-rc.0`

## Summary

`lazy()` caches its module promise permanently — including rejections. Once a chunk download fails, that failure is sealed in:

- **Client:** an error-boundary "retry" flow can never work. `Errored`'s `reset()` remounts the boundary's children, but the lazy wrapper still holds the cached rejected promise, so every remount re-throws the original download error until a full page reload. Browsers deliberately do *not* cache failed dynamic imports (a later `import()` of the same specifier re-fetches), so the cached rejection also diverges from platform behavior.
- **Server:** `p` is module-scoped and shared across requests (intentionally, so `_lazyAssets` reuse works). A single transient import failure — momentary FS/network hiccup, dev-server race — marks the shared promise errored and poisons **every subsequent SSR request** for the process lifetime.
- **Both:** the internal `p.then(mod => { comp = () => mod.default })` has no rejection handler, so every failed load additionally surfaces as an unhandled rejection.

## Reproduction (client)

```jsx
let fail = true;
const Page = lazy(() =>
  fail ? Promise.reject(new Error("download failed")) : import("./Page")
);

<Errored fallback={(err, reset) => (
  <button onClick={() => { fail = false; reset(); }}>Retry</button>
)}>
  <Page />
</Errored>
```

Clicking Retry re-throws `download failed` forever — `fn` is never called again even though the failure condition is gone. The same shape with a flaky network (chunk 503s once) leaves the route permanently broken without a hard reload.

## Suggested fix

Clear the cached promise on rejection so the next preload/render retries, and route both the render path and `preload` through one `load()` so they dedupe. Client sketch (`dist/solid.js` shape):

```js
function lazy(fn, moduleUrl) {
  let comp;
  let p;
  const load = () => {
    if (!p) {
      const cur = p = fn();
      cur.then(mod => {
        comp = () => mod.default;
      }, () => {
        if (p === cur) p = undefined;
      });
    }
    return p;
  };
  const wrap = props => {
    if (sharedConfig.hydrating) comp = _lazyHydrationLookup(comp, moduleUrl);
    let local = comp;
    if (!local) {
      const cur = load();
      local = createMemo(() => cur.then(mod => mod.default));
    }
    // ... unchanged
  };
  wrap.preload = load;
  wrap.moduleUrl = moduleUrl;
  return wrap;
}
```

Notes on the shape:

- `wrap` captures `const cur = load()` instead of reading closure `p`, so a rejection-triggered clear can't race a render in flight — the failing render still throws (through the memo's rejected promise) into the nearest boundary, and the *next* mount retries.
- The added rejection handler also fixes the unhandled-rejection leak as a side effect.
- Server-side is the same pattern: capture `cur`, set `cur.error`/`cur.errored`, clear `p === cur` on rejection so the next request retries, and read `cur` (not closure `p`) in the wrap's memo/asset paths.

## Context

Found porting TanStack Router's `lazyRouteComponent` to delegate to `lazy()` on the 2.0 rc line — its download-retry contract (error boundary → retry → re-import) is impossible against a cached rejection without wrapping `lazy()` in an external retry state machine. We're currently carrying this as a pnpm patch (with tests exercising the retry flow) in [TanStack/router#8081](https://github.com/TanStack/router/pull/8081); happy to turn it into a PR against the source if the approach looks right.

Possibly related: #2945 tracks a v1 symptom of the same missing rejection handling (`renderToStream` never closing when a lazy module promise rejects).

webdevsamran:
## Summary

Clears the cached promise in `lazy()` on rejection, allowing dynamic imports that fail due to transient network or chunk-load errors to be retried on subsequent renders/preloads without requiring a hard page reload or permanently poisoning SSR.

## Problem

In both client (`packages/solid/src/render/component.ts`) and server (`packages/solid/src/server/rendering.ts`) implementations of `lazy()`, the promise returned by `fn()` was cached permanently in closure variable `p`.

If `fn()` rejected:
1. **Client:** An Error Boundary retry flow (e.g. `<Errored fallback={(_, reset) => <button onClick={reset}>Retry</button>}>`) could never recover. Resetting/remounting the boundary simply re-observed the cached rejected promise without invoking `fn()` again.
2. **Server:** Module-scoped lazy promises that rejected would poison all subsequent SSR requests for the process lifetime.
3. **Unhandled Rejections:** The internal `.then()` calls lacked a rejection handler, emitting unhandled promise rejections.

## Solution

1. In both client and server `lazy()`, attach a rejection callback to the active promise `cur` that clears `p` (`if (p === cur) p = undefined`).
2. Subsequent `load()` / `preload()` invocations will call `fn()` afresh.
3. If `fn()` succeeds on retry, `comp` / `resolved` is populated normally.
4. Added regression unit test in `packages/solid/web/test/lazy.spec.tsx` verifying that `lazy().preload()` retries `fn()` after an initial rejection.

Fixes #2999

### DIFF
--- a/packages/solid/src/render/component.ts
+++ b/packages/solid/src/render/component.ts
@@ -359,13 +359,27 @@ export function lazy<T extends Component<any>>(
 ): T & { preload: () => Promise<{ default: T }> } {
   let comp: (() => T | undefined) | undefined;
   let p: Promise<{ default: T }> | undefined;
+  const load = () => {
+    if (!p) {
+      const cur = (p = fn());
+      cur.then(
+        mod => {
+          comp = () => mod.default;
+        },
+        () => {
+          if (p === cur) p = undefined;
+        }
+      );
+    }
+    return p;
+  };
   const wrap: T & { preload?: () => void } = ((props: any) => {
     const ctx = sharedConfig.context;
     if (ctx) {
       const [s, set] = createSignal<T>();
       sharedConfig.count || (sharedConfig.count = 0);
       sharedConfig.count++;
-      (p || (p = fn())).then(mod => {
+      load().then(mod => {
         !sharedConfig.done && setHydrateContext(ctx);
         sharedConfig.count!--;
         set(() => mod.default);
@@ -373,7 +387,7 @@ export function lazy<T extends Component<any>>(
       });
       comp = s;
     } else if (!comp) {
-      const [s] = createResource<T>(() => (p || (p = fn())).then(mod => mod.default));
+      const [s] = createResource<T>(() => load().then(mod => mod.default));
       comp = s;
       onCleanup(() => (comp = undefined));
     }
@@ -392,7 +406,7 @@ export function lazy<T extends Component<any>>(
         : ""
     ) as unknown as JSX.Element;
   }) as T;
-  wrap.preload = () => p || ((p = fn()).then(mod => (comp = () => mod.default)), p);
+  wrap.preload = () => load();
   return wrap as T & { preload: () => Promise<{ default: T }> };
 }
 
--- a/packages/solid/src/server/rendering.ts
+++ b/packages/solid/src/server/rendering.ts
@@ -556,12 +556,17 @@ export function createResource<T, S>(
 export function lazy<T extends Component<any>>(
   fn: () => Promise<{ default: T }>
 ): T & { preload: () => Promise<{ default: T }> } {
-  let p: Promise<{ default: T }> & { resolved?: T };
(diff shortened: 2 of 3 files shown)


ryansolid:
Landed on main in 77003419, combined with #2955. The rejected promise is no longer cached forever — `preload()` and the next SSR request call `fn()` again. Thanks.

Issue #2883, opened 2026-07-14
Bundle size audit: pay-for-use tree-shaking for @solidjs/signals subsystems


ryansolid:
## Background

Solid 2.0's library size has grown substantially over 1.x, driven almost entirely by `@solidjs/signals` absorbing the async model (transitions, optimistic lanes, pending sources/status rails, boundaries, projections, stores). Current measurements (all uniform: esbuild bundle + minify + gzip, production defines, Jul 2026):

**Full runtime (whole packages):**

| | Minified | Gzip |
|---|---|---|
| Solid 1.9.14 (core + store + web) | 47.4 KB | ~18.1 KB |
| Solid 2.0 beta (signals + solid + web) | 89.5 KB | ~33.5 KB |
| Svelte 5.56 (full client runtime) | 102.6 KB | 36.8 KB |
| React 19.2 + ReactDOM client | 193.4 KB | 60.3 KB |

**Tree-shaken minimal app (render + one signal):**

| | Minified | Gzip |
|---|---|---|
| Solid 1.x | 10.6 KB | 4.2 KB |
| Solid 2.0 | 31.6 KB | 12.6 KB |
| Svelte 5 (`mount`) | 39.7 KB | 14.9 KB |
| React (no shaking) | 193.4 KB | 60.3 KB |

Excluding the web runtime, the library proper is ~2.2–2.5x its 1.x size. We compare favorably against Svelte 5 and React today, but the minimal-app floor (12.6 KB gzip) means a hello-world app loads most of the async machinery whether it uses it or not.

## Root cause of the high floor

1. **`dist/prod.js` is one flat rollup bundle.** Once an app imports anything, all top-level side effects in the file are retained — including the `GlobalQueue._*` hook installs that wire subsystems together. Module-level tree-shaking can't separate subsystems that live in the same output chunk.
2. **Direct cross-subsystem calls.** Some feature machinery is invoked directly from hot core paths (e.g. store proxy traps), making it reachable from `createStore` even when the feature is unused.

## Proposed direction

- Split the published dist into chunks (rollup code splitting; `index` re-exports from feature chunks). All packages already declare `sideEffects: false`, so bundlers will drop unused chunks. Needs care around the dev/prod/node output triplet and `__DEV__` replacement.
- Convert remaining direct couplings to the existing nullable-hook pattern (`GlobalQueue._propagateAffects`, `_settleAffects`, `_clearOptimisticStore` already demonstrate it), installed by the feature module on import.
[...]

PR #3018, opened 2026-08-19
"Why did this run" dev attribution (prototype)

brenelz:
## What is this

A prototype answering "why did this scope re-run?" — dev-only observability for the reactive graph, exploring how Solid could make accidental tracking **visible** (dev attribution), **loud** (hot/wide-scope warnings), **measurable** (timings + cost attribution), and **checkable** (a static rule sketch). Opening as WIP for discussion; API names, defaults, and packaging are all up for debate.

Everything is off by default, dev-build only, and contributes **zero bytes to prod bundles** (the tree-shake guard from #2883 passes).

## The core mechanism

Every value commit stamps its node: a `setSignal` write, a memo's changed value (with its own upstream causes), an async landing, or a `refresh()`. When a computation re-runs, its previous dep links are diffed against those stamps — the newer ones are the causes, and they chain to the root write:

```
[why-run] effect "title-effect" ran (run 2, 0.42ms)
  ← memo "label" changed (#6)
    ← signal "notifications" write (#5) 1 → 2
  deps changed: +"b" -"a" (2 total)
```

Surface: `DEV.attribution.enable({ log, stacks, hotRuns, hotTime, wideDeps })`, plus `subscribe()` (structured `RerunEvent` stream), `history()`, `why(node)`, `subscriptions(node)`, `costs()`, `format(event)`.

## Layers on top

- **Hot-scope warnings** via the existing diagnostics channel, once per window:
  - `HOT_SCOPE_RERUNS` — N runs per window (default 60/1000ms), naming the leaking signal.
  - `HOT_SCOPE_TIME` — self-time budget per window (default 8ms/1000ms), catching the few-but-expensive scope that run counts miss.
- **Wide-scope warning** — `WIDE_SCOPE_DEPS` when a scope's dependency count reaches a limit (default 30, creation runs included; re-warns on 50% growth). Store property nodes now carry dev names (`store.count`) so the evidence reads as paths.
- **Subscription diffing** — `RerunEvent.depsAdded/depsRemoved/depCount` per run: when a helper edit changes what a distant scope subscribes to, the next re-run prints it.
- **Timings** — per-run `selfMs` (child-subtracted via a timing stack, so nested recomputes don't double-count) and `totalMs`. `costs()` aggregates two ways: by scope (ranked by self-time, with `wastedMs` = time spent on unchanged-value runs — the recoverable number) and by root write (`downstreamMs` = total re-run time each write caused — the fan-out cost per write site).
- **Static rule prototype** — `lint/reactive-helpers.ts` (TS compiler API; the repo has no eslint, an ESLint rule would wrap this): flags signal/store reads in un-annotated helpers, with fixpoint propagation up the module call graph (the "helper drift" case). Declaration sites: `/** @reactive */`, a `Reactive<...>` return type, or being passed to a tracking primitive. Components (PascalCase) exempt.

## Implementation notes reviewers may care about

- **Cost when disabled** is one boolean check per recompute/write; prod builds fold everything out.
- **rollup `tryCatchDeoptimization`** retains any function referenced inside a `try` even behind a folded `__DEV__` guard — the `asyncWrite` lane-branch stamp sets a local flag inside the try and stamps after the catch, or the attribution module leaks into prod (~1KB).
- **dev↔attribution import cycle**: `attribution.ts` uses `emitDiagnostic`; `DEV` exposes `attribution` through a getter because under bundler transforms the cyclic binding is `undefined` (not a TDZ throw) at `DEV` construction time.
- Attribution stamps only on **committed changes** (equality-cutoff runs leave no stamp) so a later unrelated re-run can't mis-attribute.

## Tests

25 new tests (16 runtime in `tests/attribution.test.ts`, 9 static in `tests/lint-reactive-helpers.test.ts`); solid-signals suite green (1287) including the size guard.

## Open questions

- Naming/shape of `DEV.attribution` and whether warnings should be on by default in dev.
- Whether store path names should be hierarchical (`todos.0.done`) rather than one segment.
- Where the static rule should live (eslint-plugin-solid vs in-repo) and whether `Reactive<T>` should be an exported type.


ryansolid (commented):
This is the right capability to be building — "why did this run" is the most-asked reactive debugging question there is, and several of the design calls here are exactly right: stamping only committed changes so equality-cutoff runs can't mis-attribute, the `wastedMs` vs per-write `downstreamMs` split, and the disabled-cost discipline (the `tryCatchDeoptimization` and import-cycle notes show the prototype was actually built, not just sketched). Encouraging direction overall. That said, a few things need to change shape before this can land.

**1. Transition/lane blindness is the correctness blocker.** The stamps don't distinguish an overlay recompute from a committed one. A node recomputing in a transition lane and again at commit records two runs with two timings; an optimistic recompute that lands back on the same committed value gets counted toward `wastedMs`. For a tool whose entire job is truthful blame, systematically mis-blaming transition and optimistic code — the hardest code to debug, i.e. exactly where people will reach for this — is disqualifying. Runs need a lane/transition tag, and the hot/wasted aggregations need to account for it, before the numbers can be trusted.

**2. Extract a hook surface instead of weaving into `recompute`.** ~50 lines through the most safety-critical function in the library is a permanent maintenance tax: every future core change has to keep attribution semantically honest or the tool silently lies (we've touched that exact code twice this month). Follow the `GlobalQueue._recomputeLane` precedent — a handful of narrow dev-only hook points in core, with all attribution logic living externally. Core's obligation becomes "call the hook with true facts," not "maintain attribution semantics."

**3. Default thresholds will cry wolf.** `HOT_SCOPE_RERUNS` at 60/1000ms flags every effect driven by an animation-frame signal — a fully legitimate pattern that lands at exactly 60/window. On the open question: warnings should not be on by default in dev until the false-positive story is proven out; opt-in enable answers it cleanly and costs nothing.

**4. The lint rule should move out of this PR entirely.** `/** @reactive */` and an exported `Reactive<T>` type are a language-convention commitment, not an observability feature — that deserves its own design discussion in eslint-plugin-solid, where the rule would actually ship. Keeping the prototype in `packages/solid-signals/lint/` couples an unsettled convention to this PR's fate.

**5. Packaging: decide the devtools relationship early.** The structured `RerunEvent` stream is precisely the substrate solid-devtools wants. If core exposes only the hook surface and the attribution engine lives in a dev module both `DEV.attribution` and devtools consume, we get one mechanism instead of two competing ones later. That also answers the "where does this live" open question: engine outside core, hooks inside.

Happy to iterate on the hook-point design — the set you need (recompute begin/end with create flag, committed derived change, root write, refresh, async landing) looks small and stable.

ryansolid (commented):
Follow-up review of the refactor commit — this addressed the structural asks properly, not cosmetically.

**Hook surface: right.** The 7-point `AttributionHooks` interface with `attrHooks !== null` call sites is exactly the `_recomputeLane` shape, core is down to passing raw facts, and codifying the try-block/`tryCatchDeoptimization` rule in the interface docs means future call-site authors can't re-learn it the hard way.

**Phase semantics: right.** `phase` + `held` on every run, `wastedMs` restricted to plain/non-held/unchanged runs, overlay time segregated into `overlayMs`, and the `recomputeEnd` site capturing posture before the lane restore is the correct moment. That was the correctness blocker and it's genuinely resolved.

**One remaining finding: the store node naming is eager.** `name: __DEV__ ? "store." + String(property) : undefined` runs on every store property-node creation in dev builds, engine installed or not — it's the one place the PR breaks its own "one boolean check when disabled" discipline, and it lands on the hottest store path (node creation). Gate it on `attrHooks !== null`, or better, drop the eager name entirely and derive it lazily in the engine at format time — the engine receives the node, and the property is recoverable from it. (Ignore the current CodSpeed red for now — the runner is having infrastructure issues today — but this one is worth fixing on principle regardless of what the benchmarks say.)

With that, the remaining open questions (engine packaging vs devtools, warning defaults) are genuinely just discussion items — the structure now permits the right answer to each.

brenelz:
**Follow-up: candidate runtime warnings on this hook surface**

Now that the engine is decoupled behind `attribution-hooks.ts`, each additional warning is a small consumer of hooks that already fire. Beyond the three in this PR (`HOT_SCOPE_RERUNS`, `HOT_SCOPE_TIME`, `WIDE_SCOPE_DEPS`), these are the ones I think earn their noise — each is a counter or single comparison on an existing hook, dev-only, zero prod bytes:

| Code | Trigger | Suggested fix in message | Cost |
|---|---|---|---|
| `UNSTABLE_MEMO_OUTPUT` | memo commits a referentially-new but shallow-equal value N runs in a row | return stable refs / pass `equals` | sampled shallow-compare on `derivedChanged` |
| `UNGATED_EFFECT` | effect fn fires K× with `value === prevValue` (value ≠ `undefined`) | front compute with a memo / `equals` | one compare in `runEffect` |
| `WIDE_WRITE` | one write notifies ≥ N subscribers | `createSelector` / `createProjection` for keyed questions | counter in `insertSubs` |
| `REBUILD_CHURN` | scope disposes + recreates children ≥ N×/window | `<For>` / `<Show>` / memo the expression | counter on dispose path |
| `LEAKED_OBSERVER` | live subscriptions whose owner is disposed | subscriptions belong under owners; check hand-rolled bridges | dev sweep at flush boundary |
| `HOT_SOURCE` | write freq × subscriber count exceeds budget | isolate/throttle the hot source | falls out of `costs()` aggregation |

Rationale for the set: `UNSTABLE_MEMO_OUTPUT` catches the fan-out amplifier (wide memo + fresh object output) that today is only findable by profiling; `UNGATED_EFFECT` catches the "dep changed, answer didn't" case since effect nodes are deliberately `_equals: false`; `WIDE_WRITE`/`HOT_SOURCE` blame the write side where the scope-level warnings blame the reader; `REBUILD_CHURN` is the runtime net under the two structural shapes `wrapConditionals` can't reach (`.map`, JSX-returning helpers); `LEAKED_OBSERVER` targets the ownerless-bridge leaks reported from Solid-stores-in-React hybrids.

Design constraints carried over from this PR's review: all phase-aware (overlay/optimistic runs never count toward waste-based triggers), all once-per-node-per-window, all through `emitDiagnostic` with codes so CI and agent harnesses can consume them, and all behavioral thresholds opt-in behind `enable()` until the false-positive story is proven — I'd only consider `LEAKED_OBSERVER` for default-on in dev, since it has essentially no legitimate positives.

Not proposing to land these here — just want agreement on which codes are worth speccing before I split them into follow-up PRs.

ryansolid:
Landed on `next` in 3d2c21fe — thanks @brenelz! The shape here ended up exactly right: core reports bare facts through the null-checked hook surface, all semantics live in the engine, perf warnings ride the existing diagnostic channel, and `DEV.attribution` carries the pull-based queries. That same hook surface is now the documented substrate for external devtools.

Two additions in the landing commit:

1. **Byte-parity fix** — the lane-path async landing smuggled a `devChanged` flag out of a `try` block, and rollup's tryCatchDeoptimization retained it into prod (~9 B on the simple-app floor). `asyncEnd` now fires unconditionally outside the try and the engine derives committed-ness from its `asyncStart` snapshot (`_value`/`_time`/`_pendingValue`). Prod bundles are byte-identical to HEAD and every size scenario passes.

2. **Always-on graph-size diagnostics** — `HUGE_FAN_OUT` / `HUGE_FAN_IN` warn at 2000 live edges per node, counted on link/unlink. Unlike the opt-in attribution warnings these are always on in dev; `WIDE_SCOPE_DEPS` remains the lower-threshold, named-sources version for when the engine is enabled.

Everything is documented in the diagnostics RFC (`documentation/solid-2.0/08-dev-diagnostics.md`), including the `DEV.attribution` API and the devtools hook contract.

ryansolid:
Good list — and it's the first real stress test of the hook surface, which it mostly passes: four of the six need **zero new hook sites**. That's the property to protect as these get specced. My take per code:

**Spec now (pure engine-side):**

- **`UNSTABLE_MEMO_OUTPUT`** — best of the set. The wide-memo + fresh-object amplifier is real and today only findable by profiling. No new hooks needed: snapshot `el._value` at `recomputeStart`, sampled shallow-compare at `derivedChanged`. The suggested fix text (stable refs / pass `equals`) is exactly actionable. 👍
- **`UNGATED_EFFECT`** — I verified the premise: effects are `_equals: false` at the node level, so `createEffect(() => count() > 5, fx)` refires `fx` on every count change even when the boolean didn't flip. Also pure engine-side (snapshot at `recomputeStart`, compare at `recomputeEnd`), and the `value !== undefined` guard is the right noise filter. 👍

**Yes, but implement differently:**

- **`WIDE_WRITE`** — don't add a counter in `insertSubs`; that's a new core site. The graph-size diagnostics that landed with this PR (3d2c21fe) already maintain a live `_subCount` on every node in dev, so this is one comparison inside the engine's existing `write` hook. It does earn its place next to `HUGE_FAN_OUT` — static fan-out that never writes is harmless, written fan-out is the problem — but spec the two thresholds together so they don't double-fire on the same node.

**Defer, needs design first:**

- **`REBUILD_CHURN`** — the only one requiring a genuinely new hook (dispose path), and the false-positive story is unsolved: a keyed `<For>` receiving fresh server objects each poll legitimately disposes and recreates rows with the same runtime signature as the unmemoized `.map` mistake. Needs a spec that distinguishes wholesale rebuild of identical children from data that actually changed — and the common shapes are already compiler-warning / lint territory.
- **`LEAKED_OBSERVER`** — pushback here, including on default-on. "Live subscription whose owner is disposed" collides with the observation lifecycle settled in #3024: unowned auto-dispose nodes legitimately live without owners and are *supposed* to stay subscribed until last-reader-out. The real bug signature is narrower — a dead (`REACTIVE_DISPOSED`) node still present in a `_subs` chain — and even that needs a global dev node registry or sub-chain walks at flush boundaries, both with real dev-mode cost. Worth a design issue, not a PR.

**Skip:**

- **`HOT_SOURCE`** — redundant twice over: `HOT_SCOPE_RERUNS` already fires on the downstream scope *and names the hot write in its cause chain*, and `costs()` already ranks writes by downstream re-run time (`WriteCost`). A third channel blaming the same node adds noise, not information.

Your carried-over constraints (phase-aware, once-per-window, `emitDiagnostic` with codes, opt-in until the false-positive story is proven) are all right. The one discipline to add: **no new core hook sites without a fight** — the moment a warning needs to touch `core.ts` or the graph/dispose paths, it gets the same scrutiny this PR got. On that basis: split `UNSTABLE_MEMO_OUTPUT` and `UNGATED_EFFECT` into follow-up PRs whenever you're ready, `WIDE_WRITE` after reconciling with `HUGE_FAN_OUT`, and file the other two as design issues.

brenelz:
This is kinda what I thought. `UNSTABLE_MEMO_OUTPUT` seemed like the potential best one. Jus wasn't sure about the shallow compare cost.

`UNGATED_EFFECT` I wasn't sure of. Should it gate by default?, tho I'm guessing its intentional. I thought maybe it was less valuable because usually the signal, memo you depend on is equality check gated.

I can work on creating a few prs for valuable ones.

PR #3025, opened 2026-08-19
feat: expose server function call observers

birkskyum:
## Summary

- update `@dom-expressions/runtime` to `0.50.0-next.44`
- expose `observeServerFunctionCalls` and its types through the existing `@solidjs/web/server-functions/client` re-export
- cover request and response observation through the built Solid Web bundles

## Context

This prepares Solid 2 for [dom-expressions#570](https://github.com/ryansolid/dom-expressions/pull/570) and enables the server-function inspector in [solid-vite-plugin#314](https://github.com/solidjs/solid-vite-plugin/pull/314).

The lockfile was prepared from the simulated `0.50.0-next.44` changeset release. CI installation will remain blocked until that version is published.

## Testing

- `@solidjs/web` build
- Solid and Solid Web type generation
- server-function extension suite, 6/6 tests

PR #3027, opened 2026-08-20
feat(web): protect server functions from cross-site requests (CSRF)

birkskyum:
## Summary

- update `@dom-expressions/runtime` to the simulated `0.50.0-next.44` release
- expose the CSRF configuration through `@solidjs/web/server-functions/server`
- verify cross-site requests are rejected before dispatch and same-origin requests still work
- mark existing synthetic browser requests as same-origin

## Prerequisite

Depends on [dom-expressions#571](https://github.com/ryansolid/dom-expressions/pull/571).

This is intentionally separate from [#3025](https://github.com/solidjs/solid/pull/3025). The lockfile uses a simulated `0.50.0-next.44` package containing both pending runtime changes and should be refreshed after that version is published.

## Testing

- Solid Web build
- Solid, Solid Signals, and Solid Web type generation
- Solid Web server suite: 31 files, 282 passed, 2 skipped

### DIFF
--- a/.changeset/protect-server-functions-csrf.md
+++ b/.changeset/protect-server-functions-csrf.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/web": patch
+---
+
+Protect server-function requests from cross-site calls by default.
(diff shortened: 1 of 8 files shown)


ryansolid:
Landed on next as e9008937 (squashed locally, merged with the #3025 observer tests in the shared transport shim — it now forwards prepared Requests and stamps Sec-Fetch-Site). The implementation (dom-expressions pr-571) ships with the next runtime publish. Thanks!

PR #3031, opened 2026-08-21
perf(solid): hint lazy JS/CSS from preload

v-sayapin:
## Summary

This follows up on [dom-expressions#573](https://github.com/ryansolid/dom-expressions/pull/573), which added `modulepreload` links for the entry's static import graph.

The remaining gap was lazy route modules. On the server, `lazy().preload()` started the dynamic import but did not register any client assets. JS and CSS links only appeared once the component rendered, so route preloading could not make those assets available to the initial document.

`preload()` now also registers the module's full static JS/CSS closure with the SSR asset pipeline.

When a router knows the matched route chain before rendering, those modulepreload and stylesheet links can be included in the first shell, letting the browser discover the whole chain together.

This keeps the existing behavior around hydration and streaming:

- `registerModule` stays in the render path because its hydration key is only known there
- `NoHydration` still registers CSS, but skips client modulepreloads
- asset resolution never delays or replaces the import promise; resolver failures only warn
- async resolution keeps the boundary that initiated it
- manifest results are resolved once per module per request and shared with the render path

Assets are registered again when the component renders so CSS keeps the correct boundary attribution. Duplicate links are removed by the existing asset registry.

Only static JS and CSS imports are included. Dynamic imports remain demand-driven, so code splitting is unchanged.

This targets `next`, where the Solid 2.0 SSR asset pipeline lives.

Media assets such as fonts and images are intentionally out of scope. They can be added separately once `ResolvedAssets` has a typed hint format for `as`, `type`, and `crossorigin`.

Added unit coverage for:

- preloading assets without rendering the component
- keeping `registerModule` in the render path
- sharing one manifest resolution across repeated `preload()` calls and a later render
- `$$moduleUrl`, async resolution, boundary restoration, and resolver failure handling

Added end-to-end streaming coverage for:

- a preloaded route that never renders still contributing its modulepreload and stylesheet links to the shell
- the same flow under `NoHydration` contributing CSS without client JS

Validation:

- Full test and typecheck run with Turbo `--force` — 27/27 tasks passed, 0 cached
- Solid — 19 files, 556 tests passed
- Solid Web browser — 40 files, 416 tests passed
- Solid Web server — 30 files, 282 tests passed, 2 skipped
- Solid Web hydration — 20 files, 137 tests passed
- TypeScript, Prettier, and `git diff --check` passed

### DIFF
--- a/.changeset/lazy-preload-assets.md
+++ b/.changeset/lazy-preload-assets.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+`lazy().preload()` now registers the module's client JS/CSS as head hints on the server, not just its import. Registration used to wait for the component to render, so a router warming a matched route imported the module without emitting any `<link>` for it. The import is never gated on asset resolution, the per-request resolution cache is shared with the render path, and `registerModule` stays with the render — its hydration key is only known there.
--- a/packages/solid-web/test/server/ssr-stream.spec.tsx
+++ b/packages/solid-web/test/server/ssr-stream.spec.tsx
@@ -17,6 +17,7 @@ import {
   Dynamic
 } from "@solidjs/web";
 import {
+  NoHydration,
   createEffect,
   createMemo,
   createRenderEffect,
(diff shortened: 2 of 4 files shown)


ryansolid (approved):
Vetted locally against the current dom-expressions runtime (the linked pre-publish state): full bar green — solid server 312, web server 287, solid 561, web browser 420, hydration 139. The helper extraction is a faithful move of the f14e4ec8 livelock fix (cache upgrade-to-value, boundary restore), registerModule correctly stays with the render since only it knows the hydration key, and the NoHydration CSS-only path matches the render semantics. Nice work.

PR #3042, opened 2026-08-23
Backfill in-flight value when latest() companion is created mid-transition

### LINKED ISSUE #3041: latest() returns the committed value instead of the in-flight value on first use during a held transition
### Describe the bug

Hi I am trying to learn SolidJS for the first time with 2.0, and I ran in to what seems to either be unintuitive / inconsistent behavior of latest() or a bug.

https://stackblitz.com/edit/solidjs-templates-93va2vbn?file=src%2FApp.tsx

```tsx
import { createEffect,  createSignal,  createMemo,  Loading,  isPending,  latest } from 'solid-js';

// Self-resolving fake fetch so the demo needs no network.
const fakeFetch = (name: string) =>
  new Promise<{ name: string }>((r) => setTimeout(() => r({ name }), 1000));

  const [query, setQuery] = createSignal('pikachu');
  const pokemon = createMemo(() => fakeFetch(query()));
  const pending = () => isPending(() => pokemon());

  // Un-commenting this works around the issue
  /*
  createEffect(
    () => latest(() => query()),
    () => {}
  ); // warm the companion at mount
  */

  return (
    <div>
      <button onClick={() => setQuery('pikachu')}>pikachu</button>
      <button onClick={() => setQuery('charizard')}>charizard</button>
      <button onClick={() => setQuery('gengar')}>gengar</button>

      <Loading fallback={<div>skeleton</div>} on={query}>
        <div>Current: {pokemon().name}</div>
        {/* latest(query) is NOT evaluated during initial load: the
            short-circuit means it is first called DURING the first transition. */}
        {pending() && (
          <p>
            Now Loading: <b>{latest(() => query())}</b>
          </p>
        )}
      </Loading>
    </div>
  );
}
```

Clicking between the buttons seems to, somewhat non-deterministically, display the previous pokemon's name instead of the upcoming pokemon's name.

### Expected behavior

I would expect `latest(() => query())` to always show the latest value -- and it does, if the shown work-around is commented out.

### Screenshots or Videos

_No response_

### Platform

Linux, Firefox

### Additional context

LLM indicated the issue as being this:

In `@solidjs/signals` (`dist/dev.js` of `@solidjs/signals@2.0.0-rc.1`):

- `latest()` → `latestRead(el)` reads a per-source **latest-value companion**
  (`el._latestValueComputed`), created lazily by `getLatestValueComputed` on first use:

  ```js
  // dev.js:4096
  function getLatestValueComputed(el) {
    if (!el._latestValueComputed) {
      ...
      el._latestValueComputed = optimisticComputed(() => read(el));
      ...
    }
    return el._latestValueComputed;
  }
  ```

  Note: **no backfill.** The companion's compute `read(el)` is first executed (or forced
  by `latestRead`'s `prepareComputed`) *after* the pending write was already processed.
  Inside the held transition's render context (`stale = true` for render effects, set at
  dev.js:2587-2589), `read(el)` returns the **committed** value.

- When a write is processed during a transition, the companion is updated **only if it
  already exists** (`syncCompanions`, dev.js:4030):

  ```js
  function syncCompanions(el, value) {
    if (el._pendingSignal) updatePendingSignal(el);
    if (el._latestValueComputed) setSignal(el._latestValueComputed, value);
  }
  ```

  The app's banner is gated, so the companion does **not** exist at write time on the
  first transition; it is created afterwards with the committed value and never learns
  the in-flight write.

- Compare `getPendingSignal` (dev.js:3909), which **does** backfill:

  ```js
  function getPendingSignal(el) {
    if (!el._pendingSignal) {
      el._pendingSignal = optimisticSignal(false, { ownedWrite: true });
      el._pendingSignal._parentSource = el;
      if (computePendingState(el)) setSignal(el._pendingSignal, true); // backfill
    }
    return el._pendingSignal;
  }
  ```

  This is why `isPending()` is correct on first use during a transition while
  `latest()` is not.

### Suggested framework fix

In `getLatestValueComputed`, backfill the current in-flight value when the companion is
created, mirroring `getPendingSignal` — e.g., if the source currently carries a pending
write (`el._pendingValue !== NOT_PENDING` or the equivalent optimistic-lane read),
initialize the companion to that value rather than only to `read(el)`.

brenelz:
Fixes #3041

## Problem

`latest(() => query())` first called *during* a held transition returned the committed value instead of the in-flight one. The reporter's case: a "Now loading: {latest(query)}" banner gated on `isPending()` showed the previous pokemon's name.

`getLatestValueComputed` creates the latest-value companion lazily with `optimisticComputed(() => read(el))` and no backfill. Under an optimistic lane with a stale (render-effect) reader, `read(el)` resolves to `_value`, so the companion captures the committed value. `syncCompanions` only pushes a pending write into companions that already exist at write time — so a companion born after the write never learns it. `getPendingSignal` already backfills on creation, which is why `isPending()` was correct on first use while `latest()` wasn't.

## Fix

Seed the companion from `_pendingValue` on creation (guarded by `!hasActiveOverride`, since overrides are already authoritative through `read()`'s A17 path), mirroring `getPendingSignal`.

## Test

`latest-lazy-companion-backfill.test.ts` — lazy-creation case (failed before, passes now) plus a pre-warmed control. Full solid-signals suite green (1336 tests).

### DIFF
--- a/.changeset/latest-lazy-companion-backfill.md
+++ b/.changeset/latest-lazy-companion-backfill.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+`latest()` now returns the in-flight value when first called during a held transition. The latest-value companion is created lazily, and a write processed before its creation was never pushed into it, so the first `latest()` read inside a transition (e.g. a pending banner gated on `isPending()`) showed the committed value until the next write. The companion now backfills the pending value on creation, matching `isPending()`.
--- a/packages/solid-signals/src/core/verdict.ts
+++ b/packages/solid-signals/src/core/verdict.ts
@@ -338,6 +338,10 @@ function getLatestValueComputed<T>(el: Signal<T> | Computed<T>): Computed<T> {
     setContextInternal(null); // Detach from owner so it isn't disposed with effects
     el._latestValueComputed = optimisticComputed(() => read(el));
     el._latestValueComputed._parentSource = el; // Parent-child lane relationship
+    // Backfill an in-flight write (mirrors getPendingSignal): the companion is
+    // created lazily, possibly after the write was processed (#3041).
+    if (el._pendingValue !== NOT_PENDING && !hasActiveOverride(el))
+      setSignal(el._latestValueComputed, el._pendingValue as T);
     if (__DEV__) devTrackCompanionOwner(el);
     setContextInternal(prevContext);
     setPendingCheckActive(prevCheck);
--- a/packages/solid-signals/tests/latest-lazy-companion-backfill.test.ts
+++ b/packages/solid-signals/tests/latest-lazy-companion-backfill.test.ts
@@ -1,0 +1,78 @@
+/**
+ * #3041: latest() first called during a held transition returned the committed
+ * value. The latest-value companion is created lazily; syncCompanions only
+ * pushes an in-flight write into companions that already exist, and the
+ * companion's initial compute under a stale render effect sees _value.
+ * getLatestValueComputed now backfills _pendingValue, mirroring getPendingSignal.
+ */
+import {
+  createMemo,
+  createRenderEffect,
+  createRoot,
+  createSignal,
+  flush,
+  isPending,
+  latest
+} from "../src/index.js";
+
+it("latest() first read during a held transition sees in-flight value (#3041)", async () => {
+  let resolve!: (v: string) => void;
+  const fetch = (n: string) =>
+    new Promise<string>(r => {
+      resolve = r;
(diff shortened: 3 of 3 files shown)


ryansolid:
Landed on next as 35858660 with your authorship credited (Co-authored-by) — the stage-3 internals reshaping that merged today (`_x` extension split) meant this needed a manual port rather than a rebase, so we carried it across for you: same backfill at companion creation, mirroring getPendingSignal, your tests included and verified failing pre-fix. #3041 is fixed by it. Thanks for the fix and the clean repro!

PR #3043, opened 2026-08-23
fix(signals): suspend stale readers on an uninitialized memo held by another transition

brenelz:
## Summary
- A render effect that switches onto an async memo still in flight under a *different* transaction hit the "stale reader shows the committed value, don't entangle" path in `read()`. An uninitialized memo has no committed value, so the reader got `undefined` as if settled, direct-committed outside both transactions, and never re-ran when either landed — stuck on its old value forever.
- Repro shape (s.olid.uk playground report): a global `cache[k] ??= createMemo(async …)` shared by several inserts, `setA` then `setB` 300ms later where B's new key is the memo A already put in flight. Worked on beta.10, broke in beta.11.
- Fix: drop `owner !== el` from the `STATUS_UNINITIALIZED` throw branch so plain memos suspend the way firewall-backed store reads already did. The reader registers as a reporter and the two transactions merge and reveal together.
- Regression bisected to 7d4d0c3a: its (correct) `needsPendingCommit` gate removed an accidental pending-node push that `initTransition` had been stamping with `_transition`, which was what entangled the reader before.

## Test plan
- [x] New `tests/stale-read-uninitialized-cross-transition.test.ts` — fails without the fix; includes a control showing an *initialized* held memo still shows its committed value without entangling
- [x] `@solidjs/signals` suite: 1336 pass
- [x] `solid-js` suite: 561 pass
- [x] Original DOM repro (three `<div>` inserts under `<Loading>`) and variants end `foo1 foo1 bar1`

Note: 5 solid-web tests (`frames-behavior-claims`, `server-functions-csrf/extensions`, `welcome-status-loaded/streamed`) fail identically with and without this change on the published dom-expressions `0.50.0-next.43` — the committed `link:../dom-expressions-dr2/...` overrides in `pnpm-workspace.yaml`/lockfile from 515ff561 only resolve on a machine with that checkout.

### DIFF
--- a/.changeset/stale-read-uninitialized-cross-transition.md
+++ b/.changeset/stale-read-uninitialized-cross-transition.md
@@ -1,0 +1,16 @@
+---
+"@solidjs/signals": patch
+---
+
+Suspend stale readers on an uninitialized async memo held by another
+transition. A render effect that switched onto an async memo still in flight
+under a *different* transaction took the "show the committed value, don't
+entangle" path — but an uninitialized memo has no committed value, so the
+reader was served `undefined` as if settled and left stamped into neither
+transaction. It never re-ran when either landed and stuck on its old value
+(e.g. a globally cached async memo shared by two inputs updated 300ms apart).
+Such reads now throw NotReady like the equivalent firewall-backed read already
+did, registering the reader with the active transaction so the two
+transactions merge and reveal together. Regression from the
+`needsPendingCommit` gate in 7d4d0c3a (beta.11), which removed the accidental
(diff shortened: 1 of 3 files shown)


ryansolid:
Landed on next as 54ecdb44 with your authorship credited (Co-authored-by) — manually ported over today's stage-3 internals merge (the read() site moved to the `_x` error layout). Same one-condition fix: uninitialized memos held by another transition now suspend stale readers instead of serving undefined, merging the transactions at settle. Your tests included and passing. Thanks — the bisect to the needsPendingCommit gate made this an easy review.

PR #3045, opened 2026-08-23
perf(signals): skip empty companion walks

### LINKED ISSUE #3038: 2.0.0-rc.1 | flush cost scales with total materialized store signals (async-companion walk in sync-only apps)
### Describe the bug

In a sync-only app, every update of a store-producing computed pays a walk over the store's entire materialized-signal set — cost is O(materialized signals) per store update per flush, regardless of what changed.

Mechanism (from @solidjs/signals rc.1 source):

- Every store signal node is created with `firewall = <the store's computed>` and prepended to `firewall._child` — one node per `(object, key)` any reader ever touched in a tracked scope. A large keyed record projection therefore accumulates a child per leaf its subscribers read.
- After **every** update of that computed, `clearStatus()` calls `GlobalQueue._updateChildCompanions(el)`, which walks the entire `_child` chain checking `child._pendingSignal || child._latestValueComputed` — the `isPending()` / `latest()` verdict companions.
- In an app that never calls `isPending()`/`latest()`, no companions exist: every check is false and the walk is pure overhead. The hook is installed unconditionally at module load, so the `!== null` guard never helps a bundled app.

Where we hit it: a flow-graph library with 1,600 nodes / 1,599 edges as keyed record projections, each row read by its own wrapper component (~10^5 materialized leaves across two records). Dragging ONE node — one memo re-run, five DOM attribute writes — cost **17.6 ms per mousemove** (prod build, visible tab). The JS Self-Profiling API attributed ~60% of flush samples to `updateChildCompanions` at every scale we measured, while run counters confirmed our derives did O(changed) work. Restructuring so leaf signals hang off per-row computeds (~30 children each) dropped the same drag to **1.37 ms**, confirming the attribution.

Standalone measurement (script in the repo below): one keyed record projection, 1,600 rows, per-row memos so the derive is O(changed). The workload is identical in every scenario — 60 flushes, each writing one leaf of one source row; only the number of leaves readers touch differs:

| readers | ms/flush (prod) | ms/flush (dev) |
| --- | --- | --- |
| none (~0 signals/row) | 0.28 | 0.36 |
| narrow (~4 signals/row) | 0.77 | 0.64 |
| wide + deep (~20 signals/row) | 1.50 | 1.60 |

5.4× for the same data change, growing with reader width; nested reads amplify it because each intermediate object materializes its own node.

Suggested fix: maintain a companion count (or has-companions flag) on the firewall, updated at companion creation/disposal, and skip the walk when zero. Sync-only apps then pay nothing; async apps walk only where companions actually exist.

https://github.com/dsnchz/solid-flow/tree/repro/solid-rc1-issues (script: `repro/companion-walk.mjs`, writeup: `repro/ISSUE-1-companion-walk.md`)

```sh
git clone https://github.com/dsnchz/solid-flow && cd solid-flow
git checkout repro/solid-rc1-issues
bun install
node --conditions=browser repro/companion-walk.mjs
```

(The script itself only depends on `solid-js@2.0.0-rc.1` — it runs the same from any directory with that package installed.) To see it at app scale: `bun run dev`, open `http://localhost:3000/?example=StressTest&x=40&y=40`, drag a node while profiling — `updateChildCompanions` dominates the flush.

<details>
<summary>Context pack for automated analysis</summary>

`AGENTS.md` at the root of the repro branch is a self-contained context pack: commands with expected outputs and the source coordinates inside `@solidjs/signals` `dist/dev.js` (searchable internal names: `updateChildCompanions`, `clearStatus`, `signal(v, options, firewall)` prepending to `firewall._child`, `GlobalQueue._updateChildCompanions` installed unconditionally at module load), plus the app-scale profiling evidence and the sub-store restructuring that confirmed attribution (17.6 ms → 1.37 ms per drag move with identical DOM output).

</details>

### Expected behavior

Flush cost for a single-leaf write should scale with what changed, not with how many leaves readers have ever touched — or at minimum the companion walk should be skipped when no `isPending()`/`latest()` companions exist below the computed.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS 15 (Darwin 24.6.0)
- Browser: Chrome (numbers above from node; browser profiling via JS Self-Profiling API)
- Version: solid-js 2.0.0-rc.1 / @solidjs/signals 2.0.0-rc.1

### Additional context

Found migrating `@dschz/solid-flow` (a SolidJS port of React Flow / Svelte Flow) to 2.0. We've since restructured into per-row sub-store projections and reached parity with our Solid 1.x version, so this doesn't block us — filing because the cost shape (per-update, proportional to total materialized signals) will hit any sync-only app with one large store and many readers, and it's invisible without a profiler.

dawNotPoi:
## Summary

- track whether a firewall child has ever materialized an async companion
- skip the child-chain companion walk while no `isPending()` or `latest()` reader exists
- keep the walk enabled once either companion is created, with regression coverage for both paths

Fixes #3038

- `pnpm --filter solid-js build`
- `pnpm --filter solid-js test`
- `pnpm --filter solid-js test-types`
- Prettier check for the changed files
- Re-ran the issue benchmark across no-reader, narrow, and wide/deep store shapes; the no-reader path no longer scales with the materialized child chain

### DIFF
--- a/.changeset/calm-signals-walk.md
+++ b/.changeset/calm-signals-walk.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Improve store update performance when async companion signals are not observed.
--- a/packages/solid-signals/src/core/async.ts
+++ b/packages/solid-signals/src/core/async.ts
@@ -710,7 +710,7 @@ export function clearStatus(el: Computed<any>, clearUninitialized: boolean = fal
   // Update pending signal for isPending() reactivity (companions only exist
   // once the verdict layer created them, which installs the hooks).
   if (el._pendingSignal || el._latestValueComputed) GlobalQueue._updatePendingSignal!(el);
-  if (el._child && GlobalQueue._updateChildCompanions !== null)
+  if (el._hasChildCompanions && GlobalQueue._updateChildCompanions !== null)
     GlobalQueue._updateChildCompanions(el);
   if (el._notifyStatus) el._notifyStatus();
 }
@@ -751,7 +751,7 @@ export function notifyStatus(
       el._error = error;
     }
     GlobalQueue._updatePendingSignal !== null && GlobalQueue._updatePendingSignal(el);
-    if (el._child && GlobalQueue._updateChildCompanions !== null)
+    if (el._hasChildCompanions && GlobalQueue._updateChildCompanions !== null)
       GlobalQueue._updateChildCompanions(el);
   }
 
--- a/packages/solid-signals/src/core/core.ts
+++ b/packages/solid-signals/src/core/core.ts
@@ -297,7 +297,7 @@ export function recompute(el: Computed<any>, create: boolean = false): void {
       el._pendingSources !== undefined ||
       el._pendingSignal !== undefined ||
       el._latestValueComputed !== undefined ||
-      el._child !== null
+      el._hasChildCompanions
     )
       clearStatus(el, create);
     // _optimisticLane is only ever assigned by engine paths.
@@ -562,6 +562,7 @@ export function computed<T>(
     _value: (loading ? options!.loadingValue : undefined) as T,
     _height: 0,
     _child: null,
+    _hasChildCompanions: false,
     _nextHeap: undefined,
     _prevHeap: null as any,
     _deps: null,
@@ -622,6 +623,7 @@ export function createEffectNode<T>(
     _value: undefined as T,
     _height: 0,
     _child: null,
+    _hasChildCompanions: false,
     _nextHeap: undefined,
     _prevHeap: null as any,
     _deps: null,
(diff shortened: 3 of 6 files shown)


thedanchez:
Nice! Hey @ryansolid, does this PR look good? Fingers crossed that it does 🙏🏾

ryansolid:
Thanks @dawNotPoi — this was the right fix and the right analysis. An equivalent gate landed on next as part of the stage-3 core work that just merged (debc22b9): same sticky companions-exist flag set at the two companion-creation sites, gating the same three walk sites. The one difference is storage — next packs it into the existing `_config` bitfield rather than adding a `_hasChildCompanions` field to every computed/effect literal, which matters because node literal shape/size is a hot-path concern (several recent regressions traced to literal growth).

Your walk-counting regression test approach was solid; the shipped gating is covered by equivalent tests. Closing as superseded by the merged fix — #3038 is now closed. Appreciate the contribution, and sorry for the racing-you-to-it timing.

PR #3046, opened 2026-08-24
refactor(web): move server function runtime into `@solidjs/web`

birkskyum:
## Summary

- move the server-function implementation and declarations into `@solidjs/web`
- expose Solid-owned protocol metadata and flash helpers from the core entry
- build the client, server, and rich-argument entries from Solid-owned source
- inject the server-function protocol into the generic dom-expressions frame adapters
- move the complete protocol test suite into Solid
- preserve the existing public API and generated package output

The runtime source is synchronized with [dom-expressions `next` at `496be4ae`](https://github.com/ryansolid/dom-expressions/tree/496be4ae/packages/runtime/src/server-functions).

## Why

`dom-expressions` is the framework-neutral rendering layer. Server functions are not rendering primitives. They define Solid's client/server protocol, including reference registration, request encoding, dispatch, errors, redirects, revalidation, CSRF, progressive enhancement, and observability.

The compiler and public API already identify this as Solid functionality through `@solidjs/web/server-functions`. Keeping the implementation in dom-expressions makes Solid's protocol an implicit responsibility of a lower-level renderer and exposes Solid-specific concepts to every dom-expressions consumer.

This change aligns implementation ownership with the existing public boundary. Generic framing and serialization primitives remain in dom-expressions, while Solid supplies the server-function protocol adapter.

The dom-expressions copy remains temporarily so this can land without an atomic cross-repository release. Its removal is prepared in [ryansolid/dom-expressions#577](https://github.com/ryansolid/dom-expressions/pull/577). This PR has been tested against both the current dom runtime and #577, so no third migration PR is required.

## Verification

- full workspace build
- 1,042 `@solidjs/web` tests passed, 2 skipped
- `@solidjs/web` build, declarations, and type tests against the current dom runtime
- the same build, declarations, type tests, and runtime tests against dom-expressions #577
- ESM import and packaged-type integration tests
- npm package dry run

## CodSpeed

CodSpeed could not find a successful run for the actual `next` base at `8a44c9e`, so it compared this branch to older `3dbf12b` across different runtime environments. The reported changes are all signals and store benchmarks, which this PR does not modify.


birkskyum:
Superseded by 
- https://github.com/solidjs/solid/pull/3049

Issue #2963, opened 2026-08-01
2.0.0-beta.28: <Show> when={isPending || !!promise} or when={promise} causes stale reads


dangkyokhoang:
### Describe the bug

Might be relevant issues so I create 1 tracking issue:

1: `<Show when={isPending(state) || !!state().promise}>` inside a `<>` fragment — Shows evaluate with stale state: mutually exclusive conditions both become true, and resolved data doesn't show until next submit. Works correctly with `<div>` or `<Show when={isPending(() => !!state().promise)}>`.
2+3: `<Show when={state().promise}>` — two bugs: (1) Show's memo is async-shape aware, so receiving a thenable creates an implicit transition that defers signal updates — promise is invisible during loading. (2) Second click causes stale accessor error that halts the entire reactive system.

On the 2+3, though I might be expected to write `<Show when={!!state().promise}>` instead, I think this should still be considered a bug. The mechanism of this issue might be that `<Show when>` wraps promise in a memo and the rest is related to https://github.com/solidjs/solid/issues/2802.

Because of how the createMemo integrates deeply with the transitions and the reactive system, which causes it to have some unexpected behaviors or "rules" users should be aware of, I think solid should not try to implicitly wrap `when` value in a memo. And as you commented in https://github.com/solidjs/solid/issues/2957, it seems solid by default might heuristically decide to wrap other component props in memos as well. I think this work (wrapping things in memos) should be on the user, not the framework. Less burden on the framework, and more predictable behavior for users.

Edit: Or if solid should wrap some props in memos because benchmarks say so, maybe use `sync: true` for those memos?

### Your Example Website or App

https://stackblitz.com/edit/nitrojs-nitro-azkekile?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the link above
2. Open example page
3. Select issue 1 or 2
4. Click submit button
5. Wait and click submit button again

### Expected behavior

Should be no issue.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

ryansolid:
Some of this is by design. Some of this isn't providing good enough errors. I will see what we can do here.

ryansolid:
Went through all three reported behaviors — one was a real (and sneaky) bug, now fixed on `next`.

**1. The stale sibling-`Show` (fragment works with `<div>`, breaks without) — real bug, fixed.**

[...]

PR #3050, opened 2026-08-25
fix(signals): replay pending-gated readers stranded by a transaction's silent value promotion

brenelz:
## Summary

Follow-up to #3041 — the `latest()` companion backfill itself is fixed in rc.2 (verified against the published tarball), but the reported "still one value behind sometimes" turned out to be a distinct, long-standing bug (reproduces on rc.1, rc.2, and `next`).

**Symptom:** content whose async read is short-circuited behind a pending gate —

```tsx
{isPokemonPending() ? "loading..." : pokemon().name}
```

— lands showing the **previous** committed value and never corrects until the next write. A flat effect that reads `pokemon()` unconditionally is fine, and user-tier `createEffect` is unaffected, which is why it only bites "sometimes".

## Root cause

When the pending gate flips off during the landing flush, the render effect re-runs and reads the async memo **for the first time**. That read happens under the isPending companion's optimistic lane (`laneAsyncPending` assigns the source's `_optimisticLane`, and settle keeps it), so `laneReadsCommitted` serves the committed old value. The staged `_pendingValue` then promotes via `commitPendingNode`, which deliberately never re-notifies — stranding the reader on the old value.

The engine already has a replay contract for exactly this shape (gated subs recorded at read time, replayed by `finalizePureQueue` after commit — #2963), but the `owner === el` branch of `laneReadsCommitted` only recorded when **no** transaction was active, so transaction landings fell through the hole.

## Fix

In `laneReadsCommitted`, whenever a committed-view read hides a staged `_pendingValue`, record the reader into `(activeTransition ?? globalQueue._batch)._gatedSubs`. The existing replay re-runs it right after promotion, within the same synchronous flush — the effect sequence goes `pending → idle:old → idle:new` with no paint in between.

## Tests

- `packages/solid-signals/tests/pending-gated-landing-replay.test.ts` — signals-level regression pinning the exact value sequence (fails before the fix with the effect stuck on `idle:v0`).
- `packages/solid-web/test/ispending-gated-landing.spec.tsx` — the DOM shape through `<Loading>` (failed before the fix: stuck on `pikachu` after the `charizard` transition landed).
- `packages/solid-web/test/latest-ispending-issue-3041.spec.tsx` — control spec pinning the full #3041 app shape (banner + content across sequential transitions and rapid mid-flight clicks), which passed before and after.

Full suites green: solid-signals 1351, solid-web 425, solid 561.

### DIFF
--- a/packages/solid-signals/src/core/optimistic.ts
+++ b/packages/solid-signals/src/core/optimistic.ts
@@ -235,18 +235,33 @@ function laneReadsCommitted(el: OptimisticNode, owner: OptimisticNode, c: Comput
     el._x?._overrideValue !== undefined ||
     !!el._x?._optimisticLane ||
     !!((owner as Computed<any>)._statusFlags & STATUS_PENDING)
-  )
+  ) {
+    // The committed view hides a staged in-flight value that will promote
+    // silently (commitPendingNode never re-notifies). gatedRead records plain
+    // signals for replay at commit; async memos are excluded from it by the
+    // `_fn` check and reach here instead — a lane-assigned source whose async
+    // already settled (laneAsyncSettled keeps _optimisticLane) served its
+    // committed value to a reader that never re-ran after the landing, so a
+    // pending-gated branch stayed one value behind permanently (#3041
+    // follow-up). Record the reader under the same replay contract.
+    if (el._pendingValue !== NOT_PENDING)
+      (activeTransition ?? globalQueue._batch)._gatedSubs.add(c);
     return true;
+  }
   if (owner === el && stale && (c as Computed<any>)._x?._parentSource !== el) {
-    // The committed view can hide a same-tick ambient write (a lane member —
-    // even just an isPending companion flip — puts the reader "under a lane"
-    // against unrelated plain writes). With no transaction the write commits
-    // at THIS flush's end with no re-delivery, so record the reader for
-    // replay at commit — the same contract gatedRead provides when a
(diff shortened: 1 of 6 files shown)


brenelz:
Pushed a second commit (04893aa9) covering a related "one value behind" path found while verifying: with a real task gap after a landing (human click timing), the next transition's `latest()` banner showed the previous transition's target — e.g. charizard → gengar → charizard showed "Now Loading: gengar".

Two cooperating defects in the shadow companion:
1. The shadow auto-disposes when its gated reader unmounts at a landing, but `_latestValueComputed` keeps pointing at the corpse; the next sync write lands in it and a later read revives it from the committed view. `getLatestValueComputed` now recreates a disposed shadow (re-running the #3041 backfill).
2. `optimisticWrite` equality-compared against a dirty node's stale `_value`, swallowing a write that returned the shadow to that stale value.

Regression spec drives the exact click sequence with macrotask gaps through `<Loading>` in solid-web. All suites green (signals 1351, solid-web 426, solid 561).

ryansolid:
Merged to next as 7f51871b + de5a5892 (rebased across the dom-expressions absorption — package paths moved to `packages/signals`/`packages/web` — with a changeset added, 67173988). Validated on the absorbed layout: both regression suites, full signals (1351) + web (599), and the size guard all green; treeshaking boundaries unaffected. Great find on both defects — the landing-flush first-read hole and the disposed-shadow revival were subtle. Thanks @brenelz!

PR #3052, opened 2026-08-25
fix: preserve wrapped module-level server exports

birkskyum:
Addresses the compiler bug documented in [RFC 10](https://github.com/solidjs/solid/blob/next/documentation/solid-2.0/10-server-functions.md#compiler-implications) and originally reported in [solidjs/solid-start#2271](https://github.com/solidjs/solid-start/issues/2271).

>**One pre-existing bug to fix regardless:** with a *module-level* `"use server"` directive, a wrapped export (`export const x = wrapper(async () => ...)`) is silently dropped from the client build — only direct function exports become references. This gets more visible with `GET(fn)` as the blessed idiom. Minimum: a diagnostic; better: extract the inner function and preserve the call in client output.

Module-level `"use server"` exports that wrap their function in calls such as `GET(async () => ...)` were omitted from the client build because only direct function initializers were traced.

This change:

- extracts a single function nested inside generic call wrappers
- preserves the wrapper chain in both server and client output
- retains only wrapper imports and local dependencies required by the client
- continues to exclude server-only implementation code
- does not special-case `GET` or any other wrapper

Adds compiler fixtures for nested wrappers, local dependencies, development metadata, and server-only import pruning. Updates the documented compiler behavior and includes a patch changeset.

Testing: `@solidjs/compiler` test suite, 3,924 passing.

### DIFF
--- a/.changeset/preserve-module-server-wrappers.md
+++ b/.changeset/preserve-module-server-wrappers.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/compiler": patch
+---
+
+Preserve wrapper calls around module-level server function exports.
--- a/documentation/solid-2.0/10-server-functions.md
+++ b/documentation/solid-2.0/10-server-functions.md
@@ -33,7 +33,7 @@ export async function addTodo(title: string) {
 
 Two verified compiler behaviors anchor everything below:
 
-1. **Function-level directives round-trip wrapper calls.** `export const getData = GET(async (id) => { "use server"; ... })` compiles by swapping only the function expression — the surrounding `GET(...)` call survives in both server and client output. A pure-runtime wrapper needs no compiler support.
+1. **Wrapper calls round-trip.** `export const getData = GET(async (id) => { "use server"; ... })` compiles by swapping only the function expression, so the surrounding `GET(...)` call survives in both server and client output. Module-level directives preserve wrapped exports the same way and keep only the wrapper dependencies needed by the client.
 2. **Anything referenced only inside a `"use server"` body never reaches the client.** The extraction replaces the body with a reference, and the directive pass’s orphan-scoped dead-code elimination removes now-unused imports and bindings — schema libraries, database handles, helper imports all vanish from client output. **The directive boundary is itself the privacy mechanism.**
 
 One architectural fact worth stating: **a wrapper wraps the *reference*, not the registered function.** `registerServerReference(id, fn)` registers the raw inner function for HTTP dispatch before any wrapper runs, so wrapper-position code can only affect the client transport and the in-process callable — never HTTP dispatch. This is why anything that must run on the dispatch path (validation, auth, logging) belongs *inside* the body, and why the declaration surface (`GET`) is transport-only.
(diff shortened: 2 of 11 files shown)


ryansolid:
Reviewed and verified locally — CI's compiler-binaries workflow didn't trigger on this PR, so I ran the full suite on the branch myself: cargo tests plus all 3,924 fixture tests pass, and I probed the transform directly with a few adversarial inputs. Nice work; this matches the "better" option RFC 10 sketched (extract the inner function, preserve the wrapper call in client output), the dependency preservation correctly follows transitive locals and prunes server-only imports, and function-declaration dependencies survive via the normalization to const initializers.

Two residuals I confirmed while probing, tracked as follow-up rather than blocking:

1. A wrapper receiving **two** function arguments (`export const getUser = withCallbacks(async id => id, err => log(err))`) still reproduces the original bug: the export is silently dropped from the client build (`valid: true`, no diagnostic, empty module). RFC 10's minimum bar for untraceable exports is a diagnostic — that should now be straightforward on top of this structure, since `nested_function_count(init) > 1` is exactly the detectable case.

2. Wrapper args pull transitively-referenced module locals into the client bundle — including env-derived values (`const secret = process.env.API_SECRET; const config = { secret }` ships to the client when `config` is a wrapper arg). That's inherent to evaluating the wrapper client-side, but it deserves loud documentation for module-level `"use server"` authors, since bundlers inline env vars.

Merging. Thanks @birkskyum!

ryansolid:
Follow-up: we've reversed this in 2e527443 — not back to the silent drop this PR rightly flagged, but to a compile error. Wrapped module-level exports are now rejected outright, with the error directing to the function-level directive form.

Merging this was my call and it was too hasty — the bug report was correct (silently dropping exports was unacceptable), but the fix direction locked in an invariant violation we only fully worked through afterwards. Module-level `"use server"` promises that none of the module's code runs on the client, and that each export *is* the registered server function. Transplanting the wrapper call into the client build breaks both: wrapper code (and its reachable locals — see #3062's env-var example) executes client-side, while HTTP dispatch still invokes the raw registered function, so the wrapper silently doesn't apply to real calls. There's also no honest way to decide "which argument is the function" inside an arbitrary call — that's the pattern-matching `"use server"` was adopted to end.

The supported composition point is the function-level directive, where the wrapper runs in shared code around each side's reference:

```js
import { GET } from "@solidjs/web/server-functions";
export const getUser = GET(async id => {
  "use server";
  return db.users.find(id);
});
```

Plain aliasing and separate declaration/export under module-level directives are unaffected. The invariant and rationale are now recorded in RFC 10 (compiler contract + alternatives-considered). Thanks for surfacing the original hole — the end state is strictly better for it.

ryansolid:
Final resolution for this one, after two more design rounds (none of it released, so no ecosystem impact):

1. **This PR's transplant approach was reverted.** Extracting the inner function and hoisting the wrapper call into the client build had three problems we couldn't live with: it executes server-module code on the client (module-level files promise the opposite), the wrapper never applied to HTTP dispatch (registration grabbed the inner function before the wrapper ran, so `withMeta`/`GET` semantics silently differed between SSR and real network calls), and deciding "which argument is the function" reintroduces the pattern-matching that the positional `"use server"` directive was adopted to end.

2. **We briefly replaced it with a compile error** forbidding wrapped module-level exports outright — then realized that threw away the sound part of your report to enforce a syntactic reading of an invariant that was always about *values*.

3. **What shipped instead (2f01f23e): export-value registration.** In a module-level `"use server"` file, the server build now registers each export's *evaluated value* whole:

   ```js
   "use server";
   export const getUser = withValidation(schema, async id => { ... });
   ```

   registers `withValidation(schema, ...)` — the composed function — so the wrapper applies to **every call path**: HTTP dispatch and in-process SSR calls alike (which the transplant never achieved). The client build emits bare `createServerReference(id)` exports, so wrappers, schemas, and helpers are server-only by construction. The compiler never inspects the initializer's shape; `registerServerReference` throws at module eval if the value isn't a function, so a stray non-function export fails the server boot loudly instead of being silently dropped. Anonymous default expressions (`export default withDelay(fn, 400)`) now register too.

   One intentional asymmetry: client-transport declarations like `GET` are client-side API and have no meaning *inside* a module-level file — declare transport at the consumption site or use a function-level directive.

Thanks for surfacing this — the wrapped-export gap was real, and the fix ended up simpler and stronger than what any of the intermediate designs proposed. Full rationale recorded in RFC 10 (compiler contract + alternatives-considered).

PR #3059, opened 2026-08-25
feat(web): expose server function invocation metadata

### LINKED ISSUE #3058: RFC 10 follow-up: expose declaration metadata to invocation hooks
## Problem

The server-function metadata channel is available to client transport policy through `prepareRequest(init, { id, meta })`. The server-side `wrapInvocation` context currently exposes `id`, `args`, `event`, `request`, and `direct`, but not declaration metadata.

This leaves external middleware and policy packages without a stable way to select per-function behavior. They must either repeat policy inside every function body or key global policy by generated function ids. The registered HTTP implementation is the raw function, so `getServerFunctionMetadata(getServerFunction(id))` does not recover metadata attached to the server reference.

## Proposed scope

Make declaration metadata available during server invocation, for both HTTP dispatch and direct SSR calls. For example:

```ts
wrapInvocation(run, {
  id,
  meta,
  args,
  event,
  request,
  direct
});
```

This may require server-side metadata bookkeeping keyed by function id. The metadata should remain local declaration data and should not be serialized over the wire.

## Why core

This does not ask core to implement middleware chains, authentication, or validation. It completes the mechanism that lets external packages build those policies while preserving direct-call and HTTP-dispatch parity.

Related specification: [RFC 10](https://github.com/solidjs/solid/blob/next/documentation/solid-2.0/10-server-functions.md#the-runtime-solidjswebserver-functions).

birkskyum:
## Summary

- add declaration `meta` to the server `wrapInvocation` context
- keep metadata available for HTTP and direct SSR calls, including built-in declaration wrappers
- update RFC 10 and add regression coverage

## Why

Client `prepareRequest` already receives declaration metadata. Server policy hooks could only select behavior by generated function ids. This completes the same mechanism without serializing metadata or adding middleware policy to core.

Closes #3058.

## Testing

- `pnpm --filter @solidjs/web test:server` (290 passed, 2 skipped)
- `pnpm --filter @solidjs/web types`
- `pnpm --filter @solidjs/web test-types`


ryansolid:
Reviewed — this completes the declare-on-function/react-in-hook symmetry cleanly: `prepareRequest` and `wrapInvocation` now read the same declaration bag, and keying the map at reference creation works because `withMeta`/`GET` mutate the metadata object in place (`Object.assign`), so the captured entry can never go stale under normal composition.

One behavior worth noting for the record, not blocking: `live()` copies the bag (`{ ...meta, live: true }`) and this PR re-points the map at that copy. In the blessed declaration order (`live` outermost, composed in one expression) that's exactly right — `wrapInvocation` sees the final declaration including `live: true`. The unidiomatic case — calling `withMeta` on the *original* reference after `live()` was already declared — would mutate the original bag while the map holds live's copy. Nobody should write that, and the docs already prescribe declaring live outermost, but if it ever bites we can make `live` share the bag rather than copy it.

Test coverage hits both dispatch paths with merged `withMeta` + `GET` metadata, and CI's web suite ran it green. Merging — thanks @birkskyum!

ryansolid:
Update: I merged this too hastily, without Ryan's approval — that's on me, not on the PR. It's been reverted in 2c90ae6b while the design gets decided properly. Apologies for the whiplash, @birkskyum.

The design discussion surfaced a real objection to the shape, worth capturing here:

Reference decorations (`withMeta`, `GET`) are **call-path-scoped** — they mean something when a call goes *through* the decorated reference. That semantics holds on the two paths where it's genuine: the browser calls through the client wrapper (feeding `prepareRequest`), and initial-SSR direct calls go through the server-side decorated reference. But HTTP dispatch is different: the wire carries only an id, and the handler dispatches to the raw registered function — the decorated reference was never in the call path. Keying a module-level map by id fabricates declaration semantics for a call that never passed through the declaration.

That's not just aesthetics: the map fills as a side effect of module evaluation, so if a server build never imports the decorated export (a handler only needs the registry), the entry is empty and any `wrapInvocation` policy keyed on `meta` is silently off. Policy that depends on tree-shaking outcomes is a footgun, especially for auth-shaped metadata. The `live()` overwrite (last declaration evaluated wins the id slot) has the same accidental-contents character.

The direction that survives the objection: per-function *server policy* metadata should be **registration-scoped** — declared where the dispatch path structurally guarantees it exists (attached at `registerServerReference` time, or an equivalent the handler can't dispatch without). Then `getServerFunctionMetadata(getServerFunction(id))` — the read your issue correctly noted is broken today — just works, `wrapInvocation`'s context doesn't need to grow, and `withMeta` stays purely a call-path transport decoration. Two channels because they're two different things.

If you're up for reworking it in that shape, #3058 is reopened for the design.

PR #3080, opened 2026-08-27
feat(web): let an app supply the fetch server function calls go through

frenzzy:
Follow-up to #3076, and the userland half of the addressing question in [#3072](https://github.com/solidjs/solid/discussions/3072).

`configureServerFunctionsClient({ fetch })` replaces the function the transport sends every server-function request with — retries, telemetry, a test double, or a route of the app's own. `null` restores the global.

### Why a seam and not a second address

An app that wants a cache-shaped url — an argument in a path segment, a tenant in a prefix — can't get one from a second built-in address format: that has to be taught to every gate that recognises a call (the runtime, the plugin's dev middleware, the generated dispatch gate, the router's action-url interception). It doesn't have to be one. The handler takes a web `Request`, so an app route that rewrites into the canonical address dispatches like any other call:

```js
// GET /api/stories/42/2
export async function GET({ request }) {
  const [, , , story, page] = new URL(request.url).pathname.split("/");
  return handleServerFunctionRequest(
    new Request(new URL(serverFunctionUrl(getStory.id, [story, Number(page)]), request.url), request)
  );
}
```

That already works on `next`. The client's side of it didn't exist — `prepareRequest` sees everything about a request except the part that addresses it — and this adds it, without any gate downstream learning a second format.

### The shape

Always `(address, init)`, the address relative to the document, so an ordinary fetch wrapper drops in and `parseServerFunctionUrl` reads the id back out for telemetry.

One behaviour change: the path taken when call observers are installed used to hand the global `fetch` a `Request` instead. Leaving that difference would mean a wrapper written against the documented shape misrouting the moment devtools attach. Observers now see a reconstruction of the dispatched request, built without a streaming body — reconstructing one consumes it before the send can use it.

Three things a type can't say, so the doc does: forward `init` (the call's `signal` rides on it; dropping it voids the caller's abort and a live source's teardown), keep the call same-origin, hand the response back unread.

### Checked by hand

Every dispatch path goes through the seam with a `string` address the id parses back out of: plain POST, GET-declared reads, the long-url POST fallback, bound `action.with` calls, an explicit `base`, and `live()` including reconnect. A retrying wrapper re-sends string, FormData, Blob and File bodies, which handing over a `Request` could not have done.


ryansolid (commented):
Reviewed with the dispatch map in hand: the funnel claim checks out. Every call shape — plain POST, `GET`-declared reads, the long-URL POST fallback, and `live` including its reconnect — reaches the wire through `fetchServerFunction` → `createRequest`, and after this change that function has exactly one exit. The observer-path normalization is right too: a wrapper written against the documented shape must not misroute the moment devtools attach, and read-only observers seeing a reconstructed `Request` instead of the dispatched object costs nothing.

The design also lands where the addressing question settled in #3072: the app owns its URL shape by rewriting *into* the canonical address, and no downstream gate learns a second format. Good.

One change request, one suggestion:

**Type the seam as what it promises.** The option is `typeof globalThis.fetch`, but the contract this PR documents — "always called as `(address, init)`" — is narrower and is the whole point of the seam. As typed, a wrapper must handle `Request | URL` inputs it will never receive, which is why the specs need `address as string` casts. Something like

```ts
fetch?: ((address: string, init: RequestInit) => Promise<Response>) | null;
```

keeps every existing fetch-shaped function assignable (wider parameters, contravariance) while letting a hand-written wrapper use the address without casting. The type is the second place the guarantee lives; right now it contradicts the doc.

**Doc nit:** "what a wrapper owes the transport" names `init.signal`, but the per-call options from `invoke` (`keepalive`, `priority`) ride on the same init now that the invocation channel standardizes them. Rather than growing the list one field at a time, state it as: forward `init` wholesale — everything the transport decided about the call travels on it. Worth one more line that the seam is the *client transport's* exit: in-process server-side calls (`invoke` on the server) never fetch and never see it.

CodSpeed is the same fork artifact as on #3088 — not yours.

frenzzy:
Both applied: the type is `((address: string, init: RequestInit) => Response | Promise<Response>) | null`, and the doc says forward `init` wholesale instead of naming `signal`.

Changes since your review, mostly from an edge-case pass:

- The request rebuilt for call observers was built from the live `init`, so a streaming body got consumed before the send could use it — a call that worked without devtools failed with them attached. It is rebuilt without a streaming body now, and that is the one thing here a test pins by itself.
- I added runtime guards for a non-callable option and for what the wrapper answers with, then removed them. The type rules out both, and one ran on every request.
- Dropped the server-side no-op configurator I had added: `configureServerFunctionsClient` has no server mirror on `next` either, so that gap is not this PR's to close.
- Reverting each part of the change one at a time showed `null` still reset the option with the normalisation removed — `config.fetch || fetch` already answers for it — so that went too, along with a test that died on exactly the same reverts as its neighbour.

ryansolid (approved):
Shape verified against the roadmap: Stage 8 carriers stay response-framings over POST so the live path (reconnects included) rides this same seam; single-flight leans only on the response-unread contract this doc already states; and the batching designs keep one-call-one-fetch (caller-side coalescing; a batch fn is itself a plain server function). Merging once checks pass — the earlier red check was a size-budget breach from an unrelated store fix on next, ratcheted separately.

PR #3084, opened 2026-08-28
fix(web): skip server-injected head-manager tags in the hydration claim walk

### LINKED ISSUE #3081: `useHead` + shell-authored `<head>` children: SSR order breaks whole-document hydration
### Describe the bug

When a full-document shell component renders its own children inside `<head>` AND anything
registers a tag through `useHead`, SSR emits the managed tag FIRST — ahead of everything the shell
rendered there:

```html
<head><meta charset="utf-8" data-dh="charset"><meta name="shell-owned" content="first"><!--$-->…
```

The shell's compiled head traversal is positional: it takes `head.firstChild` expecting its own
first node, and walks hydration markers from there. The prepended managed tag shifts every read by
one, the walk ends on a null read —
`TypeError: Cannot read properties of null (reading 'nextSibling')` — and hydration for the WHOLE
document dies with it: reactivity halts (`[REACTIVITY_HALTED]` in dev), and the demonstrated
post-hydration client write never reaches its bindings. Dev also reports the shell's own dynamic
head child as an unclaimed server-rendered node.

Commenting out the single `useHead` call makes the identical build hydrate clean — one line is the
whole difference. Reproduced on 2.0.0-rc.3 and on 2.0.0-rc.0 with its contemporaneous compiler
stack (production builds and dev alike), so this is not a recent regression; it surfaces for
anyone whose head registrations coexist with shell-authored head children — e.g.
`@tanstack/solid-router` since it adopted the useHead registry, which is how we hit it (every page
dead on hydration).

### Expected behavior

Tags registered through `useHead` and head children authored directly by the shell component
coexist: the document hydrates, whatever order the server chooses to emit them in.

### Reproduction

<https://github.com/TylerRick/solid-usehead-shell-head-hydration-repro> — clone, `pnpm install`,
`npx playwright install chromium`, `pnpm build`, `pnpm serve`, `pnpm check` (or just open
http://localhost:5602). vite 8.2.1 + vite-plugin-solid 3.0.0-next.27, solid-js/@solidjs/web
2.0.0-rc.3, `renderToStream` on the server, `hydrate(…, document)` on the client, production
build.

The heart of it — a component inside `<head>` registers one tag; the shell owns the rest
(`src/App.tsx` in the repo, imports and entry points included there):

```tsx
function HeadRegistrar() {
	useHead({ tag: 'meta', props: { charset: 'utf-8' }, key: 'charset' });
	return null;
}

const [started, setStarted] = createSignal(false); // set from the client entry after hydrate()

	return (
		<html lang="en">
			<head>
				<meta name="shell-owned" content="first" />
				<HeadRegistrar />
				<Show when={true}>
					<meta name="shell-owned-dynamic" content="marker-range" />
				</Show>
				<title>head manager prepend repro</title>
				<HydrationScript />
				<script type="module" src="/entry-client.js" />
			</head>
			<body class={started() ? 'started' : ''}>
				<main>
					<h1>hello</h1>
					<p id="status">{started() ? 'started' : 'waiting'}</p>
				</main>
			</body>
		</html>
	);
}
```

Observed (rc.3, production build): `#status` stays `waiting`, `<body>` never gains `.started`, and
the console carries the `nextSibling` TypeError. With the `useHead` line removed: clean hydration,
no console output, canary lands.

### Platform

solid-js / @solidjs/web 2.0.0-rc.3 (also reproduced on 2.0.0-rc.0 with `@solidjs/vite-plugin`
3.0.0-next.28 / `babel-preset-solid` 2.0.0-rc.0), vite 8.2.1, vite-plugin-solid 3.0.0-next.27,
Chromium (Playwright) and plain Chrome.

brenelz:
Fixes #3081.

## The bug

When a full-document shell authors its own `<head>` children and anything registers a `charset`/`base` tag through `useHead`, the server splices that registration as a **prelude** immediately after the `<head>` open tag — a deliberate placement constraint — landing it AHEAD of every shell-authored head child:

```html
<head><meta charset="utf-8" data-dh="charset"><meta name="shell-owned" content="first"><!--$-->…
```

The compiled shell's head traversal is positional (`getFirstChild`/`getNextSibling`), so the prepended tag shifts every read by one, the walk ends on a null read (`TypeError: Cannot read properties of null (reading 'nextSibling')`), and hydration for the whole document dies with it — which is how `@tanstack/solid-router`'s useHead registry killed every page.

## The fix

`getFirstChild`, `getNextSibling`, and `getNextMatch` now step over server-**injected** head-manager tags while hydrating: elements in `document.head` carrying `data-dh` but no `data-dhf`. The check is gated on `sharedConfig.hydrating`, so non-hydration template-clone walks stay on the fast path.

The `data-dhf` exemption is the load-bearing detail: when a title is registered, the server rewrites the shell's own static `<title>` **in place** (winner text in the element, original stashed on `data-dhf`). That element keeps its position in the walk and must still be claimed — and both the byte-rewrite path and the streamed `$dha` "t" op always leave the stash, while injected tags never carry it.

A client-side skip also fixes this for any injection order the server (or a future placement rule) chooses, rather than pinning the server to one splice position.

## Tests

Regression coverage joins the document-shell suite (the existing whole-document pattern tests):

- **server** (`test/server/document-shell.spec.tsx`): pins the injected-before-shell emission order and the in-place title rewrite, and records the chunks as an artifact.
- **client** (`test/hydration/document-shell.spec.tsx`): replays those bytes into the jsdom document, hydrates against `document`, and asserts the post-hydration signal write reaches its bindings, node identity is preserved, and the rewritten `<title>` is claimed (same element, winner text) rather than skipped.

The client test reproduces the issue's exact TypeError without the fix, and the full `packages/web` hydrate (147) and server (332) suites pass with it.

### DIFF
--- a/.changeset/usehead-shell-head-claim-walk.md
+++ b/.changeset/usehead-shell-head-claim-walk.md
@@ -1,0 +1,14 @@
+---
+"@solidjs/web": patch
+---
+
+Skip server-injected head-manager tags in the hydration claim walk. A
+`useHead` charset/base registration is spliced as a prelude right after the
+`<head>` open tag — ahead of every head child a full-document shell authored
+itself — and the compiled shell's positional head traversal died on the
+shift: a null `nextSibling` read that halted hydration for the whole
+document. `getFirstChild`/`getNextSibling`/`getNextMatch` now step over
+elements in `document.head` that carry `data-dh` without `data-dhf` while
+hydrating; the `data-dhf` exemption keeps the server's in-place static
+`<title>` rewrite claimable, since the rewrite always stashes the original
+text there and injected tags never do.
--- a/packages/web/src/client.ts
+++ b/packages/web/src/client.ts
@@ -1752,7 +1752,8 @@ export function getNextElement(template) {
 export function getNextMatch(start: Node, elementName: string): Element;
 
 export function getNextMatch(el, nodeName) {
-  while (el && el.localName !== nodeName) el = el.nextSibling;
+  while (el && (el.localName !== nodeName || (sharedConfig.hydrating && isInjectedHeadTag(el))))
+    el = el.nextSibling;
   return el;
 } /** Hydration-walk primitive; not for hand-written code. @internal */
 export function getNextMarker(start: Node): [Node, Array<Node>];
@@ -1778,8 +1779,27 @@ export function getNextMarker(start) {
   return [end, current];
 }
 
+// Head-manager tags the server INJECTED around the shell's own markup — the
+// charset/base prelude spliced right after `<head>`, registry tags spliced
+// before `</head>` — never correspond to compiled shell markup, so the
+// positional head walk must step over them: one prepended prelude tag shifts
+// every subsequent read and hydration of the whole document dies (#3081).
+// Shell-authored tags the server rewrote IN PLACE (the static <title> given
+// the winner text) keep their position and must still be claimed; the
+// rewrite always leaves the `data-dhf` stash, which injected tags never
+// carry.
+function isInjectedHeadTag(node) {
+  return (
+    node.nodeType === 1 &&
+    node.parentNode === document.head &&
+    node.hasAttribute("data-dh") &&
+    !node.hasAttribute("data-dhf")
+  );
+}
+
 export function getFirstChild(node, expectedTag) {
-  const child = node.firstChild;
+  let child = node.firstChild;
+  while (child && sharedConfig.hydrating && isInjectedHeadTag(child)) child = child.nextSibling;
   if ("_SOLID_DEV_" && isHydrating() && expectedTag && child?.localName !== expectedTag) {
     const isMissing = !child || child.nodeType !== 1;
     console.warn(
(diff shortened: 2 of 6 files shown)


ryansolid:
Thanks for jumping on this — the diagnosis is exactly right, and you independently found the load-bearing detail (`data-dhf` distinguishing the in-place title rewrite from inserted tags), which the landed fix keeps.

The reason we went a different route: `getFirstChild`/`getNextSibling` are **dev-only** codegen. The compiler gates them on `dev` (`packages/compiler/src/dom/children.rs` — `if !self.hydratable || !self.dev` falls back to raw walks), so production hydratable output is plain `parent.firstChild.nextSibling…` member chains with no helper in the path. A skip inside the helpers fixes dev but leaves production broken — and #3081 was reproduced on production builds (the hydrate test config compiles `dev: true`, which is why the suite here couldn't surface the gap).

8c48a2ed lands the build-mode-independent version: `hydrate()` normalizes the head before any claiming — the registry-inserted leading run (`data-dh` without `data-dhf`) moves to the end of `<head>`, where inserted tags are already harmless — so dev helpers and prod raw chains both walk exactly the shell's authored children. Closing as superseded; appreciate the quick turnaround on this one.

PR #3088, opened 2026-08-28
fix(web): fail a call on a response the runtime did not write

### LINKED ISSUE #3087: A response the runtime did not produce resolves the call to undefined instead of failing
### Describe the bug

The client transport treats a response as a failure only when it carries the runtime's own error header or a 5xx status:

```js
// packages/web/server-functions/src/client.ts
const failed = response.headers.has(ERROR_HEADER) || response.status >= 500;
```

So a non-2xx answer that the runtime did not produce is decoded as if it were a result. Its body is not an encoded payload, the decode yields nothing, and the call **resolves to `undefined`** — no throw, no console output, no status anywhere the caller can see. `const story = await getStory(1)` reads as "the function returned nothing" rather than "the request never reached it".

The set of responses this covers is wider than it looks. It includes the handler's *own* refusals, which carry no error header:

| answered by | status | call resolves to |
| --- | --- | --- |
| handler's method allowlist (GET to a function that never declared `GET`) | 405 | `undefined` |
| handler's origin gate (cross-site) | 403 | `undefined` |
| handler's malformed-arguments answer | 400 | `undefined` |
| a route the handler is not mounted on | 404 | `undefined` |
| an auth proxy's login page | 401 | `undefined` |
| a CDN or WAF block | 403 | `undefined` |

The 5xx clause is documented as a defensive fallback — *"Proxies may omit the protocol error header on 5xx responses."* The same reasoning applies to everything above: those responses are equally not the runtime's, and equally not a result.

### Steps to reproduce

```sh
mkdir sf-4xx && cd sf-4xx && npm init -y
npm i @solidjs/web@2.0.0-rc.3
node repro.mjs
```

`repro.mjs`:

```js
import { AsyncLocalStorage } from "node:async_hooks";
globalThis[Symbol.for("solid.RequestContext")] = new AsyncLocalStorage();
const { createServerReference } = await import("@solidjs/web/server-functions/client");
const { registerServerFunction } = await import("@solidjs/web/server-functions/server");

registerServerFunction("getStory", async () => ({ title: "Async Solid" }));

const answers = {
  "404 (handler not mounted there)": new Response("<h1>Not found</h1>", {
    status: 404,
    headers: { "content-type": "text/html" }
  }),
  "401 (auth interstitial)": new Response("<html>login</html>", {
    status: 401,
    headers: { "content-type": "text/html" }
  }),
  "403 (CDN/WAF block)": new Response(null, { status: 403 })
};

for (const [label, response] of Object.entries(answers)) {
  globalThis.fetch = async () => response.clone();
  let outcome;
  try {
    outcome = "resolved " + JSON.stringify(await createServerReference("getStory")());
  } catch (error) {
    outcome = "threw " + String(error).slice(0, 50);
  }
  console.log(label.padEnd(34), "->", outcome);
}
```

Output:

```
404 (handler not mounted there)    -> resolved undefined
401 (auth interstitial)            -> resolved undefined
403 (CDN/WAF block)                -> resolved undefined
```

The handler-produced 400/403/405 behave the same; I reproduced those against a build of `next` (`12fa7295`) rather than the published rc, since two of them only exist there.

### Expected behavior

A response the runtime did not encode should fail the call, with the status on the error, the way a 5xx already does.

The two can be told apart without a new header: every encoding path stamps `X-Server-Function-Format` — natural encodings, the JSON fast path and the codec alike (`getHeadersAndBody` / `encodeResult` / `serializedResponse`) — so a non-2xx carrying neither that nor the error header is one nothing in the runtime wrote. That keeps `respond(value, { status: 400 })` resolving as a value, which is the pattern the validation guidance in RFC 10 points at.

I have a patch for this and will open it as a PR.

### Environment

| | |
| --- | --- |
| `@solidjs/web` | 2.0.0-rc.3 (published), and `next` @ `12fa7295` |
| Node | v24.19.0 |
| OS | macOS (darwin 25.6.0) |

frenzzy:
Fixes #3087.

A call counted as failed only when the response carried the runtime's error header or a 5xx. Everything else non-2xx went to the decode path, where a login page or an empty 405 yields nothing — so the call resolved to `undefined`, reading as "the function returned nothing" where the request had not reached it. That set includes the handler's own refusals:

| answered by | status | was | now |
| --- | --- | --- | --- |
| method allowlist (undeclared read) | 405 | `undefined` | throws, `error.status === 405` |
| origin gate (cross-site) | 403 | `undefined` | throws |
| malformed `?args=` | 400 | `undefined` | throws |
| a route the handler isn't mounted on | 404 | `undefined` | throws |
| an auth interstitial | 401 | `undefined` | throws |
| a CDN or WAF block | 403 | `undefined` | throws |

### The rule

A response at 400 or above with no body format is not a result. It fails the call with the status on the error — **undecoded**, because the body belongs to whoever answered, and **before the passthrough** redirects and revalidation use, because a refusal can carry a `Location` of its own and the passthrough would hand it back as control flow.

One marker is enough: every path that encodes a response stamps the body format, thrown results included. `BodyFormat.Void` closes the one that had none — a function that returned nothing — so all three of these still resolve:

```ts
respond({ field: "required" }, { status: 400 });   // → { field: "required" }
respond(undefined, { status: 400 });                // → undefined
return new Response(null, { status: 404 });         // → null
```

The middle one used to disagree with the last, which reaches `null` through the JSON path.

### Limits

**A 2xx is not judged.** `fetch` follows redirects, so an auth interstitial usually arrives as its login page at 200, and a missing route on a static host as an SPA `index.html` at 200 — indistinguishable from a void result by header alone. Confirmed in a browser: `redirect: "follow"` gives `200 basic`, `redirect: "manual"` an opaqueredirect at `status: 0`.

**Redirects stay with the passthrough.** A 3xx only reaches the transport where something opted out of following one, and treating it as a refusal broke a verbatim `X-Content-Raw` response carrying a `Location` — the documented way to answer a call with a response of your own.

**Two neighbouring gaps stay open**, both older than this change and worth their own issues if you agree they matter: `live()`'s retry classifier reads 400–499 as definite, so a foreign `408` or `429` now ends a source permanently where it used to close silently; and `decodeResponse` still answers `undefined` for the responses the transport now throws on.

### Notes

The repro in #3087 runs against the published `2.0.0-rc.3`, so the old behaviour is reproducible without a local build.

Checked across every shape the runtime can answer with: the only non-2xx reaching the wire without a format header is an `X-Content-Raw` passthrough, which an integration's `responseHandler` claims before the check.


ryansolid (commented):
Reviewed against the runtime's encode paths — the analysis holds, and this catches a real gap. The 405 shape in your table is one I hit myself building the `invoke` test suite: `GET` against an undeclared method decoded to `undefined` and cost real diagnosis time that `error.status === 405` would have saved.

Verified the load-bearing claim from the source: the format header is genuinely the universal discriminator. All seven natural encodings stamp it in `getHeadersAndBody`, the codec stamps at the serialized-stream response, the JSON fast path stamps, and — the corner I went looking for — a function returning a raw `Response` over the runtime wire doesn't bypass it either: the handler merges status/headers and re-encodes the body through `encodeResult`, so the caller's non-2xx raw `Response` arrives stamped and still resolves as a value. The only unstamped non-2xx answers really are the handler's own gate refusals and foreign infrastructure, which is exactly the set that should fail.

Two corners worth a line in the spec header, neither blocking:

- `respond(undefined, { status: 4xx })` — you called this out already; throwing is strictly more informative than resolving `undefined` with a status the caller can't see, since the transport never exposes the response on the success path.
- A bare `X-Content-Raw` non-2xx (verbatim passthrough, no single-flight header) now fails at a runtime-less call site. Integrations are unaffected — the `responseHandler` seam sees the response before this check — and a transport that can't decode a payload reporting failure over minting `undefined` is the right default, but it is the one *runtime-produced* shape the new clause catches.

The CodSpeed failure is a fork artifact (it fails identically on your other PR and passes on `next`) — not yours, and a one-clause change on the fetch path has no benchmark surface.

Ready to go from my side once you mark it ready for review.

frenzzy:
Your review landed on the version before a rework, so the corner you approved has moved.

`respond(undefined, { status: 4xx })` no longer throws. My premise was wrong: `encodeResult`'s void branch was the one encode path with no format header, so a void result with a status looked exactly like a bare refusal — and it disagreed with `new Response(null, { status: 404 })`, which resolves to `null`. `BodyFormat.Void` marks that branch now, so both resolve and the invariant you verified holds without an exception.

Also since then:

- The check ran after the `Location` / `X-Revalidate` passthrough, so a foreign `403 + Location` still resolved. It runs before it now.
- A refusal with `application/x-www-form-urlencoded` threw a `URLSearchParams` with no `.status`, which `live()` reads as retryable. Foreign responses are not decoded at all now.
- I had widened the check to 300 and up, on the reasoning that an SSO interstitial answers with a `Location`. A browser never surfaces a 3xx here — `fetch` follows redirects, `manual` gives an opaqueredirect at status 0 — and it broke a raw `X-Content-Raw` passthrough carrying a `Location`. Back to 400 and up.
- Testing for the error header as well as the format turned out to ask the same question twice: every path that tags a thrown result encodes it too. One marker now.

Your `X-Content-Raw` corner is pinned both ways: the non-2xx raw passthrough failing, and `responseHandler` claiming a response before the check.

ryansolid (approved):
Green on the rebased base. The earlier CodSpeed flags were cross-environment noise on signals store-utility benchmarks this transport change cannot reach.

PR #3091, opened 2026-08-28
Patch mode default-on: channel hardening rounds 6–10.14, static manifests, node-driven delivery

ryansolid:
# Patch channel: default-on flip + full hardening line

**Scope:** `next..patch-hardening-r6` (33 commits). This PR carries the complete patch-mode line since round 6: the DEFAULT-ON compiler flip, ten audit-hardening rounds (all findings closed), static read manifests, the node-driven delivery architecture, two size passes, and the #3123 landing-consumption integration. The runtime audit backlog is EMPTY — nothing paused, nothing open except P3 diagnostics polish.

### Default flip (patch mode ON)
Both compilers (Babel + Oxc/Rust) emit `patchDriver` bindings by default; opt out with `patchDriver: false`. All fixture corpora regenerated; parity tier `dom-nopatch` fences the opt-out. Server entries carry `patchDriver`/`rowProof` stubs so compiled DOM modules link under Node.

### Node-driven delivery (the architecture change)
Patch delivery moved from queue-driven to NODE-driven: each channel owns a detached `deliveryEffect` (bare node, one static source, no root/owner bookkeeping) bumped by a channel signal. Emissions are signal writes — they inherit the engine's scheduling, transition stamping, and hold semantics instead of reimplementing them queue-side. Raw payload fast path (`pc.np`) serves fresh backings scan-free on the hot path. dbmon: mount 6.6 vs classic 15.1 ms, tick 2.1 vs 7.9, partial 0.6 vs 1.2 (2.3×/3.8×/2×).

### Static read manifests
Compilers emit each template's read set (`registerPatch(record, fn, keys)`); manifests are interned per template (WeakMap, prefix trees for deep paths) and shared by reference across rows. Closes the prod-soundness hole (getter demotion probes exactly the declared envelope) and the round-7 mount regression (~0.1 ms/1000 rows).

### Hardening rounds 7–10.14 — the invariant harness
Every audit finding closed harness-first (RED test, then fix). The big structural rules now pinned:

- **A patch body never reads an accessor raw** — admission scans, adoption re-scans, and deferred demotion (getter arrivals convert consumers to per-entry render effects reading their OWN manifest envelope; failed computes never commit; errors route per-entry to the registrant's boundary).
- **Deliveries mirror classic effects, lane for lane** — optimistic writes are lane-timed, held boundaries defer INTO their owner queues (value AND structural, with live resync at release), late registrants under held commits get identity-aligned resyncs, transition merges canonicalize stamps (no duplicate or lost deliveries).
- **Emission is primitive-owned** — one bump primitive owns consumer gating, ancestor bubbling, and transaction-scoped dedup; no seam gets to skip ancestors.
- **#3123 integration (final round)**: equal landings never flash committed state through value patches (optimistic families always take the override-composing proxy read — the payload fast path is gated to non-optimistic families); contradicting landings notify authoritatively — one regular-timed delivery coalesced with the adoption's emission plus a row-ops resync at the landing.

### Observability (dev-only, zero prod bytes — verified byte-identical)
Attribution cause chains flow through deliveries (`patchEmit`/`patchDispatch`/`patchDelivered`/`patchStructural`), channel-side `HUGE_FAN_OUT`/`WIDE_WRITE` twins, synthetic structural rerun events.

## Size

Two dedicated size passes (recording-proxy deletion + single-mode dispatch; shape consolidation). Costs are tiered pay-for-use:
- value tier (any app with one eligible template): ~1.5 kB brotli over classic
- list tier (driver-eligible store lists): ~2.1 kB on top — the tier the dbmon wins accrue to
- All scenario budgets green post-rebase; the audit-hardening total on the value tier is ~1.9 kB, measured and annotated per round in `.size-limit.js`.

## Verification

- `packages/signals/tests/store/patch-invariants.test.ts` — the channel-contract harness (axis products: registration mode × backing shape × lane × timing × consumer lifecycle)
- `packages/web/test/for.patchinvariants.spec.tsx` — driver-level invariants (deep paths, holds, selections, cross-family)
- Full signals (1,496) + web (709) suites green; compiler parity 108/108 with patch on; size-limit all scenarios green
- Audit trail: `packages/signals/AUDIT-BRIEF-R6.md` (per-round findings + fixes), `packages/signals/DESIGN-PATCH-CHANNEL.md` (architecture + decisions)


ryansolid:
Post-round-6 + default-flip dbmon (octane harness, alternating 30-iter rounds, midday machine — medians in ms):

| op | octane | solid (this branch) | ratio (octane/solid score) |
|---|---|---|---|
| mount | 4.00 | 6.40 | 0.63–0.67x |
| tick | 1.60 | 1.80 | 0.83–0.85x |
| tick_partial | 0.70 | 0.50 | 0.93x |
| remount | 4.10 | 4.80 | 0.77–0.79x |
| sort | 2.00 | 2.30 | 0.74–0.84x |
| unmount | 1.60 | 1.70 | high variance |

Tick at 1.8 matches the recording+probe expectation measured when A1 landed (1.8 vs 1.7 with no demotion check vs 1.9 full-scan) — the round-6 hardening added no regression beyond the already-priced probe cost. Partial-tick diff-skip ratio 0.28–0.30 (better than octane's 0.32). Semantic gate green both rounds.

ryansolid:
Patch-mode worth-proof: same branch, same runtime, same fixture source — the ONLY difference is the compiler flag (three-way alternating, 30 iters, medians ms):

| op | octane | solid patch (default) | solid classic (`patchDriver: false`) | patch speedup |
|---|---|---|---|---|
| mount | 4.10 | 6.40 | 14.10 | 2.2x |
| tick | 1.60 | 1.90 | 6.70 | 3.5x |
| tick_partial | 0.70 | 0.50 | 1.10 | 2.2x |
| remount | 4.30 | 4.80 | 9.10 | 1.9x |
| sort | 2.10 | 2.20 | 2.60 | 1.2x |
| unmount | 1.70 | 1.70 | 2.20 | 1.3x |

Patch mode is what puts this next to octane: classic sits 4.2x behind on the full tick; patch is within ~19%. The ~1.5–3.6 kB bundle cost in the size scenarios buys this column.

(Incidentally the flip surfaced in the harness itself: the fixture's opt-out env passed empty compiler options, which now mean patch-ON — fixed to explicit `patchDriver: false` in the octane repo before measuring.)

ryansolid:
Round 7 pushed — all nine audit findings fixed, harness-first. Three commits:

1. **Red invariant harness** (committed failing, before any fix): channel contracts, driver throw-atomicity matrix over every build entry point, lane-interleaving with an effect-parity oracle, consumer-list lifecycle at emission/drain/merge boundaries, and imports⊆documented-surface contract tests per generate mode. 10 red / 8 green against the audited tree — each red reproduced a finding; the greens document invariants that already held (initial-construction atomicity, staged-update atomicity, deep resync atomicity, universal import discipline).
2. **Runtime fixes** for P1-2/3/5/6 and P2-2/3/9 (see AUDIT-BRIEF-R6.md §Round 7 for the per-finding approach).
3. **P1-1 + P1-4**: static read manifests from BOTH compilers (hoisted `_mf$` arrays, byte-parity, full fixture regen — 4321 oxc + 153 babel tests green) replacing runtime recording, which provably cannot see untaken ternary branches; `Renderer`/README/type now document `patchDriver` with contract tests pinning compiled imports to documented surfaces.

Also new: PINV-1..3 per-flush channel-ledger assertions in the `__TEST__` invariant infra — registration accounting, stamp retention, and queue drainage are now checked at every quiescence point in every test, not per-scenario.

Perf, quiet machine: dbmon tick **2.1 ms** vs round-6's 1.9 — the +0.2 is the honest price of sound accessor demotion across branches and nested chains (first sound cut measured 2.5; manifest interning, prefix-tree probing, and leaf inlining recovered the rest — two profiler passes attached to the work, not guesses). Partial tick still beats octane (0.5 vs 0.7). Mount's intern misses were profiled to zero after array hoisting; final mount number needs one quiet-machine confirmation (a parallel build was loading the box during the last sweep).

Gates: 32/32 monorepo tasks, size scenarios ratcheted with dated notes (net round-7 cost: ~+0.3 kB store apps, ~+0.6 kB patch-list apps), changesets included.

ryansolid:
Quiet-machine confirmation sweep (evening, 80% idle, both orders, 30 iters):

| op | octane | solid (round 7 final) | round-6 reference |
|---|---|---|---|
| mount | 4.10–4.20 | **6.20–6.40** | 6.30–6.40 |
| tick | 1.60 | **2.00** | 1.90 |
| tick_partial | 0.70 | **0.50–0.60** | 0.50 |
| remount | 4.00–4.30 | 4.60–4.70 | 4.80 |

Mount fully recovered — the `_mf$` hoisting eliminated the intern misses exactly as profiled. The final price of prod-sound accessor demotion across ternary branches and nested chains is **+0.1 ms on the full tick (~5%)**, zero on mount, and partial ticks still beat octane. Classic (patch off) remains 6.7 on the same tick for scale.

That closes round 7 end to end: 9/9 findings fixed harness-first, gates green, perf within noise of the pre-audit numbers with the soundness holes closed.

ryansolid:
Round 8 pushed — 8/8 findings fixed, harness-first again (the new invariant tests committed red where reproducible). Highlights beyond the straight fixes:

- **The P2 duplicate-apply finding had a deeper root cause**: `patchableRaw` served `pb ?? v`, leaking uncommitted transition drafts to mid-transition mounts. Initial applies now read the COMMITTED view, and generation-stamped drains skip consumers that initialized from an entry's state (transition releases exempt themselves — their late consumers saw the pre-commit view). The fold path was verified *unreachable* for the reported interleaving — the walk queues value entries before structural ops by design, and a test now pins that — but the cross-queue optimistic window was real.
- **The structural-capture fix nearly shipped a 3.7 kB regression**: binding captured records needs raw→proxy wrapping, and importing `wrapNext` into the channel would retain the entire trap/write engine in store-less bundles that merely compile a `rowProof` list. The size gate caught it; the wrap now rides a hook `createTarget` installs (if no store ever created a target, no raw can need resolving).
- **Tentative ancestor bubbles are two-sided**: lane-timed forced entries for in-flight visibility plus settle-held twins, so reverts restore ancestor expressions and landings show landed truth — the harness asserts both directions.

Everything else as reported: manifest deep-probe at admission, decimal-key ineligibility in both compilers (byte-parity fixtures regenerated), `rowProof` on the `createRenderer` surface with the contract test extended to the whole patch tier, and the universal `patchDriver` commit phase untracked.

Gates: 32/32 tasks, 4,323 compiler + 153 Babel + 468 signals + 71 driver tests, size ratcheted (+~0.3 kB per tier with dated notes). dbmon identical to round-7 finals: mount 6.4 / tick 2.0 / partial 0.5, quiet machine, both orders.

ryansolid:
Round 9 pushed — 11/11 findings addressed, harness-first, with three findings materially refined during verification:

- **The gen-stale finding's ambient case self-corrects** (eager adoption swaps the committed backing at walk time, so pre-flush mounts read the new state — pinned by a test). The REAL stale window was mid-transaction, and the fix is semantic: the skip rule now applies only to entries emitted from committed-VISIBLE state, `patchableRaw` serves the held view for masked targets, and the mount invariant test anchors to what an untracked proxy reader sees rather than to any implementation detail.
- **The optimistic-getter finding uncovered a missing seam, not just a missing probe**: tentative reconciles never emitted on the record's own channel at all — effects saw the tentative view, patches didn't. They now emit the view at lane timing, the optimistic drain probes payloads for accessors, and demotion there runs its re-drives IMMEDIATELY (the global render queue is stashed in flight; deferral would postpone visibility to settle).
- **The NaN/unstable-getter fallback fix is structural**: the manifest IS the read set, so the effect fallback's compute pass reads the declared envelope directly and never executes the body — write-free by construction, applied to both web and universal drivers.

The rest as reported: per-queue forced-stamp clearing + merge repair, isWrappable bind guards, server-entry `patchDriver`/`rowProof` (notSup/identity, matching the `template` stub class), function-intermediate probes (accessor carriers demote conservatively), safe-integer-only manifest keys in both compilers, and unchanged reconciles no longer bubble. The attribution CI flake is fixed two-sided: the harness mutes its expected demotion warnings, and the attribution test now counts its own diagnostic's warns instead of the process-global total.

Gates: 32/32 tasks, 4,325 compiler tests, size gates ratcheted with dated notes. dbmon 6.3 / 2.1 / 0.6 — parity with rounds 7–8.

**Standing note for the next design pass** (also in the brief): rounds 7–9's P1s are dominated by hand-replicated visibility rules at channel seams. I'm prototyping node-driven delivery (one hidden node per patched record, compiled bodies unchanged) as the by-construction answer before the next audit round — if per-record nodes hold dbmon, that eliminates the timing-divergence class entirely.

ryansolid:
Review note from #3161 (For + siblings hydration desync, fixed in 91e300a7): the regression came from the list seam's lazy `mapArray` creation minting the list's hydration id scope at insert time instead of source position. The fix makes a hydrating `<For>` create eagerly, which covers the classic path this PR inherits — but once this PR arms the driver by default, driver-ENGAGED hydration takes a different road: `driveList` claims rows positionally by `_hk` and mints its own `listOwner` child id at insert time. Before flipping, the parity matrix should pin the `for-then-siblings` fragment shape (now in `test/harness/scenarios.tsx`) with a store-backed, stamped-row list that actually engages — verifying siblings after a DRIVEN list keep their keys on both the engage and decline roads.

ryansolid:
Heads-up for anyone continuing work on this branch: two review commits landed directly on `patch-hardening-r6` — do **not** force-push over them; fetch and rebase local work on top.

- **3e12ffdb** — P1 fix found while reviewing the #3123 integration: the third landing posture (a CONTINUATION landing echoing a replayed keyed add, delivered to a driven list) was untested. Red harness first: the row-ops channel served the bare landed base (the wipe's mid-reckoning snapshot) plus per-edit replay drafts — states classic readers never render. Fix: wipe + replay notify as ONE reckoning — `notifyOptimisticWrites` suppresses write-site row-ops frames while `replaying`, and `consumeOverridesNext` emits the landing notification after the replay half via `emitLandingConsumption`, carrying the `optimisticView`-composed snapshot. The changeset for the landing-integration entry was extended to cover it.
- **22268722** — test-only: pins the fourth posture (settle-drain reckoning). Entangled retainers settle together — one action completing early strips no mask — and the joint settle lands the channel on committed truth. Already green; now locked.

Full signals + web suites green on the new head, size inside budgets. The branch is also one commit behind `next` (#3162, web-only preload merge — no interaction); a trivial merge when convenient.

ryansolid:
Structural audit round landed on top of the review commits (no force-push — fetched and built on `3e12ffdb`/`22268722` as asked). Three root causes behind the six findings:

- **Sweep mechanics (holds + window + quadratic, one mechanism)**: the late-registrant sweep is rebuilt on registration sequence numbers — entries stamp `sq = ++pc.rq`, items stamp the watermark at emission, so late entries are a suffix of the append-ordered live list and the sweep is an O(#late) tail scan (was indexOf-per-entry). The window is fixed at both edges (in-snapshot entries and mid-drain registrants excluded), and held owner queues defer via `deferIntoQueue` — never through the hold. Contract refinement pinned by test: plain stores emit at the fold, so pre-flush registrants ride the snapshot with baseline-correct real ops — the sweep's genuine audience is lane items and stash windows.
- **Visible-view resyncs**: every drain-resolved structural `next` (resync forms, held releases, late sweeps) resolves through `visibleStructRows` — optimistic families read the override-composing proxy, never bare committed backing. This also closes the seam your one-reckoning suppression leans on at *partial* settles (the `pb ?? v` note in 3e12ffdb holds for joint settles; an aborted retainer with a surviving sibling re-arms overrides that a committed read would drop). Deleted-slot ticks coalesced past a shrink are gated out on both snap and resync paths.
- **Superseded work**: `emitLandingConsumption` now bumps a structural generation (`pc.sg`); items stamp it at emission and both drains skip stale-generation items — transition-held row/slot ops can't replay over a landing's resync at settle.

**Audit of 3e12ffdb**: sound — `replaying` is try/finally-safe with per-edit throw isolation, and the composed-snapshot emission is correct (composed with the sg bump above). Added a fifth-posture pin (aborted-retainer survivor, oracle-anchored at classic parity) alongside your fourth.

**Hydration ask**: `for-then-siblings-driven` is in the parity matrix — store-backed rows whose template provably emits `rowProof` under the harness's flags (attribute-only binding; text holes disqualify the purity proof, verified against the compiler). The update pushes a row through the engaged driver; invariants pin sibling node identity across the driven structural update. Static-row scenario remains the decline road.

Also merged `next` (#3162, trivial). Gates: 39/39 invariant harness, signals 1,504, web 719, parity matrix 153/153, size green (two ratchets with notes: +27 B store-family, +41 B list tier — stamps and gates).

ryansolid:
Follow-up round landed (`ada689c7`). The two P1s, the two P2s, and two findings that turned out to be **upstream** of this branch:

- **Slot registrants (P1)**: `registerSlotPatchNext` was indeed missing the sequence stamp — the suffix scan read `sq 0` and broke immediately. And you were right about the test: `every()` over an empty tick list passed vacuously, *and* the construction was doubly wrong (slot ticks only emit from the reconcile walk; direct setter writes never produce them). The test now requires the surviving slot's resync to arrive and emits through `reconcile`.
- **Back-to-back landings (P1)**: chased through five interleavings (spaced/same-microtask × echo/non-echo × blind/until-gated) with frame-level classic/channel logs. The channel-side mechanism is closed: `emitLandingConsumption`'s row emission composed its snapshot **at emission time from the mid-reckoning draft** — a parked or superseded landing's topology could reach the DOM while classic held the previous view. The new `emitRowOpsLanding` hook keeps the lane (a reverting action's stash must not own the landing's notification — that's why the regular queue is wrong here; verified by a red 10.14 test when I tried it) but resolves at **drain time** through `visibleStructRows`, which reads exactly what classic renders at that moment. Parity by construction.
- **In the process, two upstream findings** (pinned as `it.fails` in `createOptimisticStore.test.ts`, channel-independent — bare projection, no actions, no consumers, classic effects only):
  1. A second continuation landing arriving in the same microtask chain is **swallowed** — committed truth permanently loses the landed row (`confirm({id:0}); confirm({id:2})` → final committed `[0]`, not `[0,2]`).
  2. Downstream: an `until()`-gated action waiting on the swallowed echo **wedges forever** (authoritative truth never carries the row).
  Both sit in the #3123 continuation-reckoning machinery — flagged with minimal repros rather than unilaterally reworked, since that seam is actively yours. I suspect your P1's "classic remains on the previous topology until action settlement" observation and these are the same root: the second landing's commit is deferred/lost while the channel (pre-fix) showed its draft.
- **P2s**: late-row resyncs dedup per drain (was item-count × consumer-count); the superseded-work gate now re-resolves standalone slot ticks against the live view instead of dropping them (row items still drop — the landing resync covers those).

Gates: 1,505 signals + 2 expected-fail pins, 719 web, size green — store-family app actually DOWN 50 B (the emission-time composition deleted), patch tiers +61/+111 B ratcheted with notes.

ryansolid:
Fold-audit round landed (`a4c439b7`). All four P1s closed, harness-first (four red repros, then fixes):

- **Per-index held-slot defers**: exactly as flagged — `deferIntoQueue`'s shared `hq` flag collapsed a multi-slot batch to its first index behind a hold. Slot items now dedup per (entry, index) via `entry.hqs`; row items keep the single-run dedup since their resync reads live truth.
- **Resync-then-stale-ops double-build**: the late sweep now runs once at DRAIN END against a per-channel highest-emission watermark. Entries at or below `maxRq` rode some item's snapshot and received real, baseline-sound ops; the sweep covers exactly `(maxRq, winEnd]` — the audience no snapshot reached — still a suffix tail scan. This also structurally removes the item-local repetition (your P2 from last round's shape).
- **Staged raw mutation**: `stagedApply` recursed into children read through the draft proxy — for SHALLOW stores that read returns the raw row, and the merge mutated committed truth in place: visible before the reveal, notified to no one. Descent is now gated on the child being a live draft proxy (`stagedChild`); raw children replace their slot wholesale, so the write parks and reveals like any other.
- **Reveal coverage**: both fold-site row-ops emissions gated on `t.fam?.opt !== true` — correct for override materializations (lane channel), wrong for STAGED TRUTH folding at the reveal. A root array retained only through a descendant's override isn't `overlaid`, so the settle loop's resync never reached it and the gate silenced the fold's own ops. `t.sf` (set in `ensurePB` when an optimistic-family draft is written under the authoritative posture — exactly the staging bracket) exempts staged folds, and a slot-diff twin ticks changed slots, since staged reveals bypass every walk.

Also on the branch since your audit point: the rebase onto the pushed fold commit with the channel integration re-derived for fold semantics (landing-time machinery deleted — staged truth rides held-write semantics, the reveal rides the settle resync loop; the round-10.5 ancestor gate your rewrite reverted is re-applied), and the CodSpeed −11% root-caused and fixed (`68087b57`: the eager parent-slot repair ran ungated in patch-less apps — privatization re-cloned every adopted interior backing per reconcile and the identity swaps caused keyset/deep bump storms; now gated on `hasPatches()`).

Gates: 1,511 signals + 2 expected-fail pins (the same-microtask swallow and wedged `until()` persist under fold — still upstream), 719 web, size green with four ratchets (sweep bookkeeping is most of it). Confirmed on your note: the consecutive-landing divergence is resolved by the #3164 integration.

ryansolid:
Redesign landed (`b296640d`) — per the trigger set two rounds ago, the five findings at `a4c439b7` are closed by replacing the structural coverage machinery, not by a fifth round of point fixes.

**The structural version chain**: every emission stamps a per-kind version (`sv` rows, `svs` slots — separate consumer lists, separate chains); the channel tracks the VISIBLE version (bumped when items enter the live queue — commit-coincident emissions immediately, stashed items at their `releaseBatch`, lane items at emission). An entry initializes its applied version to the visible one — exactly what its registration read covered — and applies an item's payload only on an unbroken chain (`av === svAt − 1`: the item's baseline IS the entry's last-seen state). Anything at-or-below skips; any gap takes exactly ONE resync at the END of the flush, after every queue. Deleted wholesale: emission-snapshot membership, the maxRq/winEnd watermarks, the per-item sweeps, the drain-generation dedup, the rq/sq stamps.

Against your findings: (1) and (2) and (4) have no mechanism left to be wrong in — no watermark, no per-queue sweep, no per-item repetition. (3): the settle loop skips staged-fold targets; the fold's own emission carries the reveal, pinned with a single-rebuild assertion. (5): `repairAncestorSlots` at registration fixes the target's ancestor chain — gated to stale aliases of the SAME child, map-resolved. That gate matters: the ungated first cut wrote tentative rows' backings into committed truth (all four optimistic equivalence failures during development were this one root cause — a tentative row registering mid-flight has no committed slot, and absent/different slots must never be written).

Two contract notes: held-window registrants IMPROVE — their chain connects, so they receive real baseline-sound ops at release instead of an identity rebuild (10.13 pin updated to accept either sound form). Revert-form resyncs resolve COMMITTED raw only — never a lingering draft backing (row escape materializes one), never the composing proxy.

Gates: signals 1,511 (+2 expected-fail upstream pins), web 719, parity 153, size green with three ratchets (17.15/19.6/27.4 — flat redesign cost; measured honestly, this is +230 B on the value tier, bought to stop the ~150 B/round accretion this class was generating). Your adversarial repros against the new head would be the confirmation that matters.

ryansolid:
Fold audit 2 landed (`a65c3ca1`) — all five closed. Your root-cause was exact: the visible-version init was unsound for mounts inside the writing transition.

- **Transition-aware `av` init (P1)**: a consumer mounting inside the writing transition reads the SPECULATIVE view, so its baseline now covers the stashed emissions (`av = sv` when `activeTransition` is set at registration); ambient mounts keep the visible-version init and receive the stashed ops at release. The bac/abc replay is pinned by a repro asserting every frame the mid-transition mount sees equals the speculative order.
- **One-channel reveals (P1)**: the fold-site emission now applies the walk's split — aligned windows (same length) are value replacements and ride slot ticks only; length changes ride row ops only. Pinned with a shallow-reveal repro asserting zero non-null row ops at the reveal.
- **Target-keyed staged identity (P1)**: staged-reveal diffs resolve rows through the family map before matching — the fold re-seats retained rows' raws, and raw-keyed matching rebuilt rows whose proxies never changed.
- **P2s**: held releases fast-forward `entry.av` (no redundant follow-up resync), and all `| 0` coercions are gone from version arithmetic (2^53 doubles; no wrap suppression).

Gates: signals 1,513 (+2 expected-fail upstream pins), web 719, parity 153, size green (two 20 B ratchets).

Standing merge position, per the thread with Ryan: correctness rounds keep closing, but the branch's PERFORMANCE case is currently unproven — the dbmon driver column can't be measured (the babel-preset fixture path never engages `driveList`; the octane-era numbers we'd been comparing against turned out to come from an orphaned stale server), and classic shows a consistent ~5–10% tax needing a profile-driven gate-consolidation pass. Those are the blockers before this merges; next session is dedicated to them.

ryansolid:
Perf verification complete (`6aa9ad93`) — the octane bar is met on the update paths, and the earlier regression readings were harness ghosts.

**The sweep** (30 iter, servers verified fresh):

| op | octane-tsrx | classic (next) | driver (branch) |
|---|---|---|---|
| mount | 4.40 | 16.40 | **7.10** |
| tick | 1.80 | 6.10 | **2.20** |
| tick_partial | 0.80 | 1.30 | **0.50** |
| remount | 4.70 | 9.90 | **5.00** |
| sort | 2.20 | 2.90 | **2.50** |
| unmount | 1.70 | 2.50 | **0.30** |

- **The version chain kept the wins**: driver matches the historical ledger (6.6/2.1/0.6) within noise. vs octane: tick 1.22×, partial and unmount AHEAD, remount 1.06×; mount holds the known 1.6× gap.
- **Mount gap named** (profiled): store-model cost — `createTarget`/`wrapNext`/accessor scans ≈0.6 ms per mount plus GC pressure; DOM costs match octane. The next lever is lazy target creation at bind time, an increment, not a blocker.
- **The classic "regression" dissolves under controlled conditions**: classic-vs-classic with the same native compile is ≤5% on the bench and at PROFILE-PARITY on tick totals (306 vs 311 ms over 50 ticks). Two harness ghosts caused the earlier readings: an orphaned months-old preview server on :5200 serving a stale build (every earlier "solid" column was that ghost — it answered current asset URLs with HTML fallbacks), and the babel-preset fixture path never stamping `$ll` (the driver column silently ran classic; `solid-compiled` with explicit `patchDriver` is the real driver fixture — its attribute-only rows are by design, text holes disqualify the purity proof).

With this, both merge blockers from the perf thread are resolved: the driver column demonstrates octane-class update throughput on the branch head, and classic carries no profile-visible tax. Remaining open: the two upstream `it.fails` pins (reckoning swallow/wedge — not channel), and the mount-gap increment as post-merge work.

ryansolid:
**Ship ruling** (`46930bec`): per Ryan, this ships as-is pending the final audit pass. The documented driver contract for this release is deep keyed stores via `reconcile` — the benched configuration (octane-class on update ops: tick 1.22×, partial/unmount ahead; mount 1.6× with the gap attributed to store wrapping).

Deferred to a future RC, recorded in the brief:
- `keyed={fn}` engagement (accessor-row binding + `param().member` compiler grammar — today it declines to classic by defensive pin)
- Shallow logical-key retention (a `keyFn` seam in `identityOps`; shallow retention is reference-keyed by contract today)
- The mount levers: row-value multiplexing over the list channel (est. −1 to −1.5 ms dbmon mount), channel shape diet, same-shape admission-scan fast path
- The two upstream `it.fails` pins (reckoning swallow/wedge — #3123/#3164 machinery, not channel)

State for the final audit: head `46930bec`, all suites green (1,513 signals + 2 expected-fail upstream pins, 719 web, 153 hydration parity), size budgets green, CodSpeed pending on the fresh push. The last two audit rounds (version chain at `b296640d`, fold-audit-2 at `a65c3ca1`) are the ones not yet re-reviewed.

ryansolid:
Fold audit 3 landed (`eae73227`) — both P1s closed, and both of your meta-observations were correct: the transition flag was the wrong discriminator, and my own test was hiding the double.

- **Held-queue version init**: the baseline now keys off `queueIsHeld(entry.q)` at registration, not `activeTransition` — parked actions clear the execution-time flag while held-boundary registrants still read speculative state, so the flag-based init handed them committed baselines and replayed stashed row ops over speculative DOM. White-box pin: in a parked window, a held registrant initializes `av === sv`, an ambient one `av === svv` (and receives the stash at release, which is sound for its committed init read).
- **Reorder classification**: aligned windows now check reference MOVEMENT, not just length — an equal-length staged reorder is structure (one row-ops event, zero slot ticks, pinned). Slot ticks fire only when every differing slot is a non-wrappable replacement.
- **The double you flagged**: the reveal now marks itself consumed on BOTH branches (slot ticks are a reveal too), and the settle loop consumes the mark — and the shallow repro asserts the row channel is *completely silent* at an aligned reveal. You were right that the old assertion's `filter(o => o !== null)` was masking exactly the null-resync event that proved the double.

Gates: signals 1,515 (+2 expected-fail upstream pins), web 719, size green (one 24 B ratchet, createStore tier 15.3). Head for the next pass: `eae73227`.

ryansolid:
Fold audit 4 landed (`3dc92f8a`) — all three P1s plus the primitive-reorder refinement, and your parked-ambient finding exposed the root simplification: **adoption commits eagerly** (only notifications batch), so *every* reader's init read includes every emitted walk state — parked windows, held boundaries, and ambient mounts alike. The visible version now bumps at emission for all structural emitters, which deletes the deferred-visibility machinery *and* last round's held-queue registration special case entirely. There is no reader anywhere that can replay stashed row ops it already rendered — the whole replay class is gone, not patched. (Two stash-window pins updated to the corrected contract: late registrants are owed nothing; the deleted-slot gate now asserts on the early consumer.)

- **`rf` epoch-stamped**: reveal marks carry the fold epoch and expire with their flush — the boolean lingered on descendant-retained roots outside `overlaid`, where the settle loop never visits to consume it, and suppressed a later revert's resync.
- **`rf` only on proven emission**: `emitSetterRowOps` returns whether ops were emitted and the slot branch counts ticks — a no-op staged fold can no longer suppress the only revert resync a driven list needed. Pinned: identity-aligned staged restatement, then revert, list lands committed.
- **Primitive permutations are structure**: same multiset + moved positions = row ops, zero slot ticks (classic keys primitive rows by value — reorders move rows, preserving identity/focus). Pinned.

Gates: signals 1,517 (+2 expected-fail upstream pins), web 719, size green (two ratchets: 15.4/27.55 — the multiset check and epoch plumbing). Head for the next pass: `3dc92f8a`.

ryansolid:
Fold audit 5 landed (`2fadc59e`) — the last P1 closed.

**Primitive rows retain by value**: `buildRowOps` admitted only wrappable rows to the occurrence-aware key queues, so `["a","b","c"] → ["c","b","a"]` emitted `sources: [-1,-1,-1]` and rebuilt every row. Primitives now key by their value on both the queue-build and match sides — pure permutations match every moved value to its old index (pinned: a staged primitive permutation's ops carry no −1 sources), mixed moves/replacements retain the moved values and rebuild only genuine replacements, and duplicate occurrences were already sound through the existing queues. `undefined` rows still rebuild (undefined is the no-key sentinel).

One workspace note: the worktree currently carries unrelated in-progress edits to `core.ts`/`verdict.ts` (a #3166 revert) that break `latest-pending-probe-mid-flight` and fold-3164's store-path test — verified failing with my change stashed, so they're not from this round. I committed exactly the fix + pin and left those files untouched.

Store suites 530 green (+2 expected-fail upstream pins). Head: `2fadc59e`.

ryansolid:
Correction to my previous comment's workspace note: the `core.ts`/`verdict.ts` modifications were **not** in-progress upstream edits — they were torn-checkout debris from my own tooling. Fingerprinting showed the working tree mixed across commits (verdict.ts at pre-#3166 content, core.ts at the `893b8f95` state, docs/budgets at other vintages) — the signature of two concurrent `git checkout`s racing: a backgrounded benchmark job (detach → bench → return) overlapped foreground git work in the same worktree.

Restored to HEAD: working tree clean, and both previously-failing suites pass (`latest-pending-probe-mid-flight`, fold-3164 store path). Full signals suite on the clean tree: 1,517 passed (+2 expected-fail upstream pins), 126 files. No committed state was ever affected — the fold-audit-5 commit (`2fadc59e`) contained exactly the primitive-retention fix and pin.

Process note taken: no more backgrounded jobs that perform `git checkout` in the shared worktree.

ryansolid:
**Held out of rc.5** (decision 2026-09-01): an audit probe found the patch channel tearing through the `until()` flip-entanglement hold — a patch consumer can observe the confirmed world mid-hold while value-channel bindings beside it correctly hold the old world. Repro is pinned on this branch as an `it.fails` test.

A second structural review identified the root pattern: the channel re-decides visibility at emission/delivery time with private bookkeeping (counters, transaction stamps, transition-keyed stashes), so every new scheduler behavior (merges, boundary holds, parked windows, now steals) has needed a hand-written mirror. Plan for the rc.6 window, before default-on ships:

1. **Value deliveries read through the masked seam** (same hold resolution the store's traps use) + snapshot compare — mid-hold wakes then deliver nothing by construction. The raw-read fast path is already gated off for optimistic families, so the benchmark tick path is unaffected.
2. **Structural held-ops stash moves with the scheduler** — near term the steal mirrors `_heldPatches` like `mergeTransitionState` already does; the rc.6 shape attaches held ops to nodes so merge/revert/steal semantics are inherited instead of mirrored.

CI is otherwise clean at `13cb4450`: signals 1524 passed (exit 0), web 720 passed, size/treeshake gates green, all prior audit P1s closed.

ryansolid:
Closing as superseded: the architecture direction changed away from the push-based patch channel entirely. Value delivery went to graph-native regions (pull-based, riding the deep witness — `region-delivery` branch, external audit in progress), and list structure is headed to a unified keyed For that owns rows and DOM placement in one persistent structure (design: `packages/web/DESIGN-UNIFIED-FOR.md` on `unified-for`) — no delivery seam at all, which is what ten hardening rounds here kept fighting. #3229 removes the channel and driver from next (they were opt-in-off, so behavior-neutral). This branch and the node-delivery prototype branch stay on the remote as the historical record and the independent performance baselines they are.

PR #3103, opened 2026-08-29
fix(web): import the types the server-functions declarations reference

brenelz:
## Problem

Every `GET`/`live`/`createServerReference`-wrapped server function types as `any` for consumers of the published `@solidjs/web` package. Downstream this silently kills inference through anything built on server functions — e.g. a `GET`-wrapped `queryFn` makes TanStack Query's `data` come out `any`.

## Cause

`server-functions/src/client.ts` and `server.ts` only **re-export** `ServerFunction` and `ServerFunctionMetadata` (`export type { ... } from "./shared.js"`), which creates no local binding — yet `GET`/`live`/`createServerReference` and several interfaces reference those names in annotations. Both files are `// @ts-nocheck`, so nothing flagged the dangling names; declaration emit shipped them unresolved:

```
types/server-functions/client.d.ts(161,82): error TS2304: Cannot find name 'ServerFunction'.
```

(visible with `skipLibCheck: false`). Under the usual `skipLibCheck: true`, TypeScript treats the unresolved reference as an error type, which behaves like `any` at every call site.

## Fix

Add local `import type { ServerFunction, ServerFunctionMetadata } from "./shared.js"` to both files, with a comment on why the re-export alone isn't enough. Types-only — no runtime change.

Verified: after `pnpm run types`, the emitted `types/` and `types-cjs/` declarations carry the import and `tsc --skipLibCheck false` no longer reports TS2304 in them; in a consuming app, `GET`-wrapped functions infer their real return types again (checked against the solid-templates `fullstack-tanstack` template, where `useQuery(...).data` went from `any` back to the server function's return type).


ryansolid:
Good fix — the dangling references were real, and resolving them surfaced a latent declaration bug your CI run caught: the client `live()` reference was typed as returning `Promise<LiveSource<T>>` (via `ServerFunction`'s call signature), but it actually hands back the reconnecting iterable synchronously. While the declaration chain collapsed to `any`, nothing noticed; the moment your fix made the types real, `web`'s test-types failed on the live-retry spec.

I pushed 4aa25e95 onto this branch typing the client reference truthfully (`LiveServerFunction` — sync call, same `id`/`url` identity). Isomorphic consumers are unaffected since they `await` the call and awaiting the plain iterable is identity. CI should be green now; thanks for the PR — it hardened more than it set out to.

Issue #3012, opened 2026-08-18
createMemo(fn, { transparent: true }) during hydration shifts sibling hydration ids → unclaimed DOM + REACTIVITY_HALTED


TylerRick:
A memo created with the (internal) `{ transparent: true }` option inside a component while `sharedConfig.hydrating` is true causes every element AFTER the creating component to fail its claim:

```
Hydration tag mismatch for key "2": expected <pre> but found [object HTMLElement]
[REACTIVITY_HALTED] ... TypeError: Cannot read properties of null (reading 'nextSibling')
Hydration completed with 1 unclaimed server-rendered node(s): <pre _hk="3" ...>
```

The unclaimed nodes stay visible but no client scope owns them — bindings in them are permanently dead. In an app the failure is silent-but-fatal: a footer rendered after such a component keeps its SSR appearance and never reacts to state.

## Minimal reproduction

vite 8.2.1 + vite-plugin-solid 3.0.0-next.27, solid-js / @solidjs/web 2.0.0-rc.0, `renderToStream` on the server, `hydrate` on the client. `src/App.tsx` is the whole story:

```tsx
import { createMemo, createSignal } from 'solid-js';

function Inner(props: { disabled: boolean }) {
	const pending = createMemo(() => false, { transparent: true });
	return (
		<nav>
			<button id="continue" disabled={props.disabled || pending()}>
				Continue
			</button>
		</nav>
	);
}

export default function App() {
	const [checked, setChecked] = createSignal(false);
	return (
		<>
			<label>
				<input
					type="checkbox"
					checked={checked()}
					onChange={(e) => setChecked(e.currentTarget.checked)}
				/>{' '}
				done
			</label>
			<Inner disabled={!checked()} />
			<pre id="after">sibling after Inner: {String(checked())}</pre>
		</>
	);
}
```

Observed:

- With the transparent memo: the `<pre>` fails to claim, hydration reports it unclaimed, the REACTIVITY_HALTED TypeError fires, and the checkbox no longer enables the button.
- Remove `{ transparent: true }` (or the memo entirely): hydration is clean and everything works.
- The memo's value never changes and is even never read in the broken variant — creating it is enough.

<details>
<summary>The rest of the repro project (4 small files) — <code>pnpm install && node server.mjs</code> → http://localhost:5599</summary>

`package.json`:

```json
{
	"name": "solid2-hydration-repro",
	"private": true,
	"type": "module",
	"scripts": { "dev": "node server.mjs" },
[...]

PR #3105, opened 2026-08-29
fix(compiler): align lone spread hydration ids

kavhnr:
## Summary

A hydratable reactive lone spread used different compiler paths on the server and the client.

Minimal case:

```tsx
const attrs = () => ({ class: "example" });

<div {...attrs()}>
  <span />
</div>;
```

The DOM compiler sends the reactive spread through `mergeProps`. A function source makes `mergeProps` create a memo, and that memo uses a hydration child ID. The SSR lone-spread fast path sent `attrs()` directly, so the server did not create the same memo. The old server element IDs were `0` and `1`, while the client expected `0` and `2`.

This change sends a hydratable reactive lone spread through the normal SSR `mergeProps` path. The existing props thunk lets `ssrElement` allocate the element key before it evaluates the merge. The order is now element `0`, merge memo `1`, and following element `2`, which matches the client.

The direct lone-spread path is still used when SSR is not hydratable. It is also still used when the spread expression is static or has an `@static` marker. Thus, the change does not add a merge cost to those paths.

The Babel and native compilers use the same rule. The compiler tests check both outputs. A shared web harness scenario also renders and hydrates the case with each compiler. It checks the server keys, hydration warnings, DOM node reuse, and an update after hydration.

Related work:

- solidjs/solid#3012 contains the same dynamic-spread trigger, but its fix addressed server memo ownership.
- https://github.com/ryansolid/dom-expressions/pull/540 added key-first deferred SSR merges. Its lone-spread fast path still bypassed that path.
- solidjs/solid#3049 moved both compiler implementations into this repository.

A patch changeset is included for `@solidjs/babel-plugin` and `@solidjs/compiler`.

All listed commands passed.

- `pnpm install --frozen-lockfile`
- `pnpm --filter @solidjs/compiler exec vitest run __tests__/spread-hydration.test.js` — 1 file, 10 tests
- `pnpm --filter @solidjs/babel-plugin test` — 20 files, 153 tests
- `pnpm --filter @solidjs/compiler test` — Rust tests in both feature modes; 29 Vitest files, 4,328 tests
- `pnpm --filter @solidjs/compiler lint` — Clippy passed with warnings denied
- `pnpm --filter @solidjs/web test` — 58 client files, 41 server files, and 24 hydration files; 1,217 tests passed and 2 skipped
- `JSX_COMPILER=babel pnpm --filter @solidjs/web test` — the same 1,217 tests passed and 2 skipped
- `pnpm build` — 24 tasks passed
- `pnpm typecheck` — 15 tasks passed
- `pnpm test` — 32 tasks passed
- `pnpm test:integration` — 12 tasks passed, including imports and packaged types
- `pnpm exec prettier --check packages/babel-plugin/src/ssr/element.ts packages/compiler/__tests__/spread-hydration.test.js packages/web/test/harness/scenarios.tsx packages/web/test/server/spread-hydration.spec.tsx .changeset/fix-lone-spread-hydration-ids.md`
- `rustfmt --edition 2024 --check packages/compiler/src/ssr/transform.rs`
- `git diff --check`

The full repository `cargo fmt --check` also reports existing format differences in `packages/compiler/src/directives/transform.rs` and `packages/compiler/src/refresh/transform.rs`. This pull request does not change those files. The changed Rust file passes its file-scoped format check.


ryansolid:
Thanks for the diagnosis here — the id drift and the root cause (the client-side `mergeProps` memo consuming a hydration id the server's fast path never allocates) were exactly right, and the reproduction scenario was very useful.

We ended up landing the inverse fix in 52306662: rather than teaching the hydratable SSR output to defer a merge until after the element key, the client stops merging entirely. A lone spread has nothing to merge — both DOM generates now pass the accessor straight to `spread()`, which resolves a function props source inside its own tracking scopes (the Solid 1 shape). No memo is minted on either side, so the ids agree, the server output stays untouched, and every lone spread gets cheaper rather than every hydratable one getting more expensive.

Your harness scenario lives on in the fix (with the button now claiming `_hk=1` — no reserved gap). Closing as superseded; thanks again for the report and the careful analysis.

Issue #3095, opened 2026-08-29
A null-body status turns into a 200 carrying an unexplained Internal Server Error


frenzzy:
### Describe the bug

A server function that answers with a null-body status — `204`, `205`, `304` — fails with a generic *Internal Server Error* instead, and the failure carries no hint of what was wrong.

The transport always writes a body for a scripted caller, then applies the author's status to it (`packages/web/server-functions/src/server.ts`, `encodeResult`):

```js
if (value === undefined) {
  headers.set(BODY_FORMAT_HEADER, BodyFormat.Void);
  return new Response(null, { status, headers });   // fine
}
...
return new Response(JSON.stringify(value), { status, headers });   // throws for 204/205/304
```

`new Response(<body>, { status: 204 })` is a `TypeError` by construction, so the encode throws, the generic handler catches it, and the caller gets a sanitized error at HTTP 200.

Only the `undefined` branch escapes, which makes the rule an accident of the value rather than of the status: `respond(undefined, { status: 204 })` answers a real 204, and `respond({ a: 1 }, { status: 204 })` answers a 200 with an error.

Returning `new Response(null, { status: 204 })` — no body anywhere in sight — fails too, because the transport still encodes its own body before applying the status:

```
!! Response( string, status: 204 ) -> Response constructor: Invalid response status code 204
```

**Nothing surfaces the cause, in either mode.** `sanitizeServerError` replaces the `TypeError` rather than wrapping it, so even with `NODE_ENV=development` the payload is `"Internal Server Error"` and the stack starts inside the sanitizer:

```
Error: Internal Server Error
    at sanitizeServerError (.../server-functions/dist/server.js:1030:10)
    at dispatch (.../server-functions/dist/server.js:1288:20)
```

The words `204`, `status code` and `TypeError` appear nowhere in that response. An author who writes a plausible line gets a server error pointing at framework internals.

The runtime already holds this exact scenario to be forbidden. `encodeResult` on `next` carries the rule in a comment:

> By the time a result is being encoded the function has already run — side effects committed — so a failure HERE must never escape into dispatch's catch, where it would be sanitized and reported as the function itself throwing (a phantom error over a call that succeeded).

That is precisely what happens. The `try` around the JSON attempt swallows the first `TypeError`, and the codec path immediately below it throws the same one outside the guard:

```js
try {
  if (isJSONSafe(value)) {
    ...
    return new Response(JSON.stringify(value), { status, headers });   // throws, caught
  }
} catch {
[...]

PR #3113, opened 2026-08-30
fix(web): sanitize failures that escape through the result graph

frenzzy:
### The bug

`sanitizeServerError` guards the one road a thrown error takes out of dispatch. A failure can also escape through the **result graph** — a rejected promise, an async iterable that throws, a stream that errors — where it reaches the codec as a value to *encode* rather than as a throw, and never meets the sanitizer.

The leak is the one that sanitizer's own comment describes: *"an ORM/driver error's failing query, connection string, or bound params included"*. Measured, the same driver error thrown four ways:

```
upfront    500 | leaked: (nothing)
iterable   200 | leaked: hunter2, 10.0.0.5, SELECT * FROM users, abc123
promise    200 | leaked: hunter2, 10.0.0.5, SELECT * FROM users, abc123
stream     200 | leaked: hunter2, 10.0.0.5, SELECT * FROM users, abc123
```

Worse than the thrown case: the head is already committed, so it is a `200` carrying no error tag. Any streamed answer, deferred value or live source reaches it.

**Two encoders, not one.** The frames flight sink encodes its outcome with `serializeStream`, not `serializeResponseStream`, so a rejection nested inside a flight-data slice leaked there even after the response path was closed — same build, same process. It needs a response routed through frames (a mutation whose result is markup) and the failure nested one level inside a slice, which is the ordinary shape of a cache entry. A rejection at the *top* of a slice throws into dispatch's catch and is sanitized, which is why it looked covered.

Filed as #3116.

### The fix

Wrap the channels before either encoder sees them.

Not the rejection — it has not happened yet. Not the Errors already in the graph either: an `Error` reached as a **value** was never thrown, so it is data and the author's to ship. Only the shapes through which a future failure can arrive.

Each container is recorded before its children are walked, so a cycle resolves to the container being built. Containers that changed nothing are passed through by reference, so identity and dedup survive for the codec. The walk covers plain objects, arrays, `Map` and `Set`.

Unchanged by design: the wire format (no new node types, nothing asked of the peer), `markSafeError` as the escape hatch, an `Error` returned as a value, cycles and shared references, and dev builds.

### Known limits, deliberate

- A channel held by a **class instance** — its own properties are not ours to rebuild (private fields, getters, invariants).
- A channel behind an **accessor** — invoking one is not ours to do.
- Symbol-keyed properties.

Each is a place the walk declines to reach rather than an oversight, but they are holes, and worth your judgement: the alternative seam is one layer down in the serializer, where `.stack` stripping already lives and is already global — *"stripped from everything serialized outside development"* — which would cover SSR payloads too, at a much larger blast radius. #3116 lays out both.

### What review found in this fix

Three rounds, and each defect is now a test. Recording them because the last one was a crash on a shape that works today:

- **A cycle blew the stack.** Containers were recorded in the `WeakMap` *after* their children were walked, so a self-reference recursed until the stack gave out — and the `RangeError` escaped into dispatch's catch as a 500, on a shape seroval encodes natively as a back-reference.
- **`Map` and `Set` were not walked**, so a rejection in either shipped raw while the changeset claimed otherwise.
- **A null-prototype object was rebuilt as a plain object**, changing its node type from `NullConstructor` to `Object` — so "the wire format is unchanged" was not quite true.
- **The walk invoked accessors**, calling every getter once here and again at encode time, and turning a *throwing* getter into a 500 — the phantom error over a call that succeeded that `encodeResult` goes out of its way to avoid.
- **The `ReadableStream` branch was dead**, since a stream is async-iterable on every server runtime and the iterator branch claimed it first. It now runs ahead of that branch, takes the reader on first pull rather than eagerly, and guards chunk values.
- **An earlier attempt claimed `Error` via a codec plugin.** A plugin *wraps* rather than replaces, so the wire carried a node the peer had no plugin for, and every sanitized error broke — including the ordinary thrown one. That is why the spec now round-trips through the client instead of grepping the body.

### Tests

`server-functions-failure-sanitization.spec.tsx` — the channels, the escape hatch, and the shapes that must not change. Checked by mutation:

| mutation | result |
| --- | --- |
| guard not applied | 5 failed |
| `markSafeError` hatch ignored | 1 failed |
| guard also rewrites Errors that are values | 1 failed |
| frames sink left unguarded | 1 failed |
| `Map` not walked | 1 failed |

Full suite green on current `next`: 44 files, 431 passed, 3 expected fail, 2 skipped.


frenzzy:
Moving this to draft — the fix as pushed is wrong, and in a way that matters. Review caught it; I have reproduced it.

A plugin does not replace the value, it **wraps** it: the wire node is a seroval plugin node, not the plain `Error` node I claimed.

```
;0x0000007b;{"t":25,"i":0,"s":{"t":13,...,"m":"Internal Server Error",...},"c":"solid/server-function-failure"
```

`t:25` is a Plugin node and the client has no matching plugin, so decode throws. That breaks the **ordinary** thrown-error path, not just the one I was fixing — `sanitizeServerError` returns an unbranded generic `Error`, so the plugin claims that too:

```
plain-throw     THREW: Je | "Seroval Error (step: 3)"     (was: Error "Internal Server Error")
returned-error  THREW: Je | "Seroval Error (step: 3)"     (was: resolved)
```

So the sentence in the commit message and changeset — *"leaves the wire shape a plain Error node — the peer needs no matching plugin, and nothing about the protocol changes"* — is false. `serializer-decode.ts` states the contract plainly: plugins "Must match on both peers."

My tests did not catch it because they were vacuous: the helper swallowed the decode failure and returned `""`, and every assertion was `not.toContain(secret)` — which an empty body passes. Negative assertions only, no round trip through the client. That is my mistake and a fair advertisement for the point #3112 makes about markers that look green.

Reworking it. The premise I got wrong is worth stating, because it also invalidates the reasoning in the PR body: I claimed a pre-serialization walk "cannot work, because the rejection has not happened yet". Wrong — the walk does not need the rejection, only the **channel**, and a promise, async iterable or stream is present as an object before serialization. Wrapping it so its future rejection is sanitized is entirely possible, preserves the wire shape exactly, and leaves an `Error` that is genuinely a value alone.

ryansolid:
Reviewed with repros against the branch, found three gaps, and pushed the fixes as a maintainer edit (3d0bbb58) since this closes an active leak — hope that's alright. Your two follow-up commits landed while I was mid-review, so the edit is rebased on top of them; the descriptor walk survives, with the copy pre-registered (see below).

**`Map` and `Set` passed through unguarded.** Both fail the plain-object prototype check, while the codec serializes their entries with full promise support — so a rejecting promise inside a returned `Map` or `Set` still shipped the driver error verbatim:

```
map-leak     200 LEAKED: hunter2, 10.0.0.5, SELECT * FROM users, abc123
set-leak     200 LEAKED: hunter2, 10.0.0.5, SELECT * FROM users, abc123
```

Both containers are now rebuilt along channel paths, keyed on the exact constructor to match the codec's own dispatch (a subclass falls through to the passthrough, and the codec refuses it regardless).

**Cyclic results blew the stack.** The comment said the WeakMap terminates cycles, but containers registered their copy only *after* walking children, so a cyclic result — which the codec happily encodes as a ref node — recursed forever, and dispatch reported the stack overflow as a sanitized 500. A working response shape on `next` became an opaque failure on this branch. Copies are now registered before the walk, so a cycle lands on the guarded copy and round-trips intact (`value.self === value` on the client, covered by test).

**The non-async-iterable `ReadableStream` branch re-enqueued chunks raw.** Unreachable in Node (its streams are async-iterable, so the iterator branch claims them and guards step values — worth knowing that branch is the live one), but in a runtime where it isn't, a channel nested in a chunk would skip the guard. Chunks are now guarded like step values.

The flight-sink catch was a good one — that serializer split is exactly the kind of second road this whole PR is about. Suite is green locally across all three web configs; merging once CI agrees.

frenzzy:
Apologies — I force-pushed over your maintainer edit. 3d0bbb58 is restored, authorship intact, and my duplicate of it is gone.

What happened: I was mid-rework when your push landed, hit a rejected push, and checked whether the commit was upstream. It wasn't on `next`, so I read a commit authored by you on my fork as something automated rather than as a maintainer edit — I did not think to check the PR timeline, where your comment had been sitting for four minutes. Then I rewrote the branch. Entirely my error, and the wrong instinct: an unexpected commit on a PR branch is a reason to read the thread, not to reach for `--force`.

The branch is now your edit on top of my four, which is what you pushed. I had independently written the same three fixes while yours was being erased; I have dropped mine — yours is the one that stands, and the constructor-keyed `Map`/`Set` check is better than what I had, since mine would have claimed subclasses the codec then refuses.

Verified on the restored branch against current `next`: 44 files, 432 passed, 3 expected fail, 2 skipped. A cyclic result round-trips (`value.self === value`), a rejection in a `Map` is sanitized, and a null-prototype object keeps its `NullConstructor` node rather than gaining `Object.prototype`.

One loose end from my side, take it or leave it: the `markSafeError` wrapper I put on #3095's authoring error in the first commit turns out to be inert — that error is encoded as a value and never routed through `sanitizeServerError`, so removing it changes no test. It is noise in a security fix; happy to strip it in a follow-up if you'd rather the branch not carry it.

Issue #3120, opened 2026-08-30
Recent server-function fixes landed without tests that pin what they established


frenzzy:
### Describe the problem

Several recent server-function fixes landed without a test that pins the *property* they established, so the next change can undo them and the suite stays green. This is not a complaint about any one commit — it is a pattern worth naming, because every instance below was found by deliberately breaking the runtime and watching nothing fail.

**Guarded by #3111** (four of them, with the mutation that proves each):

| invariant | what silently undoes it |
| --- | --- |
| `Sec-Fetch-Site` is authoritative when present — `same-site` and `none` never reach the origin matcher | letting `same-site` fall through, the natural repair for a broken subdomain deployment |
| `Origin`/`Referer` matchers refuse what they do not match | a matcher that answers `true` — the function and array forms had no test at all |
| the redirect mask covers exactly the statuses `fetch` follows | narrowing the set; only `302` was exercised, so `301`/`303`/`307`/`308` were free |
| the error header's bound is applied to the source, not the encoding | slicing the encoded form — it still passes a length assertion, which is all the existing tests check |

**Guarded by #3121**: server-function ids are keyed on identity rather than position (7c551ae2, #3109). Every fixture in `packages/compiler/__tests__/directives/` is a *single fixed source* with frozen expected output, which pins the id **format** for that source but cannot express a differential property — no fixture has a second version. The pre-fix positional scheme passed an identically shaped suite; the fix regenerated 134 fixture and harness files. So reintroducing position goes green after exactly the fixture regeneration a developer performs reflexively when snapshots fail, and the failure mode is the silent one: an old client's call reaching a different function.

What it needs is two compilations of the same file — append a function, delete the first, reorder, edit a body — asserting every surviving name keeps its id. Also worth pinning: the ordinal suffix for a repeated name (`<name>-<hash>-<n>`) has no fixture at all, and it is the one place position still decides.

**Weak rather than missing**, for the record: nothing decoded a sanitized error *through the client* — `response-invariants.spec.tsx` asserts the header and never the body — which is how a broken first attempt at #3116's fix passed the whole suite while breaking every sanitized error end to end. #3113 adds that round trip.

### Why it keeps happening

The specs here are excellent at pinning *outputs* — the frozen fixtures, the wire-shape assertions. The gaps are all *relationships*: same input, two builds; same call, two callers; the same value read twice. Those need a test that does the thing twice and compares, and that shape is rare in this suite.

### What would help

[...]

Issue #3110, opened 2026-08-30
A stale server-function id is an undistinguishable 404, so version skew cannot be recovered


frenzzy:
### Describe the problem

When a client calls a server function whose id no longer exists on the deployment it reaches, the answer is a bare `404` that says nothing about why:

```
unknown id -> status 404 | headers: {"cache-control":"no-store","vary":"Sec-Fetch-Site, Origin, Referer"} | body: (none)
client sees        : Error | "Server function call failed with status 404" | status: 404 | own keys: ["status"]
```

The dev build puts `Unknown server function: <id>` in the body; production sends nothing. Nothing on the response distinguishes *this address is not part of this build* from a CDN 404, a stale proxy rule, or a mistyped route.

So an integration cannot act on it. The one recovery that works for version skew — tell the page it is stale and reload it — needs to know that the failure is a skew failure, and that fact is not on the wire.

This is the ordinary consequence of deploying: tabs stay open, bundles stay cached, rolling deploys serve two versions at once. Every function that is genuinely removed or renamed reaches this path, and today it surfaces to the user as a generic failed call.

### Steps to reproduce

```sh
mkdir sf-skew && cd sf-skew && npm init -y
npm i @solidjs/web@next
node repro.mjs
```

`repro.mjs`:

```js
import { AsyncLocalStorage } from "node:async_hooks";
globalThis[Symbol.for("solid.RequestContext")] = new AsyncLocalStorage();
const srv = await import("@solidjs/web/server-functions/server");
const cli = await import("@solidjs/web/server-functions/client");

srv.registerServerFunction("known", async () => "ok");

const wire = await srv.handleServerFunctionRequest(
  new Request("http://localhost/_server/data/c8cb6025-9", {
    method: "POST", body: "[]",
    headers: { "Sec-Fetch-Site": "same-origin", "X-Server-Function-Instance": "i" } }));
console.log("unknown id -> status", wire.status,
  "| headers:", JSON.stringify(Object.fromEntries(wire.headers)),
  "| body:", wire.body ? JSON.stringify(await wire.text()) : "(none)");

globalThis.fetch = (a, i) => {
  const r = new Request(new URL(a.toString(), "http://localhost"), i);
  r.headers.set("Sec-Fetch-Site", "same-origin");
  return srv.handleServerFunctionRequest(r);
};
try {
  await cli.createServerReference("c8cb6025-9")();
} catch (e) {
[...]

Issue #3109, opened 2026-08-30
A server function's id encodes its position in the file, so appending one re-points existing addresses


frenzzy:
### Describe the bug

A server function's production id is `<xxhash32(root-relative path)>-<ordinal>` — the file it lives in, plus its position among that file's server functions. The name is appended in development only:

```rust
// packages/compiler/src/directives/transform.rs
fn create_id(&mut self, name: &str) -> String {
    let base = format!("{}-{}", self.hash, self.count);
    self.count += 1;
    match self.env {
        Env::Development => format!("{base}-{name}"),
        Env::Production => base,
    }
}
```

So in production an id says *where a function sat*, never *which function it is*. Adding a server function to an existing file renumbers the others, and every client already holding the old numbering now addresses different code.

That would be tolerable if it produced a 404. It does not: the shifted ids all exist, so the call dispatches — to the wrong function, with the old client's arguments, answering `200`.

### Evidence

Ids straight from the compiler, same file path, three ordinary edits:

```js
const { transformDirectives } = require("@solidjs/compiler");
const opts = { filename: "/app/src/api/users.ts", root: "/app", mode: "server", env: "production" };
const ids = code => transformDirectives(code, opts).functions.map(f => `${f.name}=${f.id}`).join("  ");
const fn = n => `export async function ${n}(x) { "use server"; return "${n}"; }`;

console.log("baseline         ", ids([fn("getUser"), fn("deleteUser")].join("\n")));
console.log("append at the end", ids([fn("getUser"), fn("deleteUser"), fn("archiveAccount")].join("\n")));
console.log("delete the last  ", ids([fn("getUser")].join("\n")));
console.log("reorder          ", ids([fn("deleteUser"), fn("getUser")].join("\n")));
```

```
baseline          deleteUser=c8cb6025-0  getUser=c8cb6025-1
append at the end archiveAccount=c8cb6025-0  deleteUser=c8cb6025-1  getUser=c8cb6025-2
delete the last   getUser=c8cb6025-0
reorder           getUser=c8cb6025-0  deleteUser=c8cb6025-1
```

Numbering runs in reverse source order, so inserting at the *top* of a file is safe while **appending at the bottom — the ordinary way to add a function — shifts everything**. Renaming a function changes nothing, because the production id never carried the name.

Read the first two rows together: `c8cb6025-1` meant `getUser` before the deploy and `deleteUser` after it.

End to end, against the runtime:

```js
[...]

PR #3121, opened 2026-08-30
test(compiler): pin that a file edit moves no server-function id

frenzzy:
Server-function ids are the addresses deployed clients already hold, so keying them on identity rather than position (7c551ae2, #3109) is a wire contract — and nothing currently guards it.

The fixture suite cannot: every fixture in `__tests__/directives/` is a **single fixed source** with frozen expected output, so it pins the id *format* and has no second version to compare against. The positional scheme passed an identically shaped suite; the fix regenerated 137 files. Reintroducing position goes green after exactly the regeneration a developer performs reflexively when snapshots fail — and the failure it lets through is the silent one: an old tab calling `getUser` reaching `deleteUser` with a 200.

This is the differential half. Compile two versions of one file and require every surviving name to keep its address:

- a function appended at the end
- a function inserted at the top
- the first function deleted
- the two reordered
- a body edited

Plus the two consequences worth stating rather than discovering:

- **a rename is a new address** — the old id belongs to nothing, so an old client gets a clean 404 (#3110) instead of another function;
- **among functions sharing a descriptive name, position still decides** (`<name>-<hash>-<n>`), which no fixture covers today.

### Checked by mutation

Reverting `create_id` to `format!("{}-{}", self.hash, self.count - 1)` fails 4 of the 7:

```
expected 'c8cb6025-1' to be 'c8cb6025-0'
expected 'c8cb6025-0' to be 'c8cb6025-1'
expected [ 'c8cb6025-0', 'c8cb6025-1' ] to not include 'c8cb6025-1'
expected 'c8cb6025-0' to match /^handler-[0-9a-f]{1,8}$/
```

The other three pass under both schemes — inserting at the top, deleting the first function, and editing a body never moved an id even positionally. Worth keeping for exactly that reason: most edits look safe, which is why the bug survived. Deleting the first function is the surprising member of that set, and it is surprising only because the old numbering ran in reverse.

Compiler suite green: 29 files, 4325 passed. Tests only, no changeset.

Raised as a gap in #3120.

### DIFF
--- a/packages/compiler/__tests__/directives-id-stability.test.js
+++ b/packages/compiler/__tests__/directives-id-stability.test.js
@@ -1,0 +1,86 @@
+// Server-function ids are keyed on IDENTITY, not position (#3109).
+//
+// The id is the address a deployed client already holds, so an edit that
+// moves one silently re-points a call: before the fix the id was
+// `<hash>-<ordinal>`, appending a function renumbered the rest, and an old
+// tab calling getUser reached deleteUser with a 200. The fixture suite
+// cannot see that — every fixture is a single fixed source, so it pins the
+// id FORMAT and no fixture has a second version to compare against. This
+// is the differential half: compile two versions of one file and require
+// every surviving name to keep its address.
+
+const path = require("path");
+
+const compilerDir = path.resolve(__dirname, "..");
+const { transformDirectives } = require(compilerDir);
+
+const ROOT = "/app";
+const FILENAME = "/app/src/api/users.ts";
+const RUNTIME = "@solidjs/web/server-functions";
+
+/** The extracted functions, as the production build emits them. */
+function compile(source) {
+  const result = transformDirectives(source, {
+    filename: FILENAME,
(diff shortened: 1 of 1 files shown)


frenzzy:
Superseded by d6a4a52f — closing.

Your `directives-id-scheme.test.js` covers this better than mine did. Both were rebuilt against a `create_id` reverted to `<hash>-<ordinal>`:

```
directives-id-scheme.test.js     6 failed | 1 passed
directives-id-stability.test.js  4 failed | 3 passed
```

Deriving the ids from an independent implementation catches a change to the hash, the path derivation *or* the format, where my differential version only notices when an edit happens to move something — and three of my cases pass under both schemes, since inserting at the top, deleting the first function and editing a body never moved an id even positionally. That is a weaker signal for the same regression.

The README line is the part I would not have thought to write, and it is probably the most durable of the three: a fixture regeneration that changes an id is a protocol change, not churn. That is the sentence that stops the next person waving it through.

PR #3124, opened 2026-08-30
fix(web): pull a streamed result behind a demand gate

### LINKED ISSUE #3118: A streamed result has no backpressure: one slow consumer buffers it all in server memory
### Describe the bug

A streamed result has no backpressure. The response stream is built with no `pull` and no queuing strategy, and every codec node is enqueued the moment it is parsed, so the producer runs as fast as it can resolve regardless of whether anyone is reading.

```
consumer read      : 6 chunks
producer yielded   : 17224 items
ran ahead by       : 17218 items
heap growth        : 18.8 MiB

(five runs: 17,180-17,333 items, 18.4-19.9 MiB)
```

That is one consumer reading six chunks over ~300 ms. On a large or infinite stream — a live source, a paginated read, anything backed by a cursor — a single slow client buffers the entire result in server memory. Application code cannot see it happening, and nothing bounds it.

Cancellation is fine: dropping the consumer does stop the producer. It is only the reading-too-slowly case that runs away.

### Steps to reproduce

```sh
# Run against a build of the `next` BRANCH. The published `next` dist-tag is
# 2.0.0-rc.4, which predates the `<endpoint>/data/<id>` address (#3094) and
# answers 404 to every request below.
node repro.mjs
```

`repro.mjs`:

```js
import { AsyncLocalStorage } from "node:async_hooks";
globalThis[Symbol.for("solid.RequestContext")] = new AsyncLocalStorage();
const srv = await import("@solidjs/web/server-functions/server");

let produced = 0;
srv.registerServerFunction("firehose", async function* () {
  while (produced < 100000) {
    produced++;
    yield { n: produced, payload: "x".repeat(200) };
    await new Promise(resolve => setImmediate(resolve));
  }
});

const response = await srv.handleServerFunctionRequest(
  new Request("http://localhost/_server/data/firehose", {
    method: "POST",
    body: "[]",
    headers: { "Sec-Fetch-Site": "same-origin", "X-Server-Function-Instance": "i" }
  })
);

const before = process.memoryUsage().heapUsed;
const reader = response.body.getReader();
for (let i = 0; i < 6; i++) {
  await reader.read();
  await new Promise(resolve => setTimeout(resolve, 50));   // a slow client
}
console.log("consumer read    :", 6, "chunks");
console.log("producer yielded :", produced, "items");
console.log("heap growth      :", ((process.memoryUsage().heapUsed - before) / 1048576).toFixed(1), "MiB");
await reader.cancel();
process.exit(0);
```

Output on `next` (`e2b21041`) — the numbers above.

### Expected behavior

A producer that nobody is reading waits, the way a `ReadableStream` normally makes it wait.

### Options

1. **Gate the source pull on consumer demand** — #3124. I wrote the sentence below expecting this to be a question for the seroval seam rather than a small edit, and that was wrong: the runtime already installs an iterator wrapper described as *"the only seam where a dropped consumer can stop the producer"*, and the same seam lets a slow one slow it. A read drives `pull`, `pull` releases one source pull, teardown releases a parked one. Measured over 200 idle event-loop turns, the producer advances by 1 instead of tracking the turn count.
2. **Give the stream a queuing strategy and respect `desiredSize`.** Smaller: keep the push shape but stop pulling the source while the queue is over its mark. Bounds the memory without inverting the pump.
3. **Cap the queue and fail loudly past it.** Turns an invisible memory leak into an error — worse for the well-behaved slow client, better than an OOM.
4. **Document it**: a streamed server-function result is buffered at the producer's pace, so do not stream something unbounded to a client you do not control. Honest, and it leaves the hazard in place.

(1) turned out to be both, so (2)-(4) are moot unless #3124 is the wrong shape.

### Related

A failing test for this is in #3112 (counted in event-loop turns, so it means the same thing on any machine).

frenzzy:
Closes #3118.

The response stream is built with no `pull` and no queuing strategy, and every codec node is enqueued the moment it is parsed, so the producer runs as fast as it can resolve regardless of whether anyone is reading. One slow consumer buffers the whole result in server memory — unbounded, and invisible to application code.

### The seam was already there

The iterator wrapper the runtime installs describes itself as *"the only seam where a dropped consumer can stop the producer"*. The same seam lets a **slow** consumer slow it: a read drives `pull`, and `pull` releases exactly one source pull. Ending the stream releases a parked pull too, or the source would be stranded.

`desiredSize > 0` means "fewer than one chunk queued" — the stream takes no queuing strategy, so it runs on the default high-water mark of 1. That default is what sets the depth, and it is load-bearing, so the comment says so.

### Measured

Consumer reads one chunk, then idles for 200 event-loop turns:

```
before:  producer at 200      (~17,000 items and +19 MiB in the issue's wall-clock shape)
after:   producer at 1
         reading 21 chunks advances it to 20
         cancelling stops it within the pull in flight
```

Counted in turns rather than wall-clock, so the assertion means the same thing on any machine. Throughput for a fast consumer draining a 20k-item stream is unchanged (~91 ms on both branches, within noise), and a non-streamed result never enters the wrapper.

### Scope, stated plainly

The gate sits on the source the runtime wraps — the result itself. An async iterable **nested** inside the result (`{ items: rows() }`) is pumped by the codec directly and is not gated: measured at 200 items over 200 idle turns, against 1 for the top-level shape. That is the ordinary shape, so it is worth knowing this fixes half the problem. Filed as #3125 rather than widened here, because this PR had already grown two review-found defects and adding surface without another pass seemed the wrong trade.

A consumer that abandons a stream without cancelling now leaves the producer parked rather than running it to completion. That is inherent to backpressure and the consumer's side of the WHATWG contract, but it is a behaviour change.

And the limit worth knowing before treating this as a memory guarantee: **the gate only bites if whatever writes the response to the socket propagates backpressure into `desiredSize`.** An adapter that pipes `Response.body` does; one that drains it eagerly — `for await (… ) res.write(chunk)` ignoring `write`'s return, or buffering before writing — does not, and against that adapter this is a no-op observable only in-process. I have not measured a real adapter, so I am not claiming either way; the tests here prove the runtime-side property, which is the part this repository owns.

### What review found

Three rounds, and the third found the sharpest one.

**The gate was checked before `finished`.** Teardown landing while a pull was in flight stranded the codec's pump: the release it fires finds nothing parked, the in-flight pull resolves, the pump asks for the next item, `wantsMore()` is false — 0 after close, null after error — and it parks on a resolver nobody will ever call. `push()` never returns. One token to fix, and it mirrors the previous round exactly: that defect stranded the source, this one stranded the pump.

That half is **not guarded by a test**, and I would rather say so than imply otherwise. The failure is a leaked pending promise inside seroval's pump with no outward symptom — the source still runs its `finally`, the response still ends, nothing observable differs. It was found with instrumented park/release counters, and that measurement is the evidence it rests on.

**The tests did not park at all.** Both read in a tight loop, so a read request was always pending, `desiredSize` never dropped, and the producer never reached the gate. Deleting `pull()` outright — removing the entire reopening mechanism — left all eight green, including the one whose comment claimed it would catch a deadlock. The new test pauses between reads, which is what puts the producer on the gate, and reads to completion behind a deadline so a gate that never reopens fails fast instead of hanging.

Earlier rounds. The first shape had a regression I introduced: `onDone` and `onError` set `closed` directly rather than through `teardown()`, so a pull parked on the gate was stranded — `desiredSize` is 0 after close and null after error, both failing the `> 0` check, so the gate never reopened and the source's `finally` never ran, leaking generator and cursor cleanup once per failed request. Every path that ends the stream now runs `finishSource()`.

Also corrected: an unreachable `!streamController` guard, a stale `#3118 (open)` line in the spec header, a changeset that claimed more than the test asserts, and a liveness assertion of `> 1` sitting under a comment claiming it tracked the reads.

### Tests

The `test.fails` marker #3112 put on this gap comes off, following the convention the other two gaps used when they closed. Alongside it: the producer resumes as the consumer reads and stops when it leaves (which pins the cancel half; the pausing test is the liveness one).

And the regression above: a codec failure landing while a pull is parked must still close the source.

Checked by mutation:

| mutation | result |
| --- | --- |
| gate forced permanently open | `expected 200 to be less than 50` |
| `finishSource()` removed from the codec-end paths | `expected false to be true` |
| `pull()` gutted, so the gate never reopens | `the gate parked and never reopened` |

The second guard took two attempts. The first version passed with the fix removed — the source was nested, so nothing ever parked, and a nested iterable never reaches the gate. A top-level source needs no sibling branch: the deferred failure rides *inside* a yielded chunk, so the pump enqueues it, asks for the next item, parks, and only then does the failure land. `produced` is asserted alongside the cleanup so that mistake cannot quietly recur.

Full suite green: 45 files, 456 passed, 2 skipped; the spec is stable over repeated runs.

### A later correction

Review caught two things in this description. The line about a test that "drains to completion" described a test commit `f2824f6` had already replaced, and the claim that the resume test was the liveness assertion was wrong — it reads in a tight loop, so nothing ever parks, which is the same blindness the third round found in the original pair. The pausing test is what catches a gate that never reopens; the resume test pins the cancel half. Both the body and the test's own comment now say so.

### DIFF
--- a/.changeset/stream-demand-gate.md
+++ b/.changeset/stream-demand-gate.md
@@ -8,7 +8,13 @@ every codec node is enqueued the moment it is parsed, so the producer ran
 as fast as it could resolve whether or not anyone was reading: one slow
 consumer buffered the whole result in server memory, unbounded and
 invisible to application code. The consumer's reads now drive `pull`,
-which releases one source pull at a time — measured over 200 idle
-event-loop turns, an async generator advanced by one step instead of
-running away. Teardown releases a parked pull, so an aborted or cancelled
-stream still ends.
+which releases one source pull at a time, so an unread stream stays near
+the queue size instead of running away.
+
+Scope: the gate sits on the source the runtime wraps, which is the
+result itself. An async iterable nested inside the result — `{ items:
+rows() }` — is pumped by the codec directly and is not yet gated. Ending
+the stream releases a parked pull, so an aborted, cancelled or failed
+stream still closes its source; a consumer that abandons a stream without
+cancelling it now leaves the producer parked rather than running it to
+completion.
--- a/.changeset/stream-demand-gate.md
+++ b/.changeset/stream-demand-gate.md
@@ -1,0 +1,14 @@
+---
+"@solidjs/web": patch
+---
+
+Pull a streamed server-function result behind a demand gate (#3118). The
+response stream was built with no `pull` and no queuing strategy, and
+every codec node is enqueued the moment it is parsed, so the producer ran
+as fast as it could resolve whether or not anyone was reading: one slow
+consumer buffered the whole result in server memory, unbounded and
+invisible to application code. The consumer's reads now drive `pull`,
+which releases one source pull at a time — measured over 200 idle
+event-loop turns, an async generator advanced by one step instead of
+running away. Teardown releases a parked pull, so an aborted or cancelled
+stream still ends.
--- a/packages/web/server-functions/src/server.ts
+++ b/packages/web/server-functions/src/server.ts
@@ -1676,6 +1676,20 @@ export function serializeResponseStream(value, codecOptions, signal) {
   value = guardFailures(value);
   let closeIterator = null;
   let closed = false;
+  // Demand gate. seroval's pump pulls the source as fast as it resolves and
+  // enqueues every node the moment it is parsed, so without this a slow
+  // consumer never slows the producer: the whole result accumulates in the
+  // stream's queue, in server memory, unbounded. The consumer's reads drive
+  // `pull`, which releases one source pull at a time.
+  let streamController = null;
+  let releaseDemand = null;
+  const wantsMore = () => !streamController || streamController.desiredSize > 0;
+  const awaitDemand = () => new Promise(resolve => (releaseDemand = resolve));
+  const supplyDemand = () => {
+    const resolve = releaseDemand;
+    releaseDemand = null;
+    if (resolve) resolve();
+  };
   let cancelSerialize = null;
   let onAbort = null;
   const teardown = () => {
@@ -1684,6 +1698,7 @@ export function serializeResponseStream(value, codecOptions, signal) {
     if (onAbort) signal.removeEventListener("abort", onAbort);
     if (cancelSerialize) cancelSerialize();
     if (closeIterator) closeIterator();
+    supplyDemand();
   };
   if (
     value !== null &&
@@ -1711,8 +1726,12 @@ export function serializeResponseStream(value, codecOptions, signal) {
         // torn down before the codec opened the value (abort raced the
         // codec load): close the source immediately, never pull
         if (closed) closeIterator();
+        const step = () => (finished ? { done: true, value: undefined } : it.next());
         return {
-          next: () => (finished ? Promise.resolve({ done: true, value: undefined }) : it.next())
+          // Waits for the consumer to want a chunk before pulling the next
+          // one; `finished` is re-read after the wait, since teardown can
+          // land while it is parked.
+          next: () => (wantsMore() ? Promise.resolve(step()) : awaitDemand().then(step))
         };
       }
     };
@@ -1722,6 +1741,7 @@ export function serializeResponseStream(value, codecOptions, signal) {
     // the top of shared.js), and a ReadableStream start may return a
     // promise — reads wait for it, so the stream's contract is unchanged
     async start(controller) {
+      streamController = controller;
       if (signal) {
         if (signal.aborted) {
           teardown();
@@ -1792,6 +1812,9 @@ export function serializeResponseStream(value, codecOptions, signal) {
           }
         }
       });
+    },
+    pull() {
+      supplyDemand();
     },
     cancel() {
       teardown();
--- a/packages/web/server-functions/src/server.ts
+++ b/packages/web/server-functions/src/server.ts
@@ -1745,9 +1745,14 @@ export function serializeResponseStream(value, codecOptions, signal) {
         const step = () => (finished ? { done: true, value: undefined } : it.next());
         return {
           // Pulls straight through while the queue has room, and parks
-          // until a read makes room when it does not. `finished` is re-read
-          // after the wait, since teardown can land while it is parked.
-          next: () => (wantsMore() ? Promise.resolve(step()) : awaitDemand().then(step))
+          // until a read makes room when it does not. `finished` is checked
+          // FIRST: teardown can land while a pull is in flight, and the
+          // release it fires then finds nothing parked — so a gate checked
+          // first would park the next pull on a resolver nobody will ever
+          // call, stranding the codec's pump. `finished` is re-read after
+          // the wait for the same reason from the other direction.
+          next: () =>
+            finished || wantsMore() ? Promise.resolve(step()) : awaitDemand().then(step)
         };
       }
     };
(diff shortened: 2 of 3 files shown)


frenzzy:
CI is red only on CodSpeed, and I do not think this change can be the cause — evidence rather than assertion:

**The regressed benchmarks cannot reach the changed code.** All three are `packages/signals/tests/store/utilities.bench.ts::merge`, and that file imports exactly one thing:

```js
import { createStore, merge, omit } from "../../src/index.js";
```

`packages/signals/src`. This PR touches three files — `packages/web/server-functions/src/server.ts`, a spec, and a changeset — none of which is reachable from that import graph. There is no edge from a server-function response stream to a store merge.

**CodSpeed says so itself**, in the report: *"Different runtime environments detected — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."* All three regressions are the same benchmark family at 27-31%, which is the shape of a runner change rather than a code path.

**Those benchmarks are not steady.** Two consecutive local runs, same branch, same build, no code change between them:

```
merge  2,147,292 hz -> 2,235,289 hz   (+4.1%)
merge  1,884,921 hz -> 2,058,621 hz   (+9.2%)
merge  1,416,841 hz -> 1,451,197 hz   (+2.4%)
merge  2,771,735 hz -> 2,876,216 hz   (+3.8%)
```

Local wall-clock is a different measure from CodSpeed's simulation, so this does not explain 28% on its own — I am offering it as context, not proof. The import graph is the argument.

For contrast, #3113 changed the same file and reported 136 untouched benchmarks.

Everything else is green, including `check`. Happy to rebase onto current `next` to force a fresh benchmark run against a matching baseline if that is the quickest way to clear it — say the word rather than me pushing noise at your CI.

frenzzy:
One honest caveat about this PR that came out of looking at the nested case (#3125), worth raising here rather than leaving for someone to find later.

**The gate only bites if whatever writes the response to the socket propagates backpressure into `desiredSize`.** An adapter that pipes `Response.body` with `pipeTo`/`Readable.fromWeb(...).pipe(res)` does. One that drains the body eagerly — `for await (const chunk of response.body) res.write(chunk)` ignoring `res.write`'s return, or buffering to a `Buffer` before writing — does not, and against that adapter the gate is a no-op and only the in-process tests observe it.

I have not measured a real adapter, so I am not claiming either way. The tests here prove the runtime-side property, which is the part this repository owns; whether it survives the last hop is a question for whichever adapter a deployment uses, and worth knowing before anyone treats this as a memory guarantee.

If it turns out the common adapters buffer, the fix is still correct but the issue's framing ("one slow client buffers the whole result in server memory") would be understating where the buffering actually happens — it would move from the stream's queue to the adapter's.

frenzzy:
Re-ran it: the regression was the runner, not the change.

I amended to a new SHA with byte-identical content (same tree hash, `96e2fff8`), which is the only way to make CodSpeed compare again — the branch was already on the current `next` tip, so there was nothing to rebase onto and the baseline already matched.

```
before (3f5f240):  degrade performance by 28.51%   -- 3 regressed, 133 untouched
after  (6e3f010):  will not alter performance      -- 136 untouched
```

Same code, same base, opposite verdict. That matches the report's own "Different runtime environments detected" note, and the import graph said as much beforehand: the three regressed benchmarks were `packages/signals/tests/store/utilities.bench.ts::merge`, which imports only `packages/signals/src` and cannot reach a change to `packages/web/server-functions`.

All checks green now.

ryansolid:
Merging past the CodSpeed red: the three flagged benchmarks are `merge` store benchmarks in `packages/signals`, which this web-only change doesn't touch, and CodSpeed's own report warns the comparison crossed different runtime environments. Verified the full web suite locally on this branch (671 passing). Thanks — the demand gate lands exactly on the seam the wrapper already documented, and the park-release discriminator test is appreciated.

Issue #3133, opened 2026-08-30
flattenArray drops needsUnwrap when a fragment follows an accessor, universal renderers receive a raw memo in insertNode


antoinevanwel:
### Describe the bug

`flattenArray` in `@solidjs/signals` (`packages/signals/src/boundaries.ts`, the `Array.isArray(child)` branch) overwrites its `needsUnwrap` flag with the nested call's result instead of OR-ing it:

```ts
if (Array.isArray(child)) {
  needsUnwrap = flattenArray(child, results, options);
}
```

So under `doNotUnwrap`, a function child (a `<For>`/`<Repeat>`/memo accessor) followed later at the same level by an array child that contains no functions (a fragment) resets the flag to `false`. `flatten` then returns the plain `results` array with the accessor still inside it instead of the resolving wrapper.

`@solidjs/web` never sees this because its `insertExpression` has a function branch. A custom renderer built on `@solidjs/universal` does not: `insert(parent, () => props.children, null)` takes the array branch and passes the raw memo to the host `insertNode`, which crashes (in our case a native host tried to read `node.id` from a function). It appears whenever a component's children mix a `<For>` with a fragment-returning component that comes after it, and looks intermittent because it depends on child order and on which branch of a fragment exists at the time.

Fix is one line:

```ts
needsUnwrap = flattenArray(child, results, options) || needsUnwrap;
```

### Your Example Website or App

Inline repro below; it only needs `@solidjs/signals` and `@solidjs/universal` 2.0.0-rc.4 (`bun run repro.ts`, no bundler).

### Steps to Reproduce the Bug or Issue

```ts
import { createRoot, createMemo, flatten } from "@solidjs/signals"
import { createRenderer } from "@solidjs/universal"

// 1. flatten alone: a fragment after an accessor drops the flag.
createRoot(() => {
  let accessor = createMemo(() => "from memo")
  let ok = flatten([accessor, "text"], { skipNonRendered: true, doNotUnwrap: true })
  let bad = flatten([accessor, ["fragment"]], { skipNonRendered: true, doNotUnwrap: true })
  console.log("accessor then string   ->", typeof ok === "function" ? "wrapper (correct)" : ok)
  console.log("accessor then fragment ->", typeof bad === "function" ? "wrapper (correct)" : bad)
})

// 2. through a universal renderer: insertNode receives the memo itself.
type N = { tag: string; children: N[] }
let seen: unknown[] = []
let r = createRenderer<N>({
[...]

PR #3142, opened 2026-08-30
Fix 3133 flatten array

nickshiro:
## Summary

fix bug from #3133

flattenArray overwrote the accumulated needsUnwrap flag with the reslut of each nested array. When an accessor was followed by a static nested array, later recursive call returned false, causing flatten to return raw accessor instead of a resolving wrapper. Universal renderers could consequently receive accessor function in insertNode.

Also adds regression coverage for: accessor followed by nested static array and accessor inside a nested array followed by another static fragment.

Before applying the fix, both new regression tests failed because flatten returned an array containing the raw accessor instead of a function wrapper.

  After applying the fix:

```bash
  pnpm --filter @solidjs/signals test
```

 Result:
  Test Files  115 passed (115)
  Tests       1431 passed | 3 skipped (1434)

also:

```bash
pnpm --filter @solidjs/signals types
pnpm exec prettier --check packages/signals/src/boundaries.ts packages/signals/tests/flatten.test.ts .changeset/fix-flatten-nested-accessor.md
git diff --check
```

all completed great.


ryansolid:
Thank you for taking this on! Unfortunately the branch was cut from `main` (our PR template still says that — sorry, it's stale for the 2.0 line, where everything targets `next`), so the diff replays 2,283 files of divergence and can't be merged as-is. The one-line fix itself is right, and it has landed in #3149 with regression tests at both the signals and web layers — you're credited in the changeset. One finding from review worth sharing: the crash was not universal-only; the 2.0 DOM renderer fails on the same shape (the 1.x insertExpression function branch no longer exists), so the fix mattered even more than the issue suggested. Closing this in favor of #3149 — hope to see you in the next release notes again.

PR #3143, opened 2026-08-31
fix(signals): clear a node's transition stamp when its pending value commits

### LINKED ISSUE #3140: 2.0.0-rc.4: flush() guard trips from the asyncWrite write-back on a screen with several async memos (no minimal reproduction)
### Describe the bug

On `2.0.0-rc.4`, the dev build throws `Potential Infinite Loop Detected.` from the `flush()` guard while a browser sits on one screen of our app. The page keeps working. The error reaches `window.onerror`, which is how our end-to-end suite catches it.

The guard is `flush()` counting its own iterations, in `@solidjs/signals` `dist/dev.js` around line 1882:

```js
let count = 0;
while (scheduled$1 || activeTransition) {
  if (++count === 1e5) throw new Error("Potential Infinite Loop Detected.");
  globalQueue.flush();
}
```

So something reschedules work 100,000 times inside one flush.

Every occurrence carries the same three frames, and none of ours:

```
at flush                  (@solidjs/signals dev, the guard above)
at asyncWrite             (@solidjs/signals dev)
at result.then.syncError  (@solidjs/signals dev)
```

`asyncWrite` and the `syncError` continuation put it in the write-back of a settled async computation.

**I am not claiming this is a framework defect.** The guard lives in the framework, so it throws from there whoever caused the runaway. #2843 is the same message and its cause was application code. I am filing because I cannot attribute it after several days of work, and because of the production half below.

**In production the guard is not there.** `dist/prod/core/scheduler.js` has the same loop with no counter:

```js
while (scheduled || activeTransition) {
    globalQueue.flush();
}
```

Whatever the cause, in a production build this does not throw. It spins. Pegged CPU, frozen tab, no error.

What I am asking for: which application patterns can drive `asyncWrite` to reschedule like this, and whether the guard could name the node it kept rescheduling instead of only reporting that it happened. The second one would let anyone hitting this attribute it themselves.

https://github.com/PT-Perkasa-Pilar-Utama/testate

The app is public, but I have no minimal reproduction and I want to be honest about that up front rather than waste your time. The screen is a database grid at `/projects/:slug/adapters/:id/tables/:table`. It holds four async computations built on `createMemo(async ...)`: a page of rows whose query a person controls through filters, sort, cursor and page size, plus the adapter, its schema, and a write-session presenter that refreshes the first one. It is the only screen in the product with that many, and the only one whose async memo takes a query that changes as fast as someone can click.

Relevant source: `apps/web/src/lib/async.ts`, `apps/web/src/features/data/grid.presenter.ts`, and the write-up in `docs/upstream-solid-flush-loop.md`.

I cannot give you a reliable recipe, and this is the honest state of it:

1. Run the full browser suite: `bun run e2e`. The suite drives a seeded instance through snapshots, checkouts, fixture imports and a job stream, then a crawler visits about thirty screens and clicks every control on each one.
2. About one run in four turns red on the grid screen with this error.

Counting the message in a run log instead of waiting for a test to fail: **4, 0 and 0** occurrences across three identical runs. So it fires more often than the red runs suggest, and single-run counts prove nothing about whether a change helped.

What does **not** reproduce it:

- A dedicated stress spec that drives the same screen's sort, filter, paging and read/write switch twenty times over with no wait between clicks, with and without a query the API rejects. Eight clean runs across three shapes, including one that walks a live server-sent-event stream first.
- The crawler run on its own, with no other project before it. Three clean runs.

The difference between a clean crawler run and a red one is only what ran before it in the same browser context. That points at accumulated state across thirty screens rather than anything the grid does by itself.

### Expected behavior

I expected the flush to drain. Failing that, I expected the thrown error to tell me which computation kept rescheduling, so I could decide whether the fault was mine. Today the message says a loop happened and nothing about what looped, and the only frames are internal, so an application author has nowhere to start.

### Screenshots or Videos

None that would help. The stack above is the whole observable signal.

### Platform

- OS: macOS 26.6.2
- Browser: Chromium via Playwright 1.58.2 (browser build 1208)
- Version: `solid-js` and `@solidjs/web` `2.0.0-rc.4`, `@solidjs/signals` `2.0.0-rc.4` transitively. Bun 1.4.0, Vite dev server.

### Additional context

**Prior art.** #2843 is the same message on `2.0.0-beta.15`, caused by `isPending(() => latest(asyncMemo))` read in a user effect with no render effect subscribed to that memo. We use neither `isPending` nor `latest`, and that one was fixed in #2838 well before rc.4, so I do not think it is our case. It is the reason I am not calling this a framework bug.

**Version history here.** It happened on rc.3 as well. rc.4's note about a flush loop spinning forever on a pending store read (#3068) did not end it.

**What I ruled out.** The grid's only pure helper on that path. The grid's table-change effect, whose handler writes nothing its own source reads. Two server-sent-event effects elsewhere in the app, whose source value does not change when their list refreshes.

**What I tried that did not fix it.** I rewrote the data layer onto core primitives: `refresh()` on a memo instead of a version counter bumped to force re-runs, and a `createMemo` for a derived list that had been rebuilding a new array on every read. Both changes are right on their own merits and both are kept. Neither ended the loop.

**Diagnostics.** The `reactivity-diagnostics` skill shipped with rc.4 lists `UNSTABLE_MEMO_OUTPUT` as a pattern that makes the scheduler re-run more than it should. That is the closest description of anything we do, since an async memo returns a fresh promise per run by construction. Our suite dropped every console warning until yesterday, so no run has ever reported one. It keeps them now, and I will add a comment here if a run produces a diagnostic code.

Happy to run instrumented builds, patch a local copy of `dist/dev.js` with whatever logging you want, or try a branch. Tell me what to capture and I will capture it.

snowfluke:
Closes #3140.

## The bug

`_transition` is cleared in exactly two places: `optimistic.ts`, for nodes in `_optimisticNodes`, and one async settle in `async.ts`. Everything else relies on `reassignPendingTransition`, which the completing branch runs over `batch._pendingNodes`.

`commitPendingNodes` drains that list without clearing anything:

```ts
function commitPendingNodes() {
  const pendingNodes = currentBatch._pendingNodes;
  for (let i = 0; i < pendingNodes.length; i++) {
    commitPendingNode(pendingNodes[i]);
  }
  pendingNodes.length = 0;
```

and `commitPendingNode` never touches `_transition`. So a node committed by an earlier drain leaves the list still stamped. When the transition completes, `reassignPendingTransition(batch._pendingNodes)` walks a list that no longer holds it, clears nothing, and `transitionComplete` then sets `_done = true`. The node is left pointing at a finished transition.

From there `flush()` cannot end. `setSignal` re-enters `el._transition` before it reaches its own `valueChanged` check, and `CollectionQueue._checkSources` writes the same loading-boundary flag on every pass of a drain. A write that changes nothing re-arms a transition that can never complete again:

1. the drain loop sees `activeTransition`, so it calls `globalQueue.flush()`
2. `transitionComplete` returns true on its first line, `_done` being already `true`; the transition completes and `activeTransition` becomes null
3. still inside that flush, `finalizePureQueue` reaches `checkBoundaryChildren`, then `CollectionQueue._checkSources`, which writes `false` over `false`
4. `setSignal` re-enters the finished transition before discovering the write is a no-op
5. the loop condition is true again, back to 2

Nothing recomputes in that cycle. Sampled from inside the drain loop, `dirtyQueue` is empty, `scheduled` is false, both effect queues are empty, and there are no pending nodes, async reporters or lanes. Dev throws at 100,000 passes. The production scheduler runs the same loop with no counter, so it hangs instead.

There is a second face to it. In the completing branch the fresh ambient batch keeps the dead transition's containers by reference:

```ts
const fresh = createBatch();
fresh._pendingNodes = batch._pendingNodes;
```

so once a finished transition is re-armed, `initTransition`'s adoption pass iterates `batch._pendingNodes` while pushing into `activeTransition._pendingNodes`, the same array, until `RangeError: Invalid array length` in `queuePendingNode`. `mergeTransitionState` already guards this aliasing for `_optimisticNodes` and `_affectsNodes` with move-not-copy; the pending list reaches it another way.

## The change

Clear the stamp where the node leaves the transaction. This is the same clearing `reassignPendingTransition` performs at completion, extended to nodes a drain removes first.

## Evidence

- `packages/signals`: 114 files, 1432 tests, identical results with and without the change, so it breaks nothing currently covered.
- `prettier --check` clean on the touched file. The `typecheck` failures on `next` are pre-existing and in test files this PR does not touch.
- Measured in a real app, an internal tool with a data grid holding several async memos under a loading boundary. Its browser crawl visits about thirty screens; I mirrored this one line into published rc.4 and ran the crawl repeatedly, scoring runaways, `RangeError` and total page errors:

| Build | Runs meeting the condition | Runaways | `RangeError` | Page errors |
| --- | --- | --- | --- | --- |
| rc.4 | 5 of 10 | 5 of 5 | 0 | many |
| rc.4 + this change | 0 of 12 | 0 | 0 | 0 |

The middle column is the direct readout. I left a counter on `setSignal` that records a write meeting a finished stamp, without changing behaviour. Unpatched it fires in half of runs and every one of those spins; patched it never fires at all, which is what you would expect if the stamp is now cleared at the right moment rather than tolerated later.

## What I could not do

I could not reduce this to a unit test. I tried six shapes in your harness, a loading boundary over one async memo with a churning query, two memos settling out of order, `refresh()` on a memo under a boundary, an action writing a memo's input, and a four-memo arrangement mirroring the real screen. None produced a stale stamp. Whatever orders a commit before its transition completes needs a deeper boundary tree than I could construct by hand, and I would rather say so than ship a test that passes for the wrong reason.

If you would rather fix it at a different layer, two nearby options address different parts of the same story: clearing inside `commitPendingNode` itself, or making the fresh ambient batch copy its containers instead of aliasing them. Happy to rework this, and happy to run any variant against the same crawl and report the same counters.

### DIFF
--- a/.changeset/commit-clears-transition-stamp.md
+++ b/.changeset/commit-clears-transition-stamp.md
@@ -1,0 +1,17 @@
+---
+"@solidjs/signals": patch
+---
+
+Clear a node's transition stamp when its pending value commits. `_transition`
+was only ever cleared for optimistic nodes and in one async settle;
+everything else relied on `reassignPendingTransition`, which the completing
+branch runs over `batch._pendingNodes`. `commitPendingNodes` drains that list
+without clearing anything, so a node committed by an earlier drain kept
+pointing at a transition that later finished. `setSignal` re-enters
+`el._transition` before it discovers a write changes nothing, and a loading
+boundary rewrites the same flag on every drain pass, so a finished transition
+was re-armed forever and `flush()` never ended: dev threw "Potential Infinite
+Loop Detected", production has no counter on that loop and hung. The
+re-entered transition also aliases the ambient batch's containers, so
+`initTransition`'s adoption pass could push into the array it was iterating
+until `RangeError: Invalid array length`.
--- a/packages/signals/src/core/scheduler.ts
+++ b/packages/signals/src/core/scheduler.ts
@@ -843,7 +843,16 @@ export function setPatchCommitHook(fn: (batch: Transition) => void): void {
 function commitPendingNodes() {
   const pendingNodes = currentBatch._pendingNodes;
   for (let i = 0; i < pendingNodes.length; i++) {
-    commitPendingNode(pendingNodes[i]);
+    const node = pendingNodes[i];
+    commitPendingNode(node);
+    // A committed node has left the transaction, and the drain below is the only
+    // record that it was ever in it. `reassignPendingTransition` clears the stamp
+    // at completion, but only for nodes still in this list, so a node committed by
+    // an earlier drain would keep pointing at a transition that later finishes.
+    // A finished transition re-entered through that stamp cannot complete again:
+    // `setSignal` re-enters it before it discovers the write changes nothing, and a
+    // boundary rewrites the same flag on every drain pass, so the drain never ends.
+    node._transition = null;
   }
   pendingNodes.length = 0;
   storeCommitHook?.();
--- a/packages/signals/src/core/scheduler.ts
+++ b/packages/signals/src/core/scheduler.ts
@@ -845,13 +845,6 @@ function commitPendingNodes() {
   for (let i = 0; i < pendingNodes.length; i++) {
     const node = pendingNodes[i];
     commitPendingNode(node);
-    // A committed node has left the transaction, and the drain below is the only
-    // record that it was ever in it. `reassignPendingTransition` clears the stamp
-    // at completion, but only for nodes still in this list, so a node committed by
-    // an earlier drain would keep pointing at a transition that later finishes.
-    // A finished transition re-entered through that stamp cannot complete again:
-    // `setSignal` re-enters it before it discovers the write changes nothing, and a
-    // boundary rewrites the same flag on every drain pass, so the drain never ends.
     node._transition = null;
   }
   pendingNodes.length = 0;


ryansolid:
Verified and merging — thank you for the follow-through from the report to the patch. Your placement (the commit loop) and coverage are exactly right; I confirmed `commitPendingNode` has no other callers, so this closes the leak at the one funnel every commit passes through. Complementary hardening rides in #3148 on top of this: `initTransition` refusing a chased-dead transaction (covers stamps that survive via non-commit paths — merged `_done` chains, settles racing completion), the white-box regression pins for both layers, and the loop-guard attribution you asked for in the issue (your case would print `done=true, pending=0`).

Issue #3129, opened 2026-08-30
A `GET()` declaration outlives the function it was made about


frenzzy:
### Describe the bug

`GET()` records its declaration against a function **id**. `registerServerFunction` rebinds that id freely. Nothing connects the two, so a declaration made about one function can end up governing a different one — and it carries both of the things a declaration grants: GET dispatch, and the skip of the origin gate (#3114).

```js
// packages/web/server-functions/src/server.ts:653-658
const REGISTRATIONS = new Map();
const METHODS = new Map();

// :690  — rebinds the id, silently, and never touches METHODS
export function registerServerFunction(id, callback) {
  provideRPC();
  REGISTRATIONS.set(id, callback);
  return callback;
}

// :917  — records the declaration against the id, and only the id
export function GET(fn) {
  ...
  METHODS.set(fn.id, "GET");

// :2219 — dispatch consults METHODS alone
const declaredRead =
  (method === "GET" || method === "HEAD") &&
  functionId !== null &&
  METHODS.get(functionId) === "GET";
```

`declaredRead` then feeds both gates: the method allowlist at :2277 and `protectsRequest` at :2227. So the function that now answers to the id inherits GET-reachability *and* the origin-gate exemption, whatever it does.

```
declared read, bare GET             -> 200 {"which":"READ"}
after re-registration, no headers   -> 200 {"which":"MUTATION","mutated":1}
after re-registration, cross-site   -> 200 {"which":"MUTATION","mutated":2}

v1 declares GET, bare GET           -> 200 {"v":1}
v2 does NOT declare GET, cross-site -> 200 {"v":2,"sideEffect":"sent mail"}

never declared GET (control)        -> 403
cross-site POST (control)           -> 403

mutations executed over GET: 2
```

The two controls are what make it a bug rather than a curiosity: a function that never declared GET is refused, and POST to the very same id is refused. The only thing separating them is a declaration made about a function that is no longer there.

### Why it matters in production

Two ways to get there, and they are not equally likely — worth separating so the severity is not overstated.

[...]

PR #3144, opened 2026-08-31
Fix 3127 universal text

### LINKED ISSUE #3127: babel-plugin: universal renderer HTML-escapes static text (`<` becomes `&lt;`) and leaves JSX entities undecoded
**Version:** `@solidjs/babel-plugin` 2.0.0-rc.3 (rc.4 and `packages/babel-plugin/src` on `next` at 5230666 have the same code)

### Describe the bug

With `generate: "universal"`, static text children are HTML-escaped before they reach the host's `createTextNode`, and JSX entities in text are never decoded. The universal renderer has no HTML parser downstream, so the escaped string is what ends up on screen:

```tsx
const A = <item>{"<span> restyles a run"}</item>
const B = <item>{"<b> & <i>"}</item>
const C = <item>lit &lt; text</item>
```

compiles to

```js
_$insertNode(_el$, _$createTextNode(`&lt;span> restyles a run`));
_$insertNode(_el$3, _$createTextNode(`&lt;b> &amp; &lt;i>`));
_$insertNode(_el$5, _$createTextNode(`lit &lt; text`));
```

`A` and `B` show the escaping: text-mode `escapeHTML` replaces `&` and `<` only, which is why `>` survives and the output looks like a stray escaping step rather than a policy. `C` is the other half: the entity written the JSX way is passed through raw, so there is no spelling at all that produces a literal `<` in static text under a custom renderer. Dynamic text (`insert(el, () => expr)`) is unaffected, which is why this hides until a string literal with `<` or `&` shows up.

Two sites, both in the shared path used by every generator:

1. `universal/element.ts` `transformChildren` calls `transformNode(path)` with no `info`, so a static child expression takes the escaping branch in `shared/transform.ts`:

   ```ts
   const text = staticValue !== undefined
     ? info.doNotEscape ? String(staticValue) : escapeHTML(String(staticValue))
     : trimWhitespace(node.extra?.raw ?? "");
   ```

2. The `JSXText` branch of the same expression emits `node.extra.raw`, the undecoded source text.

Both are right for the DOM generator, where the string is spliced into an HTML template that the parser unescapes, and for `<script>` / `<style>` (raw-text elements, where `doNotEscape` already selects the raw path). They are wrong for universal, whose text goes straight into the host's `createTextNode`.

The universal fixture pins the current output rather than guarding against it, and does so inconsistently. `test/__universal_fixtures__/textInterpolation` expects:

```js
_$createTextNode(`&nbsp;&lt;Hi&gt;&nbsp;`)                 // <span>&nbsp;&lt;Hi&gt;&nbsp;</span>
_$createComponent(Comp, { children: "\xA0<Hi>\xA0" })       // <Comp>&nbsp;&lt;Hi&gt;&nbsp;</Comp>
_$createTextNode(`Hi&lt;script>alert();&lt;/script>`)       // <span>Hi{"<script>alert();</script>"}</span>
```

The same JSX text reaches a component decoded and an element raw, and the `injection` case is the DOM generator's XSS guard carried into a renderer with nothing to guard. The component line is the correct shape for all three.

Transform-only, no app needed:

```js
import { transformAsync } from "@babel/core"
import jsx from "@babel/plugin-syntax-jsx"
import solid from "@solidjs/babel-plugin"

const code = `
const A = <item>{"<span> restyles a run"}</item>
const B = <item>{"<b> & <i>"}</item>
const C = <item>lit &lt; text</item>
`
const r = await transformAsync(code, {
  filename: "x.jsx",
  plugins: [jsx, [solid, { moduleName: "r-custom", generate: "universal" }]],
})
console.log(r.code)
```

1. Run the script above with `@solidjs/babel-plugin` 2.0.0-rc.3.
2. Look at the `createTextNode` arguments: `&lt;` and `&amp;` in A and B, raw `&lt;` in C.
3. Render through any universal host (`@solidjs/universal` `createRenderer` with a `createTextNode` that stores the string): the six characters `&lt;` are what the host receives.

### Expected behavior

Under `generate: "universal"` text is text: `createTextNode` receives `<span> restyles a run`, `<b> & <i>` and `lit < text`. Suggested fix, universal only: `transformChildren` passes `{ doNotEscape: true }` to `transformNode`, and the `JSXText` branch uses the decoded `node.value` (Babel decodes entities into `value` and keeps the source in `extra.raw`) when `config.generate === "universal"`. `trimWhitespace` applies unchanged since `value` and `raw` differ only in the entities. DOM and SSR generators keep their behaviour; the three fixture lines above change to the decoded form.

### Platform

- OS: any (compile-time)
- Browser: n/a, custom renderer
- Version: `@solidjs/babel-plugin` 2.0.0-rc.3, `@babel/core` 7.28.x

### Additional context

Found while building a docs-style demo on a universal renderer, where element names in prose (`<span>`, `<For>`) are the normal case for static text. There is no user-side workaround short of moving every such string into a dynamic expression.

nickshiro:
## Summary

fix #3127

Universal renderers pass static text directly to the host's `createTextNode`, but compiler treated that text like DOM template HTML. Static expressions were HTML-escaped, while JSX entities remained encoded.

changeset:
  - keeps static expression text unescaped in universal element children;
  - decodes JSX entities after whitespace normalization, preserving entities like `&nbsp;`;
  - supports `universal` and `dynamic` universal transforms without changing DOM/SSR output;
  - adds fixture coverage for literal `<`, `&`, and JSX entities.

ran targeted universal fixtures:

```bash
pnpm --filter @solidjs/babel-plugin exec vitest run test/universal.spec.js test/dynamic-universal.spec.js
```
result: Test Files  2 passed (2) Tests 16 passed (16)

```bash
pnpm --filter @solidjs/babel-plugin test
```
result: Test Files  20 passed (20) Tests 153 passed (153)

and prtettier with git diff checks

### DIFF
--- a/.changeset/fix-universal-static-text.md
+++ b/.changeset/fix-universal-static-text.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/babel-plugin": patch
+---
+
+Pass decoded, unescaped static text to universal renderers instead of HTML-encoded source text.
--- a/packages/babel-plugin/src/shared/transform.ts
+++ b/packages/babel-plugin/src/shared/transform.ts
@@ -1,4 +1,5 @@
 import * as t from "@babel/types";
+import { decode } from "html-entities";
 import { transformElement as transformElementDOM } from "../dom/element";
 import { createTemplate as createTemplateDOM } from "../dom/template";
 import { transformElement as transformElementSSR } from "../ssr/element";
@@ -191,7 +192,9 @@ export function transformNode(
         ? info.doNotEscape
           ? String(staticValue)
           : (escapeHTML(String(staticValue)) as string)
-        : trimWhitespace((node.extra?.raw as string | undefined) ?? "");
+        : t.isJSXText(node) && info.decodeEntities
+          ? decode(trimWhitespace((node.extra?.raw as string | undefined) ?? ""))
+          : trimWhitespace((node.extra?.raw as string | undefined) ?? "");
     if (!(text as string).length) return null;
     const results: TransformResult = {
       template: text,
--- a/packages/babel-plugin/src/types.ts
+++ b/packages/babel-plugin/src/types.ts
@@ -143,6 +143,7 @@ export interface TransformInfo {
   fragmentChild?: boolean;
   componentChild?: boolean;
   doNotEscape?: boolean;
+  decodeEntities?: boolean;
   skipId?: boolean;
   toBeClosed?: Set<string>;
   parentResults?: TransformResult;
--- a/packages/babel-plugin/src/universal/element.ts
+++ b/packages/babel-plugin/src/universal/element.ts
@@ -274,7 +274,7 @@ function transformChildren(path: BabelPath<t.JSXElement>, results: UniversalTran
   const filteredChildren = filterChildren(path.get("children")),
     multi = checkLength(filteredChildren),
     childNodes = filteredChildren
-      .map(path => transformNode(path))
+      .map(path => transformNode(path, { doNotEscape: true, decodeEntities: true }))
       .reduce((memo: TransformResult[], child) => {
         if (!child) return memo;
         const i = memo.length;
--- a/packages/babel-plugin/test/__universal_fixtures__/textInterpolation/code.js
+++ b/packages/babel-plugin/test/__universal_fixtures__/textInterpolation/code.js
@@ -52,6 +52,10 @@ const escape3 = <>
 
 const injection = <span>Hi{"<script>alert();</script>"}</span>;
 
+const staticLessThan = <item>{"<span> restyles a run"}</item>;
+const staticAmpersand = <item>{"<b> & <i>"}</item>;
+const jsxEntity = <item>lit &lt; text</item>;
+
 let value = "World";
 const evaluated = <span>Hello {value + "!"}</span>;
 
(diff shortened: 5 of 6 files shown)


ryansolid:
Closing with a straight account and an apology: you opened this three hours before I landed 320f1f5c, I failed to re-check open PRs before implementing, and by the time I saw it the overlapping fix was on next. That's the second time this cycle I've stepped on your volunteered work (#3129's neighborhood was the first) and it's a process failure on my end, not a judgment of this PR.

On the substance, your babel half was right and matches what landed almost exactly — same TransformInfo-flag shape, same universal transformChildren site, same decode-after-trim order. What it couldn't have known about from the issue: the same class extends to string ATTRIBUTES (`normal="Search&hellip;"` reaching setProp raw), and the oxc compiler in this repo mirrors the babel plugin bug-for-bug behind a cross-mode parity suite that would have failed this diff — both halves needed to move together, which is why it went in-house for the release.

Your fixture cases were better than mine — the report's exact three shapes — so they're adopted on next in 63c7cb58 with your co-authorship, and they now run through both compilers' parity union. Thank you, and sorry again; if you're still up for contributing, ping me on an issue and I'll flag it yours before an agent session can eat it.

nickshiro:
@ryansolid I'm very pleased with the constructive dialogue and the transparency of the development process. Nothing terrible has happened - just think of it as further confirmation of the hypotheses, since we arrived at them independently. Have a great day!

PR #3148, opened 2026-08-31
Refuse re-entering a finished transaction; attribute the flush loop guard (#3140)

ryansolid:
## Summary

Companion to #3143 (merged — @snowfluke's clear of `_transition` stamps at commit), completing #3140. Stacked on #3145.

- **The belt**: a dead `_transition` reference can still reach `initTransition` from outside the commit path — merged `_done` forwarding chains, async settles racing completion — and `setSignal` re-opens a node's stamped transaction *before* the value-equal bail, so re-activating the corpse spins the drain loop (dev threw the flush loop guard; production hung the tab). `initTransition` now refuses a transaction whose chased `_done` chain ends in `true`, as a bare return — the reporter measured that redirecting to a fresh batch re-arms the loop with a new identity each pass.
- **Guard attribution** (the reporter's explicit ask in #3140): the dev loop guard now reports what kept the loop alive — scheduled work vs an active transition, the transition's done-state and queue counts, and the last staged node. The corpse signature reads `done=true, pending=0`. Dev-only, tree-shaken from prod.
- **Regression pins**, white-box since the wild stamping race did not minimize: commit clears the stamp (pins #3143); a value-equal write to a dead-stamped node leaves `activeTransition` null (fails without the belt when a dead stamp survives by any path).
- **Size/perf**: the belt is one compare folded into the existing transaction-open check; no loops, no allocation. One size budget ratcheted 26.1 → 26.15 KB (~25 B shared across #3143, #3145, and this).

## Test plan

- [x] Resurrection pin fails with #3143 alone (planted dead stamp), passes with the belt
- [x] Full signals suite (1437) green post-rebase on #3143; size budgets pass

### DIFF
--- a/.changeset/fix-dead-transition-stamp-revival.md
+++ b/.changeset/fix-dead-transition-stamp-revival.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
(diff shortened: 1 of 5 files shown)


ryansolid:
Landed directly on next as a cherry-pick (GitHub closed this PR instead of retargeting when its stacked base branch was deleted with #3145's merge). Content unchanged; signals suite green (1437 passed).

Issue #3104, opened 2026-08-29
2.0.0-rc.4  | internal `latest()` in a `memo` vs external latest and inconsistencies


mizulu:
### Describe the bug

playground uses latest commit 95229453 

logs
```
[2026-08-29T23:05:39.611Z] [log] A pre set {m1: false, count: false}
[2026-08-29T23:05:39.611Z] [log] B post set 1 {m1_internal_latest: false, m1_external_latest: true, count: true}
[2026-08-29T23:05:39.612Z] [log] C post set 2 false
```

### Your Example Website or App

https://s.olid.uk/id/Rbwj3COqSPm4tyXIt5se_A

### Steps to Reproduce the Bug or Issue

internal latest() in a memo vs external latest on a memo 
not sure what is the expectation 

also see the issue with the latest(m1) read changing the result of 
the `latest(() => isPending(count))` in log point C 

### Expected behavior

you tell me 

### Platform

.

### Additional context

```tsx
import { render } from '@solidjs/web';
import { createSignal, isPending, createMemo, latest } from 'solid-js';






export default function App() {
  const [count, setCount] = createSignal(0);


  setTimeout(() => {


    // latest() is internal to the memo
    const m1 = createMemo(() => latest(() => isPending(count)));


    console.log("A", "pre set", {
      m1: m1(),  // false 
      count: isPending(count)  // false 
    })

    setCount(v => v + 1)  // start a transiton 


    console.log("B", "post set 1", {
      m1_internal_latest: m1(),  // false  ( memo did not detect change with internal latest read)
      m1_external_latest: latest(m1),  // true  
                                      //( if you comment this `latest(m1)` call
                                      //C  logs true instead of false  )
      count: isPending(count)  // true 
    })

    console.log("C", "post set 2", latest(() => isPending(count))) // false ( unless latest(m1) above is not called )

  })

  return (
    <div class="p-2 whitespace-pre">
      {`
      a memo of an isPending,  reads inconsistently 


    `}

    </div>
  );
}


if (typeof document !== 'undefined') {
  render(() => <App />, document.getElementById('root')!);
}

```

ryansolid:
Fixed in 90603c56 (ships in rc.5). You found two stacked defects in the verdict layer, and the "commenting `latest(m1)` changes C" observation was the key clue.

[...]

Issue #3099, opened 2026-08-29
[compiler] `validate` warns but still emits the broken template walk


Mad-Kat:
## Summary

The `validate` pass in the native compiler already detects markup the HTML parser will restructure, and its diagnostic is excellent (it prints the exact DOM the browser will build). But it only warns to the console, without a file or location, and compilation proceeds to emit the template and its positional walk as written. In a Vite dev session the warning is buried in server stdout while the browser fails with an unrelated-looking runtime crash. We would like an opt-in error mode.

## Reproduction

```tsx
export function Probe(props: { label: string }) {
  return (
    <p>
      <div>relocated?</div>
      <span>{props.label}</span>
    </p>
  );
}
```

At compile time, `validate` (default on) reports the problem correctly, but only as a console message with no filename or location, and still emits the broken output:

```
The HTML provided is malformed and will yield unexpected output when evaluated by a browser.

User HTML:
 <p><div>#text</div><span></span></p>
Browser HTML:
 <p></p><div>#text</div><span></span><p></p>
```

At runtime, the component crashes with `Cannot read properties of null (reading 'nextSibling')`, and nothing connects the crash to the compile-time warning. In a Vite dev session the warning sits in the server terminal (or a log file), not in the browser or the error overlay, so the crash is what developers debug. Depending on the shape of the relocation the walk can also survive and bind to the wrong nodes instead of crashing.

[This test case](https://github.com/DigitecGalaxus/next-yak/pull/632/changes#diff-1d4319c0f8942fd0aef94ced289a6f8229abf56de05a02632e0da27400014524) verifies that the written styles apply. The paragraph styles the box through a descendant selector. It passes unfolded (programmatic insertion) and fails folded, where the relocation silently unstyles the box.

## Proposal

An error mode for the existing validator, e.g. `validate: "error"` (keeping `true` = warn as today):

- Fail the transform with the diagnostic the validator already produces, plus filename and source location. Vite then surfaces it in the overlay at the right file, like any compile error.
- Expose the option through `@solidjs/vite-plugin`.

The current warn-and-emit behavior gives the diagnostic exactly when it is hardest to see, and ships code that is guaranteed not to match its own walk. Once the validator has fired, there is no case where evaluating the emitted template is correct.

## Why we care downstream

[...]

PR #3162, opened 2026-08-31
feat(web): add typed preload links to the SSR asset pipeline

v-sayapin:
## Summary

This reopens the work from [dom-expressions#582](https://github.com/ryansolid/dom-expressions/pull/582) in its new home after the DOM runtime was absorbed into the Solid monorepo. It also completes the typed-resource follow-up from [#3031](https://github.com/solidjs/solid/pull/3031), which taught `lazy().preload()` to register a module's full static JS/CSS closure. The remaining gap was resources whose request cannot be described by a URL alone.

The SSR asset pipeline tracked stylesheets and modulepreloads as bare URLs and inferred the link kind from the `.css` suffix. That works for JS and CSS, but leaves no way to carry fonts, images, MIME types, CORS mode, SRI, or fetch priority through the same pipeline.

Static manifests and `AssetResolver` results can now attach explicit `PreloadLink[]` entries:

```ts
preloads: [
  { href: "/assets/inter.woff2", as: "font", type: "font/woff2", crossorigin: "" },
  { href: "/assets/hero.avif", as: "image", type: "image/avif", fetchpriority: "high" }
]
```

The links are forwarded through the entry manifest, `lazy()`'s render and `preload()` paths, its synchronous `moduleUrl` path for islands, and `clientOnly()`. Under `NoHydration`, CSS and non-script preloads are kept while script preloads are skipped unless `moduleUrl` access explicitly signals that the client will load the module.

`as`, MIME type, CORS mode, integrity, referrer policy, fetch priority, and media survive `renderToString`, streaming, embedded-head and custom-sink output, as well as frame transport and late frame delivery. Script and style preloads also receive the matching split CSP nonce.

Links share the existing head resource identity set, so a manifest link and a `useHead` link dedupe in either registration order. `integrity`, `referrerpolicy`, and `fetchpriority` remain first-registration metadata rather than part of the preload identity; the browser's reuse key is determined by the URL, destination, mode, and credentials mode.

Preloads stay explicit. Raw manifest `assets` are not preloaded automatically and `dynamicImports` are not traversed, so the runtime never guesses a request destination or defeats code splitting. The integration that knows the asset type and preload policy selects the links.

The existing stylesheet/modulepreload path stays unchanged: it still uses `Set<string>`, and the typed-link array is allocated only when the first link is registered. Validation and identity dedupe happen before the document and frame output forms are built.

This also exports the six HTML preload destinations as `JSX.HTMLPreloadAs` and fetch priority as `JSX.HTMLFetchPriority` for reuse across the public types.

The second commit fixes late root assets in frames. Root asset chunks share one wire key, so each arrival previously replaced the retained record and a frame mounted later could see only the last one. The warm store now extends the retained record in place and consumes every arrival.

Responsive image preloads (`imagesrcset` / `imagesizes`, including the standard form without `href`) remain a separate follow-up.

Added runtime and type coverage for:

- typed font, image, script, style, fetch, and track preloads
- MIME type, CORS mode, integrity, referrer policy, fetch priority, media, and nonce routing
- invalid or missing destinations and empty manifest URLs
- `_base` resolution through the static import graph while `assets` and `dynamicImports` remain untouched
- dedupe with `useHead` in both registration orders and conflicting first-registration metadata
- pre-shell, late streaming, embedded-head, custom-sink, and frame delivery
- repeated late root asset chunks in both cold and already-mounted frame stores
- forwarding through `lazy()` render, `lazy().preload()`, `moduleUrl`, `clientOnly()`, and `NoHydration`
- public server/server-mock type parity

Validation:

- `pnpm exec turbo run build test test-types typecheck --filter=solid-js --filter=@solidjs/web --filter=@solidjs/h --force` — 14/14 tasks passed, 0 cached
- `pnpm test` — 33/33 workspace tasks passed
- size limits passed; client scenarios are unchanged and the full frame consumer is 11.19 kB within its 11.22 kB limit
- Prettier and `git diff --check` passed

One generator note: `pnpm --filter @solidjs/web run jsx-sync` also removes a `$key` block that is present in `jsx.d.ts` but missing from `jsx-h.d.ts`. That drift predates this branch and is intentionally not included here.

### DIFF
--- a/.changeset/typed-preload-links.md
+++ b/.changeset/typed-preload-links.md
@@ -1,0 +1,19 @@
+---
+"@solidjs/web": patch
+"@solidjs/h": patch
+"solid-js": patch
+---
+
+Add typed preload links to the server asset pipeline.
+
+Static manifests can attach `preloads: PreloadLink[]`, resolver results can carry the same shape for framework integrations, and any integration can register a link with `registerAsset("preload", link)`. The runtime preserves `as`, MIME type, CORS mode, integrity, referrer policy, fetch priority, and media attributes across string, streaming, embedded-head, custom-sink, and frame renders.
+
+`lazy()` and `clientOnly()` forward resolver-provided preload links alongside their JS and CSS.
+
+`JSX.HTMLPreloadAs` and `JSX.HTMLFetchPriority` are now exported for reuse.
+
+Preload links are explicit: manifest `assets` are not preloaded automatically. Existing stylesheet and modulepreload APIs are unchanged.
+
+Development builds warn when font or fetch preloads omit `crossorigin`, because a different eventual request mode cannot reuse that preload.
+
+Frame clients also retain and consume every late root asset record instead of dropping earlier records that reuse the same transport key.
--- a/packages/h/jsx-runtime/src/jsx.d.ts
+++ b/packages/h/jsx-runtime/src/jsx.d.ts
@@ -1066,6 +1066,7 @@ export namespace JSX {
   type HTMLFormEncType = "application/x-www-form-urlencoded" | "multipart/form-data" | "text/plain";
   type HTMLFormMethod = "post" | "get" | "dialog";
   type HTMLCrossorigin = "anonymous" | "use-credentials" | EnumeratedAcceptsEmpty;
+  type HTMLFetchPriority = "high" | "low" | "auto";
   type HTMLReferrerPolicy =
     | "no-referrer"
     | "no-referrer-when-downgrade"
@@ -1091,19 +1092,8 @@ export namespace JSX {
     | "allow-top-navigation"
     | "allow-top-navigation-by-user-activation"
     | "allow-top-navigation-to-custom-protocols";
-  type HTMLLinkAs =
-    | "audio"
-    | "document"
-    | "embed"
-    | "fetch"
-    | "font"
-    | "image"
-    | "object"
-    | "script"
-    | "style"
-    | "track"
-    | "video"
-    | "worker";
+  type HTMLPreloadAs = "fetch" | "font" | "image" | "script" | "style" | "track";
+  type HTMLLinkAs = HTMLPreloadAs | "audio" | "document" | "embed" | "object" | "video" | "worker";
 
   interface AnchorHTMLAttributes<T> extends HTMLAttributes<T> {
     download?: FunctionMaybe<string | EnumeratedAcceptsEmpty | RemoveAttribute>;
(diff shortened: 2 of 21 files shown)


ryansolid (changes requested):
Reviewed in depth — the substance is right and this completes the #3031 typed-resource thread the way we intended when dom-expressions#582 was parked. The explicit-links-only policy (no destination guessing, no `assets` auto-preload, no `dynamicImports` traversal) is exactly the stance we want, the `HTMLPreloadAs` six-destination split matches the fetch spec's preload destinations, and the first-registration-metadata vs identity split mirrors the browser's preload cache key. The frames identity-consumption + warm-store accumulation interaction also traces correctly for both mounted and late-mount paths.

Three changes needed:

1. **Trim the manifest type to our contract.** Drop `assets?: string[]` and `dynamicImports?: string[]` from `AssetManifest` (both `server.ts` and `server-mock.ts`). The runtime deliberately never reads them, and the manifest type is *our* contract, not a Vite-compat surface — we may feed this from non-Vite bundlers, and extra fields on parsed JSON pass through structurally without being in the type. Keep exactly what the pipeline consumes: `file`, `css`, `imports`, `isEntry`, `preloads`, `_base`.

2. **Rebase onto current `next`.** The size-limit failure isn't your measurement being wrong — your branch measured green, but recent landings on `next` (optimistic-store replay work, among others) share the same budgets, and the merge ref tips four scenarios over by 11–49 bytes.

3. **Re-measure and set the budgets consciously on the rebased branch** — including the frames client limit you already bumped (11.22 kB will no longer clear; CI's merge measurement was 11.27 kB).

The local `diagnostics.spec.tsx` failure you'd see on the current branch is a stale-base artifact (the validate-error compiler change landed with its test fix after your base) — the rebase resolves it.

ryansolid:
Pushed the requested changes directly (maintainer edits):

- **Merged current `next`** — this picks up the rc.5 correctness batch and the fixed `diagnostics.spec.tsx` (the local failure you'd see on the old base was a stale-base artifact of #3099 making malformed HTML a compile error).
- **Trimmed `AssetManifest`** (df453d7) — dropped `assets`/`dynamicImports` from the type in `server.ts`/`server-mock.ts`. The manifest type is Solid's own contract, not a Vite-compat surface; parsed Vite manifests still satisfy it structurally, and the runtime never read those fields. The `preload-links.spec.js` no-traversal test still pins the policy via the untyped JSON literal.
- **Reconciled size budgets** — most of the CI overage was actually `next`'s: the rc.5 batch (#3104/#3122/#3123) had drifted six scenarios over budget, now fixed on `next` itself (a4a20e1e). On top of that base, this PR's frames delta measures ~210 B (11.06 → 11.27 KB), budget set to 11.28 KB (47b1949f).

`web` (151) and `solid` (576) suites pass locally, size-limit exits clean. Thanks for the feature — the substance needed no changes.

v-sayapin:
Sorry — I force-pushed before seeing [df453d7](https://github.com/solidjs/solid/commit/df453d7) and [47b1949f](https://github.com/solidjs/solid/commit/47b1949f), so they got overwritten.
Content came out identical, but your comment wording was lost.
I have both restored locally and can send a follow-up, or leave it to you.

Issue #3182, opened 2026-09-01
[Solid 2 regression] Direct value and checked bindings overwrite pre-hydration user input


nickshiro:
### Describe the bug

In Solid 2, direct reactive `value` and `checked` bindings overwrite live
  form state that the user changed before hydration.

  The server and client initial values can match exactly, so this is not an
  SSR/client rendering mismatch.

  For example:

  ```tsx
  container.innerHTML = '<input _hk="0" value="server">';

  const input = container.firstElementChild as HTMLInputElement;

  // Simulate the user typing before the client bundle hydrates.
  input.value = "user";

  hydrate(() => {
    const [value] = createSignal("server");
    return <input value={value()} />;
  }, container);

  console.log(input.value);
  // "server"
  ```

  The user-entered `"user"` value is replaced with the initial signal value.

  The same happens with `checked`:

  ```tsx
  container.innerHTML =
    '<input _hk="0" type="checkbox" checked>';

  const input = container.firstElementChild as HTMLInputElement;

  // Simulate a user interaction before hydration.
  input.checked = false;

  hydrate(() => {
    const [checked] = createSignal(true);
    return <input type="checkbox" checked={checked()} />;
  }, container);

  console.log(input.checked);
  // true
  ```

  Interestingly, an equivalent spread binding preserves the live DOM value:

  ```tsx
  hydrate(() => {
    const [value] = createSignal("server");
    return <input {...{ value: value() }} />;
  }, container);

  console.log(input.value);
  // "user"
```

### Steps to Reproduce the Bug or Issue

1. Server-render an input with an initial value.
2. Change the live `input.value` before hydration, simulating user input.
3. Hydrate using a signal whose initial value matches the server.
4. Inspect the input value.

### Expected behavior

Initial hydration should preserve the live user-modified value. The first subsequent reactive update should apply the signal value.

### Platform

- `@solidjs/web`: `2.0.0-rc.4`
- `next`: `f051db6a`
- Native and Babel compiler paths
- jsdom 25

### Additional context

This appears to be a Solid 2 regression of #1849. That issue described input being cleared before queued input events could be replayed and was marked fixed in Solid 1.8.

The direct compiler path treats stateful DOM properties as locked and emits a direct property assignment during hydration. The spread path reaches `assignProp()`, which skips the initial property write for hydrating nodes.

Related but different: #2016 concerns keeping controlled inputs synchronized during later reactive updates. This issue is specifically about preserving browser state created before the initial hydration pass.

[...]

Issue #3167, opened 2026-09-01
Solid 2 RC.4: should static and reactive `<select value>` behave differently?


yumemi-thomas:
I was testing the reproduction from [solidjs/solid#1754](https://github.com/solidjs/solid/issues/1754) in the Solid 2 Playground.

```tsx
import { createSignal } from "solid-js";
import { render } from "@solidjs/web";

function App() {
  const [selected] = createSignal("2");

  return (
    <div>
      <select value="2">
        <option value="" />
        <option value="1">1</option>
        <option value="2">2</option>
        <option value="3">3</option>
        <option value="4">4</option>
      </select>

      <select value={selected()}>
        {(() => (
          <>
            <option value="" />
            <option value="1">1</option>
            <option value="2">2</option>
            <option value="3">3</option>
            <option value="4">4</option>
          </>
        )) as unknown as Element}
      </select>

      <select {...{}} value={selected()}>
        {(() => (
          <>
            <option value="" />
            <option value="1">1</option>
            <option value="2">2</option>
            <option value="3">3</option>
            <option value="4">4</option>
          </>
        )) as unknown as Element}
      </select>
    </div>
  );
}

if (typeof document !== "undefined") {
  render(() => <App />, document.getElementById("root")!);
}
```

On Solid 2 RC.4, I get:

- `value="2"` → empty option
- `value={selected()}` → `2`
- `{...{}} value={selected()}` → `2`

The static form appears in the generated DOM as:

```html
<select value="2">
```

Browsers do not use a `value` attribute on `<select>` to determine its current selection. The reactive forms appear to assign the live `select.value` property instead.

For comparison, the equivalent Svelte 5 code displays `2`, `2`, `2`:

```svelte
<script lang="ts">
  let selected = $state("2");
</script>

<div>
  <select value="2">
    <option value=""></option>
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
    <option value="4">4</option>
  </select>

  <select value={selected}>
    <option value=""></option>
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
    <option value="4">4</option>
  </select>

  <select {...{}} value={selected}>
    <option value=""></option>
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
    <option value="4">4</option>
  </select>
</div>
```

[...]

PR #3183, opened 2026-09-01
feat(web): support responsive image preloads

v-sayapin:
## Summary

Follow-up to [#3162](https://github.com/solidjs/solid/pull/3162), which landed typed preload links but left responsive images out. `imagesrcset` and `imagesizes` decide *which* URL the browser fetches, so a typed link that can't carry them only half-describes the request.

`<link rel="preload" as="image">` now carries both, including the standard form with no `href`:

```ts
preloads: [
  { href: "/hero.avif", as: "image", imagesrcset: "/hero.avif 1x, /hero@2x.avif 2x", imagesizes: "50vw" },
  { as: "image", imagesrcset: "/card-400.avif 400w, /card-800.avif 800w", imagesizes: "100vw" }
]
```

Dropping `href` is the spec's own advice for the source-set form — it only serves browsers without `imagesrcset` support, and there it tends to preload the wrong candidate. React does the same for the same reason. `PreloadLink` becomes a union so the type states the three legal shapes instead of leaving them to runtime checks, and `as` stays the spec's set of preload destinations exactly, since anything else translates to null and the browser ignores the link.

Candidate URLs have to arrive resolved. `_base` still joins `href`, but the source set rides verbatim — rewriting each candidate would put a srcset parser on the render path for something the producer already knows.

**Bad responsive attributes don't take the link down with them.** On a non-image destination the attribute is filtered and the preload still ships, so an integration computing `imagesrcset` for every asset keeps its script and style links. `""` and non-string values count as absent too — coercing them would emit `imagesrcset="42"`, which no browser can parse, and then build an identity out of it.

That makes validation order load-bearing. The destination decides whether a source set *is* a source, so `as` is resolved and the pair normalized before the "has a source" check. The other way round accepted `imagesrcset` on a non-image destination as the source, then filtered that same attribute away, leaving `<link rel="preload" as="script">` with nothing to fetch.

### The identity fix underneath

Its own commit: a bug in what #3162 landed, not part of this feature. Resource identity compared raw prop values, and disagreed with its own markup three ways:

- **`false` was a value, not absence.** Both attribute writers drop it, so `crossorigin={cond && "anonymous"}` emitted a second, byte-identical link.
- **`crossorigin` was compared by spelling.** It's a CORS settings attribute — three states, not a string range: absent is No CORS, `use-credentials` is Use Credentials, every other present value is Anonymous. Eight registrations of one font across five Anonymous spellings gave six links; now three.
- **Fields could forge each other.** `type: "a:media=b"` collided with `type: "a", media: "b"`, and `/loader:type=6:module` read as `/loader` plus `type: "module"`. Every free-form field is length-prefixed now, the URL included.

One shared `qualifierValue` owns the rules and applies exactly what registration applies — `as` folds ASCII-case, an empty or non-string `imagesrcset`/`imagesizes` reads as absent — otherwise a client `as="IMAGE"` or `imagesrcset: ""` mounted a second link beside the server's. Identity, the document client and the frame client's mirror all route through it. So `mountHeadResource` stops adopting across a different destination or CORS mode, and a source-set link becomes adoptable at all: no href, so it matches a null one plus the qualifiers. Before that a responsive preload duplicated on hydration.

### Dev warnings

Two conformance gaps now report instead of shipping quietly: a width descriptor with no `imagesizes` (the source size falls back to `100vw`, so the preload can miss the candidate the `` picks), and a relative candidate in a manifest source set — not gated on `_base`, since it resolves against the document URL either way. Both walk the source set the way the spec's parser does, URL and descriptors apart, so `/w,400/hero.avif` is one URL and `https://cdn.example/image,400w 1x` is a density candidate, not a width descriptor.

### Left alone

Head order is unchanged — moving an image preload ahead of render-blocking CSS could lengthen the critical path, and Chrome takes image priority from `fetchpriority`, not position. Same for a preload budget: Angular can warn past a limit because it *generates* those links, whereas this pipeline is handed explicit descriptors and can't know which resource is the LCP element. That belongs in an integration.

`sink.shell`'s meta containers are live rather than snapshots (a late registration pushes into them *and* arrives via `sink.asset`). Documented at the handoff site rather than changed while that API is experimental.

The leading commit is unrelated: it restores the `AssetManifest` note and frames budget rationale from #3162 (df453d7, 47b1949f), lost to a force-push before the merge. Rebased on `next` past the patch-channel removal (#3229).

Extended the existing suites rather than adding parallel ones. New cases cover both responsive shapes on every emission path; the pair filtered on a non-image destination and kept on an image; five Anonymous `crossorigin` spellings collapsing to one link; a delimiter-carrying value and URL not suppressing another resource; `as="IMAGE"` and filtered-empty responsive values adopting the server's link on the identity, the document client and the frame client; both warnings firing once and staying quiet for density-only art direction and commas inside a candidate URL.

- `pnpm test` — 32/33 workspace tasks. The one failure is `@solidjs/web#test` on three `element.spec.tsx` cases (#3167); those and five hydrate-config cases (#3182, ref-spread parity) fail identically on `next` at `ff96a671`, which CI has green. `packages/web` otherwise: 1578 passed / 2 skipped across its three configs.
- `size-limit` — 8/8. Frames client `11.30 → 11.40 kB`: `11372 B` against `11266 B` on `next`, +106 B for the whole branch. Simple-app floor `10.73 → 10.74`: `10731 B` against `10700 B` with only `dist/web.js` swapped while the tree-shaken bundle is byte-identical — two `head.ts` helpers this bundle never reaches shift esbuild's identifier allocation and brotli pays 31 B for the names. Layout drift, ratcheted per the file's rule; rationale in `.size-limit.js`.
- Prettier, `test-types`, `typecheck` and `git diff --check` clean. CodSpeed's `merge` swings are in untouched `packages/signals` with a runtime-environment mismatch the report itself flags; acknowledged as noise.
- Checked in Chrome against the built `dist`: the source-set form selects the right candidate and the matching `` makes no second request; two `media`-gated art-direction links fetch only the matching one; a sourceless descriptor emits nothing and a coerced one emits no garbage attribute; eight font registrations produce two links in the live DOM; and `as="script"`/`as="style"` preloads are CSP-blocked without their nonce and served with it.

### DIFF
--- a/.changeset/preload-identity-qualifiers.md
+++ b/.changeset/preload-identity-qualifiers.md
@@ -1,0 +1,13 @@
+---
+"@solidjs/web": patch
+---
+
+Canonicalize resource identity qualifiers instead of comparing raw prop values, so two declarations of one request dedupe to one `<link>` on both sides of hydration.
+
+`false` now means absent, matching both attribute writers: `crossorigin={cond && "anonymous"}` no longer emits a second, byte-identical link when the condition is false.
+
+`crossorigin` is compared by its CORS state rather than its spelling. It is a CORS settings attribute with three states — absent is No CORS, `use-credentials` (ASCII case-insensitive) is Use Credentials, and every other present value including `""`, a bare attribute and an invalid one is Anonymous — so the same font is no longer preloaded once per spelling, and the client adopts the server's link instead of mounting a second one for a request the browser already has.
+
+Qualifier values are length-prefixed, so a value containing the identity delimiters can no longer collide with a different qualifier set and silently suppress another resource (`type: "a:media=b"` and `type: "a", media: "b"` were one identity).
+
+Client-side adoption of a mount-once head resource now matches a server-emitted element on the full request identity rather than the href alone: two preloads sharing an href still differ if their destination, CORS mode, type, media or source set differ. The document client, the standalone frame client and the server all apply the same rules.
--- a/.changeset/responsive-preload-links.md
+++ b/.changeset/responsive-preload-links.md
@@ -1,0 +1,11 @@
+---
+"@solidjs/web": patch
+---
+
+Support `imagesrcset` and `imagesizes` in typed image preloads, including the standard form without `href`. Candidate URLs inside `imagesrcset` must already be resolved by the integration.
+
+The responsive pair is image-only. On any other destination the attribute is dropped and the link still ships — an integration that computes `imagesrcset` for every asset keeps its script and style preloads. An empty or non-string value counts as absent for the same reason, so a source set is never emitted as garbage the browser cannot parse. A descriptor whose only source was such a filtered attribute is dropped entirely rather than emitted as a `<link rel="preload">` with nothing to fetch.
+
+`mountHeadResource` can adopt a source-set link: it has no href, so it matches a server-emitted link on a null href plus the identity qualifiers — the rule the frame client already applied.
+
+Development builds warn when `imagesrcset` uses a width descriptor without `imagesizes` (the source size falls back to `100vw`, so the preload can miss the image the `<img>` selects), and when a manifest source set carries a relative candidate — candidates are not joined with `_base`, so they resolve against the document URL whichever base the manifest declares. That check walks the source set the way the spec's parser does, so commas inside a candidate URL are not mistaken for candidate separators.
--- a/packages/web/frames/src/frame-client.ts
+++ b/packages/web/frames/src/frame-client.ts
@@ -2061,6 +2061,17 @@ function parseFragment(html) {
 // Mirrors head.ts without importing it into the standalone frame client.
 const PRELOAD_QUALIFIERS = ["as", "crossorigin", "type", "media", "imagesrcset", "imagesizes"];
 
+// Mirrors head.ts's qualifierValue: `crossorigin` is three states, not a
+// string range, so `""`, a bare attribute and `anonymous` are one request.
+// Frame `attrs` are already strings, but the document may carry any spelling.
+function qualifierValue(name, value) {
+  if (value == null) return null;
+  if (name !== "crossorigin") return value;
+  return value.length === 15 && value.toLowerCase() === "use-credentials"
+    ? "use-credentials"
+    : "anonymous";
+}
+
 /** Attribute-compared head lookup so href/id values never need escaping. */
 function findHeadElement(selector, attr, value, qualifiers) {
   candidate: for (const node of document.head.querySelectorAll(selector)) {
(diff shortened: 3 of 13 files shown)


ryansolid (changes requested):
The overall responsive-preload shape is sound, the focused tests pass against current `next`, and the reported CodSpeed signals swings are environment noise. I found three correctness gaps to address before merge:

1. **Resource identities are still forgeable at the URL boundary** (`packages/web/src/head.ts`, `resourceIdentity`). Qualifier values are now length-prefixed, but the leading URL is not. For example, `/loader:type=6:module` with no `type` collides with `/loader` plus `type=\"module\"`, suppressing a distinct valid resource. Please encode every identity field unambiguously, including the URL, and add a collision regression test.

2. **Registration and hydration adoption canonicalize qualifiers differently.** Registration lowercases `as` and drops empty/non-string responsive attributes, while `findAssetElement` / the frame client's `findHeadElement` compare raw `as`, `imagesrcset`, and `imagesizes`. Standards-equivalent input such as `as=\"IMAGE\"` vs `as=\"image\"`, or omitted vs filtered-empty responsive values, can fail adoption and duplicate an SSR link. Please route server identity, document-client adoption, and frame-client adoption through the same canonicalization rules and pin both cases.

3. **`hasWidthDescriptor` is not srcset-aware.** Scanning the raw string misreads a valid density candidate such as `https://cdn.example/image,400w 1x` as containing a width descriptor and emits a false missing-`imagesizes` warning. Reuse the candidate/descriptor parser rather than searching the whole string.

The note about needing a matching `@dom-expressions/runtime` change appears stale and is not a blocker: that runtime was absorbed here, and the corresponding dom-expressions PR was closed in favor of the earlier Solid work. The size increases are small and justified once the correctness gaps are closed.

v-sayapin:
Done!
All three fixed in the last commit.
The branch is rebased on `next`.
Stale `@dom-expressions/runtime` note dropped.

ryansolid (approved):
> Re-reviewed against `94f8f04f` and `9dd8d0f5` — all three requested changes are resolved:
>
> 1. `resourceIdentity` length-prefixes the URL as well as every qualifier, so the identity string is unambiguous end to end; the `/loader:type=…` collision is pinned by `keeps URLs from forging qualifier fields`.
> 2. One `qualifierValue` canonicalizer (ASCII-folded `as`, filtered empty/non-string responsive attrs, three-state `crossorigin`) now runs on both sides of server identity, client `findAssetElement`, and the frame client's `findHeadElement`; the frame client's mirror is equivalent for the string/null domain it sees. Pinned in both adoption suites.
> 3. `hasWidthDescriptor` walks candidates spec-style (URL = maximal non-whitespace run, descriptors read only to the next comma), so `https://cdn.example/image,400w 1x` is one density candidate. Pinned.
>
> Verified on a local merge with current `next`: web build clean, `test-types` clean, client 704/704, server 741/741. Size deltas (+31 B floor layout drift, +29 B hydrating doc, +106 B frames client) are documented in the caps and proportionate. The destination-before-source ordering fix also closes the sourceless `as="script"` emission. Merging.
>
> — *Claude via Cursor*

PR #3190, opened 2026-09-01
fix(web): preserve reusable bound event tuples

### LINKED ISSUE #3186: Reusing a bound event handler tuple breaks non-delegated events
### Describe the bug

Reusing the same `[handler, data]` tuple on multiple elements breaks non-delegated `on*` events such as `onScroll`. The first element receives the correct event. On the second element, the handler receives the bound data in place of the `Event`.
This happens because the direct-listener path replaces `handler[0]` with an element-specific wrapper, mutating the user-provided tuple. When that tuple is reused, Solid wraps the already wrapped function again. The delegated event path, such as `onClick`, does not appear to be affected.

  ```tsx
  import { render } from "solid-js/web";
  import type { JSX } from "solid-js";

  let second!: HTMLDivElement;

  const handler: JSX.EventHandlerUnion<HTMLDivElement, Event> = [
    (data, event) => {
      console.log({
        data,
        event,
        currentTarget: event.currentTarget
      });
    },
    "shared"
  ];

  function App() {
    return (
      <>
        <div onScroll={handler}>First</div>
        <div ref={second} onScroll={handler}>
          Second
        </div>

        <button
          onClick={() => {
            second.dispatchEvent(new Event("scroll"));
          }}
        >
          Dispatch scroll
        </button>
      </>
    );
  }

  render(() => <App />, document.getElementById("app")!);
```

  Click the button to dispatch scroll on the second element.

  A regression test demonstrating the issue:

```tsx
  test("the same handler tuple can be reused on multiple elements", () => {
    const calls: Array<[string, HTMLDivElement, Event]> = [];
    let second!: HTMLDivElement;

    const handler: JSX.EventHandlerUnion<HTMLDivElement, Event> = [
      (data, event) => calls.push([data, event.currentTarget, event]),
      "shared"
    ];

    createRoot(() => (
      <>
        <div onScroll={handler} />
        <div ref={second} onScroll={handler} />
      </>
    ));

    const event = new Event("scroll");
    second.dispatchEvent(event);

    expect(calls).toEqual([["shared", second, event]]);
  });
```

  The assertion currently fails because the received values are effectively:

```tsx
  [["shared", undefined, "shared"]]
```

### Expected behavior

The tuple should remain unchanged and be reusable across elements.

  The handler installed on the second element should receive:

  1. "shared" as the bound data.
  2. The dispatched Event.
  3. The second element as event.currentTarget.

  Solid should not mutate the user-provided handler tuple while installing the listener.

### Platform

- Solid: reproduced with 1.9.15 and next at f051db6a.
- Test environment: Vitest with jsdom.
- OS: Linux.

### Additional context

The mutation happens in the non-delegated array-handler branch:

```tsx
  const handlerFn = handler[0];
  node.addEventListener(
    name,
    (handler[0] = e => handlerFn.call(node, handler[1], e))
  );
```

 The wrapper likely needs to be stored separately for each element and event instead of replacing handler[0].

nickshiro:
## Summary

fix #3186

Avoid mutating user-provided bound event handler tuples when installing non-delegated event listeners.

Previously, the direct-listener path replaced `handler[0]` with an element-specific wrapper. Reusing the same tuple on another element therefore wrapped the already wrapped function and caused the bound data to be passed in place of the `Event`.

The listener wrapper is now kept local to each element, leaving the original tuple unchanged and reusable.

Regression test and patch changeset are included.

 Built the web package:

  ```sh
  pnpm --filter @solidjs/web build
  cd packages/web
  ../../node_modules/.bin/vitest run test/event-handler.spec.tsx
  JSX_COMPILER=babel ../../node_modules/.bin/vitest run test/event-handler.spec.tsx
```

  Checked test types and formatting:

```sh
  ../../node_modules/.bin/tsc --project tsconfig.test.json
  ../../node_modules/.bin/prettier --check \
    src/client.ts \
    test/event-handler.spec.tsx \
    ../../.changeset/preserve-bound-event-tuples.md
```

  Ran the complete @solidjs/web test suites:

```sh
  ../../node_modules/.bin/vitest run
  ../../node_modules/.bin/vitest run --config vite.config.server.mjs
  ../../node_modules/.bin/vitest run --config vite.config.hydrate.mjs
```

  Results:

  - Client: 700 tests passed
  - Server: 536 passed, 2 skipped
  - Hydration: 151 tests passed

### DIFF
--- a/.changeset/preserve-bound-event-tuples.md
+++ b/.changeset/preserve-bound-event-tuples.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/web": patch
+---
+
+Keep bound handler tuples reusable across non-delegated events by leaving the user-provided tuple unchanged when installing a listener.
--- a/packages/web/src/client.ts
+++ b/packages/web/src/client.ts
@@ -614,7 +614,7 @@ export function addEvent(
   name: string,
   handler: EventListener | EventListenerObject | (EventListenerObject & AddEventListenerOptions),
   delegate: boolean
-): void;
+): EventListener | EventListenerObject | void;
 
 export function addEvent(node, name, handler, delegate) {
   if (delegate) {
@@ -624,8 +624,11 @@ export function addEvent(node, name, handler, delegate) {
     } else node[`$$${name}`] = handler;
   } else if (Array.isArray(handler)) {
     const handlerFn = handler[0];
-    node.addEventListener(name, (handler[0] = e => handlerFn.call(node, handler[1], e)));
+    const listener = e => handlerFn.call(node, handler[1], e);
+    node.addEventListener(name, listener);
+    return listener;
   } else node.addEventListener(name, handler, typeof handler !== "function" && handler);
+  return delegate ? undefined : handler;
 } /** Compiler-emitted primitive; not for hand-written code. @internal */
 export function style(
   node: Element,
@@ -2042,8 +2045,9 @@ function assignProp(node, prop, value, prev, skipRef, nodeName) {
       node.removeEventListener(name, h);
     }
     if (delegate || value) {
-      addEvent(node, name, value, delegate);
+      const attached = addEvent(node, name, value, delegate);
       delegate && delegateEvents([name]);
+      if (!delegate) return attached;
     }
   } else if (
     (hasNamespace && prop.slice(0, 5) === "prop:") ||
--- a/packages/web/src/client.ts
+++ b/packages/web/src/client.ts
@@ -2041,13 +2041,16 @@ function assignProp(node, prop, value, prev, skipRef, nodeName) {
     const name = prop.slice(2).toLowerCase();
     const delegate = DelegatedEvents.has(name);
     if (!delegate && prev) {
+      // Bound tuples keep an internal [attached listener, authored tuple]
+      // record so an unrelated spread rerun can preserve listener identity.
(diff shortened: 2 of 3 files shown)


ryansolid:
Thanks — not mutating `handler[0]` is the right direction, and your tests pass. But there's a second dependency on that mutation which makes this incomplete as-is: the non-delegated removal path in `assignProp` calls `node.removeEventListener(name, prev[0])`, which only matches today *because* the wrapper was written back into the tuple. With the mutation gone, the wrapper is unreachable at removal time.

Concrete repro (verified locally against this branch): spread-rebind `{...props}` where a non-delegated tuple handler like `onScroll` gets a new tuple — the old wrapper stays attached and both handlers fire, so each rebind leaks a listener.

Suggested shape: keep the tuple immutable, but store the created wrapper somewhere removable — e.g. on the element keyed by event name (matching the existing `_$styles`/`_$classes` expando pattern), or have `addEvent` return the wrapper so the spread path can stash it in `prevProps` for the next diff. A regression test that rebinds a non-delegated tuple through spread and asserts the old handler no longer fires would pin it.

Happy to re-review once that path is covered.

nickshiro:
I went with the approach: addEvent returns the attached listener, which assignProp stores for later removal. This keeps the listener lifecycle inside the existing props diffing flow without adding hidden state to the element. I think that's really cool and the right thing to do.

Issue #3189, opened 2026-09-01
Class handles primitive values inconsistently between static, dynamic, and array forms


nickshiro:
### Describe the bug

Primitive values accepted by `JSX.ClassValue` are handled differently depending on whether they are static, dynamic, or nested inside an array.

In particular:
- A static numeric value is emitted as a class.
- The same numeric value returned by a signal is ignored.
- A standalone `true` inside an array becomes the literal class `"true"`.

This also differs from the documented “clsx-style composition”, where standalone boolean values are ignored.

### Steps to Reproduce the Bug or Issue

```tsx
  import { createRoot, createSignal } from "solid-js";

  createRoot(() => {
    const [value] = createSignal(1);

    const staticClass = <div class={1} /> as HTMLDivElement;
    const dynamicClass = <div class={value()} /> as HTMLDivElement;
    const arrayClass = <div class={[true, false, "active"]} /> as HTMLDivElement;

    console.log(staticClass.outerHTML);
    console.log(dynamicClass.outerHTML);
    console.log(arrayClass.outerHTML);
  });
```

### Expected behavior

Static and dynamic forms of the same accepted ClassValue should behave consistently.

Since the enhanced class prop is documented as supporting clsx-style composition, standalone boolean values inside arrays should not produce a "true" class:

```tsx
  <div class="active"></div>
```

For numeric values, either:

1. Numbers should be normalized consistently in static, dynamic, CSR, and SSR paths; or
2. Unsupported top-level numeric values should be removed from the ClassValue type and handled consistently by the compiler.

### Additional context

JSX.ClassValue currently includes number and boolean at the top level:

```tsx
  export type ClassValue =
    | string
    | number
    | boolean
    | null
    | undefined
    | Record<string, boolean>
    | ClassValue[];
```

I reproduced this on the current next branch at b20bdff0.

The reproduction type-checks successfully. I also confirmed the behavior with both the native Rust JSX compiler and the Babel compiler.

nickshiro:
@ryansolid I’d be happy to work on this after our accept.

After tracing the class normalization paths for #3188, I think the consistent behavior should follow the documented `clsx-style` semantics:
  - strings and numbers produce class names;
  - standalone booleans, null, and undefined are ignored;
  - the same rules apply at the top level and inside arrays;
  - object keys are included when their values are truthy.
  
  align the compiler, client runtime, and SSR paths and add parity tests for both the native and Babel compilers.
  Does that sound like the intended behavior?

[...]

PR #3191, opened 2026-09-01
fix(web): track in-place class mutations

### LINKED ISSUE #3188: Object class values do not update after an in-place mutation with equals: false
### Describe the bug

In Solid 2, an object passed to the `class` prop does not update the DOM when that object is mutated in place and its signal uses `{ equals: false }`.

The signal correctly notifies its dependents, but the `class` runtime compares the new value against the same already-mutated object. Consequently, previously applied classes are not removed and newly enabled classes are not added.

`style` object values already handle this case by keeping a separate snapshot of the declarations applied to the DOM. The object form of `class` does not appear to keep an equivalent snapshot.

1. Render the example.
2. Inspect the target element. It initially has class="before".
3. Click Update classes.
4. Inspect the element again.

The following regression test also demonstrates the issue:

```tsx
  test("removes classes after an in-place object update", () => {
    const classes = { before: true, after: false };
    let div!: HTMLDivElement;
    let notify!: () => void;

    createRoot(() => {
      const [value, setValue] = createSignal(classes, { equals: false });
      notify = () => setValue(classes);
      div = <div class={value()} /> as HTMLDivElement;
    });

    expect(div.className).toBe("before");

    classes.before = false;
    classes.after = true;
    notify();
    flush();

    expect(div.className).toBe("after");
  });
```

The final assertion currently fails:

Expected: "after"
Received: "before"

### Expected behavior

After the signal notification, the target element should have:

```tsx
<div id="target" class="after">Target</div>
```

The before class should be removed and the after class should be added.

equals: false is documented as always notifying dependents even when the signal value retains the same identity, so DOM bindings should process the updated object contents.

### Platform

- Solid: next at f051db6a
- Test environment: Vitest with jsdom
- OS: Linux

### Additional context

className() receives the same object for both value and prev:

  prev = classListToObject(prev || {});
  value = classListToObject(value);

  Because the object was already mutated before the signal notification, both normalized values contain the new state. The runtime therefore cannot determine which classes were previously applied to the DOM.

  The nearby style() implementation avoids the same problem by storing a separate snapshot on the element:

  let applied = node._$styles;

  A similar per-element snapshot of classes applied by className() could allow in-place object updates to be diffed without removing unrelated classes.

  Related but different: #2673 reported stale dependencies for an object nested inside a class array. That issue was closed as a dependency/build problem; this reproduction consistently reaches the class update
  but loses the previous object state.

nickshiro:
## Summary

fix #3188

Object-valued `class` bindings did not update correctly when the same object was mutated in place and passed through a signal using `{ equals: false }`.

The previous value retained the same object reference, so the class diff compared the mutated object against itself and could no longer determine which classes had previously been applied.

This change stores an independent snapshot of the classes applied by `className()`. It allows in-place mutations to be diffed correctly while preserving classes added by external code.

The snapshot is also initialized during hydration so subsequent in-place updates behave consistently on claimed DOM nodes.

Added regression tests covering:

  - In-place mutation of an object-valued `class` binding
  - Preservation of classes added by external code
  - Transitions between object, string, and null class values
  - In-place mutation after hydration

  Ran:

  ```sh
  pnpm --filter @solidjs/web build
  pnpm --filter @solidjs/web types
  cd packages/web
  ../../node_modules/.bin/vitest run
  ../../node_modules/.bin/vitest run --config vite.config.server.mjs
  ../../node_modules/.bin/vitest run --config vite.config.hydrate.mjs
  ../../node_modules/.bin/tsc --project tsconfig.test.json
  ../../node_modules/.bin/prettier --check \
    src/client.ts \
    test/class.spec.tsx \
    test/hydration/class.spec.tsx \
    ../../.changeset/track-applied-classes.md
```

  Results:
  - Client: 701 tests passed
  - Server: 536 tests passed, 2 skipped
  - Hydration: 152 tests passed
  - TypeScript and Prettier checks passed

### DIFF
--- a/.changeset/track-applied-classes.md
+++ b/.changeset/track-applied-classes.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/web": patch
+---
+
+Update object-valued class bindings after in-place mutations by tracking a separate snapshot of the classes applied to each element.
--- a/packages/web/src/client.ts
+++ b/packages/web/src/client.ts
@@ -562,27 +562,39 @@ export function setAttributeNS(node, namespace, name, value) {
 export function className(node: Element, value: JSX.ClassValue, prev?: JSX.ClassValue): void;
 
 export function className(node, value, prev) {
-  if (isHydrating(node)) return;
   // Numbers stringify like the compiler's static output (`class={1}`
   // inlines as `class="1"` in the template) so static and dynamic forms of
   // the same ClassValue behave identically (#3189).
   if (typeof value === "number") value = "" + value;
   if (typeof prev === "number") prev = "" + prev;
+  if (isHydrating(node)) {
+    // Seed applied state without touching the claimed DOM so later in-place
+    // mutations can still be diffed after hydration completes.
+    node._$classes = value && typeof value === "object" ? classListToObject(value) : undefined;
+    return;
+  }
   if (value == null || value === false) {
-    prev && node.removeAttribute("class");
+    if (prev || node._$classes) {
+      node.removeAttribute("class");
+      node._$classes = undefined;
+    }
     return;
   }
   if (typeof value === "string") {
+    node._$classes = undefined;
     value !== prev && node.setAttribute("class", value);
     return;
   }
+  // Track classes applied by className() itself. value/prev are user-owned
+  // and may be the same object on shared-effect reruns.
+  let applied;
   if (typeof prev === "string") {
-    prev = {};
+    applied = {};
     node.removeAttribute("class");
-  } else prev = classListToObject(prev || {});
+  } else applied = node._$classes || classListToObject(prev || {});
   value = classListToObject(value);
   const classKeys = Object.keys(value || {});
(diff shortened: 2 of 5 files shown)


ryansolid:
I independently validated the implementation and tests: the per-element applied-class snapshot is the right fix, including hydration seeding and preservation of externally added classes. This is good to land.

`next` has since landed #3189 in the same `className` function, so the PR now conflicts. Could you rebase onto current `next`? The semantic resolution is mechanical: keep #3189's number-to-string normalization at the top of `className`, retain your `_$classes` snapshot/object-diff path, and keep #3189's standalone-boolean filtering in `flattenClassList` below. Once rebased, the stale size failure should also rerun against the current baseline (the old failure was broad signals/app budget drift, not attributable to this web-only change).

nickshiro:
@ryansolid, thanks for your review. I went to bed yesterday and wasn't able to reply right away.

Issue #3146, opened 2026-08-31
Optimistic store truth-flight should own its transaction rather than rely on ambient adoption


ryansolid:
Follow-up (2) from #3141 / #3145.

The generator store's authoritative flight (initial run, per-yield landings, post-await draft writes) has no transaction of its own: each landing's settle path asks `resolveTransition` and otherwise joins whatever is ambient (`initTransition(null)`). #3145 closed the window in which a *stranger* transaction could be ambient at landing time, but the underlying shape remains: the flight's relationship to transactions is circumstantial (whatever is ambient or stamped at each landing) rather than declared.

Giving the truth-flight a dedicated transaction would make the semantics explicit:

- every landing of one flight belongs to the same, flight-owned transaction — entanglement with other work happens only through the documented graph-driven channels (A15), never through timing;
- the #2951 contract (a bare optimistic write during the store's own truth-flight rides the truth) gets a first-class anchor — today it routes through `fam.node._transition`, which is whatever last stamped the firewall, and #3141 showed that stamp can be a stranger's;
- `transitionBlocked`'s store-half (hold while flight pending + overrides live) stops needing to reconstruct ownership from `_optimisticStores` membership.

Needs design attention on: per-yield commit timing (each yield should still reveal on arrival per A18(1) — one flight-transaction must not batch all yields to flight end), and interaction with `refresh()`-initiated flights which already carry the refresh transaction.

ryansolid:
Fixed in fb5061c9 on next — the optimistic store truth-flight now owns a declared transaction rather than relying on ambient adoption. Ships in the next RC.

Issue #3092, opened 2026-08-29
Proposal: branded store types — deep Store<T> + StorePart<T>


brenelz:
I was thinking how to make stores less magical and was thinking through this. Not sure if its a good idea, but trying to show through types if this is a store instead of just a plain object. Maybe we want it to be transparent but was just an idea. On talking with AI I kinda realized the problem is when you pass just a part of the store around. If you are using the whole store you probably know and you can't really highlight leaf reads

Below is AI generated:

## Problem

Stores are indistinguishable from plain data in the type system. `Store<T>` is a bare `Readonly<T>` alias, and the store-ness evaporates almost immediately:

- For array stores it's gone at the **first binding**: `Readonly<Todo[]>` normalizes to `readonly Todo[]`, so `const [todos] = createStore<Todo[]>(...)` hovers with no trace of "store" anywhere.
- Nested views lose it structurally: `todos[0]`, `store.user`, a `.filter()` element, and a `<For>` item all type as plain `Todo` — even though at runtime every one of them is a live reactive proxy.
- Nothing stops a store view from flowing silently into a component typed with a plain shape (`function Test(props: { todo: Todo })` accepts `todos[0]`), after which the receiving code has no way to know it's holding reactive state.
- Type-aware tooling (lint rules for store destructuring/untracked reads, editor highlighting of store reads) has no reliable anchor: alias names don't survive `Omit`/`Pick`, generic inference, or mapped types.

## Proposal

Make `Store<T>` a **deep phantom brand**, and give nested views their own name:

```ts
declare const STORE_BRAND: unique symbol;
export interface StoreBrand {
  readonly [STORE_BRAND]?: true; // optional: labels, never rejects
}

type StoreValue<V> = [V] extends [NotWrappable]
  ? V
  : IsStoreType<V> extends true
    ? V
    : StorePart<V>;

/** A nested view INTO a store — live and reactive, no setter of its own. */
export type StorePart<T> = { readonly [K in keyof T]: StoreValue<T[K]> } & StoreBrand;
export type Store<T>     = { readonly [K in keyof T]: StoreValue<T[K]> } & StoreBrand;
```

`Store` and `StorePart` are structurally identical (mutually assignable, same brand); the distinct name exists purely so hovers tell you what you hold — the root you created, or a part of one.

### What you see

```ts
const [todos] = createStore<Todo[]>([...]);

todos          // Store<Todo[]>
todos[0]       // StorePart<Todo>          (was: Todo)
[...]

PR #3194, opened 2026-09-01
Proposal: make derived store APIs type-safe and ergonomic

GabbeV:
> [!NOTE]
> Both the implementation and this description were developed with substantial AI assistance and then manually iterated and reviewed.

## Summary

Derived stores currently accept `seed: Partial<T>` while exposing both the callback draft and the resulting store as complete `T`. That is not type-safe: required properties may be absent at runtime while TypeScript says they are present.

This PR closes that hole and adds an ergonomic alternative for projections that do not have a natural complete seed:

1. Explicit seeds must be a complete `T`.
2. Plain-object projections may omit the seed when their callback takes no draft and produces a complete `T`.

The same distinction is implemented for `createStore`, `createProjection`, and `createOptimisticStore`.

### Complete seeded form

When a seed is provided, the callback receives a real `T` draft and may mutate it, return a replacement, or return `void`:

```ts
createProjection<State>(
  draft => {
    draft.count++;
  },
  { count: 0, label: "ready" }
);
```

The seed is `T`, not `Partial<T>`, because it is immediately exposed as `T`. Callable roots are also rejected consistently because store proxies cannot represent callable objects.

Arrays and tuples continue to use this form. An empty array is normally a natural complete seed:

```ts
createProjection(() => [{ id: 1 }], []);
```

`seedLoadingValue: true` remains available only for the seeded form, where a complete seed can honestly act as commit zero.

Code that intentionally initializes a partial object through mutation can still opt out explicitly:

```ts
createProjection<State>(
  draft => {
    draft.count = source();
    draft.label = "ready";
  },
  {} as any
);
```

The cast is required because neither TypeScript nor the runtime can prove that every required property is established before observation.

### Return-only seedless form

For a plain-object root, the seed may be omitted or nullish:

```ts
createProjection(() => ({ count: 0, label: "ready" }));
createProjection(async () => await loadState());

createProjection<State>(async function* () {
  yield await loadInitialState();
  yield await loadUpdatedState();
});
```

This selects a separate callback contract:

```ts
() => T | Promise<T> | AsyncIterable<T>
```

The callback receives no draft and every resolution must provide a complete value. Runtime validation rejects missing values, primitives, arrays, and non-plain-object roots if the type system is bypassed.

Seedless arrays and tuples are rejected because the proxy target shape must be known when the store is created. They can use the seeded form with an empty or complete array instead.

A nullish placeholder is only needed when passing options:

```ts
createProjection(derive, undefined, { shallow: true });
```

## Runtime behavior

The seeded/seedless distinction is carried through the signals, optimistic, server, and hydration implementations:

- seeded projections continue to run against their supplied draft
- seedless projections start pending and adopt their first complete returned value
- each public seedless async-iterator yield is a complete value
- hydration treats the first serialized entry as a complete snapshot and later patch batches as a separate internal replay operation, without exposing a draft to the public seedless callback
- invalid seedless results enter the existing projection error path

`ProjectionOptions` contains options shared by both forms. `SeededProjectionOptions` adds `seedLoadingValue`, which is only meaningful when an explicit seed exists.

## Remaining question: setters before initialization

The writable seedless forms still return the existing `StoreSetter<T>`. Before the first result commits, however, there is no valid `T` to pass to a functional setter callback.

This PR intentionally does not introduce a provisional policy for that case. The broader issue also affects other derived writable primitives and is tracked in #3274. The earlier API discussion is in [this comment](https://github.com/solidjs/solid/pull/3194#issuecomment-5502732800).

## Testing

Coverage includes:

- inference and explicit-generic tests for seeded and seedless forms
- rejection of incomplete seeds, callable roots, and seedless arrays
- synchronous, promise, and async-iterable initialization
- invalid seedless runtime results
- server rendering and client hydration, including snapshot-plus-patch replay

### DIFF
--- a/.changeset/type-safe-derived-stores.md
+++ b/.changeset/type-safe-derived-stores.md
@@ -1,0 +1,6 @@
+---
+"@solidjs/signals": patch
+"solid-js": patch
+---
+
+Require complete explicit seeds for derived stores, preventing required properties from being absent at runtime while exposed as present in the store type. Object projections may instead omit the seed and use a return-only initializer that supplies a complete value on every resolution; array projections continue to require an explicit seed.
--- a/packages/signals/src/store/index.ts
+++ b/packages/signals/src/store/index.ts
@@ -16,12 +16,12 @@ export { isWrappable, $TRACK, $PROXY, $TARGET } from "./store.js";
 
 import type {
   NoFn,
+  NoArray,
   ProjectionOptions,
   SeededProjectionOptions,
   Store,
   StoreOptions,
-  StoreSetter,
-  SeedlessRoot
+  StoreSetter
 } from "./store.js";
 import type { Refreshable } from "../core/index.js";
 import {
@@ -47,17 +47,17 @@ export {
 /** Public createStore: plain form `(initialValue, options?)` and derived writable
  * forms `(fn)` / `(fn, seed, options?)`. */
 export function createStore<T extends object = {}>(
-  initialValue: NoFn<T> | Store<NoFn<T>>,
+  initialValue: T & NoFn<T>,
   options?: StoreOptions
 ): [get: Store<T>, set: StoreSetter<T>];
 export function createStore<T extends object = {}>(
-  fn: (() => T | Promise<T> | AsyncIterable<T>) & SeedlessRoot<T>,
+  fn: (() => T | Promise<T> | AsyncIterable<T>) & NoArray<T> & NoFn<T>,
   seed?: null,
   options?: ProjectionOptions
 ): [get: Refreshable<Store<T>>, set: StoreSetter<T>];
 export function createStore<T extends object = {}>(
-  fn: (draft: T) => void | T | Promise<void | T> | AsyncIterable<void | T>,
-  seed: NoFn<T> | Store<NoFn<T>>,
+  fn: ((draft: T) => void | T | Promise<void | T> | AsyncIterable<void | T>) & NoFn<T>,
+  seed: T,
   options?: SeededProjectionOptions
 ): [get: Refreshable<Store<T>>, set: StoreSetter<T>];
 export function createStore(first: any, second?: any, third?: any): any {
--- a/packages/signals/src/store/index.ts
+++ b/packages/signals/src/store/index.ts
@@ -6,6 +6,7 @@ export type {
   StoreNode,
   StoreOptions,
   ProjectionOptions,
+  SeededProjectionOptions,
   NotWrappable,
   SolidStore
 } from "./store.js";
@@ -13,7 +14,15 @@ export type { Merge, Omit } from "./utils.js";
 
 export { isWrappable, $TRACK, $PROXY, $TARGET } from "./store.js";
 
-import type { NoFn, ProjectionOptions, Store, StoreOptions, StoreSetter } from "./store.js";
+import type {
+  NoFn,
+  ProjectionOptions,
+  SeededProjectionOptions,
+  Store,
+  StoreOptions,
+  StoreSetter,
+  SeedlessRoot
+} from "./store.js";
 import type { Refreshable } from "../core/index.js";
 import {
   createStoreNext,
@@ -24,20 +33,32 @@ import {
 import { reconcileNextState } from "./next/reconcile.js";
 import { createStoreDerivedNext } from "./next/projection.js";
 
-export { createProjectionNext as createProjection } from "./next/projection.js";
+export {
+  createProjectionNext as createProjection,
+  createProjectionHydrationReplayNext as createProjectionHydrationReplay,
+  createStoreHydrationReplayNext as createStoreHydrationReplay
+} from "./next/projection.js";
 export { storeIsShallow, storeHasFamily, storeHasOptimisticFamily } from "./next/store.js";
-export { createOptimisticStoreNext as createOptimisticStore } from "./next/optimistic.js";
+export {
+  createOptimisticStoreNext as createOptimisticStore,
+  createOptimisticStoreHydrationReplayNext as createOptimisticStoreHydrationReplay
+} from "./next/optimistic.js";
 
 /** Public createStore: plain form `(initialValue, options?)` and derived writable
- * form `(fn, seed, options?)`. */
+ * forms `(fn)` / `(fn, seed, options?)`. */
 export function createStore<T extends object = {}>(
   initialValue: NoFn<T> | Store<NoFn<T>>,
   options?: StoreOptions
 ): [get: Store<T>, set: StoreSetter<T>];
 export function createStore<T extends object = {}>(
-  fn: (draft: T) => void | T | Promise<void | T> | AsyncIterable<void | T>,
-  seed: Partial<T> | Store<NoFn<T>>,
+  fn: (() => T | Promise<T> | AsyncIterable<T>) & SeedlessRoot<T>,
+  seed?: null,
   options?: ProjectionOptions
+): [get: Refreshable<Store<T>>, set: StoreSetter<T>];
+export function createStore<T extends object = {}>(
+  fn: (draft: T) => void | T | Promise<void | T> | AsyncIterable<void | T>,
+  seed: NoFn<T> | Store<NoFn<T>>,
+  options?: SeededProjectionOptions
 ): [get: Refreshable<Store<T>>, set: StoreSetter<T>];
 export function createStore(first: any, second?: any, third?: any): any {
   if (typeof first === "function") return createStoreDerivedNext(first, second, third);
--- a/packages/signals/src/store/next/optimistic.ts
+++ b/packages/signals/src/store/next/optimistic.ts
@@ -64,6 +64,7 @@ import {
 } from "../store.js";
 import {
   runProjectionComputedNext,
+  validateSeedlessOptions,
   validateStoreValue,
   createReplayStoreValidator,
   type ProjectionResultValidator
@@ -210,8 +211,7 @@ export function createOptimisticStoreNext<T extends object = {}>(
   const derived = typeof first === "function";
   const seeded = !derived || second != null;
   const options = (derived ? third : second) as SeededProjectionOptions | undefined;
-  if (!seeded && options?.seedLoadingValue)
-    throw new Error("seedLoadingValue requires an explicit store seed");
+  if (!seeded) validateSeedlessOptions(options);
   const derive =
     derived && !seeded ? () => (first as () => T | Promise<T> | AsyncIterable<T>)() : first;
   return createOptimisticStoreNextInternal<T>(
@@ -388,6 +388,7 @@ export function createOptimisticStoreHydrationReplayNext<T extends object = {}>(
   replaying: () => boolean,
   options?: ProjectionOptions
 ): [get: Refreshable<Store<T>>, set: StoreSetter<T>] {
+  validateSeedlessOptions(options);
   return createOptimisticStoreNextInternal<T>(
     fn,
     {} as T,
--- a/packages/signals/src/store/next/optimistic.ts
+++ b/packages/signals/src/store/next/optimistic.ts
@@ -56,11 +56,19 @@ import {
   markRawIngest,
   type NoFn,
   type ProjectionOptions,
+  type SeededProjectionOptions,
   type Store,
   type StoreOptions,
-  type StoreSetter
+  type StoreSetter,
+  type SeedlessRoot
 } from "../store.js";
-import { runProjectionComputedNext } from "./projection.js";
+import {
+  runProjectionComputedNext,
+  validateStoreValue,
+  createReplayStoreValidator,
+  type ProjectionFn,
+  type ProjectionResultValidator
+} from "./projection.js";
 import {
   bumpDeep,
   authoritativeRead,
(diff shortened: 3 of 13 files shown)


ryansolid:
Triaging this against #3092 — this PR is effectively the implementation vehicle for the branded-store proposal there (an optional, type-only source marker on `Store<T>` plus ordered overloads that let the store arm recover the source `T` before the plain arm consumes it). I did a full design assessment of that approach, posted on #3092: https://github.com/solidjs/solid/issues/3092#issuecomment-5500718406

The load-bearing findings that apply directly to reviewing this PR (all probed against the repo's `typescript@6.0.3`):

- **The ordered-overload pattern works** — a store seed binds through the fallback arm and the draft recovers the real `T` (mutable arrays/fields intact, intentional `readonly` preserved). This PR's core mechanism checks out.
- **Critical hazard to pin: the `this`-typed getter literal × a `X | Store<X>` parameter union crashes tsc** (`RangeError` in the type-node printer), not just errors. The ordered overloads avoid it for *our* signatures, but please add a type-test locking the three literal shapes (plain form, derived typed, derived untyped, each with a `get x() { return this.y }`) so a future signature edit can't reintroduce a raw union. Worth filing the minimal repro upstream regardless.
- **`snapshot()` must strip the brand** — otherwise it infers and returns a branded type for a plain deep copy. An `Unstore<T>` overload sketch is in the assessment.
- **Two more:** audit the `keyof`-driven utilities (`merge`/`omit` `Override`, `storePath` `KeyOf`) against branded inputs, and note that platform types (`Map`/`Date`/DOM) get mapped at the type level even though the runtime serves them raw (the `SolidStore.Unwrappable` hook is the escape hatch).

Net: the assessment recommends adopting during the RC window (breaking type changes are cheapest now) with those amendments. This PR already covers the hardest part (source recovery via overloads); the remaining items are the review checklist. Thanks for making the tradeoffs concrete with a working implementation — that's exactly what made the proposal assessable.

GabbeV:
@ryansolid This PR is not specifically intended to implement anything from @brenelz proposal. I think the ergonomics of such recursive mapping might need more investigation. There is probably some overlap in what they solve but they are mostly about different things. The goal of this PR is to improve the ergonomics and type safety for the primitives them selves and doesn't concern it self with later consumption of the store in the app like brenelz issue.

I think the runtime guards especially might need your consideration as that goes beyond a simple type change. The AI also claimed it found and fixed SSR bugs along the way that allowed access to the seed value. I have little clue if these fixes are correct.

ryansolid (commented):
Thanks for the clarification. I initially conflated this with #3092 because both use a marker around `Store<T>`; that was the wrong frame, so I removed that comment and reviewed this proposal independently as a derived-store seed/initialization contract.

There are valuable pieces here, but I don't think the current combined design is safe to continue as one implementation. My recommendation is to split it before further work.

## 1. Type/API blockers

**The source marker does not mean “exactly this Store<T>.”** Mapped transformations preserve it, so it contaminates transformed values:

```ts
const [source] = createStore<State>(initial);
createStore(omit(source, "mutable")); // no overload matches

const extended = merge(source, { extra: true });
const [, setExtended] = createStore(extended);
setExtended(draft => { draft.extra = false }); // extra is lost; T collapses to State
```

The marker is useful for exact source recovery, but `omit`, `merge`, built-in mapped types, and intersections need to strip or recompute it rather than carrying a stale source identity.

**The unseeded callback type is dishonest.** It remains `(draft: T) => void | T | ...`, while the runtime intentionally makes every first-run draft operation throw. These guaranteed-invalid programs compile:

```ts
createProjection<State>(() => {}); // can never initialize
createProjection<State>(draft => ({ count: draft.count })); // first read always throws
```

A value that cannot be observed as `T` should not be passed as `T`. Runtime validation cannot repair that static contract.

**Seeded overloads permit callable roots** because `NoFn<T>` is missing from the seeded forms. The draft proxy is not callable, so accepted code fails at runtime. This also contradicts the PR description.

Requiring a complete seed rather than `Partial<T>` is otherwise directionally correct: if the callback receives `draft: T`, the seed must establish a real `T`.

## 2. Client runtime blockers

**The gated draft membrane can be bypassed through exported `$TARGET`.** Internal setter machinery needs that symbol before initialization, but user code can read it too, return the internal target, pass object validation, and expose internals. The proxy also cannot hide `typeof`, `Array.isArray`, or identity. “A stable `T` draft that rejects every observable operation” is not implementable with this shape.

**Initialization is tracked outside the scheduler.** The closure-local `initialized` flag opens immediately after `write`/`wrapCommit` returns, even when truth is only staged in `_pendingValue`/store pending backing and still transaction-held or masked. That can enable the public setter/draft before the first value is observable. Core commit state must be the authority; a second state machine will drift.

**The optimistic first-landing path conflicts with current `next`.** The PR predates #3146. Its `runAuthoritative(write)` initialization branch bypasses the now-required flight-owned transaction path (`declareFlight` / `enterFlightTransition`), risking premature reveal and breaking fold/`until()` composition. This cannot be resolved mechanically during rebase.

The gates also run for every already-seeded projection, adding closure/branch overhead and changing reflection behavior where no initialization protection is needed.

## 3. Server/hydration blockers

There are real baseline SSR bugs here worth extracting: pending server proxies expose seeds through reflection/symbol operations, and rejected projections can become readable as their seed because error handling marks them ready. Those are valid hardening targets.

The proposed membrane is incomplete, though. It omits fundamental proxy traps (`getPrototypeOf`, `setPrototypeOf`, `isExtensible`, `preventExtensions`). `Object.preventExtensions()` can succeed while pending and later make `Reflect.ownKeys()` permanently throw a proxy-invariant TypeError.

Server first-value adoption also diverges from the client: `Object.keys`/`Object.assign` leave omitted symbol keys behind and erase class prototypes/methods. Either unseeded values must be restricted to plain records, or adoption must preserve the supported object contract.

Changing client-only hydration from “serve the complete seed” to “hide it and suspend unless `seedLoadingValue`” is a new public semantic, not an SSR bug fix. It needs an explicit ruling and separate release treatment.

There is also an existing synchronous hydration replay path where processing failures can escape while its promise remains pending. Useful to fix, but separate from this API proposal.

## Recommended decomposition

1. **Source recovery + complete seeded forms (types first).** Fix marker contamination, exclude callable roots consistently, and add `merge`/`omit`/mapped-type/intersection probes. Keep complete `T` seeds; don't restore `Partial<T>` under a `(draft: T)` contract.
2. **Redesign the unseeded API without a fake draft.** The honest first phase is return-only: `() => T | Promise<T> | AsyncIterable<T>`, with no initial `void`. If post-initialization draft mutation is required, expose that as an explicit second phase/updater contract rather than pretending a gated object is `T`.
3. **Extract the confirmed SSR hardening separately.** Complete all fundamental proxy traps/invariants, preserve the supported object shape, and add reflection/symbol/class/error-stream tests.
4. **Only then design hydration/optimistic integration on current `next`.** It must use scheduler commit state and the flight-owned transaction model, not parallel initialization flags.

So: the complete-seed and source-recovery goals are worth pursuing, and parts of the SSR investigation found genuine bugs. The seedless runtime/state-machine portion should be replaced rather than patched incrementally. Given the stale base and current semantic conflicts, I would not rebase this branch wholesale; split the proposal into independently reviewable pieces first.

GabbeV:
I find my self a bit confused about your response but this is how I parse it:

* The brand currently causes issues with omit, merge. I'll see what I can do about that.
* Missing `NoFn<T>` will be added.
* You agree with my finding that `seed: Partial<T>` is unsound as it currently is in the code base. Maybe a separate PR reverting seed to `T` could land first but would in my opinion be an unfortunate end state.

My goal with this draft PR is to do a API design exploration for how to have a overload where you don't need a full seed but that is also type safe given I found the `seed: Partial<T>` solution fails that.

One finding in this exploration is that arrays mostly don't have a problem with providing seed as an empty array is almost always fine and easy to provide. Tuples being the exception.

Avoiding arrays by saying they always need a seed opens up additional design space where we can let omitted or nullish seed mean uninitialized object in a way that is known both at compile time and runtime easily.

Given these constraints I see a couple of different options that keep the API type safe:

### Option 1: Optional draft `(draft: null | T) => void | T | Promise<void | T> | AsyncIterator<void | T>`
This works well for functions projections. You get a draft when it is initialized and then the only option you have is to return a T to initialize it. However for async iterators stays alive so would need to get the initialized draft later somehow. Given it is a generator `const draft = yield { /* initial object */ };` could be an option but is also a bit exotic.

### Options 2: Throwing draft `(draft: T) => void | T | Promise<void | T> | AsyncIterator<void | T>`
This is the one I chose where trying to access the draft in any way before it you have yielded/returned a real T initializing the store throws. I expect this to be the kind of runtime error you immediately hit first time you run your program explaining what you did wrong. While not as strong as a type check it seems like in the spirit of other "you are using the api wrong" assertions where it is unlikely you don't discover the issue before it is in production or something like this. Linting could potentially raise the issue earlier if desired.

### Option 3: seedless doesn't allow mutation `() => T | Promise<T> | AsyncIterator<T>`
Unfortunate limitation but doesn't need any runtime errors or exotic yielded drafts etc. The limitation can be worked around by providing a seed with a type cast if necessary. 

### Option 4: always require a `seed: T`
Avoiding this is the whole motivation of the exploration in the first place.

### Other options?
Your comment mentions phases but doesn't really specify what is meant. Are you suggesting that you are open to larger changes here with multiple callbacks or things like that? My attempt here has been to stay close to the API as it already is.

## Where we go from here
This draft is primarily intended to work out the public API. Given your much deeper context on the internals, I would appreciate your help in the final implementation of whatever we land on here. I'll work on any API level fixes and push that up but much rather leave the internals to you given I don't feel that I know what I'm doing there.

GabbeV:
This might be controversial but my honest thoughts after having tried to solve the merge/omit issue is that Store making things readonly isn't worth the downstream issues it causes in types. This PR tried to solve it in the simple case with the brand allowing me to extract the original T however that caused issues with merge/omit and still didn't solve other issues like how a store remains readonly when stored as part of another store. I choose to make Store a transparent noop for now and removed the brand as it no longer solves anything this PR is trying to solve.

GabbeV:
I’ve split the seeded client-store hydration behavior out into #3223 and removed it from this branch so it can be reviewed independently.

GabbeV:
I’ve split the server projection rejection-state fix into #3224. That PR keeps the existing proxy target and reflection/symbol behavior and only ensures rejected pending projections continue to expose their original error.

GabbeV:
One issue I missed in the current implementation is that seedless hydration replay needs a mutation phase that the proposed API currently has no way to represent.

The first replay entry is a complete snapshot and can be adopted directly as the store's backing. Later entries are serialized patches, though, so replay needs a draft to apply them to. It effectively needs two phases:

```ts
// first value
() => T

// later replay entries
(draft: T) => void
```

The public seedless overload is return-only, while the projection runner currently uses one `seeded` distinction both to decide whether the callback receives a draft and whether `void` is valid. Hydration needs to provide a draft internally for patch replay, while still requiring the first result to be a complete `T`. Passing the private `{}` backing as a seed makes those states difficult to distinguish, so the current implementation cannot model seedless hydration replay with an honest initialization contract.

How this should be fixed depends on which direction we choose for the seedless API: it could remain return-only with a separate internal replay phase, or we could design an explicit post-initialization mutation phase. I think we should defer implementing this until that API direction is decided rather than add another provisional initialization state machine here. I'm leaving this comment so the current limitation is explicit.

ryansolid:
I agree with the latest finding. The public seedless contract should remain return-only; hydration replay mutation is an internal transport phase, not a reason to expose a second public callback phase.

I don't think we should extend this implementation further as one PR. Let's split it into:

1. Complete `T` seeds and callable-root exclusions.
2. A separate decision on removing `Readonly` from `Store<T>`.
3. Seedless projections after initialization and replay are modeled as distinct internal operations.

Seedless roots should be limited to plain records. Arrays are common, but `[]` is already a natural complete `T[]` seed and gives the runtime the array target shape synchronously; tuples need their actual shape. Supporting genuinely seedless async arrays would require a separate runtime shape token, which is less ergonomic than the empty seed.

This draft has been useful, but the hydration gap, runtime breadth, and current size/performance regressions make decomposition the right next step.

GabbeV:
### Open question: setters before seedless initialization

Seedless projections do not yet have an obvious writable-store contract. Before the first result commits, there is no valid `T` to provide to the store setter's mutation callback.

The directly comparable APIs currently behave as follows:

- Solid 2's derived `createSignal<T>(fn)` / `createOptimistic<T>(fn)` overloads return `Signal<T>`, so their functional setters expose `prev: T`, even though runtime `prev` is `undefined` before the first async result. That looks like an existing type hole rather than a contract we should copy.
- Readonly async primitives avoid the question by exposing no setter.
- Seeded stores have no ambiguity: the seed is a real `T`, so the mutation callback always receives a valid draft.

Making the seedless store setter value-only would make it safe to initialize with a complete `T`, but does not provide a general solution. It removes ergonomic mutation callbacks even after the store has initialized, since TypeScript cannot make that overload appear when the first result lands. More importantly, regular derived signals already expose both value and functional setters and have the same unresolved pre-initialization `prev` problem. A value setter can therefore be a safe initialization escape hatch, but it does not define the contract for functional setters across the derived APIs.

For a seedless writable store, the options I can see all have meaningful drawbacks:

1. **Throw before initialization.** This keeps `draft: T` accurate because the callback is never invoked with an invalid value, but introduces an invisible temporal restriction. A setter reachable outside the relevant `<Loading>` boundary may throw depending only on whether loading happened to finish before an interaction, which can easily escape testing.
2. **Expose `draft: T | undefined`.** This is honest, but every setter callback must handle `undefined` forever because TypeScript cannot change the setter type after the first resolution. It also widens the setter even though store reads remain `T` by suspending while unresolved.
3. **Queue setter callbacks.** This changes when user code executes, can capture stale event state, and can detach optimistic writes from the action or transition that originally owned them.

GabbeV:
Having thought about this issue for a while I think the queue is the least bad solution, as having to handle the undefined in every setter from a derived primitive would be a pain and would probably lead to people simply using ! in TS to ignore the issue everywhere anyway. It should be rare that you are in this situation I think so the slightly unexpected queuing is probably the better trade off.

Queuing however raises another issue and that is that setSignal currently return the updated value and that can't work with queuing.

GabbeV:
@ryansolid I've updated this PR to include both the seedless overload and the change of seed from `Partial<T>` to just `T`. I still would prefer a way to use async iterator to first provide a full snapshot and then mutate but this PR would still be a improvement on both ergonomics and safety as is.

The question around prev/draft in setter before init has been left as a separate issue because of how it affects other primitives as well https://github.com/solidjs/solid/issues/3274

ryansolid:
> Stepping back from the individual pieces — I don't think this PR, #3274, and the #3261 slice can be evaluated in isolation, so I want to name what they add up to, because that's what actually needs a ruling.
>
> **The proposal as one invariant:** *every user-facing surface of a derived store or signal holds a valid `T`; there is no window in which user code touches a value that isn't real data.* The four pieces are what that invariant forces:
>
> 1. Complete `T` seeds — the seed is what the callback drafts against and what the store exposes.
> 2. The seedless overload — most async stores have no honest seed, so remove the seed rather than fake one.
> 3. Pre-init setters queue or go value-only (#3274) — a seedless store has nothing before the first result, and the invariant forbids handing the callback a stand-in.
> 4. `Store<T> = T` — merged in #3261.
>
> **The model the engine runs on today**, stated the same way: *the seed is a draft for the derive function, not data.* It's `Partial`, unobservable during the loading window (A25), the write base during that window (A25's write-visibility corollary), and anything staged during a flight loses to the landing. Internally consistent — but a different invariant.
>
> Where they genuinely diverge is two rule questions, not four API changes:
>
> - **Q1 — Is the seed data or a draft?** Complete seeds were shipped in #3258 and reverted: async object stores that need the options slot had to fabricate a throwaway `T` nobody ever observes. The seedless overload is the answer to that revert (remove the seed instead of completing it), but it covers plain records only; arrays keep `[]`, and tuples or anything needing a runtime shape stay seeded.
> - **Q2 — Does a write staged during a flight win over the landing, or lose to it?** Today it loses, and not only pre-init: a `setV(100)` during a refetch also loses to that refetch's landing (verified). Queueing the pre-init callback and applying it after the first value lands makes the write *win* — causality order, since the write is newer than the ask. Ruling that only for the first flight would be inconsistent with every other mid-flight write; ruling it generally is a real engine semantic change, plus the setter-return contract change you already noted.
>
> Everything else in the PR follows mechanically from those two answers. If both stay as ruled, the seedless overload is still a reasonable additive API (record-only, no draft, no seed) and the rest has nowhere to go. If either moves, that's a re-ruling and should be specced before code.
>
> Folding #3274 into this thread as the child question — it's Q2's first-flight face. Ruling to follow here.
>
> — *Claude via Cursor*

GabbeV:
I agree with Claude. There isn't much more to do before a ruling. https://github.com/solidjs/solid/issues/3274 is partly related and partly unrelated as it affect signals as well as stores, but it also touches this api surface and also needs a ruling to move forward.

I'm a bit surprised that causality wasn't already the rule for a refresh and then a set etc. I think that pretty clearly would be the expected behavior by most users.

Some consideration to this landing during RC might also have to be made given it is breaking but I would consider at least some of this to be addressing bugs given the types currently lie.

PR #3223, opened 2026-09-02
fix(solid): keep client stores pending during hydration

GabbeV:
## Summary

Client-sourced stores currently expose their seed during hydration unless `seedLoadingValue` is enabled. This makes a value that was not declared as commit zero observable before the client derive has run.

Keep these stores unasked through hydration so that:

- without `seedLoadingValue`, reads suspend until the first client result
- with `seedLoadingValue: true`, the seed remains visible as commit zero

This is extracted from #3194 because the behavior applies independently to the existing seeded API and does not depend on the proposed seedless store API.

From `packages/solid`:

- `pnpm exec vitest run test/client-hydration.spec.ts` (128 tests)
- `pnpm types`
- `pnpm test-types`

### DIFF
--- a/.changeset/client-store-hydration-pending.md
+++ b/.changeset/client-store-hydration-pending.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Keep client-sourced stores pending during hydration until their first client result unless `seedLoadingValue` exposes the seed as commit zero.
--- a/packages/solid/src/client/hydration.ts
+++ b/packages/solid/src/client/hydration.ts
@@ -1129,14 +1129,10 @@ function hydrateStoreLikeFn(
     return withHydrationGate(hydrated =>
       coreFn(
         (draft: any) => {
-          // Windowed (seedLoadingValue): UNASKED — a sync no-op derive would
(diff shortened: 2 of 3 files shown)


ryansolid (changes requested):
The runtime change looks directionally correct and remains independent from #3194, but the committed regression only covers a synchronous `createProjection` property read. Please add an asynchronous hydration regression through the public store path (`createStore(fn)` is sufficient) that demonstrates `<Loading>` remains pending through hydration/first flight and reveals only after the client result lands. This is the observable behavior the change is protecting.

No API concern. The measured client output is slightly smaller, so there is no size blocker.

ryansolid:
The underlying fix looks right and matches the existing memo/server-store semantics. I'm going to add the broader asynchronous `createStore(fn)` + `<Loading>` hydration regression directly, then we'll move this toward merge. No action needed from you on the earlier coverage request.

ryansolid (approved):
Added the missing async createStore/Loading hydration regression. The focused hydration suite, declarations, type tests, treeshake diagnostics, and all size scenarios pass. Hydrated output is 13–43 gzip bytes smaller.

PR #3224, opened 2026-09-02
fix(solid): preserve server projection errors

GabbeV:
## Summary

- keep pending server projection proxies errored when their async source rejects
- rethrow the original error from later property reads instead of exposing the seed
- cover the behavior with a seeded Promise projection regression test

This is extracted from #3194 so the baseline bug can be reviewed independently. It intentionally keeps the existing proxy target and reflection/symbol behavior unchanged; it does not include the broader pending-proxy membrane from that proposal.

## Test plan

- `pnpm exec vitest run test/server/ssr-async.spec.ts` (158 tests)
- `pnpm types`
- `pnpm test-types`

### DIFF
--- a/.changeset/server-projection-errors.md
+++ b/.changeset/server-projection-errors.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Keep rejected server projections errored instead of exposing their seed values.
--- a/.changeset/server-projection-errors.md
+++ b/.changeset/server-projection-errors.md
@@ -2,4 +2,5 @@
 "solid-js": patch
 ---
 
-Keep rejected server projections errored instead of exposing their seed values.
+Keep ordinary property reads on rejected server projections throwing the original error instead
+of exposing seed values, while preserving existing reflection and symbol behavior.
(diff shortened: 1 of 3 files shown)


ryansolid (changes requested):
The extracted fix is independent from #3194 and the implementation appears sound, but it changes four settlement paths while only Promise rejection is covered. Please add focused coverage for at least (1) an iterable rejecting before its first yield and (2) `seedLoadingValue` rejecting, including repeated/different ordinary property reads rethrowing the original error.

Please also narrow the changeset wording: ordinary string-key reads stop exposing the seed, while reflection/symbol behavior intentionally remains unchanged.

No API concern. The measured impact is server-only at about +38 bytes gzip; please keep the broader pending-proxy membrane out of this PR.

ryansolid:
This is a worthwhile memo/projection parity fix and the narrow implementation looks sound. I'm going to add the iterable and `seedLoadingValue` rejection regressions and tighten the changeset wording directly, then we'll move this toward merge. No action needed from you on the earlier coverage request.

ryansolid (approved):
Added the missing iterable and seedLoadingValue rejection regressions and narrowed the changeset wording. The SSR async suite, workspace build, declarations/type tests, formatting, and all size scenarios pass. The implementation remains narrow and server-only.

PR #3225, opened 2026-09-02
fix(solid): gate pending projection reads

GabbeV:
## Summary

- suspend user-symbol reads, membership checks, key enumeration, and own-property descriptor reads while a server projection is pending
- use the same `NotReadyError`/client-hole path as ordinary property reads
- continue delegating directly to the real state after resolution, without a dummy target or mutation traps

This is a baseline inconsistency found while working on #3194, but the fix is independent of that API proposal. Previously `store.name` suspended while operations such as `store[symbol]`, `"name" in store`, `Object.keys(store)`, and `Object.hasOwn(store, "name")` could inspect the pending backing directly.

Descriptor reads are included because A25 requires the seed to remain hidden from every external consumer path. The trap returns the real target descriptor unchanged after readiness, preserving the target's proxy invariants.

This is expected to compose with the rejection-state fix in #3224 once that lands.

## Test plan

- `pnpm exec vitest run test/server/ssr-async.spec.ts` (157 tests)
- `pnpm types`
- `pnpm test-types`

### DIFF
--- a/.changeset/pending-server-store-reads.md
+++ b/.changeset/pending-server-store-reads.md
@@ -1,0 +1,5 @@
+---
+"solid-js": patch
+---
+
+Suspend property, symbol, membership, enumeration, and descriptor reads on pending server projections.
--- a/packages/solid/src/server/signals.ts
+++ b/packages/solid/src/server/signals.ts
@@ -1845,9 +1845,9 @@ export function createOptimisticStore<T extends object>(
 }
 
 /**
- * Wraps a store in a Proxy that throws NotReadyError on property reads
- * while the async data is pending. Once markReady() is called, reads
- * pass through to the underlying state.
+ * Wraps a store in a Proxy that throws NotReadyError on property and
+ * structural reads while the async data is pending. Once markReady() is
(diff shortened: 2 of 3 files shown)


ryansolid:
This is directly required by A25: the seed must be hidden from every external consumer path, including symbols, membership, enumeration, and descriptor/`Object.hasOwn` reads. So the open question resolves toward adding a gated `getOwnPropertyDescriptor` trap too. #3224 also changes this proxy's state machine to preserve rejection errors; we'll merge that narrow fix first, then rebase this PR and add the descriptor/error-path coverage ourselves so the two changes compose cleanly.

ryansolid (approved):
Rebased cleanly over #3224 and added rejection coverage across ordinary, symbol, membership, enumeration, descriptor, and Object.hasOwn reads. The composed status gate now preserves the original rejection across the full pending membrane. Full solid suite (580 tests), focused SSR suite (160 tests), declarations/type tests, package build, and all size scenarios pass. Server-only; no public API or client bundle impact.

PR #3227, opened 2026-09-02
perf(signals): mapArray small-move fast path for keyed reorders

ryansolid:
## Summary

- Adds a bounded small-move pre-pass (`trySmallMove`) to `mapArray`'s keyed update path: rotations, swaps, and small displacements (≤16 displaced rows) commit as targeted moves instead of full-window rebuilds. Ineligible shapes fall through to the untouched algorithm after a few probes.
- Scoped to plain identity-keyed mode only — custom key functions and index-tracking modes never enter the fast path.
- Extracted into its own function so the main `updateKeyedMap` loop stays within JIT inlining budgets (validated: no reverse/shuffle regression).

## Validation

- New `mapArray-smallmove.test.ts` suite pins semantics at 1000-row scale: rotate, displace, swap, remove, replacement, duplicates, mixed shapes, and custom-key exclusion.
- js-framework-benchmark: swap1k −6.5%, no regressions across the main suite (create, replace, append, remove, select, clear).
- Reorder matrix (rotate/displace/reverse/shuffle): delta-cost wins on small moves; full-path shapes unchanged.

## Notes

- Changeset included (`maparray-small-move-fast-path.md`, patch — prerelease mode).
- Renderer-agnostic: benefits all keyed `For` consumers (DOM and universal); does not touch `reconcileArrays`.


ryansolid:
Parking this (not shipping in RC.6): it optimizes the reactive half of a double-reconcile (mapArray diffs items, reconcileArrays re-derives the same moves against the DOM), and we've decided to pursue deleting the double-reconcile instead — a unified keyed `For` that owns row structure and DOM placement in one persistent structure (Octane-shaped, pull-based, no message seam). Keeping this branch as the validated fallback-tier reference and the performance floor the unified approach has to beat: swap1k −6.5%, delta-cost small reorders, main suite clean.

PR #3258, opened 2026-09-03
fix(signals): require complete store seeds

GabbeV:
## Summary

Split from #3194 to isolate the seed type-safety change.

Derived store APIs currently accept a `Partial<T>` seed while exposing both the returned store and the projection draft as a complete `T`. This allows required properties to be absent at runtime even though reads and mutations are typed as if they are present.

This changes the seed to a complete `T` for `createStore(fn, seed)`, `createProjection`, and `createOptimisticStore(fn, seed)` across the signals, client, and server declarations. It also applies the existing `NoFn<T>` exclusion consistently, since a callable root cannot be distinguished from the function form by the runtime dispatch.

Existing tests that intentionally initialize from `{}` and establish the complete value through mutation now use `{} as any`. This preserves that pattern as an explicit unsafe escape hatch rather than claiming it is type-safe.

No runtime behavior is intended to change.

- Ran `pnpm types` in `packages/signals`.
- Ran `pnpm exec tsc -p tsconfig.build.json --noEmit` in `packages/solid`.
- Compiled `store.type-tests.ts` independently with strict TypeScript settings.
- Ran the focused projection, store type, and server async tests: 184 tests passed.

### DIFF
--- a/.changeset/require-complete-store-seeds.md
+++ b/.changeset/require-complete-store-seeds.md
@@ -1,0 +1,6 @@
+---
+"@solidjs/signals": patch
+"solid-js": patch
+---
+
+Require complete values for projection store seeds. `createProjection`, derived `createStore`, and derived `createOptimisticStore` no longer accept `Partial<T>`, preventing a store typed as `T` from being created without all required properties.
--- a/packages/signals/src/store/index.ts
+++ b/packages/signals/src/store/index.ts
@@ -36,7 +36,7 @@ export function createStore<T extends object = {}>(
 ): [get: Store<T>, set: StoreSetter<T>];
 export function createStore<T extends object = {}>(
   fn: (store: T) => void | T | Promise<void | T> | AsyncIterable<void | T>,
-  store: Partial<T> | Store<NoFn<T>>,
+  store: NoFn<T> | Store<NoFn<T>>,
   options?: ProjectionOptions
 ): [get: Refreshable<Store<T>>, set: StoreSetter<T>];
 export function createStore(first: any, second?: any, third?: any): any {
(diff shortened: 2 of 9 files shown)


ryansolid:
Heads up: this has been reverted on `next` in de1c8b50, and we owe you an honest accounting of how it got merged in the first place.

**Process first:** this PR was merged by Ryan's agent without his review — a process error on the agent's side, not a maintainer approval. When Ryan reviewed the merged state, he ruled against the design. We're flagging that plainly so the reversal reads as what it is: the design question was never actually decided until now, and contributors aiming at the project's rulings shouldn't read this as instability.

**The ruling on the substance:** the completeness requirement costs more than it fixes. The case that decides it: any async object store that reconciles on a non-`id` key needs the options argument, and therefore needs the seed slot filled —

```ts
createStore(async () => fetchUser(), /* complete User required here */, { key: "email" });
```

and in exactly that case the seed is pure ceremony: reads of an async derived store pend with `NotReadyError` until the first resolution, so the fabricated complete object is never observable by any consumer (we verified this empirically while reviewing). A fetch-shaped callback returning a fresh `T` never reads its draft either. The type-honesty gain is real, but only for **sync draft-reading callbacks** — and taxing every async store to fix that corner is the wrong trade.

**The concern itself is legitimate**, and we think the right home for it is your #3194 direction: a callback that takes **no draft parameter** needs no seed at all, which fixes the awkward case *and* keeps the honesty for draft-taking callbacks. That conversation is still live.

#3260 is untouched by the revert — the overload alignment (slot order, `shallow` in options, `Refreshable` derived returns) stands. Thanks for the careful work here and sorry for the whiplash; the fault is on our side of the merge button.

PR #3270, opened 2026-09-04
perf(signals): store create-floor diet — slot-signal literals + first-read dedupe

ryansolid:
## What

Two isolated, battery-vetted optimizations to the store's node-creation floor. No semantic changes, no new API, no compiler involvement — pure allocation/shape diets on the hottest store paths.

### 1. Slot-signal creation diet (O1)

Store mounts materialize one signal per touched leaf (~13 × rows on dbmon), so per-node allocation is mount cost. The generic `signal()` path charged every store node:

- an options object,
- an `equals` closure (capturing the target),
- an `unobserved` closure (capturing target + key),
- a `NodeExtension` to hold that closure,
- three post-construction expandos (`acc`/`px`/`pxv` → hidden-class transitions).

`slotSignal` bakes everything into **one pre-shaped literal**: `_host`/`_key` backrefs replace the closure captures (`_equals` is one shared function using method-call `this`; the unobserved sweep dispatches `CONFIG_SLOT_NODE` nodes to one shared hook installed by the store module), and the store's read-path cache fields are pre-shaped. The core sweep sites read the hook as a live binding — no wrapper frame.

### 2. First-read diet (O2)

The get trap's first tracked read of a key did redundant work:

- **Probe once:** the trap's accessor-ness probe verdict now threads through to `getNode`, so node creation skips the second descriptor scan (guarded: only when the probed object is the one `getNode` would scan, `pb ?? v`).
- **Look up once:** the trap's duplicate node-map lookup is hoisted — one lookup serves the hot-path check and the accessor branch.
- **Cache the first wrap:** the first tracked read routes through `serveDataKey`'s wrap-cache branch, so read #2 of an object key skips `wrapNext` entirely.

## Measured

From the isolation battery (each opt solo on clean `next`, 5 canonical fixtures, correctness gates green):

- `getNode` self-time **−23%** on warm dbmon mounts
- get-trap self-time **−19%**
- dbmon mount min **−4%**
- jfb store-fixture create/runlots improved; combined with prior work, 1k keyed creation now beats the octane reference (1.7 vs 1.9 ms)
- No regressions on the signal-only fixtures (these paths are store-only and shake out of storeless bundles)

## Size

Budgets ratcheted per convention (measured, rounded up to next 0.01 kB), notes in `.size-limit.js`:

- core floor: 7.98 → **8.00 KB** (+15 B: the config-flag dispatch on the two core sweep sites; `slotSignal` itself tree-shakes out of storeless bundles)
- `+ createStore`: 14.05 → **14.16 KB** (~105 B retained code that deletes four allocations + three hidden-class transitions per store leaf)
- hydrating + store families: 26.15 → **26.25 KB** (same bytes, full retention)

All other scenarios fit within existing limits (isPending/latest and CSR app came in *under*).

## Tests

- signals: 129 files, 1461 passed
- web: 69 files, 672 passed
- solid: 21 files, 585 passed
- treeshake budgets + size-limit: green

Changesets included for both commits (`@solidjs/signals`, patch — prerelease).

### DIFF
--- a/.changeset/slot-signal-creation-diet.md
+++ b/.changeset/slot-signal-creation-diet.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Store leaf nodes ride `slotSignal` — one pre-shaped literal with `_host`/`_key` backrefs replacing the per-node options object, equals closure, unobserved closure, and the NodeExtension that held it; the unobserved sweep dispatches CONFIG_SLOT_NODE nodes to one shared hook. getNode self-time −23% on dbmon warm mounts.
--- a/.changeset/store-first-read-diet.md
+++ b/.changeset/store-first-read-diet.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Store first-read diet: the get trap's accessor probe verdict threads through to node creation (one descriptor scan per first read, not two), the trap's duplicate node-map lookup is hoisted, and the first tracked read populates the wrap cache so the second read skips wrapNext. dbmon mount min −4%, get-trap self-time −19%.
--- a/packages/signals/src/core/constants.ts
+++ b/packages/signals/src/core/constants.ts
@@ -115,6 +115,14 @@ export const CONFIG_FRESH_READ = 1 << 16;
  * (A17). Cleared at commit (the commit IS the reveal); subscribers masked
  * during the hold are woken by finalizePureQueue's post-revert pass. */
 export const CONFIG_HELD_TRUTH = 1 << 17;
+/** SLOT node (store leaf): created through `slotSignal` with `_host`/`_key`
+ * backrefs baked into the literal. The unobserved sweep dispatches these to
+ * the ONE shared hook (`setSlotUnobserved`) instead of a per-node closure
+ * held in a per-node extension — store mounts materialize one signal per
+ * touched leaf, so per-node allocations (options object, equals closure,
+ * unobserved closure, NodeExtension) were the measured create-floor bytes
+ * (warm dbmon profile: store node machinery ~36% + GC ~29%). */
+export const CONFIG_SLOT_NODE = 1 << 18;
 
 export const STATUS_NONE = 0;
 export const STATUS_PENDING = 1 << 0;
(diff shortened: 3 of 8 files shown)


ryansolid:
**CodSpeed regression triage — `omit-mixed(100, 25)`: false positive.**

Three independent lines of evidence:

1. **No shared code path.** The bench feeds `omit()` a plain object (`Object.defineProperties` output — not a store, `$PROXY in props` is false), so it takes the descriptor-copy branch in `store/utils.ts`. That branch touches none of this PR's changes: no `getNode`, no proxy trap, no `slotSignal`. The other 135 benchmarks — including every store bench that *does* exercise the changed paths — are untouched.
2. **CodSpeed flagged the comparison itself**: "different runtime environments detected" — BASE and HEAD were measured on different runner environments.
3. **Local A/B is flat**: `next` 128.6k hz vs this branch 126.6k hz (~1.5%, within the reported rme), identical 7.0 µs minimums.

Acknowledging on the dashboard rather than chasing bytes in a path this PR doesn't modify.

ryansolid:
**CodSpeed run 2 triage (post-rebase onto `ac5159ac`): still artifact, and run 2 is itself the proof.**

The code diff was byte-identical between the two runs (the rebase picked up only the #3261/#3258 type-level changes). Yet the flagged set completely changed:

- Run 1 flagged `omit-mixed(100, 25)` at −15.6% → **cleared itself in run 2** with no code change.
- Run 2 flags `merge-merge-mixed(both 5, 1, 3, 2)` −34% and `merge-merge-static(deep 0, 100, 3, 2)` −14.6% — both of which were *untouched* in run 1.

A real regression does not migrate between benchmarks across runs of identical code. Same analysis as before also applies to the newly flagged pair: both merge `Object.defineProperties` plain objects (no `$PROXY`), taking the descriptor-copy path in `store/utils.ts` — no store machinery from this PR executes at all.

Local A/B (`next` vs branch):

| bench | `next` | branch |
| --- | --- | --- |
| merge-merge-static(deep 0, 100, 3, 2) | 151.0k hz (min 5.8 µs) | 149.5k hz (min 5.8 µs) |
| merge-merge-mixed(both 5, 1, 3, 2) | 1.385M hz (min 0.5 µs) | 1.389M hz (min 0.5 µs) |

The `merge-mixed` iteration is ~0.5 µs of wall time — at that size, simulation-mode instruction counts are dominated by code-layout/alignment shifts, which any diff in the module graph perturbs. The 134 store benches that actually exercise this PR's changed paths (node creation, trap reads, reconcile) are stable across both runs.

Recommending dashboard acknowledgement; the utilities micro-benches may want a minimum-work floor so sub-µs iterations stop tripping the gate.

PR #3275, opened 2026-09-04
perf(signals): witness coarse-read machinery — one node per record subtree

ryansolid:
## What

The coarse-read stack from the create-floor isolation battery (O5–O7), landed as **internal machinery with zero public API**. Three pieces, one semantic unit:

### 1. `witnessNext` — the coarse read (O5)

Subscribe to ONE record's deep-witness node (`dk`) and serve the raw backing. No child walk, no key-set subscription, no snapshot, no leaf nodes, no wrap caches. This is the coarse-row read model: a row effect witnesses its record and reads values raw — one node + one link per record instead of one per touched leaf.

**Not exported** from `solid-js` or `@solidjs/signals`. The consumer is the compiler-emitted coarse-row track (the unified-For scan emission); user code never calls it. Tests import it through the internal module path.

### 2. Deep-witness bubbling (O6)

`bumpDeep` now notifies witnesses on the record AND its ancestors, so one witness covers a record subtree. Gated by a live-witness counter: **apps with no `deep()`/witness subscribers pay one number check per change, never a chain walk.** Ancestor bumps are authoritative (never lane cargo) — bisected during the battery: bubbling into an optimistic root's armed `dk` mid-action created an unintended override that wedged the family's generator.

### 3. Under-witness identity (O7) — the R18 amendment

Keyed reconcile's adoption walk threads an under-witness flag: records beneath a live deep witness keep key-matched proxy identity without materializing per-record nodes — the witness IS the subscription for its subtree. The reviewer-facing semantic surface:

- **Never-wrapped records are still skipped always**, even under a witness — cost stays O(wrapped), never O(store).
- **Unwitnessed captures still detach** (R18's "subscribing buys liveness" is unchanged where no witness exists) — narrow-subpath pruning in large store diffs is byte-for-byte today's behavior.
- Since `deep()` also creates `dk` nodes, `deep()` subscribers get bubbling + subtree identity too. This is a strengthening, not a break.

All three behaviors are pinned by the new `reconcile-pruning-contract.test.ts` (uses the `__adoptVisits` probe to assert exact walk-entry counts across unwitnessed / witnessed / witness-released scenarios).

## Measured

Isolation battery (solo on clean `next`, 5 canonical fixtures, correctness gates green). Hand-compiled dbmon coarse rows vs fine-grained:

- mount **0.70x**, unmount **0.61x**, tick **0.90x**
- deep dbmon: one witness per row replaces one per nested record
- Signal-only fixtures untouched (all of it tree-shakes out of storeless graphs; the bubble gate is one counter check)

## Size

- `+ createStore`: 14.16 → **14.30 KB** (+140 B: witnessNext + bubbling + the reconcile flag)
- hydrating + store families: 26.25 → **26.31 KB** (+51 B)
- Core floor, signal-only, CSR, frames: unchanged/under

## Tests

signals 1472 / web 677 / solid 585, all green. Changesets included (patch, prerelease).

### DIFF
--- a/.changeset/witness-coarse-read-prototype.md
+++ b/.changeset/witness-coarse-read-prototype.md
@@ -2,4 +2,4 @@
 "@solidjs/signals": patch
 ---
 
-Prototype `witness()` coarse read: subscribe one record's deep-witness node and serve the raw backing — the coarse-row read model (zero leaf nodes/links/wrap caches per row). Hand-compiled dbmon rows: mount 0.70x, unmount 0.61x, tick 0.90x vs fine-grained.
+Internal coarse-read machinery (no public API): `witnessNext` subscribes one record's deep-witness node and serves the raw backing — the coarse-row read model (zero leaf nodes/links/wrap caches per row), consumed by the compiler-emitted row track, not user code. Deep-witness bumps now bubble to witnessed ancestors (gated by a live-witness counter: unwitnessed apps pay one number check per change), so one witness covers a record subtree. Hand-compiled dbmon rows: mount 0.70x, unmount 0.61x, tick 0.90x vs fine-grained.
--- a/.changeset/witness-coarse-read-prototype.md
+++ b/.changeset/witness-coarse-read-prototype.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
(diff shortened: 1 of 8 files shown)


ryansolid:
Marked draft deliberately: this machinery is inert without its consumer (witnessNext has zero callers; bubbling/under-witness identity only activate through deep()'s dk nodes). Holding the merge until the compiler-emitted coarse-row track lands — the PR exists now so the R18-amendment semantics review can happen decoupled from the unified-For review. The one standalone behavior change if merged early: records captured under a deep() subscription keep key-matched identity through keyed reconcile instead of detaching.

PR #3277, opened 2026-09-04
fix(signals): suspend uninitialized async across lanes

### LINKED ISSUE #3276: 2.0.0-rc.6 | an ternary condition  using `latest()` in the condition, will render an unresolved dependency in the truthy branch.
### Describe the bug

```
/***
 *  click the button withing the initial 5s async signal resolution
 * 
 *  case A 
 *   will show undefined, and the result when the signal resolved. 
 *  
 *  case B 
 *   hold the render until the async is resolved then render the result 
 * 
 * The difference is in the ternary condition. in A we use latest(entangle) in B we use entangle() 
 * 
 * that difference bypass the tree UI blocking for an async dependency. 
 */
```

https://s.olid.uk/id/KfbvBP17QkKuvZdgx8d6CA

a ternary condition with a latest condition, stacks the new branch into the optimistic lane of the condition. 

this might be by design. but seem like an edge case.

### Expected behavior

probably block the ui until the async signal is ready.

### Screenshots or Videos

_No response_

### Platform

.

### Additional context

```tsx
import { render } from '@solidjs/web';
import { createSignal, latest } from 'solid-js';

  const [entangle, setEntangle] = createSignal(false);
  const [asyncSignal, setAsyncSignal] = createSignal(async () => {

    await new Promise(r => setTimeout(r,5000))
    return Math.random()

  })



  return (

    <div class="p-2">
      <h3>click the button in less than 5s.</h3>
      <button class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded" onClick={() => setEntangle(v => !v)}>
        Entangle {latest(entangle)}
      </button>

      <hr />
      <h1>A</h1>{latest(entangle) ? " " + asyncSignal() : "BAD"}
      <hr />
      <h1>B</h1>{entangle() ? " " + asyncSignal() : "OK"}

    </div>
  );
}


if (typeof document !== 'undefined') {
  render(() => <App />, document.getElementById('root')!);
}

```

xia-chao:
Fixes #3276.

An uninitialized async source has no committed stale value to show. Suspend it even when the reader is running under a different optimistic lane, so a `latest()`-conditioned branch keeps its existing DOM until the source resolves.

Adds a DOM regression test covering `latest(entangle) ? asyncValue() : ...`.

Tests:
- `pnpm exec vitest run test/latest-conditional-async-issue-3276.spec.tsx test/latest-async.spec.tsx test/latest-banner-landing-gap.spec.tsx test/latest-ispending-issue-3041.spec.tsx test/ispending-gated-landing.spec.tsx test/loading.spec.tsx test/dev-warning.spec.tsx`
- `JSX_COMPILER=babel pnpm exec vitest run test/latest-conditional-async-issue-3276.spec.tsx`

### DIFF
--- a/.changeset/fix-latest-conditional-async-gate.md
+++ b/.changeset/fix-latest-conditional-async-gate.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Suspend uninitialized async values across optimistic lanes so `latest()`-conditioned branches wait for their first value instead of rendering `undefined`.
--- a/packages/signals/src/core/core.ts
+++ b/packages/signals/src/core/core.ts
@@ -1204,9 +1204,16 @@ export function read<T>(el: Signal<T> | Computed<T>): T {
         console.warn(message);
       }
       // Per-lane suspension lives with the engine (a non-null lane implies it
-      // is installed): under a lane, only same-lane pending async without an
-      // active override throws.
-      if (currentOptimisticLane === null || GlobalQueue._laneSuspends!(owner)) {
+      // is installed): under a lane, same-lane pending async without an active
+      // override throws; uninitialized async sources always suspend.
+      // A lane mismatch can preserve an already committed stale value, but an
+      // uninitialized async source has no such value. Never let that path
+      // surface `undefined` while a conditional branch is being updated.
+      if (
+        currentOptimisticLane === null ||
+        owner._statusFlags & STATUS_UNINITIALIZED ||
+        GlobalQueue._laneSuspends!(owner)
+      ) {
         if (!tracking && el !== c) link(el, c as Computed<any>);
         throw owner._x?._error;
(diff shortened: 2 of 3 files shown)


ryansolid:
Follow-up note: this merged while I was preparing a size-relocated variant, so the relocation landed as fa568d39 on top — same semantics, but the uninitialized check now lives in `laneSuspends` (optimistic module) instead of core `read()`'s throw path. The inline placement taxed every bundle including storeless floors (27-66 B across five size scenarios, which is why the Size check on this PR was red); `laneSuspends` is only reachable under a lane, which implies the engine is installed, so only bundles that retain the optimistic module pay (~8 B on isPending/latest). Your test is on `next` unchanged and green. Thanks again — sharp diagnosis and exactly the right contract.

Issue #3271, opened 2026-09-04
Bug in solid 2.rc6 store


Azravos:
### Describe the bug

A store created from a function (writable projection / createProjection), a write to a node is SILENTLY DISCARDED if a descendant of that node was written earlier in the same flush. A plain createStore(value) is correct.

The LLM proposed fix did seem to solve the issue when applied to the @solidjs/signals/../dev.js file. I've provided screenshot of this proposed fix in case it might help you to see/find the problem too.

### Your Example Website or App

no

### Steps to Reproduce the Bug or Issue

(AI generated reproduction):
```
import { createStore, createRoot, flush, snapshot, untrack, deep } from "@solidjs/signals";

const seed = () => [{ id: 1, selected: true, children: [{ id: 11, selected: true }] }];

function scenario(makeStore, order) {
  return createRoot(() => {
    const [store, setStore] = makeStore();
    untrack(() => deep(store));                      // materialize the projection
    setStore(prev => {
      const row = prev[0];
      if (order === "descendant-first") {
        row.children[0].selected = false;            // descendant
        row.selected = false;                        // ancestor  <-- discarded
      } else {
        row.selected = false;                        // ancestor
        row.children[0].selected = false;            // descendant
      }
    });
    flush();
    return JSON.stringify(snapshot(untrack(() => deep(store))));
  });
}

const plain      = () => createStore(seed());
const projection = () => createStore(() => seed(), []);
const EXPECTED   = `[{"id":1,"selected":false,"children":[{"id":11,"selected":false}]}]`;

for (const [name, make] of [["plain     ", plain], ["projection", projection]])
  for (const order of ["ancestor-first", "descendant-first"]) {
    const got = scenario(make, order);
    console.log(`${got === EXPECTED ? "PASS" : "FAIL"}  ${name}  ${order.padEnd(17)}  ${got}`);
  }
```

### Expected behavior

See reproduce

### Screenshots or Videos

<img width="2431" height="708" alt="Image" src="https://github.com/user-attachments/assets/ff47eaa9-f270-4fbf-af1e-c6d949fabeaa" />

_No response_

### Platform

Windows

javascript-unsafe:
Reproduced on `2.0.0-rc.6` and current `next`.

1. Create a writable projection with one row and one child.
2. Prime it with `deep(store)`.
3. In one `setStore`, write `row.children[0].selected = false` and then `row.selected = false`.
[...]

Issue #3089, opened 2026-08-28
Projection/derived-store recomputes are lane-captured by unrelated in-flight transitions (committed rail serves torn state)


ryansolid:
Found while validating the composition story for discussion #3085 (library layers a writable derived store over a user's store). Two related defects in how projection/derived-store recomputes interact with the transition machinery; minimal repros below reproduce on `next` (rc.3-era).

## 1. Lane capture by an unrelated transition

A projection (or `createStore(fn, seed)` derived store) whose recompute runs while **any** transition is in flight has its output lane-captured by that transition — even when the projection shares no data with it, lives in a different root, and was created after the transition parked.

```ts
// Park a foreign action (unrelated optimistic store, own root):
const leak = action(function* () {
  setLeak(d => { d[0].id = 10; });
  yield new Promise(() => {}); // a slow server call
});
leak();
flush();

// Fresh, unrelated derived store:
const [view] = createStore(draft => {
  draft.length = 0;
  for (let i = 0; i < users.length; i++) draft.push({ idx: i });
}, []);
createRenderEffect(() => view.length, () => {});
flush();
```

The initial recompute runs, but its fold never commits: the setter-exit `setSignal`s stamp `_pendingValue` under the live foreign transition, and `drainFolds` re-queues the target until a node commits — which only happens when the *foreign* transition settles. Until then (forever, for a hung request):

| Rail | `view.length` |
| --- | --- |
| committed base (`t.v`) | 0 — still the seed; recompute output never landed |
| length node `_pendingValue` | 1 |
| draft backing (`t.pb`) | 1 |

Readers hit different rails in different contexts, so the view is torn: ambient reads returned 1 in one scope and 0 in another in the same tick, and `Object.keys` reported an element while `length` read 0.

Control: plain `createStore` setter writes and plain signal writes under the same parked action commit normally — the capture is specific to the recompute/fold path.

Impact: one long-lived action (slow mutation, SSE-adjacent wait, hung fetch) silently freezes the committed state of every projection/derived store that recomputes during it, app-wide, with self-healing only at settle. This is the same under-sampled-reader-class pattern as the store-rewrite regression cluster (#3044/#3074/#3075/#3078/#3082) — projection recomputes were never pinned against the transition matrix.

## 2. Derive visibility over optimistic sources is internally inconsistent

For a derive reading an optimistic source mid-action:

- **Field reads** resolve lane-aware through node reads → **committed** values (the documented "derives are authoritative" doctrine).
[...]

PR #3278, opened 2026-09-04
fix(signals): retain projection parent writes after descendant writes

xia-chao:
## Summary

- fold pending store targets from parent to child
- prevent a descendant path copy from invalidating its parent's later fold
- add regression coverage for the reported order and all six three-write orderings

## Test plan

- `pnpm --dir packages/signals exec vitest run`

### DIFF
--- a/.changeset/fix-projection-fold-order.md
+++ b/.changeset/fix-projection-fold-order.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Retain writable projection parent updates when an earlier nested update folds in the same flush.
--- a/packages/signals/src/store/next/store.ts
+++ b/packages/signals/src/store/next/store.ts
@@ -688,9 +688,21 @@ function privatizeCommitted(target: StoreNextTarget): void {
   }
 }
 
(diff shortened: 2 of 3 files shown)


ryansolid:
Thanks for this — genuinely unlucky race: a fix for #3271 landed on `next` in fc7e626f about 40 seconds after you opened this PR. Both regression tests from this branch pass against current `next`, and your six-ordering permutation test was broader than what we'd written, so it's been adopted into the landed spec with credit (80906073).

On the approaches, for the record: the depth-sort makes parents fold before children within one drain, which does fix the reported ordering — but folds can *re-queue across flushes* (a transition-held parent defers while a child commits, and #3089-deferred folds do the same), so child-before-parent interleavings still occur between drains where a sort can't reach them, and sorting also globally reorders fold notifications. The landed fix instead detects at commit time that the container was privatized mid-batch (a descendant's path-copy already ran) and merges the draft's written keys onto it in place, which is order- and flush-agnostic.

Closing as superseded — appreciated the quality of both the fix and the tests, and the analysis on #3276 as well (that one's reopened; your #3277 is under review now).

Issue #3187, opened 2026-09-01
Dynamic creates ambiguous SVG elements in the HTML namespace during client rendering


nickshiro:
### Describe the bug

When `<Dynamic>` renders an SVG element whose tag name also exists in HTML, the element is created in the HTML namespace during client-side rendering.

For example, a static `<a>` inside `<svg>` is correctly created in the SVG namespace, while `<Dynamic component="a">` in the same parent becomes an HTML element.

This creates a CSR/SSR discrepancy:

- CSR creates an HTML `<a>`.
- Server-rendered markup is parsed by the browser as an SVG `<a>`.
- Hydration can therefore begin with the correct namespace, while client-only rendering or a later dynamic replacement produces a different DOM structure.

### Steps to Reproduce the Bug or Issue

1. Render an <svg> containing both a static <a> and <Dynamic component="a">.
2. Read namespaceURI from both elements.
3. Observe that the static element uses the SVG namespace while the dynamic element uses the HTML namespace.

The following regression test demonstrates the difference:

```tsx
  test("creates ambiguous SVG tags in the SVG namespace", () => {
    let svg!: SVGSVGElement;

    createRoot(() => (
      <svg ref={svg}>
        <a id="static" href="#target" />
        <Dynamic component="a" id="dynamic" href="#target" />
      </svg>
    ));

    const staticAnchor = svg.querySelector("#static")!;
    const dynamicAnchor = svg.querySelector("#dynamic")!;

    expect(staticAnchor.namespaceURI).toBe("http://www.w3.org/2000/svg");
    expect(dynamicAnchor.namespaceURI).toBe("http://www.w3.org/2000/svg");
  });
```

The final assertion currently fails with:

Expected: "http://www.w3.org/2000/svg"
Received: "http://www.w3.org/1999/xhtml"

### Expected behavior

`<Dynamic component="a">` inside an `<svg>` should create an element in the SVG namespace, matching:

```tsx
<svg>
  <a />
</svg>
```

Client rendering and server-rendered markup should produce elements with the same namespace.

### Platform

  - Solid: next at f051db6a
  - The same tag-set-based creation path is present in Solid 1.9.15
  - Test environment: Vitest with jsdom
  - OS: Linux

### Additional context

The client runtime determines the namespace only from the tag name:

```tsx
  SVGElements.has(tagName)
    ? document.createElementNS(Namespaces.svg, tagName)
    : MathMLElements.has(tagName)
[...]

Issue #3161, opened 2026-08-31
`<For>` followed by siblings desyncs hydration in 2.0.0-rc.4


TylerRick:
A `<For>` followed by ordinary siblings breaks hydration for everything after the list at
`solid-js` / `@solidjs/web` 2.0.0-rc.4. It is clean at 2.0.0-rc.3, and the SSR markup is
byte-identical between the two — only the client changes.

Minimal case — no custom components, no JSX-valued props, no `children()`, no `<Show>`, no
`keyed`, no metaframework (`<For each>` is of course itself a prop; the point is that nothing is
forwarded through a component boundary):

```tsx
import { For, createSignal } from 'solid-js';

export default function App() {
	const [count, setCount] = createSignal(0);
	return (
		<>
			<For each={[{ id: 1 }, { id: 2 }]}>{(row) => <div class="row">row {row.id}</div>}</For>
			<button id="bump" onClick={() => setCount(count() + 1)}>
				bump
			</button>
			<pre id="after">count: {String(count())}</pre>
		</>
	);
}
```

`renderToStream` on the server, `hydrate` on the client. The server spends top-level ids `0` and
`1` on the two rows (which get composed ids `000` and `010`) and gives the siblings `2` and `3`:

```text
<div _hk=000 class="row">row <!--$-->1<!--/--></div>
<div _hk=010 class="row">row <!--$-->2<!--/--></div>
<button _hk=2 id="bump">bump</button>
<pre _hk=3 id="after">count: <!--$-->0<!--/--></pre>
```

At rc.4 the client asks for `0` for the `<button>` and `1` for the `<pre>`:

```
Hydration key miss for "0": no server-rendered element carries this key (template: <button id=bump>bump).
Hydration key miss for "1": no server-rendered element carries this key (template: <pre id=after>…).
Hydration key miss for "300": no server-rendered element carries this key (template: <div class=row>…).
```

The practical impact is that the siblings are built detached: the `<button>` in the document is
the server's, the one carrying the handler is not in the document, and clicking it does nothing —
`count` stays at `0`. Removing the `<For>` (writing the two rows out directly) is clean at the
same version, which is what isolates it.

**Environment** — the whole chain, since hydration depends on the compiler as well as the runtime:

```
solid-js              2.0.0-rc.4
@solidjs/web          2.0.0-rc.4
@solidjs/vite-plugin  3.0.0-next.35   (via vite-plugin-solid 3.0.0-next.27)
[...]

PR #3281, opened 2026-09-04
feat(solid,web): unified For — one slot owns rows and placement, default-on via For's module graph

ryansolid:
## What

Keyed `<For>` gets a **unified slot**: one persistent structure (intrusive row chain + incremental key→row map) owns both row bookkeeping and DOM placement, replacing the mapArray + reconcileArrays double pass for engaged lists. Pull-based — an ordinary two-phase render effect reads `each()`, diffs against its own committed chain (prefix/suffix walks, middle partition + LIS), and commits placement. No delivery seam, no channel, no second diff.

## Delivery: For's own module graph — zero API, zero compiler

- `For` stamps its accessor with `$for = { each, row, keyed, impl: unifiedForSlot }` — the slot algorithm lives in `packages/solid/src/client/for-slot.ts` and travels on the descriptor.
- Web's `insert()` engages it by handing over its `SlotOps` singleton (`insertBefore`/`remove`/`createText`/`isNode`/`clear`/`tag`). The slot is platform-free; every node touch rides the ops.
- **Tree-shaking is the eligibility mechanism**: apps without `For` shake the slot entirely. Renamed imports work. Renderers that ignore `$for` (universal, today) call the accessor and get classic mapArray — the stamp is advisory. Universal adoption = passing its own ops (follow-up; its `createRenderer` contract already matches).
- No `enableUnifiedFor`, no registration, no compiler emission, no new user API. SSR untouched (server `For` never stamps).

**Every keyed For in the test corpus now runs the slot**: web 697 / solid 585 / signals 1479 / universal 43 / element 10 / html 192, all green.

## Scope and fallbacks

Engages: identity-keyed array Fors (`keyed !== false`, no fallback prop, row arity < 2). Declines to classic (pre-engage) or late-demotes (post-engage): key functions, hydration claiming, duplicate keys, function-top-level rows, empty-rendering rows, non-array subjects. Flat mode keeps mounts at mapArray economics: first fills build parallel arrays (no Rows/chain/map); the structure materializes lazily on the first partial structural op. Keyed-fn mode was built and measured (+1,290 B for an idiom we don't believe in) and deliberately excluded — banked on `opt/unified-for-keyed-flat`.

## Measured

Acceptance battery (5 canonical fixtures, semantic gates green, clean provenance) + drift-immune interleaved A/B for every flagged cell:

- **jfb-signal**: structural ops 1.2–7x faster (displace 0.40–0.44, removefirst 0.13–0.20, rotate ~0.63, insertmid 0.60–0.75, shuffle 0.80–0.86); creates at parity (1k 0.94–1.00, 10k 0.94–1.06 across pairs); clear parity. Geomean 0.63.
- **jfb store fixtures**: replace 0.77–0.94, swap 0.75–0.87, shuffle 0.61–0.64; creates parity.
- **uibench**: total 0.73 (tree 0.68, table 0.78, worst-cases 0.82; nested-For render 0.45).
- **dbmon**: no regression — battery flags reproduced only in sequential legs; interleaved A/B dead even on mount/tick/tick_partial/remount (e.g. mount 16.10/16.10).
- **Ops indirection + module-graph wiring**: interleaved dead even (mount 13.60/13.60 and 14.10/14.20; tick 5.40/5.40 and 5.70/5.60) — the monomorphic singleton inlines to nothing.

## Size

- Signals scenarios + frames: **flat**.
- Simple-app floor (no For): **+153 B** — the engagement seam only (`$for.impl` call site + domOps singleton in insert).
- For-bearing app scenarios: **+2.1–2.2 KB** — the deliberate default-on bill (slot retained via For's import). Budgets ratcheted with notes.

## Open items (follow-ups, not blockers)

- Universal `insert` wiring (pass its ops; until then universal bundles retain the slot as dead weight through For's import — flagged, small).
- Hydration claiming (H2) — hydrating inserts decline to classic today; post-hydration mounts engage.
- No runtime kill-switch: demotion covers contract exits, but there's no app-level "force classic" toggle. Say the word if we want one before release.

Changesets included (patch, prerelease).


ryansolid:
Hydration claiming (H2) ruling: NOT a merge blocker — a For present during hydration declines pre-engage and runs classic mapArray for its lifetime (today's exact path and perf, zero mismatch risk); post-hydration Fors engage normally. The gap is benefit coverage on first-paint lists in SSR apps, not correctness. H2 lands as the immediate follow-up and silently upgrades hydrated Fors when it does.

ryansolid:
**Audit response — all findings addressed on the branch** (`f864dede` + `95ee4ea2`):

- **P0 (marker collapse):** the marker now passes through the seam untouched — `undefined` = whole parent, `null` = trailing MULTI child, Node = bounded hole. Every bulk path (`flat clear`, no-survivor replace, chain batch-clear, `removeFlatDom`/demote) is additionally gated on classic's `ownsAllChildren` ruling (our first/last nodes must be the parent's first/last children), so streamed foreign nodes survive even true whole-parent clears. Regression suite covers all four paths, the trailing-sibling shape, and foreign-node survival — with engagement asserted so the tests can't pass vacuously (the audit's own `render(() => <For/>)` shape turns out to run classic; the compiled sole-child shape is the one that engages).
- **Late-demotion state loss:** empty-rendering rows (null/boolean/"" /empty fragment) now hold position with a placeholder text node instead of demoting — the typed-input repro is pinned green. Function-top-level rows still demote (dynamic content is real future work); duplicates still demote (documented limitation, classic owns them).
- **Throw mid-pass:** row fns that throw now dispose the rows built so far (they chain to the persistent slot owner) before the error rides the boundary — both `buildFlat` and the mid-window build.
- **`__unifiedForStats`:** increments are `IS_DEV`-gated; frozen in prod bundles.
- **Changesets:** the four spike-history changesets are consolidated into one describing shipped behavior.
- **Rebased** onto `next` tip (including today's store sweep and the #3282–#3284 ratchets).

**New coverage for the reconcile surface** (this was the audit's meta-point — heavily tested area deserves heavy tests): `for.unified.reconcile-parity.spec.tsx` runs a 31-transition matrix (the classic for.spec families + jfb moves) through both the slot and a **live classic oracle** (arity-2 rows decline the stamp, same semantics through mapArray + reconcileArrays), across three row shapes × three container anchors, asserting zero demotions — ~1,700 DOM assertions — plus a differential section rendering both modes off one signal through a cumulative no-reset sequence with DOM equality asserted at every step.

Suites: web 728 / solid 585 / signals 1490, size budgets green (the P0 guards cost ~120 B in For-bearing scenarios, noted in `.size-limit.js`).

Not addressed here, per the audit's own framing: the coverage/size trade (fragment-with-expression and component rows demote today — the slot is deliberately conservative) and hydration claiming (ruled a fast-follow, not a blocker, in the thread above).

ryansolid:
**Two more landings on the branch** (`dedf642d`, `fe954591`) — both were coverage gaps that would have left most real apps on classic:

### Hydration (H2) — first-paint SSR lists now engage the slot
- **Id parity**: `For` peeks the id classic's mapArray owner would spend (via an `enableHydration()`-installed `sharedConfig.peekNextContextId`) and the slot's row parent is created with that explicit id — rows mint identical hydration keys. Proven against real server artifacts; `for-then-siblings` (the #3161 regression scenario) still passes.
- **mapArray `lazy` (@internal)**: under hydration `For` defers the classic first pass to first read so the slot claims first; the owner (id slot) is still created eagerly, so #3161's sibling-id fix is intact. Classic readers claim on first read with identical ids — ordering of user code now matches CSR.
- **Reversible claims**: the hydrating fill records the registry keys its templates consume; a demote mid-fill hands them back and classic's re-run (same ids) claims the same server nodes. The "never strand a claim" invariant is satisfied by making claims reversible rather than by predicting row shape — no probe, no double-run.
- **Fill commit is a claim pass**: zero DOM writes unless server/client mismatch (leftover rows removed, key-missed fresh rows inserted).
- **Tree-shakeable**: everything above lives in `for-slot-hydration.ts`, installed by `enableHydration()`; CSR bundles carry only null-guarded hook calls (CSR 15.68 → 15.48 KB after the split; floor dropped too once For's direct id-formatter import went behind the hook).
- Scope (v1): whole-parent lists. Anchored holes stay classic under hydration.
- 9 server→client scenarios via the real harness: basic reorder with server-node identity, text rows, mismatch both directions with exact warning counts, **demote mid-fill with zero warnings**, empty, trailing hole staying classic, nested engagement, and a For passed through a wrapper.

### Hole seam — lists passed through `props.children` engage
`<Table><For/></Table>` compiles the parent's hole to `insert(el, () => props.children)`, which previously hid the stamp → classic. The outer insert's compute now engages the slot when the resolved value is the `$for` accessor. The slot is created inside that compute, so a children change tears it down (hole-mode cleanup removes rows; prior classic content is cleaned via `cleanChildren` first, preserving the multi-placeholder invariant). A post-engage demote flips a per-hole flag and bumps a lazily-created signal so the hosting effect re-runs on its classic path — no second insert fighting for the hole. `children()` introspection and fragment children stay classic by design. Works under hydration too.

**Perf**: interleaved on frozen dists (pre-hydration vs current): mount 14.10/13.10 vs 14.00/12.60, tick 5.20/4.90 vs 5.20/4.80 (32 pairs), tick_partial identical — the seam's per-insert null check is invisible.

**Suites**: web 734 / server 750 / hydrate 174 / solid 585 / signals 1490 / universal 43. Size budgets ratcheted with notes (floor +~110 B for the seam living in insert; app scenarios +67–147 B; hydrating scenarios carry the hooks module, CSR shakes it).

Ready for the next audit round — the invariants I'd want checked hardest: reversible claims under hydration, and the hole handoff/demote under dynamic children.

ryansolid:
**CodSpeed on `8baed803` — 2 improved, 1 regressed, 133 untouched.** Context for the flag:

These three permutation benches render via a wrapper accessor (`insert(container, () => <For/>, null)`), so until the hole seam landed they ran classic and CodSpeed saw no change. They are now the first true slot-vs-classic instruction-count comparison on permutation shapes:

- `reverse: 1000 rows` **+75%**, `mount-clear-cycle: 1000 rows` **+21%**.
- `shuffle: 1000 rows (Fisher-Yates)` **−10.8%** — the adversarial case: a full random permutation has LIS ≈ 2√n (≈63 of 1000), so the slot's O(n log n) LIS pass buys almost nothing while classic's udomdiff skips that work.

**Why this is a measurement-model artifact rather than a user regression:** CodSpeed simulates against jsdom, where `insertBefore` is cheap JavaScript — the LIS's JS cost is visible and its DOM-move savings are not. In a real browser DOM moves dominate: on this exact op the slot wins (jfb-signal shuffle 0.80–0.86, uibench adversarial worst-cases 0.71–0.82, drift-cancelled A/B). Recovering the instruction count would mean skipping LIS on full-window permutations, which *adds* DOM moves and makes the browser number worse — the wrong trade. Recommending acknowledgement on the dashboard; a bounded micro-pass on the full-window bookkeeping (Map/stamp overhead, no change to move counts) is a reasonable follow-up if we want the flag gone.

ryansolid:
> **Third-pass audit (`8baed803`) — scope since last audit: H2 hydration claiming, the `{props.children}` hole seam, anchored-hole hydration.**
>
> ### P1 — nested lists strand claims on a mid-fill demote (the "never a stranded claim" invariant fails)
>
> Reproduced with a temporary harness scenario (not committed): outer keyed list whose rows each contain a nested `<For>`, where a LATER outer row is `<Show>`-rooted (heterogeneous rows — `cond ? <Special/> : <li/>` is a very common shape). Outer engages, row x (and its nested slot) claim, row y returns a function → outer demotes mid-fill.
>
> Result: text correct, but **4 hydration key-miss warnings, only 1 of 3 server `<span>` rows survive as server nodes** — the rest are recreated detached elements (the runtime's "subtree will not appear hydrated" path). The single-level `slot-hydrate-demote-mid-fill` scenario passes; the failure is nesting-specific.
>
> Two causes in `for-slot-hydration.ts`:
>
> 1. `record()` shadows `registry.delete` as an own property and its `finally` does `delete reg.delete`. A nested slot's `record()` (row build → inner `insert` → inner engage → inner fill, all synchronous inside the outer build) installs its own shadow and then **restores the prototype method**, silently ending the outer log. Every key the outer's remaining rows consume after the first nested list is unlogged → not handed back on `restore()`.
> 2. Even with (1) fixed, a nested slot that COMMITTED (`commitFill` drops its `hydLog`) has made its claims permanent — they are in no log. When the outer demotes and disposes it, classic's re-run re-engages the nested lists, which key-miss on the already-consumed ids.
>
> Fix direction: a single module-level recording stack rather than per-slot shadows — the outermost `record()` installs the shadow once, every active log on the stack receives each deletion, nested `commitFill` only clears its OWN entry, and outer `restore()` hands back everything claimed beneath it (including nested slots' committed claims). Then add the nested+demote scenario to `for-slot-scenarios.tsx` (`engaged: 4, demoted: 1, warnings: 0`, identity on `span`).
>
> ### P2 — hole seam + hydration + demote: classic re-enters with `current = []`
>
> `client.ts` hole seam: under active hydration `keep = []`, so after a mid-fill demote the hosting effect's classic path runs `reclaimRegion([], …)` (returns `[]` — `first` is undefined) and `insertExpression(parent, inner, [], marker)`. Claimed rows are already children so they reposition rather than duplicate, but on a server/client mismatch the leftover server rows are invisible to classic and survive. The direct-insert lateClassic passes `initial = undefined` for whole-parent holes so `claimInitial` re-derives the range — the hole path has no equivalent. Not covered by `slot-hydrate-through-children` (no demote). Worth a scenario (through-children + Show-rooted row + `isServer` mismatch) even if the ruling is "accept".
>
> ### Size — the floor now pays
>
> `.size-limit.js`: floor (no For) 10.73 → 11.03 KB (+0.30), CSR app 12.97 → 15.62 (+2.65), hydrating app 17.61 → 20.57 (+2.96), store app 26.43 → 29.45 (+3.02). The floor cost is the engagement seam living in `insert`'s effect (domOps singleton, `$for` guards, `holeGen`/`holeClassic`, `slotRegion`, `contains`) — every bundle carries it, For or not. The notes document it honestly, but the changeset's "apps without For tree-shake it entirely" should be qualified: the *algorithm* shakes; ~300 B of seam does not. Hydration added ~0.8 KB on top of the previously audited ~2.1 KB.
>
> ### Verified fine
>
> - `mapArray` `lazy` option: `computed(fn, LAZY_OPTIONS)` → `REACTIVE_LAZY` → recompute on first read via `prepareComputed`; the internal owner still spends the id slot at creation (#3161 fix preserved). Internal-only, documented `@internal`.
> - Id parity: `peekNextContextId` (installed by `enableHydration()`, CSR-shaken) peeks before the eager `create()` consumes; `createOwner({ id })` accepts the explicit id; keyed:true rows mint one owner per row in order on both paths. `slot-hydrate-nested` confirms parity for the happy path.
> - `demote()` runs in the effect phase, so the hole seam's `holeGen[1](…)` write is not an owned-scope write; `holeGen` is created lazily only for holes that see a `$for`.
> - Hole-mode cleanup removes rows (flat or chained) — a children change tears down cleanly; `marker` passes through untouched (P0 wipe fix intact).
> - `engage()` declines `marker === null` under hydration; `commitFill` inserts back-to-front anchored on `slot.end` (whole-parent: append; anchored: `<!--/-->`), removes only region leftovers still in `parent`.
> - `__unifiedForStats` exported from `solid-js` (client + server stub) — a **new public export**, dev-only counters. Flagging per your rule; `@internal` in the doc comment but it is on the package's public surface.
>
> ### Status
>
> CI green except CodSpeed (shuffle −10.8% is the jsdom-instruction-count artifact explained above; reverse +75%, mount-clear +21%). Branch is **CONFLICTING** with `next` on `packages/web/src/client.ts` and `scripts/size/.size-limit.js` after #3183 — needs a rebase before merge regardless.
>
> Recommendation: fix P1 (it turns a common row shape into degraded hydration with warnings), decide P2, rebase, then merge.
>
> — *Claude via Cursor*

ryansolid:
**Audit round 3 — addressed on the branch** (`e5a87f26`, rebased onto `next` at `f587361c`; the rebase over #3183 is done).

**P1 (nested claims stranded on mid-fill demote) — fixed.** Both causes were as diagnosed. Per-slot registry shadows are replaced by one module-level recording stack in `for-slot-hydration.ts`: the outermost `record()` installs the shadow once, every active log receives every deletion, an inner `commitFill` drops only its own log, and an outer `restore()` hands back everything claimed beneath it — including nested lists' already-committed claims, which classic's re-engaged nested lists mint again with the same ids. Scenario `slot-hydrate-nested-demote` (outer list, nested lists per row, Show-rooted later row): **5 engagements, 1 demote, zero warnings, every `<span>` is its server node.** (5 rather than 4: the Show-rooted row's `<li>` template runs its own hole insert before the outer sees the function and demotes — all five claim cleanly on the re-run.)

**P2 (hole seam + hydration + demote) — fixed, with one ruling.** Two problems surfaced: (a) `keep = []` — the hosting effect now keeps the claimed region as `current`; (b) more importantly, the deferred `holeGen` re-run was landing *after* `hydrate()` flipped the flag and **cloning instead of claiming**. A demote inside a hydrating fill now re-enters classic synchronously within the hydration window (`insert(parent, () => listFn(), marker, current, options)` under the hosting owner — the invoking wrapper cannot re-engage, which is also why the direct seam never loops; verified empirically). `holeGen` is `ownedWrite` (the bump may fire inside an owned scope). Scenario `slot-hydrate-through-demote-mismatch`: rows are server nodes, 1 demote. **Ruling pinned:** the leftover server row survives with the runtime's honest "1 unclaimed server-rendered node" report — that is classic's own claim-pass behavior on the identical mismatch (it never removes server leftovers), so this is classic parity, not a slot defect. The hosting effect's real-range `current` means a later children change cleans it.

**Public surface:** `__unifiedForStats` is no longer a package export — the counters ride `DEV.unifiedFor` (solid-js's existing dev diagnostics bag; `undefined` in prod). **Changeset** now says the slot *algorithm* tree-shakes while ~0.3 KB of engagement seam in `insert` is retained by every web bundle, and hydration adds ~0.8 KB to hydrating bundles only.

Verification on the rebased tip: web 760 / server 754 / hydrate 178 / solid 585 / signals 1490 / universal 43; size budgets green (+17–32 B from #3183 drift + round 3, noted in `.size-limit.js`).

Verified-fine items from the audit stand unchanged. CodSpeed shuffle ruling is in the comment above.

ryansolid:
> **Fourth-pass audit — round-3 fixes (`e5a87f26`, `f587361c`, rebased over #3183; local, not yet pushed at time of review).**
>
> ### P1 (nested claim recording) — fixed, verified
>
> The module-level recording stack is correct: the outermost `record()` installs the shadow once, every active log receives every deletion, `logs.pop()` is LIFO-safe because nesting is synchronous, an inner `commitFill` nulls only its own `hydLog` (the array object stays on the stack), and an outer `restore()` hands back nested committed claims. Re-set of an already-restored key (inner demote → inner classic re-claim → outer demote) is an idempotent `Map.set`. `slot-hydrate-nested-demote` pins it.
>
> ### P2 follow-up — the synchronous hydration demote leaves list residue on a later children swap (reproduced)
>
> `client.ts` hole seam, hydrating branch of the demote thunk: `runWithOwner(holeOwner, () => insert(parent, () => listFn(), marker, current, options))`. That nested `insert()` has its **own** `current` closure. The hosting effect's `current` stays frozen at the claimed region, and `holeClassic = true` with no `holeGen` bump means the hosting compute never re-runs to re-take ownership. Classic `insert` never removes its nodes on dispose — it relies on the enclosing effect's `current` — so any row the nested insert adds after hydration is invisible to the hosting effect.
>
> Probe (temporary harness scenario, reverted): `<ListShell>{cond() ? <For each={items()}>{heterogeneous rows}</For> : <p>none</p>}</ListShell>`, hydrate (engage → demote mid-fill → sync classic), then `setItems([...,"d"]); flush(); setCond(false); flush()` → **`noned`** — the `<p>` lands before a stranded `<li>d</li>`. The same scenario with homogeneous rows (no demote; `current = region` stale but ownership-guarded) gives `none`, and the CSR demote path (`holeGen` re-run) is covered by `for.unified.children.spec` and is fine. So this is specific to the sync re-entry.
>
> Fix: don't spawn a nested `insert()`. Re-enter with the hosting effect's *own* inner-effect shape so it writes the shared `current` — factor the existing classic block into a local `classic(value)` closure (`effect(() => (hydrationRt && (current = reclaimRegion(current, parent, marker)), withInsertionParent(parent, () => normalize(value, current, multi))), inner => { current = insertExpression(parent, inner, current, marker); host && tagHost(current, host); }, …)`) and call `runWithOwner(holeOwner, () => classic(() => listFn()))` from the hydrating demote. Same owner and lifetime as the nested insert, but `current` stays honest. Then pin: through-children + Show-rooted row, post-hydration append, flush, children swap → `none`.
>
> ### Design note — mismatch handling is now asymmetric, and the direct path is silent
>
> `slot-hydrate-mismatch-fewer` / `slot-hydrate-trailing-mismatch-fewer` expect the leftover server row REMOVED with `warnings: 0`; `slot-hydrate-through-demote-mismatch` expects it KEPT with the runtime's 1 unclaimed-node report and is pinned as "classic parity". If classic's claim pass leaves leftovers and reports them, then the slot's `commitFill` is *stricter* than classic (good: no residue) but *quieter* (it repairs a server/client mismatch without a dev report). Hydration mismatches are bugs the developer should hear about; suggest `commitFill` warn in dev when it removes leftovers or inserts key-missed rows — one message, parity with the runtime's unclaimed-node report — rather than silently normalizing. Not blocking; a ruling.
>
> ### Verified fine
>
> - `__unifiedForStats` off the package surface; rides `DEV.unifiedFor` (dev only). Note `Object.assign(_DEV!, …)` mutates `@solidjs/signals`' `DEV` singleton, so signals' `DEV` also grows `unifiedFor` — harmless, just cross-package.
> - `holeGen` is `ownedWrite` — the hydrating demote bump path no longer exists, but the CSR bump can still fire inside an owned scope; correct.
> - Stale `current = region` after a successful hydrating engage: `reconcileArrays` (isLive / `after = marker || null`), `cleanChildren` (`parentNode === parent`), and `ownsAllChildren` (first/last identity) are all ownership-guarded — a later children change reconciles correctly (probe: `none`, no residue).
> - Size ratchet: floor 11.07 (+0.04 incl. #3183 drift), CSR 15.66, hydrating 20.59 — within the previously discussed envelope.
>
> Recommendation: fix the sync re-entry `current` leak (P2) before merge; the mismatch-report question can follow.
>
> — *Claude via Cursor*

ryansolid:
**Round-3 follow-up addressed** (`91b0f946`, pushed — the earlier round-3 commits were pushed at `f587361c` shortly after the audit ran).

**P2 residue leak — fixed as suggested.** The hosting effect's classic inner-effect block is now a shared local `classic(value, prev)`; the hydrating demote calls it under the hosting owner instead of spawning a nested `insert`, so classic writes *this* insert's `current` and a later children change cleans everything classic appended. Scenario `slot-hydrate-through-demote-residue` pins the exact repro (hydrate → demote → append `d` → swap children) → `none`.

**Design note — taken.** `commitFill` now emits one dev warning when it repairs a server/client mismatch ("Hydration mismatch in `<For>`: N unclaimed server row node(s) removed, M client row node(s) inserted — DOM repaired, server and client should render the same list"). The count covers element rows only: primitive rows currently re-create their text node (fresh text swaps in for the server's — correct DOM, not a mismatch), and adopting server text nodes for primitive rows is the follow-up that makes that path zero-write. Scenario warning counts updated accordingly (mismatch-fewer 1, mismatch-more 2 = key-miss + repair, trailing-mismatch-fewer 1).

Verification: web 760 / server 755 / hydrate 179 / solid 585 / signals 1490 / universal 43; budgets green (+9/+12 B).

ryansolid:
> **Fifth-pass audit — `91b0f946` (round-4 follow-up; local, not yet pushed).**
>
> ### P2 residue — fixed, verified
>
> The hosting effect's classic inner-effect block is now a shared `classic(value, prev)` closure that writes THIS insert's `current`; the hydrating demote calls `runWithOwner(holeOwner, () => classic(listFn, undefined))`. `normalize` unwraps `listFn` through `flatten` (no `$for` check on that path, so no re-engage), `prev = undefined` matches a first fill's options, and the normal path is byte-for-byte the previous block. `slot-hydrate-through-demote-residue` pins append → flush → children swap → `none`. Re-ran the hydrate (28 files / 179), unified-For (6 / 56) and server (78 / 755) suites on the branch: green.
>
> ### Mismatch report — taken, one gap noted
>
> `commitFill` now warns once in dev when it removes or inserts ELEMENT row nodes; scenario counts updated (mismatch-fewer 1, mismatch-more 2 = runtime key-miss + repair, trailing-mismatch-fewer 1). The text-row exclusion is honest about why (primitive rows currently re-create their text node on every hydrating fill, so counting them would warn on the happy path) — but it also means a **genuine text-row mismatch is silent**, and the happy path for `{item => item}` lists is a DOM write pass, not the zero-write claim the module header promises. Fine as a documented follow-up (adopt server text nodes for primitive rows, then drop the exclusion); flagging so it doesn't get lost.
>
> ### Minor — direct-path lateClassic under anchored-hole hydration
>
> `insert`'s direct seam re-enters classic with `initial = marker !== undefined ? [] : undefined`. For whole-parent holes `undefined` lets `claimInitial` re-derive the range; for ANCHORED holes `[]` throws the bounded region away, so classic's claim pass reconciles against nothing — element rows recover through the registry, but a primitive row in a demoting heterogeneous list has no positional text node to adopt (`normalize`'s "raw primitive in an array during a claim pass" bail). Passing the slot's `region` back (`region ?? []`) would give classic the same range the slot had. Not covered by a scenario (`trailing`/`bounded` × demote); low frequency, but it is the one remaining asymmetry between the two demote paths.
>
> ### Status
>
> Remote is still at `8baed803`; the rebased branch (`e5a87f26`, `f587361c`, `91b0f946`) hasn't run CI. Push, let CI + CodSpeed run (expect the shuffle flag to persist — the jsdom instruction-count artifact already explained), then merge. No blocking findings from this pass.
>
> — *Claude via Cursor*

ryansolid:
**Remote state correction + the minor asymmetry, closed.** The remote was *not* at `8baed803` — `f587361c` was force-pushed right after round 3 ran, and `91b0f946` fast-forwarded on top (`git ls-remote` confirms `91b0f946` at the time of the r3 follow-up audit; the audit's fetch was stale). Now at **`af30bda1`**.

**Minor asymmetry — taken** (`af30bda1`): the direct seam's `lateClassic` now passes `region ?? []` for anchored holes, so a hydrating demote hands classic the bounded region the slot had — a primitive row in a demoting heterogeneous list adopts its positional text node, same as the hole seam. Whole-parent still re-derives via `claimInitial`. +2–7 B, ratcheted.

**Follow-up recorded** (not in this PR): adopt server text nodes for primitive rows in the hydrating fill, then drop the element-only exclusion from the mismatch report — closes both the silent text-row mismatch and the text-row write pass.

CI is running on `af30bda1`. Shuffle flag persists under the standing ruling; everything else green locally (web 760 / server 755 / hydrate 179).

ryansolid:
**Rebased over the #3187 revert (`d2b50e91`) — and it simplified the hole seam** (`0c262a4b`, force-pushed).

- The revert removed insert's insertion-parent tracking; the hole seam and the shared `classic()` helper drop their `withInsertionParent` wrappers. No slot behavior depended on it.
- **Checked ourselves against the revert's semantics.** Eager creation at component time + refs firing at creation is exactly the slot's row model (rows build in the compute, refs fire then — same as mapArray), so the revert brings classic back in line with the slot rather than the other way round. `Dynamic`-rooted rows still demote — `Dynamic` returns a memo, so the row's top level is a function regardless of eager/deferred *element* creation — pinned as such, with the DOM asserted synchronously after `render()`.
- **That last assertion caught a real gap:** the CSR hole demote deferred the classic re-run through the lazy `holeGen` signal, leaving the hole **empty for a microtask** (a sync `querySelector` after `render()` saw no rows; classic shows them synchronously). The hydration path already re-entered classic synchronously via the shared `classic()` effect, and nothing about that required hydration — so the deferred path is gone entirely. One path for CSR and hydration: no `holeGen`, no `ownedWrite` signal, no empty-hole microtask.
- **Budgets locked in DOWN:** floor 11.07 → 10.98, CSR 15.68 → 15.61, hydrating 20.60 → 20.54, store 29.48 → 29.40 (revert + signal removal).

web 760 / server 758 / hydrate 182 / solid 585 / signals 1490 / universal 43. CI running on `0c262a4b`.

ryansolid:
**Audit round 4 — addressed** (`c805d0df`, rebased onto `next` and pushed).

**1. P0 (first-fill demote in a hole) — fixed, verified, committed.** Root cause as diagnosed: the slot's first fill runs synchronously inside `impl()`, so a first-fill demote re-entered classic *before* the hosting effect had pointed `current` at the hand-off range — classic's first run reconciled against stale nodes and was then clobbered by the post-assign. `current = keep` now precedes `impl()` and the post-assign is gone; the decline path (same gap, the orphan placeholder) is covered by the same move. The audit's regression test is in `for.unified.children.spec.tsx`: bounded + whole-parent holes with `Dynamic` rows — replace (not reorder), children change both ways, `childNodes` count, `holeClassic` sticks. Credit for the fix and the attribution work (rebuild-to-test, the `6801317b` self-notify finding) to the audit pass.

**2. CodSpeed shuffle** — standing ruling unchanged (jsdom instruction count on the LIS worst case; browser wall-time favors the slot on the same op). Ryan's call whether to acknowledge or spend a round.

**3. Rebase** — done; only `.size-limit.js` conflicted, resolved by keeping upstream's notes and re-measuring. Two caps absorb upstream drift (+16 B each: hydrating+store 29.42, CSR 15.63 — #3287 et al., not slot bytes); floor holds at 10.98.

**4.** Noted as its own issue candidate; not touched here.

Verification on `c805d0df`: web 761 / server 767 / hydrate 182 / solid 585 / signals 1496 / universal 43; budgets green.

ryansolid:
**CodSpeed `shuffle: 1000 rows` — investigated, ruled: prioritize browsers** (`a4e8b9b2`, comment-only commit at the placement loop).

**What the −10.8% is.** CPU-profiled the exact bench shape in jsdom, slot vs a classic oracle. Both arms spend ~85% of time in jsdom's DOM internals; the slot's own JS (LIS, walk, chain) is ~1%. Both arms detach the **same** number of nodes (`_remove` time identical → same move count). The difference is insertion *shape*: the slot performs ~939 LIS-minimal mid-list `insertBefore` calls; classic's udomdiff does 369 appends (`anchor === null`) + `replaceChild` swaps. jsdom computes the anchor's child index on every mid-list insert via a sibling walk — that walk is the entire delta. Real browsers don't charge for that, which is why Chrome shows the slot **winning** shuffle (0.63 vs ~0.75–0.80).

**Attempted fix and why it was reverted.** Batching runs of moved rows through a `DocumentFragment` cut live insertions from ~940 to ~60 and brought jsdom to parity (7.68 vs 7.90 ms in this harness). It also moved each live row **twice** (into the fragment, then into the parent) — Chrome **reverse 0.58 → 0.75, shuffle 0.65 → 0.78**, confirmed in both A/B orders. Reverted: a jsdom-only win that costs users 20–25% on the same ops is the wrong trade. (Fragments remain a legitimate future win for *fresh* rows only, which does not touch this bench.)

**Ruling.** The two environments want opposite placement strategies on a full permutation and no strategy satisfies both without adding browser moves. Browsers win: the slot keeps move-minimal direct placement; the jsdom CodSpeed shuffle reads ~10% behind classic *by construction* and is acknowledged as a measurement-environment artifact, with this evidence recorded at the placement loop in `for-slot.ts` so it isn't "optimized" toward jsdom later.

Everything else unchanged since round 4: all suites green, budgets green.

ryansolid:
**Browser re-verification on the rebased tip (`57e5f531`, over `next` @ `b6a90f9f` incl. #3296/#3299).**

Different from earlier batteries: the baseline here is a **separate classic `next` build** (fixtures relinked to a clean `next` worktree), not an arm-toggle of the same bundle — so the slot's extra code is in the cold path being measured. Chrome, 5 canonical fixtures + reorder harness.

**jfb-signal (priority axis):** shuffle 0.72, rotate 0.58/0.61, displace3–8 0.41–0.47, removefirst 0.19, insertmid100 0.75, append100 0.75; creates at parity (run 0.96, runlots 1.01, clear 1.01, replace 0.94). Sub-ms cells (update/select) sit on the 0.1 ms quantum.

**jfb-shallow / jfb-deep:** structural wins hold (deep: shuffle 0.80, rotate ~0.64, displace ~0.68; shallow: swap 0.57, append100 0.60). Sequential legs flagged deep `reverse` 1.18 and sub-ms swap/remove; **interleaved ABBA** resolves reverse to parity-to-win (slot min 1.13/1.10 vs base 1.15/1.15) and shuffle/displace win in both orders — the flag was leg drift.

**dbmon-deep interleaved, 32 pairs, cold mount:** base 13.40 med / 12.20 min vs slot 13.50 / 12.40 — inside the quantum. tick slot-faster (5.50/5.00 vs 5.70/5.20), tickPartial identical (1.30/1.20), remount slot-faster (9.20/8.80 vs 9.50/9.20).

Suites: web 765 + hydrate 182 + server 774, solid 585, signals 1503, universal 43 — all green. Size: one 73 B ratchet on the store-app scenario, attributable to #3296's store-module bytes (present on `next`), not the slot.

ryansolid:
> **Response to the 2026-09-07 external audit (8 P1 / 4 P2) — fixed in `06db6791`.** Every behavioral claim was reproduced against the tip before touching code; one finding is disputed with a citation, one became moot.
>
> **P1-2 (the substantive one) — no more shape demotion.** Reproduced: a component row returning `<Show>`/a conditional ran once speculatively, then all rows re-ran through classic (4 calls for 3 rows), and a late dynamic row remounted its siblings. Rather than the compiler/narrow-eligibility direction, the slot now handles these as **dynamic rows** using classic's own cost model: the row is built once (owned, untracked) and its function is resolved by the slot's compute, *tracked* — exactly the `flatten` read `insert`'s effect performs for such rows today (`client.ts:2434`), moved into the slot. No per-row effect, no marker nodes (mount cost: one field on the row), no double invocation, no remount cliff; a flip splices only that row's range, reusing positional text nodes with a `.data` write. A NotReady thrown from a row's resolution parks the built plan and the retry reuses the rows — mapArray parity; rebuilding would re-fire per-row async and never settle. Pinned: component rows run exactly N times, sibling node identity survives a flip, late dynamic rows keep siblings, `<Dynamic>` rows reorder as the same elements, NotReady settles with one invocation per row.
>
> **P1-1** — rows now live under For's *creation* owner (`$for.owner`, where mapArray's internal owner lives); the insert cleanup disposes it explicitly. Pinned with a creator/inserter provider probe against an arity-2 classic oracle (both read `creator`).
> **P1-3** — list end anchor is the committed tail's `nextSibling` while it is still ours (classic's contiguity rule); `a b <hr> c` → `a b c <hr>`, across fill/replace/structural.
> **P1-4** — removes are parent-guarded; a migrated row node is left in its new parent (flat and chain modes).
> **P1-5** — `buildParts` disposes the row owner when the row fn throws (cleanup registered before the throw runs; matches classic under `<Errored>`).
> **P1-6** — the hydrating fill commit adopts positional server text nodes for primitive rows; `slot-hydrate-text-rows` now asserts text-node identity.
> **P1-7 — moot.** With no shape demote, *nothing can demote mid-fill* (duplicates are only detectable on a later partial op; a non-array subject demotes before any row is built). The claim `record`/`restore` stack is deleted outright rather than made per-registry.
> **P1-8 — disputed.** `map.ts:83` `const indexes = map.length > 1`: mapArray itself keys index-signal minting off `Function.length` and calls `this._map(item)` with no index at arity ≤ 1. `(item, index = 0) => …` gets a broken index in classic today; the slot mirrors classic's contract exactly. Mutating `Function.length` between creation and first map is the only slot-specific nuance.
> **P2-1** — fresh-key duplicates are caught through the existing probe map (`-1` entries), zero extra allocation; they demote like other duplicates.
> **P2-2 / P2-3** — rulings, now explicit in the changeset: empty rows hold a placeholder text node; mismatch repair removes leftovers and warns once (classic leaves them and reports at hydration end — both are wrong for a shifted list; repair+warn is the chosen behavior).
> **P2-4** — the slot's node type is opaque (`SlotNode = object`); core's public declarations carry no DOM types, and the one direct DOM read moved behind `ops.owns`.
>
> Still open, by design: late duplicate-identity keys demote (remount) — classic tolerates them; rare, tracked as a follow-up.
>
> **Cost.** +~850 B brotli on For-bearing bundles (structural: resolution, parked-plan reuse, anchors, guards; ~80 B recovered by dedupes, the rest isn't golfable), +61 B floor (four `domOps` methods). Chrome, interleaved pre/post on jfb-signal and jfb-deep reorder: parity (reverse min 0.52/0.55 → 0.53/0.55; shuffle/displace/removefirst identical). Suites: web 778, hydrate 182, server 774, solid 585, signals 1503, universal 43.
>
> — *Claude via Cursor*

ryansolid:
> Closing: the unified-For campaign is wound down in favor of targeted work on the existing list primitives.
>
> Across #3281 and #3308 the design got architecturally cleaner each round (one engine, no seam, `mapArray` as the oracle), but every audit still surfaced a fresh batch of P1s in ownership, hydration, and key semantics, and the paired benchmarks never showed a general win that justified the size or the coordination risk — the takeaway being that a parallel list system is where complexity caught us every time.
>
> What was kept: the Tier-1 DOM-lane benches and semantic gates, the oracle-harness approach for differential testing against `mapArray`, and the `UNSTABLE_LIST_IDENTITY` diagnostic. The targeted outcome of the same investigation is #3227 (small-move fast path inside `mapArray`). The branch stays in place as reference.
>
> — *Claude via Cursor*

PR #3288, opened 2026-09-05
fix(babel-plugin,compiler): assign bare variable refs inside ref arrays (#3285)

### LINKED ISSUE #3285: [2.0.0-rc6] Variable ref inside a ref array is never assigned
### Describe the bug
Doesn't work at all, elementRef never becomes assigned. But this is fully allowed by the type system and compiles
```
const MyComp = () =>
{
    let elementRef: HTMLDivElement | undefined
    
    return (
       <div ref={[
           elementRef
        ]}>
       </div>
    )
}
```
Works as intended
```
const MyComp = () =>
{
    let elementRef: HTMLDivElement | undefined
    
    return (
       <div ref={[
           el => elementRef = el
        ]}>
       </div>
    )
}
```

So I'm guessing either this is a bug or it simply shouldn't compile if it's not meant to be supported.

See description

Just put a ref variable inside the ref array like ref={[myRef]} and it will never be assigned.

### Expected behavior

Either a compile error if this isn't intended or the ref element to be successfully assigned

### Screenshots or Videos

_No response_

### Platform

- OS: [Windows]
- Browser: Chrome
- Version: [152]

### Additional context

_No response_

xia-chao:
Closes #3285

## What was broken

`<div ref={[el]} />` compiled without errors and passed type-checking, but `el` was never assigned the element:

```jsx
let el;
render(() => <div ref={[el]} />, document.body);
el; // undefined ❌
```

while the callback form worked:

```jsx
let el;
render(() => <div ref={[(node) => (el = node)]} />, document.body);
el; // <div> ✅
```

## Why

- A plain `ref={el}` works: the compiler emits a callback that assigns the element to `el` when it's not callable.
- `ref={[el]}` took a different branch — the array was passed through untouched (wrapped as `_$ref(() => [el], element)`), with nothing done to the `el` inside.
- At runtime, ref arrays are flattened and each entry is **called** with the element (`f && f(element)`). A bare variable evaluates to `undefined` at mount time, so it gets silently skipped and nothing is ever assigned.

The runtime only sees the evaluated *value* (`undefined`), not the binding, so this can't be fixed at runtime — the compiler has to turn the bare variable into something callable.

## The fix

Both compilers (the babel plugin **and** the native Rust compiler) now transform ref array literals element by element:

| Element in ref array | Behavior |
| --- | --- |
| bare `let` variable `ref={[el]}` | lowered to an assignment callback, same contract as non-array refs: read the target once; if it currently holds a function, call it with the element; otherwise assign the element |
| member expression `ref={[obj.el]}` | same treatment |
| nested array `ref={[[el]]}` | recursed |
| spread `ref={[...refs]}` | left as-is (already a value) |
| callback `ref={[fn]}`, `const`/module bindings, globals (`undefined`) | left untouched — assigning to a global like `undefined` would throw in strict mode |
| falsy slots `null` / `false` | untouched, still short-circuit at runtime |

This also preserves the existing `let cb = someFn; <div ref={[cb]} />` semantics — `someFn` gets **called** with the element rather than `cb` being overwritten.

## Tests

- **9 new behavioral specs** in `packages/web/test/ref-array.spec.tsx` (the issue's case plus multiple variables, callbacks, mixed arrays, member expressions, nested arrays, falsy slots, mutable/`const` function bindings) — green with **both** the native compiler and `JSX_COMPILER=babel`
- new `refArrays` babel-plugin fixture, reused as an oxc fixture — **zero parity diff** between the two compilers' output
- full suites green: babel-plugin **257 tests**, compiler **5362 JS tests** + **55 cargo tests** (default / no-default-features / tsrx), `clippy -D warnings` clean

## Changeset

Patch bumps for `@solidjs/babel-plugin` and `@solidjs/compiler`.

### DIFF
--- a/.changeset/ref-array-lval-assignment.md
+++ b/.changeset/ref-array-lval-assignment.md
@@ -1,0 +1,15 @@
+---
+"@solidjs/babel-plugin": patch
+"@solidjs/compiler": patch
+---
+
+Assign bare variable refs inside `ref` arrays. `ref={[el]}` type-checked and
+compiled but `el` was never assigned: the array branch passed the array
+straight through, and the runtime only *calls* ref-array entries (a bare
+variable evaluates to `undefined` at mount, so it was silently skipped).
+Bare identifiers and member expressions inside a ref array are now lowered
+to assignment callbacks — read the target once, call it with the element
+when it holds a function, otherwise assign the element — matching the
+non-array `ref={el}` contract. Nested arrays are recursed, callback refs
+and const/module refs pass through untouched, and falsy slots
+(`null`/`undefined`/`false`) keep short-circuiting.
--- a/packages/babel-plugin/src/dom/element.ts
+++ b/packages/babel-plugin/src/dom/element.ts
@@ -32,7 +32,8 @@ import {
   inlineCallExpression,
   hasStaticMarker,
   canChildSlotAllocateIds,
-  isFunctionShapedHole
+  isFunctionShapedHole,
+  transformRefArrayLiteral
 } from "../shared/utils";
 import { transformNode } from "../shared/transform";
 import { InlineElements, BlockElements } from "./constants";
@@ -1003,6 +1004,11 @@ function transformAttributes(
             t.isFunction(value.expression) ||
             t.isArrayExpression(value.expression)
           ) {
+            if (t.isArrayExpression(value.expression)) {
+              // Bare lval elements (e.g. `ref={[elRef]}`) become callback refs
+              // so the variable actually gets assigned (#3285).
+              transformRefArrayLiteral(path.scope, value.expression);
+            }
             results.exprs.unshift(
               t.expressionStatement(
                 t.callExpression(
--- a/packages/babel-plugin/src/shared/component.ts
+++ b/packages/babel-plugin/src/shared/component.ts
@@ -7,7 +7,8 @@ import {
   filterChildren,
   trimWhitespace,
   transformCondition,
-  convertJSXIdentifier
+  convertJSXIdentifier,
+  transformRefArrayLiteral
 } from "./utils";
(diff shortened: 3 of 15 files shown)


javascript-unsafe:
CI note: the `compare` parity check is passing. The aggregate `job` currently fails in pre-existing type checks unrelated to this PR (`@solidjs/web#test-types` cannot resolve the workspace `solid-js`/`@solidjs/web` packages in the CI typecheck environment, and the signals build reports existing errors across `packages/signals/tests/*`). The `size` check also reproduces the current `next` baseline failure: the same `ace227e` next HEAD exceeds two existing caps by 33 B and 30 B; the size apps contain no `ref` usage, so this compiler change cannot affect those bundles.

Local validation on the rebased branch: 9/9 ref-array behavior tests pass with both native Rust and Babel compilers; Babel plugin 257 tests pass; compiler JS suite 5362 passed / 28 skipped; Rust 55 tests pass; clippy is clean.

ryansolid:
> Thanks for the thorough work here — both compilers, fixture parity, a full behavioral suite. Your diagnosis is exactly right, and the implementation is careful about the edges (`const`/globals untouched, spreads left alone, the `let cb = fn` contract preserved).
>
> We're not taking the direction, though — ruling is on #3285. Short version: ref arrays are a callback list, and the bare-variable rewrite stays with the single `ref` form. Arrays don't have to be inlined, so a literal-only rewrite gives `<div ref={[el]} />` and `const refs = [el]; <div ref={refs} />` different behavior under the same type. The compiler can't be the thing that closes that gap; a lint can flag the common shape. Docs now state the rule explicitly.
>
> Closing. Appreciate it — the analysis is what made the decision easy to reason about.
>
> — *Claude via Cursor*

PR #3293, opened 2026-09-06
fix(signals): re-wake tracked readers after staged commits

### LINKED ISSUE #3291: 2.0.0-rc.6: a signal written from a ref is never observed when the element is rendered via <Dynamic> inside a <Portal>
### Describe the bug

Under `2.0.0-rc.6`, a signal written from a `ref` callback **never wakes an effect that reads it** — but only when the element is rendered through `<Dynamic>` **inside a `<Portal>`**. The write itself lands (the signal holds the element); the reading effect is simply never re-run.

Either half alone is fine. `<Portal>` alone works. `<Dynamic>` alone works. Only the combination breaks.

This worked in `2.0.0-rc.5`.

Real-world impact: it breaks every Kobalte popper (`Select`, `Popover`, `Tooltip`, …). Kobalte's popper positioner is a `Polymorphic` (which is `<Dynamic {...props} />`) rendered inside a `Portal`, and it stores the node via `ref={[setPositionerRef, props.ref]}`. The effect that starts floating-ui's `autoUpdate` reads `positionerRef()` and never sees the element, so `updatePosition()` never runs and every dropdown renders unpositioned at the top-left corner of the viewport.

### Minimal reproduction

```tsx
import { Dynamic, Portal } from "@solidjs/web";
import { createSignal, createTrackedEffect } from "solid-js";

const Wrapper = (props: Record<string, unknown>) => <Dynamic {...props} component="div" />;

  const [el, setEl] = createSignal<HTMLElement>();

  createTrackedEffect(() => {
    // rc.5: logs `false` then `true`
    // rc.6: logs `false` and is never re-run
    console.log("effect sees element:", !!el());
  });

  return (
    <Portal>
      <Wrapper ref={setEl} />
    </Portal>
  );
}
```

### What we measured

Seven shapes, same page, same build (Vite 8 + `@solidjs/vite-plugin`, client mode, `ssr: false`). Each renders an element, stores it in a signal from `ref`, and reports from a `createTrackedEffect` whether it ever sees a non-empty value.

| shape | rc.5 | rc.6 |
| --- | --- | --- |
| `<div ref={setEl} />` | sees it | sees it |
| `<Dynamic {...props} />` with ref, no Portal | sees it | sees it |
| `<Portal><div ref={setEl} /></Portal>` | sees it | sees it |
| `<Portal>` behind `<Show>` | sees it | sees it |
| `<Portal>` + plain spread `<div {...props} />` with ref | sees it | sees it |
| **`<Portal>` + `<Dynamic>` with ref** | sees it | **never** |
| same, without `<Show>` | sees it | **never** |
| same, without `omit()` on the props | sees it | **never** |

So neither `<Show>`, nor `omit()`, nor the array-ref form matters — the necessary and sufficient combination is `<Portal>` + `<Dynamic>` + `ref`.

### The write lands; the effect is not notified

Instrumenting Kobalte's popper in `node_modules` under rc.6:

```
[probe] autoUpdate effect: anchor=true positioner=false   (×4)
[probe] setPositionerRef called, element=true             ← ref callback does fire
[probe] in a microtask, signal=true                       ← the write does land
[probe] after 500 ms,   signal=true                       ← and stays
```

The effect ran four times, always reading an empty signal, and was never re-run after the write. Consequently `autoUpdate` never starts and `updatePosition()` is never called — its very first statement (`floatingEl.style.setProperty("--kb-popper-content-overflow-padding", …)`) never executes.

DOM of the positioner, same page, same interaction:

```
rc.5  style="position: absolute; top: 0; left: 0; min-width: max-content;
             --kb-popper-content-overflow-padding: 8px; z-index: 60;
             --kb-popper-anchor-width: 542px; …; transform: translate3d(727px, 422px, 0);"

rc.6  style="position: absolute; top: 0; left: 0; min-width: max-content;"
```

Only the statically declared styles survive; everything written from the effect is absent.

### Ruled out

To save whoever picks this up the same detours:

- **`hasOwn` in the rc.6 spread path** (`const r = hasOwn.call(source, "ref") && source.ref`). `omit()` returns a plain object whose `ref` **is** an own property in both rc.5 and rc.6 (`Object.hasOwn(omit(props,"as"), "ref") === true`), so the ref is not dropped there. Same for the new `hasOwn` guard in the style path: `combineStyle()` returns `{...a, ...b}`.
- **Array refs** (#3285). Kobalte passes `ref={[fn, props.ref]}`; replacing it with a single function in `node_modules` does **not** fix the breakage.
- **Async entanglement holding effects** (#3289). Leaving the page idle for 8 s changes nothing, and other reactivity on the same page works (the dropdown opens and closes).
- **`trackedEffect` itself.** Its implementation is byte-identical between rc.5 and rc.6.

### Versions

- `solid-js`, `@solidjs/web`, `@solidjs/signals`: `2.0.0-rc.5` (works) vs `2.0.0-rc.6` (broken)
- `@kobalte/core`: `2.0.0-alpha.0` and `2.0.0-alpha.1` — both affected, so the Kobalte version is not a factor
- Vite `8.x`, `@solidjs/vite-plugin` start mode, `ssr: false`, Chromium via Playwright

### Expected behavior

A signal written from a `ref` callback should notify effects that read it, regardless of whether the element is rendered through `<Dynamic>` inside a `<Portal>`.

---

*This report was written by an AI assistant (Claude) on my behalf. The measurements, the isolation table and the minimal reproduction above were produced and re-run by it on a real project; I've reviewed and am posting them. Happy to run any further experiment you'd like on the same setup.*

xia-chao:
Closes #3291

## What was broken

A signal written by a `ref` callback was not observed by a `createTrackedEffect` when the element was rendered through `<Dynamic>` inside a `<Portal>`.

The write landed in the signal, but the tracked effect ran only once and kept observing the old value.

## Root cause

The rc.6 Dynamic namespace fix made intrinsic elements lazy. In a Portal, that moves `spread()` and the ref write into the render-effect phase. The tracked effect is notified before the staged signal value is committed, but committed-visibility reads intentionally return the old value. The commit then updated the signal without waking the tracked subscriber again.

## Fix

Reuse the existing missed-wake latch when a committed-visibility reader observes a staged signal value. When the signal commits, tracked subscribers are queued again so they observe the committed value.

## Tests

- Added a regression test covering `createTrackedEffect` + `Dynamic` + `Portal` + `ref`.
- `packages/web`: `npx vitest run test/portal-dynamic-ref-3291.spec.tsx`
- Relevant signals tests: `createTrackedEffect.test.ts`, `effect-phase-writes.test.ts`, and `treeshake.test.ts`

### DIFF
--- a/.changeset/fix-tracked-effect-staged-ref.md
+++ b/.changeset/fix-tracked-effect-staged-ref.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Re-wake tracked effects after a committed-visibility read observes a staged signal value.
--- a/packages/web/test/portal-dynamic-ref-3291.spec.tsx
+++ b/packages/web/test/portal-dynamic-ref-3291.spec.tsx
@@ -1,0 +1,53 @@
+/**
+ * @jsxImportSource @solidjs/web
+ * @vitest-environment jsdom
+ *
+ * Repro for https://github.com/solidjs/solid/issues/3291
+ * A signal written from a ref callback is never observed by an effect
+ * when the element is rendered via <Dynamic> inside a <Portal>.
+ */
+import { describe, expect, test } from "vitest";
+import { createRoot, createSignal, createTrackedEffect, flush } from "solid-js";
+import { Dynamic, Portal, render } from "@solidjs/web";
+
+describe("issue #3291: ref signal not observed through <Dynamic> in <Portal>", () => {
+  test("repro: effect should see the element written from ref", () => {
+    const [el, setEl] = createSignal<HTMLElement>();
+
+    const observations: boolean[] = [];
+    let effectRuns = 0;
+    createTrackedEffect(() => {
+      effectRuns++;
+      observations.push(!!el());
+    });
+    flush();
+
+    const Wrapper = (props: Record<string, unknown>) => <Dynamic {...props} component="div" />;
+
+    const div = document.createElement("div");
+    document.body.appendChild(div);
+
+    const disposer = createRoot(dispose => {
+      render(
+        () => (
+          <Portal>
+            <Wrapper ref={setEl} />
+          </Portal>
+        ),
+        div
+      );
+      return dispose;
+    });
+
+    // The signal should now hold the element.
+    const element = el();
+    expect(element).toBeInstanceOf(HTMLElement);
+
+    // Regression assertion (#3291): the effect MUST be re-run after the ref
+    // write, and that run must observe a truthy element.
+    expect(observations).toContain(true);
+    expect(effectRuns).toBeGreaterThan(1);
+
+    disposer();
+  });
+});


ryansolid:
> Thanks for digging into this one — your root-cause read was right: the tracked effect was woken by the write, ran in the same pass with committed visibility, and nothing re-notified it once the value landed.
>
> We fixed it from the other side in b3c94be8, which is now on `next`: tracked-effect wakes (and the first run) ride the heap like every other subscriber, and `GlobalQueue._update` hands the callback to the user queue when the heap reaches it. That orders the run after the commit structurally, so no latch on the signal is needed — and the tracked special case in `enqueueSub` goes away rather than growing. The Portal + Dynamic + ref shape is pinned in `packages/web/test/dynamic.spec.tsx`, and #3187's lazy creation was reverted in d2b50e91 so `ref` fires at creation again. `createTrackedEffect` is also now marked `@deprecated` (c6c415b2).
>
> Closing as superseded. Appreciate the reproduction and the writeup.
>
> — *Claude via Cursor*

Issue #3299, opened 2026-09-07
Solid 2: `deferStream: true` is silently ignored for memos created inside lazy-loaded (code-split) components


samerset-rmn:
## Describe the bug

`createMemo(fn, { deferStream: true })` only blocks the first flush of `renderToStream` when the memo is created during the initial synchronous render pass. If the memo lives inside a component behind `lazy()`, the shell is flushed before the dynamic import resolves, so by the time the memo exists the `deferStream` branch is unreachable and the option is discarded without a warning.

This is a no-op in the place the option is most needed: with file-based routing, code splitting is the default, so every route component is lazy.

## Environment

- `solid-js`, `@solidjs/web`, `@solidjs/signals` — all `2.0.0-rc.6`
- Node 24.0.1, vite 8.1.5, macOS

## Reproduction

```tsx
import { createMemo, lazy, Loading } from 'solid-js';
import { renderToStream } from '@solidjs/web';

const after = <T,>(ms: number, value: T): Promise<T> =>
  new Promise((resolve) => setTimeout(() => resolve(value), ms));

// Simulates a code-split route chunk: the module itself arrives asynchronously.
const LazyPage = lazy(async () => {
  await after(50, null);
  return {
    default: () => {
      const data = createMemo(() => after(100, 'async content'), {
        deferStream: true,
      });
      return <h1>{data()}</h1>;
    },
  };
});

// Same module, but with the moduleUrl the bundler plugin normally injects.
const LazyPageWithUrl = lazy(
  async () => {
    await after(50, null);
    return {
      default: () => {
        const data = createMemo(() => after(100, 'async content'), {
          deferStream: true,
        });
        return <h1>{data()}</h1>;
      },
    };
  },
  undefined,
  '/src/LazyPage.tsx'
);

// Baseline: eager, no deferStream — streaming here is the documented behaviour.
const EagerNoDefer = () => {
  const data = createMemo(() => after(100, 'async content'));
  return <h1>{data()}</h1>;
};

// Control: the identical memo in an eagerly imported component.
const EagerPage = () => {
  const data = createMemo(() => after(100, 'async content'), {
    deferStream: true,
  });
  return <h1>{data()}</h1>;
};

// lazy() refuses to render without one; nothing here needs real assets.
const manifest = { resolve: () => ({ js: [], css: [] }) };

async function render(label: string, Page: () => unknown) {
  const t0 = Date.now();
  const stream = renderToStream(
    () => (
      <html lang="en">
        <body>
[...]

PR #3300, opened 2026-09-07
fix(web): warn when deferStream misses the shell

nickshiro:
## Summary

ref #3299

- Warn in development when `deferStream` is registered after the initial SSR shell has already flushed.
- Document that late-created sources (such as ones inside lazy-loaded routes) stream normally.
- Add regression test for the lazy `<Loading>` boundary case.
- Preserve the existing root-lazy `deferStream` coverage.

Late lazy boundary cannot retroactively hold an already-sent shell without delaying normal streaming for all lazy boundaries. The warning makes this limitation explicit instead of silently ignoring `deferStream`.

- `pnpm exec vitest run --config vite.config.server.mjs test/server/ssr-stream.spec.tsx`
- 117 passed
- `pnpm --filter @solidjs/web build`


ryansolid:
> Thanks for the quick turnaround on this — but the ruling on #3299 went the other way, so I'm closing this as superseded.
>
> The premise here is that a shell flushed before a lazy chunk loads is the contract, and `deferStream` created afterwards should warn. We decided instead that a `lazy()` module load is **code, not data**: the shell's "no new async discovered during the sync render" rule cannot be evaluated for a segment whose code has not run, so as of b6a90f9f the shell waits for the chunk even under a `<Loading>` — while the boundary still owns whatever data the loaded code discovers. A `deferStream` read inside a code-split route now holds the shell exactly like one in an eagerly imported component, and the test this PR adds (fallback in the shell for `Loading > lazy > deferStream`) asserts the behaviour that was the bug.
>
> The one shape where a `deferStream` still lands after the shell — a fragment whose own data resolved post-shell mounting more content — is by-design streaming (nothing can hold a shell that has gone out), so there isn't a mistake left for the warning to catch.
>
> — *Claude via Cursor*

Issue #2138, opened 2024-04-18
Cannot use istanbul ignore comments for generated code


CreativeTechGuy:
### Describe the bug

When using Solid and testing our app with coverage reporting enabled (powered by istanbul/nyc), there are some branches/functions which are uncovered, but they are not branches which exist in our code, instead they are generated by Solid. Because these code paths are generated, we cannot place a `/* istanbul ignore next */` comment before the appropriate section because it doesn't exist anywhere in our source.

A few cases where this occurs (I'm sure there's others):

* `Show`/`Match` components - If the match is never used, there's an unused function because of the `get children()` method which got generated
* Refs - It generates a ternary with `typeof _ref$ === "function"` as the condiiton. Which in most cases will always be true/false so that means there's a branch of the code which never gets hit.

All of these result in an extra uncovered function here, an extra uncovered branch there. And there's no way we can add the appropriate comments before this code to ignore it.

### Your Example Website or App

https://playground.solidjs.com/anonymous/7b917606-9c75-4695-b669-accb486afc92

### Steps to Reproduce the Bug or Issue

N/A

### Expected behavior

When Solid is generating the runtime code, it should relocate the comments appropriately so they keep the same meaning that they have in the source.

Source:

```jsx
{/* istanbul ignore next */}
<Match when={condition()}>
    <div>Hello</div>
</Match>
```

Generated:

```jsx
_$createComponent(Show, {
  get when() {
    return condition();
  },
  /* istanbul ignore next */
  get children() {
    return _tmpl$();
  }
})
```

-----

Source:

```jsx
let ref;
return <div ref={ref}>Hello</div>
```

Generated: (This should always add the appropriate ignore comments since only one branch will ever be hit.)

```jsx
let ref;
return (() => {
  var _el$ = _tmpl$();
  var _ref$ = ref;
  /* istanbul ignore next */
  typeof _ref$ === "function" ? _$use(_ref$, _el$) : ref = _el$;
  return _el$;
})();
```

### Platform

- OS: N/A
- Browser: N/A
- Version: 1.8.16

### Additional context

It's also impossible currently to write our own plugin to do this since comments will get totally deleted during compilation when placed in certain spots of the JSX.

For example, this input:

```jsx
function Counter() {
  let ref;
  /* istanbul ignore next -- 1 */
  return (
    <>
      {/* istanbul ignore next -- 2 */}
      <div ref={ref}>Hi</div>
    </>
  );
}
```
[...]

PR #3301, opened 2026-09-07
fix(compiler): preserve coverage pragmas in JSX

nickshiro:
## Summary

ref #2138

Preserve JSX Istanbul ignore comments when compiling component children, and annotate compiler-generated ref dispatch branches so they do not count toward coverage.

- Add `/* istanbul ignore next */` annotations to compiler-generated conditional `ref` dispatches.
- Add regression coverage.

- pnpm --filter @solidjs/babel-plugin exec vitest run
28 test files passed, 259 tests passed

- pnpm --filter @solidjs/babel-plugin typecheck

### DIFF
--- a/.changeset/compiler-coverage-pragmas.md
+++ b/.changeset/compiler-coverage-pragmas.md
@@ -1,0 +1,6 @@
+---
+"@solidjs/babel-plugin": patch
+"@solidjs/compiler": patch
+---
+
+Preserve `istanbul ignore` and `c8 ignore` JSX comments on generated component children getters.
--- a/packages/babel-plugin/src/shared/component.ts
+++ b/packages/babel-plugin/src/shared/component.ts
@@ -5,6 +5,7 @@ import {
   isDynamic,
   registerImportMethod,
   filterChildren,
+  getCoverageIgnoreComments,
   trimWhitespace,
   transformCondition,
   convertJSXIdentifier
@@ -20,7 +21,7 @@ type ComponentTransformResult = TransformResult & {
   exprs: Array<t.Expression | t.Statement>;
 };
 
-type ComponentChildrenResult = [t.Expression, boolean] | undefined;
+type ComponentChildrenResult = [t.Expression, boolean, t.Comment[]?] | undefined;
 
 function isSimpleOptionalMemberExpression(
   expression: t.Expression | t.JSXEmptyExpression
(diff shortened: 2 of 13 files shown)


ryansolid:
> Thanks for taking this on — half of it is exactly what we want, and the diff splits cleanly along that line.
>
> **Keep: preserving the user's own pragma.** A `{/* istanbul ignore next */}` the author wrote in JSX should survive onto the `get children()` getter the compiler emits. That is comment fidelity for a deliberate annotation, and it's the right layer for it. Two asks on this half:
>
> - Match `c8 ignore` alongside `istanbul ignore` (`/^\s*(istanbul|c8)\s+ignore\b/`). v8-based coverage is what most Vitest users run now, and it's the same regex.
> - Oxc parity. The Babel plugin and `packages/compiler` share codegen expectations; the same carry-through needs to land in the native compiler with the shared fixture before this can merge.
>
> **Drop: injecting `/* istanbul ignore next */` on the generated ref dispatch.** This is the part that touches `dom/element.ts`, `universal/element.ts`, and rewrites fourteen fixture outputs, and we don't want it — not for size, on principle:
>
> - The compiler would be emitting a specific tool's pragma into every user's output. That's tool knowledge in the wrong layer, and it doesn't stop at one vendor.
> - Coverage of compiler-generated branches is the coverage tool's problem, and the tools have been solving it: v8 coverage with AST-aware remapping attributes back through the source map to the JSX, where there is no branch. Instrumenting post-compile output with Istanbul is the configuration that produces the phantom branches.
> - Every injected comment is dev-build bytes for everyone and one more thing both compilers must mirror forever.
>
> If real demand shows up later, an explicit opt-in is the ceiling we'd consider, but we'd rather not add it speculatively.
>
> Also: `@solidjs/web#test-types` is red on this branch. It's probably the base (`next` moved on Sep 7); a rebase should tell.
>
> Once it's down to the preserve half + `c8` + Oxc parity, happy to merge.
>
> — *Claude via Cursor*

nickshiro:
- Removed all automatically injected coverage comments from generated `ref` dispatches.
 - Preserved only user-authored JSX pragmas, now supporting both `istanbul ignore` and `c8 ignore`.
 - Added Oxc parity for DOM, Universal, and SSR.
 - Added a shared fixture consumed by the Babel and native compiler regression tests.

### Thanks for your review!

ryansolid (changes requested):
> This is now exactly the shape we asked for — thanks. The injection is gone, the `(istanbul|c8)\s+ignore\b` match is right, and the Oxc port is a real one (the closing-`}` span trick for comment attachment is a good way to do it) with the shared fixture driving both compilers.
>
> One blocker: the native compiler doesn't build. `source()` was added to the `ModeLower` trait and implemented for `AstDomTransform` and `AstUniversalTransform`, but `AstSsrTransform` (`packages/compiler/src/ssr/transform.rs`, the `impl ModeLower` near the bottom) was missed:
>
> ```
> error[E0046]: not all trait items implemented, missing: `source`
> ```
>
> Same three lines as the other two impls:
>
> ```rust
> fn source(&self) -> &str {
>     self.source
> }
> ```
>
> Please also rebase onto `next` — the CodSpeed and Size failures are inherited from the base at the time you branched and will clear — and run the Rust side before pushing (`cargo build` / `pnpm --filter @solidjs/compiler test`), since the JS tests alone can't see this.
>
> Two nits, take or leave: the Babel test is in `ref-spread.spec.js`, which is about intrinsic ref/spread sources — a `coverage-pragmas.spec.js` (or the shared-fixtures runner) would be a more findable home. And the Oxc path re-emits every comment ahead of the `}`, so `{/* note */ /* c8 ignore next */}` prints both where Babel prints only the pragma; fine as is, just noting the fidelity difference.
>
> Once it builds green this merges.
>
> — *Claude via Cursor*

PR #3305, opened 2026-09-07
fix(signals): hold conditional reveals on existing async work

brenelz:
**Summary**
- Hold staged writes when a render reader first observes an unresolved async flight from an earlier flush. This keeps `show` committed as `false` with `isPending(show) === true` until the details can reveal with it, without retroactively holding the already committed `value` update.
- Check transition ownership rather than reader initialization: newly created child readers must also hold the reveal, while readers already parked in a transition must not open a second one. Fresh/reset loading boundaries still consume pending and display their fallback.
- Add coverage to the existing async-semantics and DOM loading suites, plus an `@solidjs/signals` patch changeset. No standalone scratch tests.

**Reproduction**
https://s.olid.uk/id/1xogoHinT9uZhdrpGNy7LQ

The show-line indicator in the playground uses `isPending(value)`; the DOM regression uses `isPending(show)`. The runtime bug remains independently of that typo: `show() ? details() : "hidden"` discovers the existing flight only after the reveal write.

**Verification**
- Signals suite: 1,502 passed, 1 skipped.
- Solid suite: 585 passed.
- Web suites: 704 client, 750 server, and 165 hydration tests passed; 2 server tests skipped.
- DOM loading suite with `JSX_COMPILER=babel`: 21 passed.
- Rebuilt runtimes before downstream tests; bundle-size limits, Prettier, and `git diff --check` passed.

### DIFF
--- a/.changeset/fix-conditional-async-reveal-pending.md
+++ b/.changeset/fix-conditional-async-reveal-pending.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Hold conditional reveals that first observe async work started in an earlier flush. The revealing signal now stays pending until the details can commit with it, including when the reveal creates a new child reader. Preserve fresh/reset loading-boundary fallbacks and avoid opening a second transition for readers already waiting in one.
--- a/packages/signals/src/core/scheduler.ts
+++ b/packages/signals/src/core/scheduler.ts
@@ -755,6 +755,17 @@ export class GlobalQueue extends Queue {
         // to completion accounting BY CONSTRUCTION: it never registers a
         // reporter and never counts toward the loading-boundary diagnostic.
         if ((actualError as NotReadyError)?._markVisual) return true;
+        // A reveal can discover a flight started in an earlier flush. Hold the
+        // staged writes with that reader (A15), even if the reader is new.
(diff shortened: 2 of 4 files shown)


ryansolid:
> Ruling: this is a gap, not a semantics change — the same A15 rule (a non-action write whose render surfaces async is held with `isPending` on the written signal), applied when the reader *discovers* a flight instead of starting one. Today's fresh-start path opens the transition from `handleAsync`; nothing did so for a pre-existing, never-held flight, so `show` committed `true` while the content froze on `"hidden"` with nothing reporting pending.
>
> I pushed one commit on top (`98d457f8`): identical semantics, the new check folded into the existing `actualError` branch, `currentBatch` instead of `this._batch`, and `??` for the error fallback (exact — every caller passes nothing or this node's own `_x._error`). That takes `signals: + createStore` from 18 B over the cap on Linux to 14.36 KB, under both the cap and the pre-PR baseline, so no ratchet was needed. I also checked the guards: the `_pendingNodes.length` gate is what keeps a first mount from opening a reporter-blocked transaction that would stash every effect in the flush, and `!node._transition` is load-bearing (three `createProjection.async` isPending tests fail without it). Both stay.
>
> Merging — thanks for the clean reproduction and the boundary-shape tests.
>
> — *Claude via Cursor*

PR #3308, opened 2026-09-08
feat: unified For is the keyed-For engine — all modes, no seam, mapArray as the oracle

ryansolid:
Successor to #3281, restructured per the 2026-09-07 audit findings and the plan discussed with @ryansolid. #3281 stays open as the record/comparison.

## What changed in shape

#3281 built the slot as an **opportunistic fast path**: engage when the row shape looked right, decline or demote to classic otherwise. Every non-algorithm finding across three external audits lived in that seam — ownership, demote hand-offs, hydration claim restore, dynamic rows. This branch removes the seam: **the slot is the `For` implementation for web (and universal)**. There is no engage/decline/demote, no `lateClassic`, no hole re-entry. `For` stamps `$for` unconditionally; `insert` engages unconditionally.

That requires the engine to own every `For` mode itself, and it does — with **mapArray as the specification**:

- `keyed` default (reference identity), `keyed={false}` (positional reuse, item accessor + plain index, tail append/remove), `keyed={fn}` (item accessor + index accessor), index accessors by row arity (mapArray's own `length > 1` rule)
- duplicates are legal: mapArray's chained pairing (`newIndicesNext`), so the second occurrence of a key reuses the second old row — the persistent key map is gone with it
- `fallback` as an owned empty-state row; array-like subjects duck-typed like mapArray
- empty rows render **zero nodes** (neighbor-anchored) — no placeholder
- rows live under `For`'s creation owner; contiguous list end past foreign trailing nodes; parent-guarded removes; throw-safe builds
- dynamic rows (function top level) resolved tracked by the engine's compute — classic's list-effect model — with NotReady-parked plan reuse

`For` still returns a callable that runs mapArray for `children()`, introspection, and renderers that don't engage, so mapArray stays in `For`-bearing bundles; it is the public primitive and the spec.

**Flat mode** (parallel-array first fills, chain materializes lazily) was measured rather than assumed: +5–10% on 10k create/clear without it, parity at 1k, so it stays — and it now covers every keyed mode (item/index signals ride the parallel arrays; keys are computed at materialize; survivors are judged by key). Only `keyed={false}` is chain-first.

**Hydration mismatch — detect, don't recover** (ruling): the fill commit performs no DOM writes. Server text nodes are adopted for primitive rows; leftover server rows stay; key-missed client rows land on the next update (classic parity). The runtime already reports unclaimed elements and key misses; the engine warns once for the one blind spot (text rows).

**Universal**: `createRenderer`'s `insert` engages `$for` with ops built from the renderer's own primitives — no new `RendererOptions`.

## Verification

- **Oracle harness** (`for.unified.oracle.spec`): 4 modes × 3 seeds × 220 steps off one signal, `mapArray` + `insert` directly on the other side (no `<For>`), mixed row shapes (element / zero-node / fragment / dynamic `<Show>` that flips mid-run), ops incl. shuffles, duplicates, clears, replaces, same-key re-mints. Compared after every step: DOM, per-key node retention, row-fn invocation counts, cleanup set. Ruling recorded in the harness: disposal *order* among rows removed in one step is not a For contract (mapArray's own order is an internal artifact).
- **Parity matrix** runs engine / engine-index / mapArray-oracle across 3 anchors × 3 row shapes × 31 transitions.
- **Modes spec** with the oracle beside every mode; the pre-existing classic `for.spec`, `for.nonkeyed.store.spec` etc. now run *through the engine*.
- **Hydration**: 187 scenarios incl. key-fn / indexed / by-index / fallback fills and a classic-fallback oracle.
- Bugs the oracle caught during this work (all fixed, all pinned): fill paths reading items untracked (store index writes didn't re-run), fallback row built under the wrong owner (duplicated under hydration), key-fn survivor probe by identity (re-minted objects rebuilt rows), and two wrong identity assumptions in my own tests.
- Suites: web 805, server 779, hydrate 187, solid 585, signals 1503, universal 46.

## Size, cumulative vs `next` (brotli)

| scenario | next | this branch | delta |
|---|---|---|---|
| app floor (no For) | 10,733 | 10,962 | +229 B |
| CSR app with For | 12,900 | 17,182 | **+4.28 KB** |
| hydrating app | 17,485 | 22,003 | **+4.52 KB** |

For reference, #3281 was +3.58 / +3.92 KB; the seam removal bought ~0.4 KB back and the modes/fallback/duplicates cost ~0.9 KB. Every ratchet note in `.size-limit.js` now states the cumulative figure.

## Performance (Chrome, interleaved A/B on the flagged cells; sequential battery for the rest)

- Creates at parity with classic on all five fixtures (jfb-signal runlots 16.1 vs 15.5–16.1; jfb-shallow 17.0/17.1 vs 17.9/17.3 after flat-mode key-fn; dbmon mount/tick_partial parity, tick engine-faster).
- Reorder wins match #3281 on jfb-signal (removefirst 0.19×, displace 0.47×, rotate 0.66×, shuffle 0.76×) and extend to jfb-shallow, which #3281 declined to classic (rotate 0.46×, displace 0.45×, removefirst 0.18×, reverse 0.58 vs 0.65).
- Verified: the mapArray move-plan alternative (export old→new positions, place by LIS, no second diff) recovers only 5–10% of the small-op gap — the wins come from owning the bookkeeping, not from removing the second diff. That spike is what justified keeping this shape.

## Follow-ups

- One-engine array output (teach the engine to answer a plain call with the current row values) would let `For` drop mapArray for `children()`/universal — ~200–300 B, decide after this lands.

— *Claude via Cursor*


ryansolid:
> **Update — one list implementation (`f70d5ce3`).**
>
> The row core now lives in `@solidjs/signals` (`list.ts`) and **`mapArray` is its array output** — same API and contract (values, array identity while structurally unchanged, `[fallback]` when empty, dev strict-read name, `_parentComputed` routing for row bodies). A plain call of a `<For>` accessor returns the same. The old mapArray body is deleted from the shipped package and kept only as a **test reference** (`web/test/reference/mapArray.ts`, on public API); every oracle harness now compares the engine against it. signals' own 35 mapArray tests (NotReady exception safety, store non-keyed, latest/repeated writes) run against the engine and passed unchanged.
>
> The rendered output goes through a **node layer** — the only code that touches nodes — which solid-js builds over a renderer's `SlotOps`. Web (`domOps`) and `@solidjs/universal` (`createRenderer` primitives) both consume the same layer; universal's oracle test now includes dynamic rows as the second-renderer proof.
>
> Two bugs the reference oracle caught *in the move itself*, both fixed: the array memo was created under its reader (a `<For>` builds array output lazily on first read) and died with the reader's next run; and routing row owners' `_parentComputed` through the rendered compute gave dynamic rows' memos a height above the compute — a height inversion that double-ran the compute on a flip + list change and rebuilt the pass's fresh rows. Routing is array-output only now.
>
> **Size (brotli, minimal bundle, vs `next` For-only 12,019 B):**
>
> | consumer | next | engine before move | now |
> |---|---|---|---|
> | For only | 12,019 | +3.59 KB | +3.81 KB |
> | For + mapArray | 12,026 | +4.38 KB | **+3.87 KB** |
> | mapArray, no For | +850 B | +850 B | +3.69 KB |
>
> For-only pays ~220 B for the layer boundary; For + mapArray drops ~510 B (no second implementation). mapArray-without-`For` (custom renderers without `For`, data-only mapArray) pays more because the rendered-only control flow shares the engine closure — making flat/LIS/dynamic/placement pluggable is the follow-up if that consumer shape matters.
>
> **Perf:** Chrome, interleaved pre-move vs moved on jfb-signal — reorder mins and creates identical. Suites: web 810, server 779, hydrate 187, solid 585, universal 47, signals 1503.
>
> — *Claude via Cursor*

ryansolid:
> **Response to the 2026-09-08 audit (9 P1 / 4 P2) — fixed in `6a914102`, one test per finding.**
>
> **P1-1 module-global scratch — real, the most serious one.** `bpOwner` was written *before* the row fn ran; a `<For>` nested in a row engages synchronously and clobbered it, so the outer row recorded the nested row's owner. `buildParts` now returns the owner and writes its result slots only *after* user code (LIFO-safe); the node layer's `dyn` slot likewise. Pinned: removing an outer row disposes exactly that row and its nested rows.
> **P1-2 key-fn semantics** — identity-first comparison (`row.item === next || key(row) === key(next)`), keys computed lazily by the first diff that needs them, so no key fn runs during a fill and an in-place key mutation keeps its row. Pinned against the reference (0 key calls on fill; mutated `id` keeps the node).
> **P1-3 creation-owner disposal** — the engine registers its death on the *creation* owner (not the list owner, whose cleanups fire on every bulk `dispose(false)`): a rendered list whose `<For>` owner was disposed freezes at its last value, like mapArray.
> **P1-4 two engines** — one engine per list. Rendered first → a plain call reads that engine's tracked array view (version signal bumped per commit). Called first → `impl` returns `false` and the renderer inserts the array output the classic way. Rows are built once in both orders (pinned both).
> **P1-5 migrated retained rows** — chain-wide liveness sweep on structural commits, matching classic's live-only prefix/suffix skip. Pinned against the reference. **Cost flagged below.**
> **P1-6 host tagging / initial range** — `SlotOps.placed` hook; web's host-aware inserts use per-host ops tagging `_$host`; a caller-provided non-hydrating initial range is consumed before engagement (web + universal).
> **P1-7 chain-mode hydration** — `keyed={false}` and fallback fills run the same positional adoption commit as flat fills. Pinned: `keyed={false}` text rows and a primitive fallback hydrate without duplication.
> **P1-8 extra id** — `For` consumes the server's two slots and hands both to the engine: row owner (`hid`) and the array computed (`hid2`). Pinned: a `children()`-introspected `For` followed by siblings keeps their ids.
> **P1-9 universal node inference** — `RendererOptions.isNode` (default: non-array object, which is what universal's `insertExpression` already assumed); text-data tracking guards non-object nodes.
> **P2** — fallback called with zero arguments (pinned on `mapArray`); hydration adoption no longer writes text data (server text stands, live edits survive — pinned with a differing-text scenario); falsy-fallback retention recorded as a ruling.
>
> **Perf trade-off to rule on (P1-5):** the reclaim sweep is one `parentNode` read per row per structural commit — ~20 µs at 1k rows: removefirst 0.015→0.035 ms, rotate 0.066→0.076, displace 0.07→0.09. Still 2–4× ahead of classic on every op, but it spends a third of the small-op win on a rare user action. Options: keep (exact parity), or restrict reclaim to the changed window (a migrated *prefix* row would come back only when it enters a window). Parity is in for now.
>
> **On the architectural recommendation** (share the kernel, keep a light mapArray adapter): the +43% standalone mapArray figure is the mapArray-*without*-`For` case. Measured on this branch: For-only apps +3.81 KB vs `next`, For **+** mapArray apps +3.87 KB (was +4.38 with two implementations), mapArray-only +3.69 KB (was +0.85). Making flat mode / LIS / dynamic rows / placement pluggable so the array path shakes them is the follow-up if that consumer shape matters; it's additive.
>
> Suites: web 819, server 783, hydrate 191, solid 585, universal 47, signals 1503. Cumulative size vs `next`: CSR +4.34 KB, hydrating +4.51 KB, floor +276 B.
>
> — *Claude via Cursor*

ryansolid:
> Closing: the unified-For campaign is wound down in favor of targeted work on the existing list primitives.
>
> Across #3281 and #3308 the design got architecturally cleaner each round (one engine, no seam, `mapArray` as the oracle), but every audit still surfaced a fresh batch of P1s in ownership, hydration, and key semantics, and the paired benchmarks never showed a general win that justified the size or the coordination risk — the takeaway being that a parallel list system is where complexity caught us every time.
>
> What was kept: the Tier-1 DOM-lane benches and semantic gates, the oracle-harness approach for differential testing against `mapArray`, and the `UNSTABLE_LIST_IDENTITY` diagnostic. The targeted outcome of the same investigation is #3227 (small-move fast path inside `mapArray`). The branch stays in place as reference.
>
> — *Claude via Cursor*

PR #3310, opened 2026-09-08
fix(signals): preserve parked effects when transitions merge

DerpyCrabs:
## Summary

Fixes lost render updates when a pending action merges into another transaction. Signal values commit correctly, but the DOM can remain stale because `mergeTransitionState` does not transfer the outgoing transaction’s parked effect queues.

The fix restores those effects into the scheduler and clears the outgoing stash. Includes a minimal regression test and a patch changeset for `@solidjs/signals`.

Confirmed the regression test fails on current `next` (`94fe5b46`) without this fix and passes with it. Also verified a standalone reproduction using only `@solidjs/signals` and the original application’s grid-switching and clipboard-replacement failures.

Completed validation using Node 24, matching `.nvmrc`:

- `pnpm i` — passed.
- `pnpm --filter @solidjs/compiler run build` — passed; prerequisite used by upstream CI.
- `pnpm build` — all 25 tasks passed.
- `pnpm test` — all 33 tasks passed, including workspace type tests: **4,399 tests passed, 3 skipped**.
- Prettier check of all changed files — passed.

### DIFF
--- a/.changeset/preserve-merged-transition-effects.md
+++ b/.changeset/preserve-merged-transition-effects.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Preserve queued effects when pending actions merge into another transaction, preventing stale DOM output after signal values commit.
--- a/packages/signals/src/core/scheduler.ts
+++ b/packages/signals/src/core/scheduler.ts
@@ -222,6 +222,10 @@ function createBatch(): Transition {
 function mergeTransitionState(target: Transition, outgoing: Transition): void {
   if (__DEV__ && attrHooks !== null) attrHooks.transitionMerged(target, outgoing);
   outgoing._done = target;
+  // Requeue held effects under the surviving transition so its next flush
+  // either parks them again or runs them when the merged work completes.
+  globalQueue.restoreQueues(outgoing._queueStash);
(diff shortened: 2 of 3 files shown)


ryansolid:
> Confirmed and landed as b1c4399e — thank you for the reproduction and the fix, this was a real lost-update bug. Your regression test is in verbatim and you're credited as co-author.
>
> Two small changes from your diff:
>
> - The restore moved to the merge site in \`initTransition\` (\`this.restoreQueues(outgoing._queueStash)\`) rather than inside \`mergeTransitionState\`. Same behaviour; that phrasing measures byte-neutral on the core-floor size budget where the original form was +19 B (and brotli amplified it to +93 B in the store scenario, over the cap).
> - The \`outgoing._queueStash = { ... }\` reset was dropped. After the merge the outgoing transaction is dead — \`_done\` chased, removed from \`transitions\`, never stashed into or completed again — so nothing reads that stash; the reset was dead bookkeeping.
>
> The changeset was folded straight into the rc.7 changelog instead of committed as a file: rc.7 was already versioned on \`next\` but its publish had been blocked by an unrelated CI harness issue, so this rides along in rc.7 rather than waiting for rc.8.
>
> Closing in favour of the landed commit.
>
> — *Claude via Cursor*

PR #3317, opened 2026-09-09
feat: observe tier — wiring vs checks as a build flavor, attribution engine as an entry

ryansolid:
## Summary

Splits what `__DEV__` gated into two flags and three build tiers, and moves the attribution engine out of the runtime into its own entry. Plan: `documentation/plans/observe-tier-plan.md` (decisions D1–D6). Rationale: `documentation/proposals/production-observability-sketch.md` §3.

- **Checks** (strict reads, owner-scope writes, invariants, forbidden scopes, console, dev-only error text) stay under `__DEV__` — dev builds only.
- **Wiring** (attribution hook sites, `_name` labels, graph edge counters, the diagnostics channel, the interaction frame) moves to `__OBSERVE__` / `"_SOLID_OBSERVE_"` — dev **and** a new `observe` tier.
- `__DEV__ ⇒ __OBSERVE__`, asserted at module init.

### New export condition: `observe`

Listed after `development` (dev wins when both are set). Packages ship an observe flavor only for entries that contain wiring: signals `dist/observe/` + `node.observe.cjs`, solid-js `solid.observe.*` + `server.observe.*`, web `web.observe.*`, universal `universal.observe.*`. Frames, server-functions, storage, h, html, element fall through to prod (pinned by the resolution test). Compile-time gating everywhere — measured: Rollup/Rolldown fold a runtime gate to zero but esbuild does not, so a runtime gate would have charged esbuild/webpack users and the size harness. Prod artifacts are byte-identical apart from `OBSERVE = undefined`.

### Runtime surface

- `OBSERVE` (dev + observe) = `{ diagnostics: { subscribe, capture, emit }, attribution: { install, installed, withInteraction }, subjectOf }`. `emit` accepts an explicit `ownerPath` for hosts whose owners are not signals' owners (SSR, P1).
- `DEV` (dev only) = `{ hooks, getChildren, getSignals, getParent, getSources, getObservers, report, setConsoleFooter }`.

### The engine is an entry, not a member

`@solidjs/signals/attribution` (re-exported as `solid-js/attribution`) owns `enable/disable/subscribe/history/why/subscriptions/costs/waterfalls/holds/feedback/markFlight/format/formatOrigin` and, newly, exports the record types. Core keeps only the hook slot and `withInteraction` (moved to `attribution-hooks.ts`; hooks gained `interactionStart`/`interactionEnd`). Nothing reachable from the core index imports the engine. Prod resolves an inert twin with the same `Attribution` surface, so the import needs no per-tier guard.

Every signals build now has two entries sharing one module instance: trees via `preserveModules`; flat dev/CJS builds code-split into `<name>.<ext>`, `<name>.attribution.<ext>`, `<name>-shared.<ext>`, mangled as one domain (`mangle-props.mjs` accepts comma-separated groups). `_name` is reserved from mangling — the one cross-package field.

### Size (CSR scenario, brotli)

| scenario | before | after |
|---|---|---|
| prod | 12.91 KB | 12.91 KB |
| observe | — (engine-in: 23.79 KB) | **14.20 KB** (+1.29 KB, the wiring) |
| observe + engine enabled | — | 23.91 KB (engine 9.7 KB, opt-in) |

New `csr-app-attribution.js` scenario; caps 14.30 / 24.00 KB. Prod scenarios unchanged.

### Tests

- Signals dist: `DEV`/`OBSERVE` per tier × ESM/CJS; the separately-loaded engine drives its own core through the slot (incl. code-split CJS); the engine is absent from the observe core and from prod's `attribution.js` (which imports nothing).
- solid-js: new `dist-client-artifact.spec.ts` (labels present in observe/dev, absent in prod; no dev-only checks in observe); server artifact test extended to observe.
- web: `exports-server-conditions` covers `observe`, `development`-over-`observe`, CJS pairing, and `solid-js/attribution` → `@solidjs/signals/attribution` per tier.
- All `OBSERVE.attribution.*` tests migrated to the engine import; `withInteraction` stays on `OBSERVE`.

### Breaking (pre-release)

`DEV.diagnostics` → `OBSERVE.diagnostics`; `DEV.diagnostics.setConsoleFooter` → `DEV.setConsoleFooter`; `DEV.attribution.*` → `import { attribution } from "solid-js/attribution"`. Consumers in-repo (`@solidjs/diagnostics`, skills, docs) updated. Changeset included.

### Not in this PR

Serializable event shape (`RerunEvent.node`, `ts`, `origin` unifying interaction + request) — to be specified against an external adapter prototype, alongside server-dev-build P1. Observe-tier runtime cost is unmeasured (bench plumbing exists; baseline pending). CJS removal and the `@solidjs/vite-plugin` `observe` option are separate PRs. Selecting the tier today: `resolve.conditions: ["observe"]`.

---
Co-authored with Claude via Cursor.


ryansolid:
> Review pass before merge, for the record:
>
> - **Export conditions**: `development` precedes `observe` at every branch in all four packages. Observe flavors ship exactly where wiring exists — web's server has zero gate sites (client-only flavor is right), `solid-js/refresh` has none, `solid/src/server` has one and ships `server.observe`.
> - **Gating**: every `__DEV__` → `__OBSERVE__` conversion in the engine is a hook call, `_name` label, graph-link note or `registerGraph`; `_internal` stayed dev-only; in observe `registerGraph` only stamps `_owner` (the `_signals` list and `onGraph` hook stay under `__DEV__`). No check leaks into observe.
> - **Prod artifacts**: `web.js` byte-identical to `next`; `solid.js` and signals prod differ only by `OBSERVE = undefined` (+export, +two kept JSDoc comments the minifier drops). The +59 B on the "hydrating (no stores)" size scenario is brotli reacting to the export list, not code.
> - **Full forced build + all suites** green in a clean worktree install (signals 1604, solid 594, web 712/777/165, universal 43, diagnostics 31, element 10, html 192, h 61, babel 255, test-integration, type tests). Size harness green including the two new observe scenarios (14204/14300, 23909/24000).
> - **CodSpeed's 4 flagged web benches are phantom**: run locally on both trees they're flat (`dynamic-tag update` ×3 within ±0.5%, `update-partial` −1.3% inside its ±2% rme), and structurally the benches run under `development` on both trees where dev ⇒ observe, so the wiring is identical. Same cross-environment artefact as #3309.
> - **One real blocker found and fixed on the branch** (feb425e5): the compiler's `tsrx-runtime.test.js` bundles sources with `define: { __DEV__ }` and knew nothing of `__OBSERVE__` — every runtime test threw `ReferenceError: __OBSERVE__ is not defined`. Solid CI doesn't run those (native binary); they'd have failed in the Compiler-binaries workflow the release gates on, exactly like rc.7. Added the flag to both the define and the string replace, and made the harness fail with a named message on any surviving `__FLAG__` / `"_SOLID_FLAG_"` so the next flag can't slip through.
>
> — *Claude via Cursor*

PR #3318, opened 2026-09-09
build: ESM only — remove CJS artifacts, require branches and types-cjs

ryansolid:
## Summary

All runtime packages publish ESM only and declare `engines.node >= 22.12`. Every `.cjs` artifact, every `require` branch in the exports maps, the `types-cjs/` declaration mirrors and the `sync-dual-types.mjs` pipeline are gone.

**Why now.** Node 20 is EOL (2026-04); every supported Node `require()`s ESM natively (unflagged in 22.12). Every other runtime Solid targets (Bun, Deno, workers, Tauri, SolidTV's bundled output, OpenTUI) is ESM-native and never read the CJS files. The dual build existed for Jest, which Solid 2 neither documents nor ships tooling for. Keeping it cost a build matrix a third larger and — as the observe-tier PR showed — a second set of per-tier artifacts (`node.cjs` / `node.dev.cjs` / `node.observe.cjs`) that had to be kept honest by hand.

**Resolution is unchanged.** Each `{ import: X, require: Y }` pair collapses to X's entries *at `import`'s position*, so condition order is preserved: `browser` still precedes `node`, `development` still precedes `observe`, and a node16/nodenext TS consumer still finds the server types under the `node` key before the top-level client ones. A CJS host resolves the same files through the same conditions.

## What replaces the CJS coverage

- **`packages/signals/tests/dist-artifacts.test.ts`** (renamed from `dist-cjs-artifacts`): for each tier, spawns a real Node that `require()`s the core and engine entries and checks they are the *same instance* `import()` yields (vitest's module runner would hand `import` its own copy, so the probe cannot run in-process). Also scans every shipped module for a top-level `await` — the one thing that makes `require(esm)` throw.
- **`packages/web/test/server/exports-server-conditions.spec.tsx`**: the `require` walk `@solidjs/web → solid-js → @solidjs/signals` lands on the same ESM file per tier (prod / development / observe), and the chain actually loads through a CJS `require` with `DEV` defined in dev and undefined in prod.
- **`packages/test-integration/test-imports.mjs`**: every public specifier (now including `/attribution`, `/serialization`, `/frames*`, `/rich-args`) is loaded through both `import()` and `require()`; `ERR_REQUIRE_ASYNC_MODULE` fails the run.
- **`packages/test-integration/fixtures/packaged-types/nodenext-cjs`** (was `node16-cjs`): the packed-types check for a CommonJS TS project moves to `module: NodeNext`, the setting under which TS type-checks `require()` of ESM. Passes against the packed tarballs.

## Consumer impact

- ESM apps, Vite, Vitest, Bun, Deno, workers: none.
- CommonJS Node apps: Node 22.12+; `require("solid-js")` keeps working.
- TypeScript CommonJS projects: `module: "NodeNext"` (TS 5.8+); `Node16` reports TS1479.
- Jest: needs Node 22.12+; presets that map to `.cjs` paths have nothing to map to.

## Out of scope

`@solidjs/babel-plugin` and `@solidjs/compiler`'s napi loader are build-time tooling loaded by Babel/Node tooling, not runtime artifacts; unchanged.

Size budget unchanged (ESM artifacts are byte-identical). Full `pnpm test`, `test:integration`, and `scripts/size` pass.


ryansolid:
> Verified before merge:
>
> - No stale references to `.cjs` / `types-cjs` / `sync-dual-types` / `d.cts` anywhere in tooling, CI workflows or `smoke-release.mjs`; `.nvmrc` is 24 so every workflow can `require(esm)`.
> - Export maps are a faithful collapse: at every branch `types` sits exactly where the old `{ import: { types, default } }` did, `development` still precedes `observe`, `browser` precedes `node`; no `import`/`require` keys remain, no `.cjs` paths remain.
> - Clean worktree install → forced build + types: no CJS artifacts in any `dist/`. All suites green (signals 1603, solid 594, web 712/777/165, universal 43, diagnostics 31, element 10, html 192, h 61, babel 255), type tests, `test-imports.mjs` (import + `require(esm)` walk of every public specifier), packaged-types NodeNext-CJS fixture, compiler TSRX runtime harness. Size harness byte-identical to `next` on all ten scenarios.
>
> — *Claude via Cursor*

Issue #2916, opened 2026-07-18
2.0.0-beta.20: post-action completion race permanently freezes fresh signals


brenelz:
### Describe the bug

In Solid 2.0.0-beta.20, a signal written in the microtask window after one action completes but while another action keeps the shared transition incomplete can become permanently frozen.

The first write never commits. More seriously, every later ambient write to the same signal also remains uncommitted because the signal keeps a `_pendingValue` that is no longer present in any batch or transition queue. Dev mode reports `INV-7`, confirming that the value can never commit.

### Your Example Website or App

Focused regression test against `solidjs/solid` `next` at `ec432b40`:

```ts
import { action, createSignal, flush } from "../src/index.js";
import * as scheduler from "../src/core/scheduler.js";

const tick = () => Promise.resolve();
const [y, setY] = createSignal(0);

let resolveA!: () => void;
const pA = new Promise<void>(r => (resolveA = r));
let resolveB!: () => void;
const pB = new Promise<void>(r => (resolveB = r));

const A = action(async function* () {
  yield pA;
});
const B = action(function* () {
  yield pB;
});

const aDone = A();
const bDone = B();
flush(); // stash the shared incomplete transition

resolveA();

// Write in the window where A's done() has restored activeTransition but
// its scheduled flush has not run. The internal read only makes the timing
// deterministic; the write itself is an ordinary application write.
let wrote = false;
for (let i = 0; i < 16; i++) {
  await tick();
  if (!wrote && scheduler.activeTransition !== null) {
    wrote = true;
    setY(7);
  }
}

resolveB();
await Promise.all([aDone, bDone]);

try {
  flush();
} catch {
  // Dev: INV-7, pending value at quiescence with no queued commit.
}

console.log(y()); // actual: 0; expected: 7

setY(9);
try {
  flush();
} catch {}

console.log(y()); // actual: still 0; expected: 9
```

### Steps to Reproduce the Bug or Issue

1. Start two actions that share an incomplete transition.
2. Flush so the transition is stashed.
3. Complete action A while action B remains pending.
4. Write a fresh signal after A's `done()` restores `activeTransition` but before the scheduled flush.
5. Complete action B and flush.
6. Observe that the write never commits and later writes to the signal remain frozen.

### Expected behavior

The write should either join the active transition and commit when B settles, or remain in the ambient batch and commit normally. Once the transition completes, later writes must always remain usable.

### Screenshots or Videos

Not applicable.

### Platform

- OS: macOS
[...]

PR #3319, opened 2026-09-09
fix(signals): preserve transactions reentered during finalization

DerpyCrabs:
An ambient flush can re-enter a pending transaction while finalizing, then continue committing state and applying effects as though it still owns the original batch.

We hit this in an application when an SSE reconnect triggered parallel query refetches. One refetch updated a settings projection read through `snapshot(deep(settings.pins))`, while another refetch was still pending. An unrelated signal update triggered a flush; committing the settings store notified its deep reader and re-entered the pending transaction. Once both refetches completed, the network data was current but the pins UI remained stale.

`finalizePureQueue()` can adopt another transaction through store commit hooks, boundary checks, or recomputations. Two parts of finalization currently assume that has not happened:

- The remaining commit/optimistic-reversion work can operate on the newly adopted transaction.
- The caller can proceed to effect application without returning through transition completion and parking logic.

This patch captures the batch identity before the first `commitPendingNodes()` and skips the remaining commit/reversion work if that identity changes. It also returns through the transition gate before applying effects when finalization leaves an active transition.

The regressions cover both entry points:

- A boundary check re-enters a suspended action; its writes and effects must remain held until the action completes.
- A public API reproduction using two async projections and a deep reader; the parked render must be released when the final refetch settles.

Both tests fail on current `next` and pass with this patch. The full repository tests and type checks, compiler suite, package integration checks, and TanStack Solid Query suite also pass.

### DIFF
--- a/.changeset/preserve-reentered-transitions.md
+++ b/.changeset/preserve-reentered-transitions.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Preserve pending transaction state when boundary checks re-enter a transaction during finalization.
--- a/packages/signals/src/core/scheduler.ts
+++ b/packages/signals/src/core/scheduler.ts
@@ -713,6 +713,12 @@ export class GlobalQueue extends Queue {
         }
       }
       clock++;
+      // Finalization can re-enter a pending transaction. Its effects must
+      // return through the transition gate before any apply runs.
+      if (activeTransition) {
+        scheduled = true;
+        return;
+      }
       // Check if finalization added items to the heap (from optimistic reversion)
       scheduled = dirtyQueue._max >= dirtyQueue._min;
       // Run lane effects first (for ready lanes), then regular effects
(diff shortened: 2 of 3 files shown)


ryansolid:
> Landed on `next` in [1c9e9e74](https://github.com/solidjs/solid/commit/1c9e9e74) with you as co-author — thank you for the precise diagnosis and especially the public-API reproduction (two projections + `deep()` reader), which is exactly the shape that hits real apps.
>
> What landed vs. the PR:
>
> - **Finalize guard — taken as-is in mechanism**, with one refinement: when `completingTransition` is non-null but the ambient batch was a separate object (the #2916 shape), adoption never touched the completing transaction's own containers, so it still settles them (optimistic reversion, affects release) instead of returning early.
> - **Flush guard — replaced with per-effect ownership.** Parking the whole flush also parks the effects of the write that *caused* the flush, which was computed mainline and already committed. In your second scenario that means `tick()` reads `1` while its own render stays at `0` until the unrelated transaction settles — a read/DOM split. The landed version stamps each effect's value with the transaction that computed it (`Effect._valueTransition`, added for #3322 in the commit just before) and, in a flush whose finalize entered a transaction, `runEffect` leaves runs owned by the still-held transaction queued for the next gate to park while applying everything computed mainline. Optimistic lanes are exempt by construction — they apply their own effects ahead of their transaction on purpose and never enter the ordinary queue.
>
> Both of your tests are in `finalize-reentry.test.ts` unchanged, plus two that distinguish the two approaches (they fail under the whole-flush park). Cost is ~+166 B minified in the core vs. ~+70 B for the coarse guard; we took the precision.
>
> Closing this in favour of the landed commit. It will ship in rc.8.
>
> — *Claude via Cursor*

Issue #3054, opened 2026-08-25
insert() should accept iterable collections, not only arrays


ryansolid:
Moved from [ryansolid/dom-expressions#294](https://github.com/ryansolid/dom-expressions/issues/294).

`insert` only walks arrays. A custom iterable of nodes (anything with `Symbol.iterator` that is not an array) is skipped with `Unrecognized value. Skipped inserting`. That shows up from non-JS producers (e.g. Squint) that hand JSX a collection rather than `Array.from(...)`.

This is still real. It was never taken because walking general iterables on the insert hot path has a cost, and `For` / other list helpers would need the same treatment to be consistent.

Leaving it here as an enhancement for 2.x rather than a 1.x fix.

nickshiro:
I've started working on the PR for this issue. JSX child normalization **only flattens arrays**, so non-array synchronous iterables reach the renderer as unknown objects instead of reconciliable child list. I’m tracing the DOM, universal, SSR, and hydration paths to keep their behavior aligned, while keeping <For>/mapArray out of scope since they intentionally depend on array indexing and identity diffing.

PR #3320, opened 2026-09-09
feat: support iterable JSX children

nickshiro:
## Summary

ref #3054 

Support finite synchronous iterables as JSX children.

 - Normalize iterable children into arrays before reconciliation.
 - Recursively flatten iterables nested inside arrays and other iterables.
 - Preserve reactive reconciliation semantics: nodes retained between iterable updates keep their identity.
 - Add SSR handling for iterable children, including nested iterables.
 - Add hydration coverage to ensure SSR-rendered iterable children hydrate without replacing existing DOM text nodes.
 - Add universal renderer coverage.

 - `pnpm --filter @solidjs/web exec vitest run test/runtime/insert.spec.js --reporter=dot`
  - `pnpm --filter @solidjs/web exec vitest run --config vite.config.server.mjs test/server/iterable-children.spec.tsx --reporter=dot`
  - `pnpm --filter @solidjs/web exec vitest run --config vite.config.hydrate.mjs test/hydration/iterable-children.spec.tsx --reporter=verbose`
  - `pnpm --filter @solidjs/universal exec vitest run test/runtime-insert.spec.js --reporter=dot`
  - `pnpm --filter @solidjs/signals types`
  - `pnpm --filter @solidjs/web types`

### DIFF
--- a/.changeset/iterable-children.md
+++ b/.changeset/iterable-children.md
@@ -1,0 +1,7 @@
+---
+"@solidjs/signals": patch
+"@solidjs/universal": patch
+"@solidjs/web": patch
+---
+
+Support finite synchronous iterables, such as `Set` and custom `Symbol.iterator` collections, as JSX children. Iterables are normalized into arrays for DOM and universal rendering, SSR, and hydration.
--- a/packages/signals/src/boundaries.ts
+++ b/packages/signals/src/boundaries.ts
@@ -590,6 +590,10 @@ export function createRevealOrder<T>(
   });
 }
 
+function isIterable(value: any): value is Iterable<any> {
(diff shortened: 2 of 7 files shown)


ryansolid:
> Quick note on timing so this isn't read as being ignored: this and #3342 are both compiler/runtime-behavior changes carried over from the dom-expressions backlog, and they need a real design-level review rather than a quick pass — the babel plugin and the native compiler have to move together, and the size scenarios will shift. Ryan is prioritizing rc release-stability fixes right now, so these are deliberately queued behind that work. No action needed on your side in the meantime.
>
> — *Claude via Cursor*

PR #3325, opened 2026-09-09
perf(web,signals): spread() reads a merge() proxy through its sources

ryansolid:
Rebased over #3326. What remains here is the one piece #3326 did not cover: reading a `merge()` proxy **through its flattened sources** rather than through the proxy.

A spread mixed with other attributes compiles to `spread(el, merge(statics, () => rest))`. On `next`, that source now costs one `ownKeys` trap — but for a merge proxy the trap is merge's `keys()`: a `Set` plus an own-enumerable scan (`Reflect.ownKeys` + `propertyIsEnumerable` filter) of every source; then every key's `get` is a right-to-left `in` walk of the sources. Iterating the sources directly removes all of it: own string keys per source, later sources overriding earlier (merge's own contract), `children`/`ref` excluded, each source enumerated through the same single-trap `ownKeys` helper `readShallow()` uses. `omit()` is not a merge and stays opaque (enumerated through its own filtering trap).

**Dropped from the original commit as subsumed by #3326:** the `style()` proxy branch (`style()` now receives a plain copy from `readShallow`), the `collectProps` per-source descriptor-trap fix (now the shared `ownKeys()`), and the `spread-style` bench (the `merge(static, reactive)` row of `spread-enumerate.bench.tsx` is the guard).

**Semantics pinned** in `spread-sources.spec.tsx`: later-wins, `children`/`ref`/symbols excluded, `omit()` inside a merge and alone, store key add/remove tracked, whole-object replace, own-properties-only. One documented contract: own keys per source — a key an earlier source owns and a later source merely *inherits* resolves to the earlier source (the proxy's `in` walk saw the inherited one); spread has always applied own properties only.

**Size:** ~+50 B on the app scenarios vs `next`; all limits hold.

**Perf** (Tier-1 `spread-enumerate`, 500 elements × 8 keys, median of 3 runs each, paired on the same machine):

| source | `next` (after #3326) | this PR | Δ |
|---|---|---|---|
| `merge(static, reactive)` | 602 ops/s | 1,243 ops/s | **+107%** |
| store record (untouched path) | 427 ops/s | 478 ops/s | +12% (drift; same code) |
| plain object (untouched path) | 1,462 ops/s | 1,599 ops/s | +9% (drift; same code) |

The merge row is the change; it now sits within ~25% of the plain-object floor. The two untouched rows moved together by ~10%, which is machine drift between the paired runs, not the patch — the merge row's 2× is far outside it.


ryansolid:
> Parking as draft: while writing the tests, `style={store.obj}` turned out to be identity-reactive only — `style()` (and `className()` for object values) enumerate the live object in the effect **callback**, so per-key reads are untracked (in-place store mutations don't apply; `STRICT_READ_UNTRACKED` fires in dev). Fixing that moves enumeration into the tracked half (a compute-side snapshot, as `spread()` already does), which is where this PR's `style()` change then belongs. Sequencing: the tracked-read fix first, with its cost model measured on the new Tier-1 bench, then this rebased on top. The spread half is unaffected.
>
> — *Claude via Cursor*

PR #3329, opened 2026-09-09
feat(signals): router-agnostic navigation attribution (OBSERVE.attribution.withOrigin)

ryansolid:
Stacked on #3324 (base: `observe-perf`); GitHub will retarget to `next` when that merges.

## Why

Sentry's Solid support today is shaped per router — `withSentryRouterRouting` for `@solidjs/router`, a separate integration for TanStack, the same pattern as React Router v3–v7 — because React has no framework-level notion of "a navigation". Solid 2 does, almost: a navigation is a plain write to the location, and the runtime already sees everything it costs (the hold behind route data, the re-runs, the blockers, the silence). The one thing it can't see is that the writes *were* a navigation, and to which route.

This PR adds that seam, in the engine's existing vocabulary, so the router-specific surface collapses to one call **inside the router** and every consumer (Sentry, devtools, `@solidjs/diagnostics`) reads navigations from the runtime with zero per-router code.

## The seam

```ts
OBSERVE
  ? OBSERVE.attribution.withOrigin(
      { kind: "navigation", name: match.pattern, to, from, params: match.params },
      () => setLocation(to)
    )
  : setLocation(to);
```

Twin of `withInteraction`: a declared frame whose writes are attributed to the ref, nested under the enclosing interaction when there is one (a link click), standing alone otherwise (a redirect). `OriginRef` is a discriminated union (`NavigationRef` today) so kinds can be added without the seam changing shape.

## What the engine does with it

- **Provenance**: `navigation` is a first-class `ChangeOrigin.kind` (`name`/`to`/`from`/`params`). Cause chains read `← signal "location" write … — navigation to /users/:id (/users/42) (under click on a.nav "Alice")`. A `navigate()` from an action step after a `yield` still inherits the click.
- **Holds named by route**: `HoldEvent.origin`; `SILENT_HOLD`/`LONG_HOLD` messages start from the actor — `click on a.nav (navigation to /users/:id) wrote "location" …`, or `navigation to /login wrote …` for a redirect — with `data.navigation` on the event.
- **`attribution.navigations()`**: one `NavigationEvent` per frame, settled exactly once:
  - `committed` — a plain drain took the writes (settle = end of that drain, the instant the screen had them);
  - `held` — they waited in a transition (settle = its commit; `hold` carries the `HoldEvent` when hold tracking recorded one; still settles with hold tracking off);
  - `superseded` — a later write replaced them before they landed (the user navigated again).
  
  A navigation whose write didn't survive the equality gate settles at once with `writes: 0`. `origin` on the event is the same object as `ChangeRecord.origin` / `HoldEvent.origin`, so records join by identity.
- **`feedback().navigations`**: per route pattern — count, `settledMs`/`worstMs`, `held`/`heldMs`/`silent`, `superseded`, `redirected`. The route-level table a router integration used to build itself.

## Late-bound on purpose (second commit)

Two things about the frame are read late, so the seam ships in its final shape rather than growing a second API:

- **The ref is re-read at settle.** The engine keeps the object passed to `withOrigin` and copies `name`/`to`/`params` from it again when the navigation settles (and when a hold on it is judged). A router whose match is coarse at write time — a lazy route subtree that resolves inside the hold — describes `/admin/*`, then assigns the exact pattern and params onto the same object once it knows them; the settled record, the hold's verdict and the feedback row all read the refined name. `from`/`at` are read once, at open. No `refine()` method.
- **A redirect is a hop, not a new navigation.** `NavigationRef.redirect: n` (`n >= 1`, the hop depth both routers already track — `_navigation` in `@solidjs/router`, the tx depth in TanStack) re-enters the pending navigation's frame instead of opening one. Its write replaces the pending one with the *same origin object*, so nothing is superseded; the record keeps the user's request time and interaction (so `settledMs` runs from the click, not the hop); the abandoned destination moves to `NavigationEvent.redirects`; the hold joins by identity as before. `formatOrigin` reads `navigation to /login (redirected from /users/42)`, and a chain reads `redirected from /users/42 → /login`. With nothing pending to fold onto, a `redirect` frame opens a navigation of its own.

### Hold census fix (found by the router-shaped test)

Memos compute eagerly. A router's internal `createMemo(() => isPending(location))` — Solid Router 2.0's `routingPending` has this shape — subscribes to the pending companion whether or not the app ever renders the memo, and the census counted any subscriber as "the screen acknowledged". So every Solid Router navigation would have read as acknowledged, never `SILENT_HOLD`, even with no routing indicator on screen. The census now requires a companion to **reach an effect**, through however many memos: a memo alone is not the screen. Two tests pin both directions (router memo unrendered → `SILENT_HOLD`, `latestOnly: 0`; the same memo read by a render effect → `isPending:location` acknowledged).

## Core changes

- `withOrigin` + `originStart`/`originEnd` hooks (`attribution-hooks.ts`), exposed as `OBSERVE.attribution.withOrigin`.
- **One new hook, `flushEnd`**, fired once per `flush()` drain (after the loop, outside every `try`) so the engine has the "committed and effects ran, or parked" instant for writes no transition ever held — transitions are created lazily only when async is hit, so instant navigations never produced a settle signal before. Cost: one null check per drain in observe; the site and its `drained` flag fold out in prod.
- `formatOrigin` renders `navigation to <pattern> (<to>)`.

## Verification

- **Prod dist byte-identical** for every `dist/prod/*.js` except `attribution.js` (the inert twin gained `navigations()` and the `navigations: []` table).
- New `tests/attribution-navigation.test.ts` (21 tests): stamping under an interaction, standalone, through an action step; no-engine passthrough; settle paths — committed at drain end, zero-write immediate, drain inside the frame, redirect inside an effect during a drain, held with hold attached and route-named verdict, redirect-named verdict, held with hold tracking off, superseded, feedback fold, reset on disable/enable; lazy-match refinement during a hold; redirect folding (held chain with click-relative timing and chain-named verdict, successive hops, synchronous nested hop, no-pending fallback); router-shaped census in both directions.
- `dist-artifacts.test.ts` drives a navigation end to end through the built observe artifacts and asserts `flushEnd` reached the engine.
- Full suites green: signals 1646, solid 594, web 712, diagnostics 31; `tsc -p tsconfig.build.json` clean.
- Size: prod scenarios unchanged; observe tier +30 B (14.34 → 14.40 KB limit, did not move in the second commit), observe + engine +0.78 KB then +0.26 KB (24.08 → 25.20 KB limit), all annotated in `.size-limit.js`.

## What the routers need to do

Match eagerly, then wrap the location write in the call above (both `@solidjs/router` and TanStack Solid Router can match synchronously); pass `redirect: depth` on the hop when a guard or loader redirects; assign the exact pattern onto the ref once a lazy subtree resolves. That is the whole contract. Server side, the same label wants to land on the request event so the server transaction is named by the router — that folds into the trace-context item (C1), not this PR.

## Not in this PR

- `@solidjs/diagnostics` artifact/browser bridge don't yet include `navigations` (they also omit `waterfalls`); follow-up.
- `RerunEvent` carries no navigation field — cause records already carry the origin.

Co-authored-by: Claude via Cursor

### DIFF
--- a/.changeset/attribution-navigation-origin.md
+++ b/.changeset/attribution-navigation-origin.md
@@ -1,0 +1,14 @@
+---
+"@solidjs/signals": patch
+---
+
+Router-agnostic navigation attribution: `OBSERVE.attribution.withOrigin({ kind: "navigation", name, to, from, params }, fn)`
+
+A navigation in Solid 2 is a plain write to the location; the runtime already sees everything it costs (the hold behind route data, the re-runs, the silence) but not that the writes were a navigation, or to which route. `withOrigin` is the seam where a router says so, around its write — the one router-specific line, living in the router. From it the attribution engine:
+
+- stamps the writes with a `navigation` origin (new `ChangeOrigin.kind`, with `name`/`to`/`from`/`params`), nested under the enclosing interaction — including through an action step — so cause chains read `— navigation to /users/:id (under click on a.nav "Alice")`;
+- names holds by route: `HoldEvent.origin`, `SILENT_HOLD`/`LONG_HOLD` messages that start from the navigation, and `data.navigation` on the event;
+- keeps one `NavigationEvent` per frame (`attribution.navigations()`), settled exactly once as `committed` (a plain drain took the writes), `held` (with the `HoldEvent` attached), or `superseded` (a later write replaced them before they landed);
+- folds settled navigations per route into `feedback().navigations`.
+
+One new core hook, `flushEnd`, fires once per `flush()` drain so the engine has the "committed and effects ran" instant for writes no transition held. Prod builds are unchanged (the hook site folds out; only the inert attribution twin gained the new empty queries).
--- a/.changeset/observe-node-shapes.md
+++ b/.changeset/observe-node-shapes.md
@@ -1,0 +1,29 @@
+---
+"@solidjs/signals": patch
+---
+
+Observe tier: no post-construction fields on reactive nodes.
+
+The observe build stamped `_name` on every node, `_owner` on `createSignal`
+nodes and live `_subCount`/`_depCount` edge counters on linked nodes after the
+node literal — exactly the hidden-class transitions the prod literals are
+shaped to avoid. Measured against prod on the reactivity benchmark the tier
+cost +15% overall with creation tests 2–3× under polymorphic load.
+
+- Node factories (`computed`, `createEffectNode`, `signal`, `slotSignal`,
+  `createOwner`) now have two literals selected at build time: prod, and
+  observe = prod plus its `_name` slot (`_owner` too on signals). Default
+  labels (`signal`, `computed`, `effect`, `trackedEffect`) come from the
+  literal, so the `createEffect`/`createRenderEffect`/`createTrackedEffect`
+  wrappers no longer spread a fresh options object per effect to inject one.
+  A dist test pins observe's key set to prod's plus the slots.
+- Edge counters are gone. `HUGE_FAN_OUT` is counted by the notify walk a
+  committed change already makes over its subscribers, `HUGE_FAN_IN` by the
+  recompute pass over the sources it tracks (one module counter in `link()`).
+  Both therefore fire on the work — the change / the recompute — rather than
+  on the link, once per node and again after +500 growth (a WeakMap, not a
+  node field). `WIDE_WRITE` (engine) counts the subscriber list on the write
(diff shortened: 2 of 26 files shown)


ryansolid:
> Closed by GitHub when #3324 squash-merged and `observe-perf` was deleted; a closed PR can't be retargeted. Rebased onto `next` and reopened as #3332 with the same two commits.
>
> — *Claude via Cursor*

Issue #3055, opened 2026-08-25
Merge adjacent sibling templates in a fragment into one template


ryansolid:
Moved from [ryansolid/dom-expressions#458](https://github.com/ryansolid/dom-expressions/issues/458).

A fragment of adjacent static siblings compiles to one `template()` per root:

```jsx
<>
  <div />
  <button type="button">hi</button>
</>
```

```js
var _tmpl$ = _$template(`<div>`),
  _tmpl$2 = _$template(`<button type=button>hi`);
```

Those could be a single template string (and one clone) when nothing dynamic sits between them. Reported by @milomg; playground: https://playground.solidjs.com/anonymous/6f9638af-453b-4841-b8cd-76ea1c4fc15a

nickshiro:
I’d like to work on this. I’ll start by adding a compiler fixture and investigating how to merge adjacent DOM templates while preserving Babel/Oxc parity and hydration behavior.

PR #3342, opened 2026-09-10
perf(web,babel-plugin,compiler): merge adjacent sibling templates

nickshiro:
## Summary

ref #3055 

Merge adjacent fully static DOM elements in a JSX fragment into a single template.

Previously:

  ```jsx
  <>
    <div>First</div>
    <div>Last</div>
  </>
```

  created and cloned two independent templates. It now compiles to a single multi-root template:

```jsx
  const template = _$template(
    `<div>First</div><div>Last</div>`,
    4
  );

  const view = template();
  ```
  
  ## Implementation

  - Detect maximal runs of adjacent, fully static DOM roots during fragment lowering.
  - Combine those roots into one template while explicitly closing every non-void root.
  - Preserve the external fragment shape by returning the cloned roots as a Node[].
  - Add bit 4 to the internal template() flag:
      - 1: use document.importNode
      - 2: wrapped namespace template
      - 4: multi-root template

  - Flags remain composable, so 5 represents a multi-root template that requires importNode.
  - Keep the existing single-root cloning path unchanged and select the cloning strategy when the factory is created rather than branching for every clone.
  - Implement the same lowering in both the Babel plugin and the native Oxc compiler.
  - Support both generate: "dom" and generate: "dynamic" when the selected renderer is DOM.

  The optimization is deliberately limited to safe static runs. Expressions, components, meaningful text, dynamic attributes, effects, refs, spreads, and wrapped SVG/MathML partials break the group. Hydratable
  output is unchanged.
  
  Added coverage for:

  - adjacent static fragment roots;
  - groups separated by an expression;
  - dynamic roots remaining independent;
  - adjacent void elements;
  - combined importNode and multi-root flags;
  - preservation of the fragment array shape;
  - cloning each root on every factory call;
  - HTML and SVG namespace preservation;
  - Babel/Oxc output parity.

  - Babel plugin: 255 tests passed
  - Native compiler: 5360 tests passed, 28 skipped
  - Rust feature matrix: 141 tests passed
  - Web client: 738 tests passed
  - Web server: 785 tests passed, 2 skipped
  - Web hydration: 168 tests passed
  - Web runtime through both Babel and Oxc: 9 tests passed in each mode
  - Babel typecheck passed
  - Web build and declaration generation passed
  - Full workspace build passed

### DIFF
--- a/.changeset/merge-sibling-templates.md
+++ b/.changeset/merge-sibling-templates.md
@@ -1,0 +1,7 @@
+---
+"@solidjs/babel-plugin": patch
+"@solidjs/compiler": patch
+"@solidjs/web": patch
+---
+
+Merge adjacent fully static DOM roots in JSX fragments into a single template clone.
--- a/packages/babel-plugin/src/dom/template.ts
+++ b/packages/babel-plugin/src/dom/template.ts
@@ -83,7 +83,10 @@ export function appendTemplates(path: NodePath<t.Program>, templates: TemplateRe
       raw: escapeStringForTemplate(templateText)
     };
 
-    const flag = template.isWrapped ? 2 : template.isImportNode ? 1 : null;
+    const flag =
+      (template.isImportNode ? 1 : 0) |
+      (template.isWrapped ? 2 : 0) |
+      (template.isMultiRoot ? 4 : 0);
 
     return t.variableDeclarator(
       template.id,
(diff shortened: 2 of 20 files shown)


ryansolid:
> Quick note on timing so this isn't read as being ignored: this and #3320 are both compiler/runtime-behavior changes carried over from the dom-expressions backlog, and they need a real design-level review rather than a quick pass — the babel plugin and the native compiler have to move together, and the size scenarios will shift. Ryan is prioritizing rc release-stability fixes right now, so these are deliberately queued behind that work. No action needed on your side in the meantime.
>
> — *Claude via Cursor*

PR #3343, opened 2026-09-10
chore: stale gitignore entries

nickshiro:
## Summary

Clean up the root `.gitignore` so it only contains entries that are relevant to working with this repository.

- Remove stale declaration and test artifact paths left behind by package restructuring.
  - Remove personal editor, OS, and devcontainer entries.
  - Keep generated build outputs, dependency directories, coverage files, and test caches ignored.
  - Preserve the existing Cursor distinction: personal `.cursor/*` files remain ignored, while shared `.cursor/rules/` stay versioned as project-level contributor guidance.

Editor- and OS-specific files such as .vscode/, .idea/, and .DS_Store are better handled by each contributor's global Git ignore configuration, since they reflect personal tooling rather than the project itself. I use Neovim, which generates its own artifacts - swap files, backups, undo history - and Cursor adds more too (.cursor/*), though .cursor/rules/ is kept since it's project convention, not personal tooling. Multiply the personal-tooling artifacts by every editor a contributor might use, and baking them all into the repo's .gitignore simply doesn't scale - it's a per-machine problem, better solved once, globally, than repeatedly, per project.

 - `git diff --check`.
  - Verified current build, type, coverage, dependency, and test-cache paths remain ignored.
  - Verified project Cursor files remain ignored while `.cursor/rules/` remains available to Git,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,17 +16,7 @@ packages/babel-plugin/index.js
 .pnpm-store/
 .types-build/
 
-packages/solid/src/jsx.d.ts
-packages/solid-web/src/jsx.d.ts
-packages/solid-web/src/jsx-properties.d.ts
-packages/solid-h/jsx-runtime/src/jsx.d.ts
-packages/solid-h/jsx-runtime/src/jsx-properties.d.ts
-packages/solid/web/server/server.d.ts
-packages/solid/web/src/client.d.ts
-packages/solid/universal/src/universal.d.ts
-
 # vscode devcontainers see issues #419 and #421 for context
 /.devcontainer.json
 /devcontainer
-packages/solid-web/test/harness/__artifacts__/
 packages/test-integration/tests/fixtures/downloaded/
--- a/.gitignore
+++ b/.gitignore
@@ -3,20 +3,14 @@ dist/
 lib/
 # ...but an example's source tree may legitimately have one (src/lib/api.ts).
 !examples/*/src/lib/
-.vscode/
 .cursor/*
 !.cursor/rules/
-.idea/
 coverage/
 types/
 types-cjs/
 packages/babel-plugin/index.js
-.DS_Store
 .turbo/
 .pnpm-store/
 .types-build/
 
-# vscode devcontainers see issues #419 and #421 for context
-/.devcontainer.json
-/devcontainer
 packages/test-integration/tests/fixtures/downloaded/


ryansolid (changes requested):
> Thanks — the dead-path entries are right to go: none of `packages/solid/src/jsx.d.ts`, the `packages/solid-web/*` / `packages/solid-h/*` paths, `packages/solid/web/**`, or `packages/solid/universal/**` exist in the tree anymore.
>
> Three of the removed lines aren't stale, though, and dropping them would regress contributor hygiene:
>
> - `.DS_Store` — macOS drops these anywhere Finder touches; there is one in the repo root on a maintainer checkout right now, and without the ignore it shows up as untracked in every `git status`.
> - `.vscode/` and `.idea/` — editor workspace dirs. They're not present because they're ignored; un-ignoring them invites accidental settings commits from contributors.
>
> The devcontainer lines (`/.devcontainer.json`, `/devcontainer`) are borderline — they exist for contributors who add a local devcontainer (see #419/#421) — I'd leave them too, but won't block on it.
>
> Please restore `.DS_Store`, `.vscode/`, `.idea/` and this is good to merge. (The red Size check is inherited from `next` at the time you branched, fixed since; it'll clear when the branch is updated.)
>
> — *Claude via Cursor*

nickshiro:
@ryansolid, a stray `.DS_Store` on one checkout means that one machine's global excludesfile isn't set up - that's a one-time fix (`git config --global core.excludesfile ~/.gitignore_global`), not a reason to carry the rule in every clone of the repo. By your own logic, this file should never have needed the cleanup PR in the first place - the dead .d.ts paths sat there unnoticed for who knows how long because nobody's pruning a list that only grows. .DS_Store/.vscode/.idea are the same mechanism: today it's three lines, next year it's Sublime, Zed, Fleet, whatever hits after that - each one a permanent tax on every contributor's .gitignore, versus a git config that costs you ten seconds once. The dead paths prove the growth problem is real, not hypothetical.

ryansolid:
> Merged (ebe77564) — thanks for pruning the dead paths; those removals stand.
>
> On the editor/OS entries we'll keep them in the repo's `.gitignore`: a follow-up on `next` (344ed054) restores `.DS_Store`, `.vscode/`, `.idea/` and the devcontainer lines, with the policy stated inline. We carried the same rules on the 1.x branch and the maintainers' call is that a stray artifact should never surface as untracked or land in a commit regardless of whether a contributor has a global excludesfile set up. The growth concern is fair, but the list is bounded to the handful of artifacts that actually show up in practice, not one line per editor.
>
> — *Claude via Cursor*

Issue #3239, opened 2026-09-03
The flash cookie carries the submission with no `SameSite`, no lifetime and no `__Host-` prefix


frenzzy:
## Summary

The cookie that carries a no-JS server-function outcome back to the browser is written with `HttpOnly` and `Secure` and nothing else: no `SameSite`, no `Max-Age`/`Expires`, and no `__Host-` prefix. What rides it is not a status line, it is the **submission** — a no-JS login form flashes `input` verbatim, so the user's password sits in the cookie value as plaintext JSON (the codec is deliberate about the payload being plain JSON, flash.ts:17). With no lifetime that is a session cookie at `Path=/`: any submission whose outcome is never read is re-sent on every subsequent request to the origin — scripts, images, fonts — and lands in access and CDN logs, for as long as the browser lives. With no `__Host-` prefix a sibling subdomain can set a `Domain`-scoped `flash` of its own; the Cookie header is one string, `parseCookieHeader` is last-wins, and the render shows the attacker's "result" as the outcome of the user's submission.

**This issue merges four symptoms that would otherwise be filed separately**, because the fix is one hunk region and the pieces cannot be moved independently (see Options): *flash cookie has no `SameSite`*; *flash cookie never expires / the submission rides every request*; *flash cookie is not `__Host-` / a sibling subdomain can shadow it*; *`clearFlashCookie`'s attribute set disagrees with the write, and a `__Host-` deletion cookie without `Secure` is silently rejected exactly like the cookie it meant to delete (the #3138 shape)*.

Affects `@solidjs/web@2.0.0-rc.6`.

## Reproduction

The two files are copied out of the repo at `f0f7531b`, unmodified except for one import specifier so node can run them with type stripping (the in-repo `dist/` is already patched, so it cannot be used to show the released behaviour):

```sh
git clone https://github.com/solidjs/solid && cd solid && git checkout f0f7531b
mkdir -p /tmp/r/packages/web/src /tmp/r/packages/web/server-functions/src
cp packages/web/src/cookies.ts /tmp/r/packages/web/src/
cp packages/web/server-functions/src/flash.ts /tmp/r/packages/web/server-functions/src/
sed -i'' -e 's#"../../src/cookies.js"#"../../src/cookies.ts"#' \
  /tmp/r/packages/web/server-functions/src/flash.ts
```

```js
// /tmp/r/repro.mjs — node --experimental-strip-types repro.mjs
[...]

PR #3359, opened 2026-09-11
fix(web): host-lock the flash cookie

lxsmnsyc:
The flash cookie carries the outcome of a server function call made without the client runtime, and the next render shows it as if the server had produced it. Nothing authenticates the payload, and signing it would need a deployment secret this runtime has no channel for.

- Rename the cookie to `__Host-flash`. Browsers refuse that name unless the cookie is `Secure`, `Path=/`, and free of `Domain`. `Domain` is what let any sibling subdomain set the cookie and forge a successful outcome.
- Set `path`, `secure`, `httpOnly`, and `sameSite` explicitly at the call site instead of relying on defaults.
- Make `clearFlashCookie()` meet the same rules, so the browser does not refuse the deletion and leave the outcome on every later request.
- Check the decoded payload on its shape instead of on whether the result is truthy. A call that returned `false`, `0`, `""`, or `null` decoded as if nothing had been submitted, which makes a user retry a mutation that already committed.

Tested with a new spec covering the cookie attributes, the deletion, falsy results, and payloads the encoder did not write.


lxsmnsyc:
Closing due to #3239

PR #3361, opened 2026-09-11
fix(web): stretch the flash key from the deployment secret

lxsmnsyc:
A captured flash cookie is an offline oracle for `secret`, and the option accepts any non-empty string. Under a single SHA-256 each guess cost one hash, so a short or human-chosen secret fell quickly. Recovering it means reading every flash payload and forging new ones.

- Derive the key with PBKDF2-HMAC-SHA-256 at 100,000 iterations, salted with the same domain string as before. The cap is the ceiling some edge runtimes enforce on PBKDF2.
- Document the entropy requirement on the `secret` option and show how to generate a value. Stretching raises the price of a guess. It does not make a guessable secret safe.

The derivation runs once per process per secret and only on the no-JS form path, so ordinary requests are unaffected. The derived key changes, so in-flight flashes read as no flash, the same as a secret rotation.

Tested with a spec that attacks the cookie the public API produces: the old bare digest must not open it, PBKDF2 with the documented parameters must, and neither a lower iteration count nor another domain salt does.

### DIFF
--- a/.changeset/flash-key-derivation.md
+++ b/.changeset/flash-key-derivation.md
@@ -1,0 +1,9 @@
+---
+"@solidjs/web": patch
+---
+
+The flash cookie key is now derived with PBKDF2 instead of a single SHA-256 hash.
+
+- Guessing a weak `secret` from a captured cookie is now 100,000 times more expensive.
+- Set `secret` to a high-entropy value of 32 bytes or more. The option docs show how to generate one.
+- Flash cookies in flight when you deploy read as no flash, the same as a secret rotation.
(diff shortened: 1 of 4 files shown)


ryansolid:
> Pushed 420433ee to fix the red `@solidjs/web#test-types` check: `opens(packed: Uint8Array, …)` made `packed.subarray(…)` a `Uint8Array<ArrayBufferLike>`, which `crypto.subtle.decrypt` rejects as a `BufferSource` under the current lib types. The bytes come from `new Uint8Array(n)`, so `Uint8Array<ArrayBuffer>` is the true type. No runtime change.
>
> The derivation itself looks right: PBKDF2 over the raw secret, domain string as salt, 100k cap for the edge-runtime ceiling, cached per process and only paid on the no-JS path. The spec attacking the produced cookie from both sides is a good pin.
>
> — *Claude via Cursor*

PR #3362, opened 2026-09-11
docs: add skills

lxsmnsyc:
Three skills under `.claude/skills/`, each a self-contained SKILL.md that an agent picks up on this checkout and that can be copied into another project. They cover usage patterns for people writing Solid apps, not workflows for working on this repository.

- `solid-reactivity`: the primitive signatures that changed from Solid 1.x. `createEffect` takes a compute and an effect, a function argument to `createSignal` or `createStore` makes it derived, async is a computation that throws while pending, and the 1.x names that are gone map to their replacements.
- `solid-control-flow`: Show, For, Repeat, Switch, Match, Loading, Errored, Dynamic, Portal, and Reveal, with the callback-child and keyed semantics that decide what remounts.
- `solid-server-functions`: the closure rule the compiler enforces, why the boundary is not authentication, what declaring GET gives up, error sanitization, and the deployment secret the no-JS form path needs.

Every claim is taken from the source in this repository rather than from Solid 1.x, including the removed-API map in `packages/solid/src/index.ts`.

### DIFF
--- a/.claude/skills/solid-control-flow/SKILL.md
+++ b/.claude/skills/solid-control-flow/SKILL.md
@@ -1,0 +1,147 @@
+---
+name: solid-control-flow
+description: How to use Solid's control-flow components in JSX. Use when writing or reviewing markup that needs conditionals, lists, error handling, or async fallbacks, or that reaches for Show, For, Repeat, Switch, Match, Dynamic, Loading, Errored, Reveal, or Portal. Covers the callback-child and keyed semantics that decide whether content remounts.
+---
+
+# Solid control flow
+
+Never use a ternary or `.map()` where one of these fits. They exist so the
+runtime can keep DOM across updates instead of rebuilding it.
+
+## Show
+
+Renders children when `when` is truthy, otherwise `fallback`.
+
+```tsx
+<Show when={user()} fallback={<SignIn />}>
+  <Greeting />
+</Show>
+```
+
+A function child receives the narrowed value, and `keyed` decides its shape
+and its remount behavior.
+
+```tsx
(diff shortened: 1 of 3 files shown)


nickshiro:
Coming from the PR discussion #3343  - I think this hits the same issue, just with higher stakes.

`.claude/skills/` is vendor-specific tooling integration, not project content, the same way `.vscode/` or `.cursor/*` are — except it's worse, because it's not something you can ignore and forget. It's living content that has to stay in sync with the actual API, duplicating what should already live in the docs, just reformatted for one vendor's schema. 

**And it won't stop at one.** Cursor has its own rules format, Copilot has `.github/copilot-instructions.md`, and more will show up. If this repo takes `SKILL.md`, there's no principled way to say no to the next vendor's format - and in a year there's a shelf of overlapping, vendor-specific files that all encode the same knowledge about the same API, none of which the maintainers necessarily have the tooling expertise to review for correctness.

If the underlying knowledge (signatures, control-flow semantics, server-function rules) is accurate, it belongs in the project's actual docs - vendor-neutral, human-readable, maintained by people who know the internals. Turning that into a `SKILL.md`, `.cursorrules`, or anything else is a mechanical format conversion that doesn't need core-repo review or maintenance - it's a good fit for a separate community project (or something the vendor generates itself from existing docs), not something baked into solidjs/solid.

nickshiro:
Speaking of my understanding of the field - I use Claude Code, too.

lxsmnsyc:
@nickshiro no worries, this is still a draft and isn't final yet. I'm still exploring the recipes before moving this to plugin form. Take note that the skills were meant to be plugins to be used by users, not a guide for the repo structure.

PR #3363, opened 2026-09-11
fix(compiler): name server functions by their binding path

lxsmnsyc:
Server-function ids used the nearest binding name alone, so two functions with the same name in one file were told apart by a positional ordinal assigned in traversal order. Appending a third same-named function renumbered the other two and re-pointed addresses a deployed client was still holding, which is the mis-dispatch #3109 set out to remove.

- The descriptive name is now the function's dotted binding path, so `makeA.submit` and `makeB.submit` are two names instead of one name and a counter.
- Object property keys, class names, and class member keys contribute path segments too, so `handlers.save` and `Api.refresh` name themselves.
- A key that is not identifier-shaped contributes no segment, keeping `-` and `.` out of the name so `id.split("-")[1]` stays the file hash.
- The ordinal remains only for two functions that share one path, such as two inline callbacks in the same scope.

Ids move once for nested server functions. The frozen Babel reference files for the three affected fixtures were edited by hand, id literals only.

### DIFF
--- a/.changeset/server-function-path-names.md
+++ b/.changeset/server-function-path-names.md
@@ -1,0 +1,9 @@
+---
+"@solidjs/compiler": patch
+---
+
+Server-function ids now name a function by its binding path, so two same-named functions no longer share one id.
+
+- `makeA`'s `submit` becomes `makeA.submit-<hash>` instead of `submit-<hash>`.
+- Adding a same-named function no longer re-points the ids of the existing ones.
+- Ids for functions in objects and classes pick up those names too, such as `handlers.save`.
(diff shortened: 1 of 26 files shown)


ryansolid:
> Reviewed and pushed one commit on top (0e665f2d), with Ryan's sign-off on the approach.
>
> The path rule drops a function's own name whenever an enclosing path exists, which put one shape back on the positional ordinal the PR removes:
>
> > ```ts
> > export function wire() {
> >   register(function saveHandler() { "use server"; return 1; });
> >   register(function dropHandler() { "use server"; return 2; });
> > }
> > // next:  saveHandler-<hash>  dropHandler-<hash>
> > // PR:    wire-<hash>         wire-<hash>-1
> > ```
>
> Nested function declarations had the same gap: the bubbler skips them, so they contributed no segment and same-named arrows in sibling declarations collided on `outer.s`.
>
> The commit makes a named function a named container like the others. Its name is a segment, skipped only when it repeats the segment already on the path (`const submit = function submit() {}`, bubbled `const makeA = function makeA() {}`). Results: `wire.saveHandler`, `outer.a.s` / `outer.b.s`, and the `object-property` fixture moves once more to `handlers.save.saveRecord` (frozen reference hand-edited, id literals only). Tests added to `directives-id-scheme.test.js` for each case. Full compiler suite green locally.
>
> Since nested ids move once with this PR regardless, landing the complete rule here avoids moving them twice.
>
> — *Claude via Cursor*

Issue #3181, opened 2026-09-01
2.0.0-rc.4 | async createProjection commit is not seen through a memo over the projection — leaf reader behind `createMemo(() => store.value)` is never notified again


mgarcialeniolabs:
### Describe the bug

An async `createProjection` whose settle is announced by a signal write in the **same synchronous
step** in which its promise resolves commits correctly — the DOM and direct leaf readers both see
the new value — but a leaf reader that reaches the same leaf **through a `createMemo(() =>
store.value)` indirection** is not notified, and is never notified again. Its last observation stays
the superseded value permanently.

The memo itself is behaving correctly by value: the projection reconciles in place, so
`store.value` keeps a stable identity and the memo has no reason to recompute. The problem is that
the propagation appears to stop at the memo — the downstream effect's *own* subscription to the
changed leaf is not re-evaluated on that flush either.

Found while narrowing TanStack/query#11351 (`@tanstack/solid-query` 6.0.0-rc.1, whose `data` is a
single `createProjection` node). The adapter's derive is provably correct: it runs exactly the
expected number of times, the commit lands, and an untracked read through the very same memo
returns the new value. Only the notification is lost.

### Your minimal, reproducible example

A vitest file with no TanStack code — only `solid-js` and `@solidjs/testing-library`:

```tsx
import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
import {
  Loading,
  createEffect,
  createMemo,
  createProjection,
  createSignal,
} from 'solid-js'
import { render } from '@solidjs/testing-library'

describe('async projection commit is not seen through a memo', () => {
  beforeEach(() => vi.useFakeTimers())
  afterEach(() => vi.useRealTimers())

  it('notifies a leaf reader behind a memo', async () => {
    const throughMemo: Array<boolean> = []
    const direct: Array<boolean> = []

    let committed: { a: boolean } | undefined
    let inFlight: Promise<{ a: boolean }> | null = null
    const [version, setVersion] = createSignal(0)

    // Models a cache-backed fetch: on settle, the "cache" is written and
    // subscribers are notified in the SAME synchronous step in which the
    // promise resolves.
    const fetchNow = (next: { a: boolean }) => {
      inFlight = new Promise((resolve) => {
        setTimeout(() => {
          committed = next
          inFlight = null
          setVersion((v) => v + 1)
          resolve(next)
        }, 10)
      })
    }

    function Probe() {
      const store = createProjection<{ value: { a: boolean } }>(
[...]

PR #3371, opened 2026-09-11
fix(signals): wake blocked readers after equal-value pending recovery

DerpyCrabs:
A conditional memo can stop reading a pending dependency and recover to its previous value:

```ts
const enabled = createMemo(() => !disabled() && asyncData.enabled);
```

If a dependent tries to read while this memo is pending, setting `disabled(true)` should unblock it—even if `enabled` was already `false` before loading. Currently, equality suppresses notification and the dependent can remain stuck pending.

In our app, this leaves list rows missing after a successful fetch: their rendering is blocked on a conditional event-handler dependency that is no longer needed.

Reuse the pending-settlement walk to clear inherited pending state on equal-value recovery. Retry blocked readers without rerunning unchanged dependents or clearing pending state still required through another dependency path.

Seven regression cases pass across all three build modes; four fail on unpatched `next`.

### DIFF
--- a/.changeset/settle-unchanged-pending-branches.md
+++ b/.changeset/settle-unchanged-pending-branches.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Wake blocked readers when a conditional drops pending dependencies without changing its result, while preserving pending state from other dependency paths.
--- a/packages/signals/src/core/async.ts
+++ b/packages/signals/src/core/async.ts
@@ -69,7 +69,7 @@ function clearPendingSources(el: Computed<any>): void {
 // SOURCE's subscribers, so it is retryable iff a tracked read created that
 // edge: a dep that IS the source, or one whose own pending chain carries it
 // (pending sources propagate the origin node, so this covers any depth).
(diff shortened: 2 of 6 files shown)


ryansolid:
> Thanks — the bug and the fix shape both check out: the recovered node acts as the settling source for the subtree it orphaned, riding the existing `settlePendingSource` walk, so the `_blocked` reader is woken without re-running non-blocked dependents. (For the record: status already clears when the dropped source lands; the actual defect was that the parked reader was never re-enqueued.)
>
> I pushed one commit on top (fdef2fb3):
>
> - The recovery loop now skips `source === el`. When the outgoing set holds the node's own superseded flight, the #3181 `wasPendingSource` sweep right below already runs `settlePendingSource(el)`, so the loop was walking the dependents twice for that case.
> - A comment on the `retryReaches` early return in the settle closure explaining why it deliberately does not mark the node visited (converging unchanged branches need the revisit).
> - Treeshake note updated to the measured 22,648 B; your 22,700 bound stands.
>
> Local: signals dev suite 1751 passed, prod tier at the 238-failure baseline, solid/web green. Merging once CI agrees.
>
> — *Claude via Cursor*

PR #3380, opened 2026-09-12
fix(signals): OBSERVE.exclude covers store nodes and interaction writes

ryansolid:
## What

`OBSERVE.exclude(owner)` now covers writes and stores, not only runs and diagnostics:

- **Store nodes are excluded subjects.** `registerGraph(proxy, owner)` never landed on a store: the proxy's set trap swallows writes outside a draft, so `isExcluded(storeNode)` was always `false`. The root target now keys its creating owner (a WeakMap, one set per `createStore`) and the store's nodes (value, presence, key-set, deep) copy it into `_owner` as they are created — under the same `attrHooks !== null` gate as node naming, so node creation's disabled cost stays one null check. `registerGraph` still runs for the dev tier's owner signal lists / `onGraph`.
- **Excluded writes don't count toward an interaction.** `stampWrite` routes a root write to an excluded subject into `InteractionState.excludedWrites` instead of `event.writes`.
- **An interaction that only touched the observer is forgotten.** `writes 0, runs 0, created 0, excludedWrites > 0` at settle → removed from the ring, never delivered on `subscribe("interaction")`. An idle click that wrote nothing at all is still recorded (that's a fact about the app).

## Why

Found by the Sentry spike's record-driven adapter (rewritten on the rc.8 `subscribe(type)` channel): clicking the panel's own "clear" button produced a `ui.interaction.click` transaction with `writes 1, runs 0`, even with the panel root excluded and the write made under `runWithOwner(panelOwner)`. The documented contract is that the engine records nothing about the observer's own subtree; this closes the two paths that leaked.

Adapter-side consequence, now in the doc paragraph: the store must be *created* under the excluded owner, not only written under it.

## Size

Engine scenario 26888 → 26944 B (+56 B), limit 26.92 → 26.98 KB with a note. The observe CSR scenario did not move (no observe scenario bundles stores); prod tiers are `__OBSERVE__`-gated and unchanged.

## Tests

- `observe-exclude.test.ts`: "forgets an interaction whose only writes went to the panel's own store" — fails without the change (record present, `outcome: idle`), passes with it; also asserts a click that writes the app *and* the panel is recorded with `writes 1`.
- signals 1752 / solid-js 595 / web 734 pass.

Companion router fix from the same spike run: solidjs/solid-router#604 (guard redirect in the landing flush recorded as a hop, `RouterIntegration.inflight`).


ryansolid:
> Rebased onto `next` and pushed f880165f on top: the CodSpeed regression on `fresh stores: create + first write + first commit` was real — a `WeakMap.set` per fresh store is a growing ephemeron table, ~+30% on that shape locally (1.08 → 1.45 ms; `next` ≈ gated fix at 1.10 ms). The owner record is now gated on `attrHooks !== null`, the same boundary the node stamping already draws, so the disabled cost is one null check. Consequence noted in the doc paragraph and the `storeOwners` comment: a store created before `enable()` has no recorded owner and its nodes are never excluded subjects — a panel enables first. Size note rebased (engine scenario 27090 B, cap 27.16 KB).
>
> — *Claude via Cursor*

Issue #3376, opened 2026-09-11
2.0.0-rc isPending briefly returns false during overlapping updates to an async memo chain


GabbeV:
### Describe the bug

Updating a signal twice while its async memo chain is pending makes `isPending` report `true → false → true → false`.

Both writes happen before the first false, and no further writes happen when it becomes true again.

### Your Example Website or App

https://s.olid.uk/id/_vddZGrFTleyH3BBtMLRbg

### Steps to Reproduce the Bug or Issue

1. Wait for the initial load.
2. Click Run once.
3. Watch Pending briefly become false before becoming true again.

### Expected behavior

Pending should stay true until the replacement update finishes.

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

The same case with longer delays was verified on next at 344ed054 and #3337 at 20f31433.

Replacing the two async memos with one async memo of the same total duration removes the false interval.

ryansolid:
> Reproduced and fixed in #3381 — same root cause as #3373.
>
> `b`'s first flight lands while `a` has been re-asked by the second write. The landing cleared every pending entry on `b` (including the one `a`'s re-ask propagated), so for the gap between that landing and `a`'s second landing `b` read as settled — `isPending(b)` blipped `false → true`. With the landing retiring only `b`'s own entry, `b` stays pending on `a` and the verdict stays `true` until `b`'s final answer: `false → true → false`.
>
> — *Claude via Cursor*

ryansolid:
> Fixed on `next` in d7cb4566 (#3381). Same root cause as #3373: the first-hop landing wiped the propagated pending entry for the re-asked input, so `isPending(b)` read false for the gap. The landing now retires only the node's own entry and `b` stays pending on `a` until `a`'s second flight lands.

> — *Claude via Cursor*

Issue #3375, opened 2026-09-11
2.0.0-rc Loading reveals stale content before downstream async work finishes


GabbeV:
### Describe the bug

After two updates to a Loading boundary's `on` value, it leaves its fallback before the new details are ready. The repro shows `Details: 0 → Loading... → Details: 0 → Details: 3`. The second 0 is stale: the displayed sum is already 3.

### Your Example Website or App

https://s.olid.uk/id/KASyYgRqQcqR9mASsJOqbA

### Steps to Reproduce the Bug or Issue

1. Wait for Sum and Details to show 0.
2. Click Run.
3. Loading disappears and the old Details: 0 returns for two seconds before Details: 3 arrives.

### Expected behavior

Keep the loading fallback until the new details are ready, then show Details: 3.

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at f8b40b7e (development and production) and #3337 at 20f31433.

Removing either `setCount(1)` or the second page update prevents the stale reveal.

ryansolid:
> Reproduced at the signals level and fixed in #3381. Two things were wrong, both surfaced by the second `on` change.
>
> 1. **Boundary collection.** The `on` reset clears the boundary's collected sources and rebuilds them from later notifications. Status propagation dedupes on what the effect already carries, so `details` — which the effect had been pending on since the `count` write — was never re-reported after the reset, and `details`' re-flight after `pageData` landed stayed invisible. When `pageData` settled the boundary had nothing left to wait on and revealed with the effect still pending. A collecting boundary now records every source the notifying effect is pending on.
> 2. **Transaction completion.** The transaction decided "this source's own flight is still up" by `_error.source === source`, which a later-pending input overwrites on propagation — `details` was still flying but read as settled, so the `count=1` hold committed at 5500, ahead of its answer. The test is now the source's self entry in its pending sources.
>
> Your timeline now shows the fallback from the first page change through `details`' final landing, then `Details: 3` — no stale reveal. One thing to be aware of: `page=2` now rides the `count` hold and reveals at 8500 too (`pageData`, a reader of `page`, joined the hold when its earlier landing entered it — A15). Before the fix that hold was leaking early through the `_error.source` bug. Whether an `on` reset should release the page write instead is a spec question flagged on the PR.
>
> — *Claude via Cursor*

ryansolid:
[...]

Issue #3374, opened 2026-09-11
2.0.0-rc Repeating a signal write after remounting its async reader publishes it early


GabbeV:
### Describe the bug

Remounting an async reader and then repeating the pending signal write makes the signal publish before its derived value. The repro briefly shows `Count: 1` alongside `Details: 0`.

### Your Example Website or App

https://s.olid.uk/id/U6TNbM1ORvOIBIs3c4H9hQ

### Steps to Reproduce the Bug or Issue

1. Wait for both values to show 0.
2. Click Run.
3. Count changes to 1 while Details stays 0 for another second.

### Expected behavior

Count and Details should update together.

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at f8b40b7e (development and production) and #3337 at 20f31433.

Removing either `setVersion(2)` or the second `setCount(1)` makes the values update together.

ryansolid:
> Reproduced at the signals level and fixed in #3381. Both ingredients you identified are required — the keyed remount and the same-value rewrite — and here is why.
>
> `count=1` is held while `details` is in flight; the reader inside `<Show keyed>` is the transaction's *reporter* for that flight. The remount disposes it. The new reader is served the committed `Details: 0` (correct — the frame still shows `Count: 0`) and recorded for the commit replay, but not as a reporter. The same-value `setCount(1)` then re-activates the transaction, whose completion check finds no live reporter for `details`, drops it, and commits `count=1` with the flight still up → `Count: 1` beside `Details: 0` until the landing.
>
> Fix: a reader served a pending node's committed value joins the transaction's reporters for that node when it already has an entry — it observes the flight, so it holds the write the way the disposed reader did, and dies with disposal the same way. Your timeline now publishes `Count: 1` and `Details: 1` in the same frame at 4500.
>
> — *Claude via Cursor*

ryansolid:
> Fixed on `next` in d7cb4566 (#3381). The keyed remount disposed the transaction's only reporter for `details`; the new reader was served the committed value by the A15 reveal carve-out but never joined the reporters, so the same-value `setCount(1)` found no live reporter and committed ahead of `details1`. A carve-out reader now joins the transaction's reporters for a node the transaction already holds an entry for, and dies with its own disposal like any reporter.

> — *Claude via Cursor*

PR #3381, opened 2026-09-12
fix(signals): second write while an async chain is in flight (#3373, #3374, #3375, #3376)

### LINKED ISSUE #3373: 2.0.0-rc Async memo chain publishes a stale result alongside a newer signal value
### Describe the bug

Updating a signal again while its async memo chain is pending can make one JSX expression show mismatched values. The repro displays `0 / 0 → 2 / 1 → 2 / 2`, even though both memos just forward the signal's value.

https://s.olid.uk/id/iP5gJf0HRSGuTKCEGVKrLA

1. Wait for `0 / 0` to appear.
2. Click Run.
3. See `2 / 1` appear before settling at `2 / 2`.

### Expected behavior

Both numbers should always match.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at f8b40b7e and #3337 at 20f31433.

Making either memo synchronous removes the mismatch.

ryansolid:
Fixes #3373, #3374, #3375, #3376 — the four reports from @gabbev that share one shape: **a second write arrives while an async chain started by the first is still in flight**, and the engine's hold accounting loses track of the flight.

Each was reproduced at the `@solidjs/signals` level with a manual clock (`packages/signals/tests/async-chain-supersession.test.ts`, seven pins) before fixing. All reproduce on `next` @ `344ed054`.

### #3373 / #3376 — a stale first-hop landing commits the newer signal

`a = memo(delay(count))`, `b = memo(delay(a))`. `count=1` starts `a1`; `a1` lands and starts `b1`; `count=2` arrives and starts `a2` while `b1` is still up. `a2` only changes `a`'s *status* (not yet its value), so `b` is marked pending on `a` by propagation with `b1` still current — nothing supersedes it. When `b1` lands, `asyncWrite`'s `clearStatus(b)` wiped **all** of `b`'s pending sources, including `a`. `b`'s staged 1 woke the text effect, which saw nothing pending; the transaction's reporter for `a` found nothing blocking; it committed `count=2, b=1` with `a2` in the air → `2 / 1` (#3373), and `isPending(b)` read false for the gap (#3376).

Fix: the landing uses `landStatus`, which retires only the node's **own** entry. With another source still pending the node stays derivatively pending on it; the landed value is written and held (it is still the answer for the inputs it was asked with), and the input's settle releases it — or its value change re-asks the node. Pinned both ways: `a2` landing changed (reveal `2 / 2` at `b2`) and unchanged (`b1`'s held answer reveals at `a2`'s landing, no extra flight).

Complement at registration: `handleAsync` now drops pending entries the node's inputs propagated earlier. The run that asked the flight read every input without throwing, so a still-pending input was masked for it (an active override, A17) and does not describe the new answer. Without this the A17 spec pin regressed (the override-derived flight was held by the authoritative refetch).

### #3375 — `Loading on={page}` reveals content with details still pending

Two mechanisms:

1. **Boundary collection.** The `on` reset clears `_sources` and rebuilds from later notifications. `notifyStatus` dedupes on the effect's `_pendingSources`, so a source the effect already carried across the reset (`details`, from the `count` write's flight) is never re-reported, and `details`' re-flight after `pageData2` lands stays invisible. When `pageData` settled, `_sources` emptied and the boundary revealed with the effect still pending on `details`. Fix: a collecting boundary records every source the notifying effect is pending on (`node._x._pendingSources`), gated to the `STATUS_PENDING` type — an error boundary must not collect pending sources, since `_checkSources` would hold its fallback on them.

2. **Transaction completion.** `transitionComplete` tested "the source's own flight is up" with `_error.source === source`. Propagation from an input that went pending later overwrites `_error` with the input, so `details` — still flying — read as settled and the `count=1` hold committed at 5500, ahead of its answer. Fix: the test is the source's self entry in `_pendingSources`, which only `notifyStatus`'s source path adds and the landing / supersede sweep retire.

Note for the reviewer: the pinned timeline records `page=2` riding the `count` hold (`Sum: 3 @8500`). That is A15 applying — `pageData`, a reader of `page`, lives in the hold since its earlier landing joined it — where before it was leaking early through the `_error.source` glitch. If an `on` reset should release the page write instead, that is a spec ruling, not this bug.

### #3374 — a same-value rewrite after a keyed remount publishes early

`count=1` is held with `details` in flight. `<Show keyed when={version()}>` remounts the reader; the new reader is served the committed `Details: 0` by the A15 reveal carve-out (correct — the frame still shows `Count: 0`) and recorded for replay. But the transaction's only *reporter* for `details` was the disposed reader. The same-value `setCount(1)` activates the transaction; `transitionComplete` finds no live reporter, drops the source and commits `count=1` with `details1` still up → `Count: 1` beside `Details: 0` for a second.

Fix: a carve-out reader joins the transaction's reporters for the node **when it already has an entry** (`heldFromStale`, one Map lookup). It observes the flight (A15: async work observed by a reader settles as one unit with the writes that asked it) and dies with disposal like any reporter (`reporterBlocksSource` sees the dep the read linked). INV-3 is preserved as "entries open only from queue notification" — a boundary-consumed flight has no entry and stays consumed; the internals doc is amended accordingly. A chain's intermediate memo never reaches the carve-out (the read re-pulls it and enters the transaction), and a node that is both an origin and pending on an upstream re-ask blocks through its own flight until it lands — both pinned.

## Size

Core floor `22,648 → 22,780` (+132 B minified) after a golf pass (from +259 on the first cut). Brotli scenarios +32…+73 B; nine `.size-limit.js` caps bumped with notes; `size-limit` exits 0.

## Verification

- signals 1758/1758 (7 new pins), solid 595/595, web 734/734; `tsc` clean on `src/` and the new test (remaining `tsc` errors are pre-existing test files on `next`).
- Every `spec-async-semantics` pin holds, including A17 (the registration clear exists for it).
- Changeset: `.changeset/fix-async-chain-second-write.md` (`@solidjs/signals`, patch).
- Docs: `INTERNALS-ASYNC-STATE.md` — `landStatus` vs `clearStatus`, `transitionComplete`'s own-flight test, INV-3's join site, the carve-out paragraph's "does not entangle" qualified.

The one edit with blast radius beyond the four repros is the `transitionComplete` own-flight test; suites are clean, but it is the line to watch on CI/CodSpeed.

— *Claude via Cursor*


ryansolid:
> Pushed a82d2e3c on top: the A15 question flagged in the description is now ruled and implemented.
>
> **Ruling (2026-09-12):** a `Loading` boundary's `on` reset ends the hold on writes that only its readers observed. A reader registered as a reporter while the boundary showed content stops blocking once the boundary flips to its fallback; if nothing outside the boundary consumes the flight, the hold is over. A reader outside the boundary still holds it.
>
> **Mechanism:**
> - `reporterBlocksSource` walks the reporter's `_queue._parent` chain; a reporter behind a collecting pending-type boundary (`_collectionType & STATUS_PENDING && !_initialized`) is not live. Exact, because a collecting boundary consumes new notifications — the only reporters under it are the stale forwarded ones.
> - The flush judges only the active transaction, so the reset itself wakes parked transactions (`wokenTransitions`, deduped). The flush re-enters a woken transaction from the `finally` of an idle pass (`!scheduled`), so it adopts no unrelated ambient work; a wake with other live reporters simply re-parks.
>
> **#3375 pin now:** `0: Loading | Sum: 0` → `3000: content | Details: 0` → `4000: Loading... | Sum: 1 | Sum: 2` → `5500: Sum: 3` → `8500: content | Details: 3`. The two `Sum` publishes at 4000 are one synchronous drain: the ambient `page=1` commit, then the released `count=1` in the idle pass after it — no tear (fallback is showing). A companion pin keeps the hold when an outside reader also observes `details`.
>
> While writing that companion pin I found a pre-existing #3322 artifact, unchanged by this PR: an effect computed mainline in the pass a stamped recompute later pulls into the transaction publishes its stale slot at commit *before* the contested re-derive publishes the right value (`Sum: 1` then `Sum: 2` in the same frame). The note in `recompute` says the stale value is never published; here it is. Tracked separately; the pin asserts around it.
>
> Floor: +135 B on top (22,780 → 22,915); size caps bumped with notes. Existing tests untouched; signals 1759, solid 595, web 734.
>
> — *Claude via Cursor*

PR #3395, opened 2026-09-12
docs: correct createContext guidance for server rendering

### LINKED ISSUE #3390: createContext JSDoc recommends module-scope state for app-wide state; that is shared across requests on the server
`packages/solid/src/client/core.ts` (next, 344ed054), in the `createContext` JSDoc, lines 64–66:

> If you want truly app-wide state, **don't use Context** — a module-scope signal/store *is* a global. Context is for scoping state to a subtree; that's why a Provider is required.

Taken on its own this is fine for a client-only app, but with `ssr: true` a module is evaluated once per server process, so a module-scope signal or store is shared by every request. A store derived from a server function at module scope leaks one user's data into another's render, and the module-scope signal has no owner to dispose it.

The 2.0 docs take the opposite stance for exactly that reason (Components and JSX → Context: "Prefer this over a module-scope signal or store. Module-scope state has no owner, and on the server one module instance is shared across requests"; the State management guide has a worked example of the failure). The reference page for `createContext` is generated from this JSDoc, so the two currently contradict each other on the same site.

Suggested wording:

> Context is for state that belongs to a subtree, which includes app-wide state in an app that renders on the server: a value created inside a component is created once per request, and the Provider owns and disposes it. Module scope is shared by every request in the same process, so reserve it for constants.

Related, smaller JSDoc items found by type-checking the docs' generated examples:

- `packages/web/server-functions/src/server.ts` line 3239, `handleServerFunctionRequest` `@example`: `import { handleServerFunctionRequest } from "@solidjs/web/server-functions"`. That subpath is the client entry; the function is exported from `@solidjs/web/server-functions/server`.
- `createContext` / `useContext` `@example` blocks render `<TodoList />` where `TodoList` is a `// ...` stub with no return value, so the example fails `tsc` as written. A `return null;` (or `return <ul />;`) in the stub fixes it.
- `useContext` has `@description https://docs.solidjs.com/reference/component-apis/use-context`, a 1.x URL.

JohannesGezachew:
Closes #3390 (except the `useContext` URL, see below).

### `createContext` guidance

The JSDoc recommended a module-scope signal or store for app-wide state. Under `ssr: true` a module is evaluated once per process, so that state is shared across requests, and the 2.0 docs advise the opposite. Since the reference page is generated from this JSDoc, the two contradicted each other. This uses the wording from the issue.

### Smaller items from the same issue

- `createContext` / `useContext` `@example`: the `TodoList` stub had no return value, so the examples fail `tsc` as written. Added `return null;`.
- `handleServerFunctionRequest` `@example`: imported from `@solidjs/web/server-functions`. In `packages/web/package.json` that subpath resolves to `server-functions/dist/client.js` under the `browser` condition, while `./server-functions/server` maps unconditionally to the server build, so the example now imports from `@solidjs/web/server-functions/server`.

### Not done

The fourth item, `useContext`'s `@description https://docs.solidjs.com/reference/component-apis/use-context` being a 1.x URL, is left alone: every `@description` in the 2.0 source points at `docs.solidjs.com/reference/...`, including `create-context` and `children` directly around it, and the repo has no other docs host, so I could not tell what the intended 2.0 target is. Happy to update this one, and its neighbours if they are stale too, once you name the URL.

Comment-only changes; no runtime code is touched.

### DIFF
--- a/packages/solid/src/client/core.ts
+++ b/packages/solid/src/client/core.ts
@@ -61,9 +61,11 @@ export interface Context<T> extends ContextProviderComponent<T> {
  *   static fallback (theme, locale, frozen config). Outside any Provider,
  *   `useContext` returns `defaultValue`.
  *
- * If you want truly app-wide state, **don't use Context** — a module-scope
- * signal/store *is* a global. Context is for scoping state to a subtree;
- * that's why a Provider is required.
+ * Context is for state that belongs to a subtree, which includes app-wide
+ * state in an app that renders on the server: a value created inside a
+ * component is created once per request, and the Provider owns and disposes
+ * it. Module scope is shared by every request in the same process, so reserve
+ * it for constants.
  *
  * @param defaultValue optional default; only meaningful for primitive
  *   fallbacks. Omit for any context carrying reactive state.
@@ -87,6 +89,7 @@ export interface Context<T> extends ContextProviderComponent<T> {
  * function TodoList() {
  *   const [todos, { addTodo }] = useContext(TodosContext); // typed as TodosCtx
  *   // ...
+ *   return null;
  * }
  * ```
  *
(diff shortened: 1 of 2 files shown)


ryansolid:
> Merged as e02157a4 — thanks, and the reasoning on the `server-functions/server` subpath was exactly right.
>
> — *Claude via Cursor*

Issue #3386, opened 2026-09-12
`dynamic()` / `Dynamic` with a string tag: no `runHydrationEvents()`, namespace ignores parent


ryansolid:
**Version:** 2.0.0-rc.8 (`@solidjs/web`)

Two correctness gaps in the string-tag branch of `dynamic()` (`packages/web/src/index.ts` ~366–374), which `Dynamic` uses:

```ts
case "string":
  const el = sharedConfig.hydrating
    ? getNextElement()
    : createElement(component as string, untrack(() => (props as any).is));
  spread(el, props);
  return el;
```

## 1. Hydration event replay never runs

Compiled JSX emits `_$runHydrationEvents()` after any element with event handlers (`babel-plugin/src/dom/element.ts` ~323). `dynamic()` binds handlers through `spread(el, props)` and never calls it. Events queued by the hydration header script for a `<Dynamic component="button" onClick=…>` are replayed only if some *other* compiled element on the page happens to call `runHydrationEvents()` after it hydrates; if the interactive elements are all `Dynamic`, queued clicks are dropped until `_$HY.done`. `@yak/solid` had to add the call in its own element renderer.

**Fix:** call `runHydrationEvents()` after `spread` when hydrating (cheap — it's a no-op when nothing is queued).

## 2. Namespace is chosen from the tag alone

`createElement(tag)` picks the namespace via `SVGElements.has(tag)` / `MathMLElements.has(tag)`. Tags that exist in both HTML and SVG — `a`, `script`, `style`, `title` — are always created as HTML, so `<svg><Dynamic component="a">` produces an HTML anchor inside an SVG tree. The compiler resolves this from the parent at build time; the runtime path has no parent context. This is why `@yak/solid` carries an `ambiguousSvgTags` set and routes those four tags differently.

**Fix options:** (a) accept the namespace from the mount context — `insert`/the parent element is known at spread time (`el.parentNode` after insertion is too late; a context or an explicit `ns` option would work); (b) at minimum document the limitation and expose the set so libraries don't hard-code it.

## Related

- Perf: with a constant string tag `dynamic()` still allocates a factory memo and a per-instance memo to track a tag that can't change — tracked in #3387 and the perf tracker #3389.
- Surfaced by the `@yak/solid` audit ([DigitecGalaxus/next-yak#644](https://github.com/DigitecGalaxus/next-yak/pull/644)).

ryansolid:
[...]

PR #3396, opened 2026-09-12
fix: run hydration events after dynamic() spreads a string tag

JohannesGezachew:
Addresses part 1 of #3386.

Compiled JSX emits `runHydrationEvents()` after any element carrying event handlers — `packages/babel-plugin/index.js` (`info.topLevel && config.hydratable && results.hasHydratableEvent`), which produces the `_$spread(el, props, false); _$runHydrationEvents();` pair visible in the hydratable fixtures.

The string-tag branch of `dynamic()` binds handlers through `spread()` but never calls it, so events the hydration script queued for a `<Dynamic component="button" onClick=…>` are replayed only if some other compiled element hydrates after it and makes the call. When every interactive element is a `Dynamic`, queued events are dropped until `_$HY.done`.

This calls `runHydrationEvents()` after `spread()` while hydrating. It is cheap: the function returns immediately unless `sharedConfig.events` has an unqueued entry. The `hydrating` flag is captured before `spread()` so the new branch cannot see a different value from the one that selected `getNextElement()`.

### Not included

Part 2 of the issue (ambiguous SVG/HTML tags `a`, `script`, `style`, `title` taking the namespace from the tag alone) is left out: you list two options there, (a) accepting a namespace from the mount context and (b) documenting the limitation and exposing the set, and that is your call to make rather than mine.

### Testing

New spec `packages/web/test/hydration/dynamic-hydration-events.spec.tsx`: it queues a real click on the server-rendered button the way the hydration script does, hydrates a `<Dynamic component="button" onClick=…>`, and asserts the handler ran. Against the previous runtime it fails with `expected +0 to be 1`; with this change it passes.

`pnpm test` for `@solidjs/web` on macOS 26 / Node 26, all three configs green:

| config | result |
|---|---|
| default | 77 files, 734 tests |
| server | 84 files, 791 passed, 2 skipped |
| hydration | 29 files, 169 tests |

`prettier --check` is clean on both changed files.

### DIFF
--- a/.changeset/dynamic-run-hydration-events.md
+++ b/.changeset/dynamic-run-hydration-events.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/web": patch
+---
+
+Call `runHydrationEvents()` after `dynamic()` spreads props onto a string tag, so events the hydration script queued for a `<Dynamic>` element are replayed instead of being dropped until `_$HY.done`.
--- a/packages/solid/src/client/core.ts
+++ b/packages/solid/src/client/core.ts
@@ -61,9 +61,11 @@ export interface Context<T> extends ContextProviderComponent<T> {
  *   static fallback (theme, locale, frozen config). Outside any Provider,
  *   `useContext` returns `defaultValue`.
  *
- * If you want truly app-wide state, **don't use Context** — a module-scope
- * signal/store *is* a global. Context is for scoping state to a subtree;
- * that's why a Provider is required.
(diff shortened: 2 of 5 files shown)


ryansolid:
> Merged as 1af28a16 (#3386 part 1). Capturing `hydrating` before the `spread()` was the right call. Part 2 (namespace from the mount context vs. documenting the ambiguous tags) stays open on the issue for a ruling.
>
> — *Claude via Cursor*

Issue #3379, opened 2026-09-11
2.0.0-rc.8 isPending stays false while an async memo loads an optimistic value


GabbeV:
### Describe the bug

With two async memos downstream of an optimistic value, `isPending` stays false while loading the optimistic update. It correctly becomes true during the revert.

### Your Example Website or App

https://s.olid.uk/id/oQ6V-U3oTRyzeG7Pm_rVYA

### Steps to Reproduce the Bug or Issue

1. Wait for initialization.
2. Click the button.
3. Pending stays false until the view becomes 1 / 1.
4. During the revert, Pending correctly becomes true.

### Expected behavior

Pending is also true while loading the optimistic value.

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Bypassing the async copy memo and reading value() directly in the delayed memo makes the indicator work correctly.

Reproduced against next at 344ed054 and #3337 at 20f31433.

ryansolid:
> Confirmed and fixed in #3400 — thanks for the reproduction, and for the "reading `value()` directly works" observation, which pointed straight at it.
>
> Root cause: `isPending` indicators ride a small *child* lane of the node they report on, so they can flush while the node's own async is in flight — but that child lane is parented from the node's lane at the moment it is created. When `copy`'s pending status propagated onto `details`, the engine refreshed `details`' indicator companion *before* it put `details` on the optimistic lane, so the companion's lane was created with no parent. Your `Pending` effect depends on both `details` and the companion; with no parent/child relation between the two lanes, they merged, and the indicator's update waited in the held lane until the 1000ms derivation landed — exactly the window you were trying to show a spinner for. Reading `value()` directly avoided it because `details` joined the lane before its companion was refreshed. The lane is now assigned first.
>
> — *Claude via Cursor*

ryansolid:
> Landed on `next` in 9da7f0a8 (#3400) — see the root cause above. Ships in the next RC.
>
> — *Claude via Cursor*

PR #3397, opened 2026-09-12
fix(signals): give the latest() shadow companion ownedWrite (#3378)

### LINKED ISSUE #3378: 2.0.0-rc.8 latest read throws an owned-scope write error when an action resumes
### Describe the bug

Toggling a conditional JSX branch while an action is pending, then restoring it as the action resumes, throws `REACTIVE_WRITE_IN_OWNED_SCOPE` and halts reactivity.

The reactive setters are called from the event handler and action. The failing write is internal to Solid: `syncCompanions -> setSignal` updates the memo's latest companion.

https://s.olid.uk/id/ac2B3KcWSk-Jko5Xf7I8og

1. Open the repro.
2. Click the button.
3. After 400ms, the owned-scope write error occurs.

### Expected behavior

The conditional branch displays `Latest: 1` without throwing.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Replacing `latest(copy)` with `latest(count)` avoids the error. Using Show instead of the JSX conditional also avoided it in the tested variant.

Reproduced against next at 344ed054 and #3337 at 20f31433.

brenelz:
Fixes #3378

## Root cause
`latest()`'s shadow companion was the one internal companion created without `ownedWrite`. A companion sync can run from inside a computation: a transition-held memo recompute pulled mid-tick by a reader creating or refreshing its `latest()` shadow reaches `syncCompanions` -> `setSignal(shadow)` with `context` set to the pulling node. The `isPending` signal companion already carries the flag, so only the shadow write tripped the dev owned-scope guard and halted the app. Production never threw (the guard is dev-only) and the write path is unchanged.

The issue's shape: a JSX branch reading `latest(memo)` is toggled off while an action is pending (its shadow goes dormant with the reader) and restored as the action resumes. The restored reader runs before the memo's heap slot, so the fresh shadow's first compute pulls the still-dirty memo.

## Change
- `getLatestValueComputed` creates the shadow with `{ ownedWrite: true }`, matching `getPendingSignal`.

## Tests
- `packages/signals/tests/latest-shadow-owned-write.test.ts`: both paths (a fresh shadow created mid-pull, and a live shadow pulled current by a mounting reader). Both fail on `next` with `REACTIVE_WRITE_IN_OWNED_SCOPE`.
- `packages/web/test/latest-shadow-owned-write-issue-3378.spec.tsx`: the issue's JSX shape with an action; reproduces the halt on `next`.
- Full signals and web suites pass.

### DIFF
--- a/.changeset/latest-shadow-owned-write.md
+++ b/.changeset/latest-shadow-owned-write.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+Give the latest() shadow companion the `ownedWrite` flag its isPending companion already carries (#3378). A companion sync is internal plumbing that can run from inside a computation — a transition-held memo recompute pulled mid-tick by a reader creating or refreshing its latest() shadow — and the dev owned-scope write guard halted the app on the shadow write. Toggling a JSX branch that reads `latest(memo)` off while an action is pending and restoring it as the action resumes threw REACTIVE_WRITE_IN_OWNED_SCOPE.
--- a/packages/signals/src/core/verdict.ts
+++ b/packages/signals/src/core/verdict.ts
@@ -364,7 +364,7 @@ function getLatestValueComputed<T>(el: Signal<T> | Computed<T>): Computed<T> {
     setPendingCheckActive(false);
     const prevContext = context;
     setContextInternal(null); // Detach from owner so it isn't disposed with effects
-    lvc = optimisticComputed(() => read(el));
+    lvc = optimisticComputed(() => read(el), { ownedWrite: true });
     ext(el)._latestValueComputed = lvc;
     el._config |= CONFIG_HAS_COMPANIONS;
     markFirewallChildCompanions(el);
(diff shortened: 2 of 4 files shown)


ryansolid:
> Merged as 632e45c9 — the existing note in `collectPending` already named this exact failure class ("a shadow created without ownedWrite"), so this closes the last companion missing the flag. Thanks.
>
> — *Claude via Cursor*

ryansolid (approved):
> Reviewed against an independent root-cause of #3378 that landed on the same one-line change (`optimisticComputed(() => read(el), { ownedWrite: true })`), so the fix is corroborated: `recompute` restores the caller's `context` before its held-branch `syncCompanions`, so a pull-recompute of the memo from inside an effect writes the `latest()` shadow with that effect as `context`; the pending signal already carried `ownedWrite`, the shadow was the one companion that didn't. Dev-only guard, no behavior change in prod, verdict layer only (no core-floor cost).
>
> The two signals variants (fresh shadow / live shadow pulled current) plus the JSX-shape web test cover more than the issue's exact route, so I've dropped the overlapping pin from the companion PR (#3400 now carries only #3379) and referenced this PR in the internals doc's `syncCompanions` notes.
>
> — *Claude via Cursor*

Issue #3014, opened 2026-08-18
SSR spread ignores omit()'s hidden keys when props are a merge proxy — omitted props leak into the HTML and the leaked handler is re-bound on hydration


TylerRick:
### Describe the bug

When a component's props are a merge proxy — which they are as soon as a caller mixes static JSX props with a dynamic spread — spreading the result of `omit(props, …)` onto an element leaks the omitted keys during SSR. The client-side spread respects `omit()`'s key-hiding proxy; the SSR spread serializer does not.

Worse than the stray attributes: an omitted event handler leaks too, and hydration re-binds it as a native listener, so it is invoked with the raw `Event` instead of whatever the component's own handler would pass.

```tsx
function Field(
	props: { name: string; label: string; value: string; onChange: (v: string) => void } & Record<
		string,
		unknown
	>,
) {
	const rest = omit(props, 'name', 'label', 'value', 'onChange');
	return (
		<label>
			{props.label}
			<input
				name={props.name}
				value={props.value}
				onInput={(e) => props.onChange(e.currentTarget.value)}
				{...rest}
			/>
		</label>
	);
}

export default function App() {
	const [value, setValue] = createSignal('a@b.c');
	// getters, like a form library's fieldProps() — this dynamic spread is what makes the
	// callee's props a merge proxy
	const fieldProps = () => ({
		get value() {
			return value();
		},
		onChange: (v: string) => setValue(v),
	});
	return (
		<>
			<Field label="Email address" name="email" placeholder="you@example.com" {...fieldProps()} />
			<pre id="state">signal: {value()}</pre>
		</>
	);
}
```

Server output — `label` was omitted from `rest`, and is on the input anyway:

```html
<label _hk="1">
	<!--$-->Email address<!--/-->
	<input name="email" value="a@b.c" label="Email address" placeholder="you@example.com" />
</label>
```

Then type one character into the field. The leaked `onChange` fires with the `Event` object, `setValue(event)` puts it in the signal, and the next render dies:

```
[REACTIVITY_HALTED] An uncaught error halted the reactive system. No further updates will be processed.
TypeError: Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'.
    at reconcileArrays (@solidjs_web.js:250:16)
    at insertExpression (@solidjs_web.js:1579:9)
```

In the app where we hit this, it showed up as form fields whose submitted value was the string `[object Event]`.

Two things make it go away, which is what points at the merge proxy + SSR spread combination:

[...]

PR #3401, opened 2026-09-13
fix(signals): treat a plain merge() result as an ordinary source (#3384)

### LINKED ISSUE #3384: `merge()` re-flattens through a stale `$SOURCES` and discards the object's own properties
**Version:** 2.0.0-rc.8 (`@solidjs/signals` / `solid-js`)

## Repro

```js
import { merge, omit } from "solid-js";

// A merged props object (plain-object form: no source is a $PROXY).
const props = merge({ href: "/x", title: "t" }, { $active: true, children: "hi" });

// Copy it the way a library might (Reflect.ownKeys → includes the $SOURCES symbol),
// dropping `$active` and adding an own `class` getter.
const out = {};
for (const key of Reflect.ownKeys(props)) {
  if (key === "$active") continue;
  Object.defineProperty(out, key, Reflect.getOwnPropertyDescriptor(props, key));
}
Object.defineProperty(out, "class", { get: () => "yak-abc", enumerable: true, configurable: true });

const final = merge(out, { rel: "noopener" });
final.class;    // undefined   (expected "yak-abc")
final.$active;  // true        (expected undefined — it was removed)
```

<details><summary>Full repro (the yak `copyProps` logic verbatim, both `merge` and `omit` paths)</summary>

```js
// Repro: yak's copyProps copies merge()'s $SOURCES symbol, so a downstream
// merge() flattens to the ORIGINAL sources and bypasses yak's filtering.
import { merge, omit } from "solid-js";

const skip = k => typeof k === "string" && (k.charCodeAt(0) === 36 || k === "class" || k === "theme");

// yak PR copyProps (verbatim logic)
function copyProps(props) {
  const out = {};
  for (const key of Reflect.ownKeys(props)) {
    if (skip(key)) continue;
    const d = Reflect.getOwnPropertyDescriptor(props, key);
    if ("value" in d && d.enumerable) out[key] = d.value;
    else Object.defineProperty(out, key, d);
  }
  Object.defineProperty(out, "class", { get: () => "yak-abc", enumerable: true, configurable: true });
  return out;
}

// Call site: <StyledLink {...linkProps} $active children="hi" />  → compiled createComponent(StyledLink, merge(linkProps, {...}))
const linkProps = { href: "/x", title: "t" };
const props = merge(linkProps, { $active: true, children: "hi" });
console.log("props is plain object with $SOURCES:", Object.getOwnPropertySymbols(props).length > 0);

const forwarded = copyProps(props);
console.log("forwarded keys:", Reflect.ownKeys(forwarded).map(String));

// Target component body: <a {...props} rel="noopener" /> → spread(el, merge(props, { rel }))
const final = merge(forwarded, { rel: "noopener" });
console.log("final.class   =", final.class, "   (expected 'yak-abc')");
console.log("final.$active =", final.$active, "   (expected undefined)");
console.log("final keys    =", Object.keys(final));

// Same thing through omit() in the target (rest-props pattern), then merge:
const rest = omit(forwarded, "title");
const final2 = merge(rest, { rel: "noopener" });
console.log("via omit: class =", final2.class, " $active =", final2.$active);
```

</details>

## Cause

`packages/signals/src/store/utils.ts` `merge()` (~112–193, plain-object path): when a source carries `$SOURCES`, `merge` flattens *through* it to the original sources instead of treating the object as a source in its own right. Any own property added or removed after the fact is ignored. `omit()` had the same shape of problem and was fixed in #3014; `merge()` still trusts `$SOURCES` unconditionally.

## Why it matters

`$SOURCES` is an enumerable-by-`Reflect.ownKeys` symbol on a plain object that looks like ordinary props. Anything that copies props generically (`Reflect.ownKeys`, `Object.getOwnPropertySymbols`, structured clones of descriptors) inherits it, and the next `merge` downstream silently resurrects removed keys and drops added ones. `@yak/solid`'s PR #644 does exactly this (`copyProps`), so `$`-prefixed style props leak onto DOM elements and the computed `class` is lost whenever the target component spreads `{...props}` — we confirmed it against rc.8.

## Options

1. Make `$SOURCES` non-copyable: define it non-enumerable *and* have `merge` verify the object's own keys still match the flattened sources before trusting it (cheap: compare key count).
2. Treat an object with own data/accessor properties that aren't in its `$SOURCES` as opaque (no flattening) — the #3014 approach.
3. Drop `$SOURCES` flattening entirely and accept one extra level in nested merges.

(1) or (2) keeps the current fast path for the common compiler-emitted case.

ryansolid:
Fixes #3384.

## Problem

`merge()`'s plain-object form recorded its flattened sources on the result under the `$SOURCES` symbol, and a later `merge()` flattened *through* them instead of reading the object. But that result is a real object callers copy or mutate afterwards:

- descriptor copies via `Reflect.ownKeys` carry the symbol along (this is what `@yak/solid` PR #644's `copyProps` does, so `$`-prefixed props leaked onto DOM elements and the computed `class` was lost whenever the target spread `{...props}`);
- `{...props}` copies carry it too (enumerable symbol);
- `@solidjs/html` (`tagged-jsx.ts`) calls `mergeProps(props, spread)` and then assigns later props and a `children` getter onto the result before `createComponent` — a component's own `merge(defaults, props)` then dropped them.

No cheap check validates the recorded sources (a same-key value overwrite is invisible to key counts or owner identity), so the plain form can't be trusted at all.

## Fix

The plain-object result records nothing and is treated as an ordinary source by a nested `merge`. Only merge **proxies**, whose writes are no-ops, are still flattened through — that is also all `mergeSources()` (used by `spread()`) ever returned, so no consumer changes. `omit()` proxies keep answering `$SOURCES` with `undefined` (#3014).

Side effect worth knowing: `merge(defaults, props)` where `props` is a plain merge result covering every default now returns `props` directly — flattening used to defeat the "all keys covered → return the source" shortcut, so that common component pattern always allocated before.

## Tests

Four new tests in `utilities.test.ts`; three fail on `next`:
- descriptor copy of a merged object (the issue repro) — copy is read, `$active` gone, `class` present;
- `{...props}` copy with a deleted and an overridden key;
- html-style post-merge mutation (assign, add, `defineProperty children`);
- guard: merge proxies still flatten (`mergeSources(outer)` is the flat list), plain results yield `undefined`.

Suites: signals 1769, solid 595, html 200, web dom 735 / server 790 / hydrate 169 — all green.

## Performance

Interleaved A/B, old vs new `dist/prod` in one process, 41 alternating rounds, ratio new/old:

| path | ratio |
|---|---|
| `merge(defaults, props5)` construct | 0.97 |
| `merge(props12, {children})` construct | 0.96 |
| nested `merge(inner, {c})`, not covered | 1.05–1.10 |
| nested `merge(defaults, inner)`, covered | **0.52** |
| getter reads through 1 / 2 / 3 levels | 0.84 / 1.01 / 1.06 |
| proxy construct / read, `omit` | 0.96–1.03 (unchanged path) |

The one regression is the non-covered nested plain merge: it walks the intermediate result (bound getters, re-bound once more) instead of the original sources. The covered case, which is the `merge(defaults, props)` shape, no longer allocates at all.

### DIFF
--- a/.changeset/fix-merge-plain-result-not-flattened.md
+++ b/.changeset/fix-merge-plain-result-not-flattened.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+`merge()` no longer flattens through a plain-object merge result. That form is a real object callers may copy (`Reflect.ownKeys` descriptor copies, `{...props}`) or mutate afterwards (`@solidjs/html` assigns props and a `children` getter after spreading), and a later `merge` tunnelled back to the original sources through the `$SOURCES` symbol — resurrecting removed keys and dropping added or overwritten ones (#3384). The plain result now records no sources and is read like any other object; only merge *proxies*, whose writes are no-ops, are still flattened. A side effect: `merge(defaults, props)` where `props` is itself a plain merge result that covers every default now returns `props` directly instead of always allocating.
--- a/packages/signals/src/store/utils.ts
+++ b/packages/signals/src/store/utils.ts
@@ -81,11 +81,12 @@ function resolveSource(s: any) {
 
 const $SOURCES = Symbol(__DEV__ ? "MERGE_SOURCE" : 0);
 /** @internal The flattened sources behind a `merge()` PROXY, or undefined.
- * Only the proxy form: its writes are no-ops, so the sources are the whole
- * truth. merge()'s plain-object form also records `$SOURCES` (so nested
- * merges flatten), but it is a real object callers may mutate afterwards
- * (html's tagged templates assign props after spreading) — those own writes
- * live on the object, not in the sources, so it must be read directly. */
+ * Only the proxy form carries sources: its writes are no-ops, so they are
+ * the whole truth. The plain-object form is a real object callers may copy
+ * (descriptor copies, `{...props}`) or mutate afterwards (html's tagged
+ * templates assign props and a children getter after spreading) — what is
+ * on the object is the truth there, so it records nothing and every consumer,
+ * a nested merge included, reads it directly (#3384). */
 export function mergeSources(o: any): any[] | undefined {
   return o != null && o[$PROXY] === o ? o[$SOURCES] : undefined;
 }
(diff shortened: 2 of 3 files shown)


ryansolid:
> Reset the branch to `9692c81a` (the original fix), dropping the two getter-record commits (`fded3b2d`, `7b508752`).
>
> Local `utilities.bench.ts` (merge cases, mean time vs `next`): v1 geomean 1.004× — faster on small merges (0.85–0.94×), slower only on wide, deep re-merges (`deep 0,100,3,2`, `covered 100,25,5`). The record variants moved that cost onto every plain result carrying a getter: geomean 1.16×, slower than v1 on 20 of 24 cases, and none of the deep cases recovered to `next`. The remaining CodSpeed regressions on this PR are the inherent price of not trusting a plain result's origins; a design that makes nested merges cheaper than `next` (compiler-marked `this`-free props passing through unbound) is a separate change.
>
> — *Claude via Cursor*

PR #3402, opened 2026-09-13
feat(web): trace context — getTraceContext(), W3C traceparent, OBSERVE.server.trace (C1)

ryansolid:
Follows #3398 (C0, merged as `af94f674`) — rebased onto `next`, one commit.

## What

The server runtime now reads the **W3C Trace Context** half of the HTTP exchange once per request and exposes it:

```ts
import { getTraceContext } from "@solidjs/web";

const trace = getTraceContext();
// { traceId, spanId, parentId?, sampled?, state?, baggage?, entries: { traceparent: "00-…" } }
await fetch(url, { headers: { traceparent: trace!.entries.traceparent } });
```

- **Continues** an incoming `traceparent` (version `00`; later versions read as `00`; `ff`, all-zero ids, wrong lengths, uppercase → ignored), **originates** a trace when none came in. `tracestate`/`baggage` carried verbatim.
- One object per request — repeated reads, the derived events direct SSR-time server-function calls run under, and every bundle copy (the record is parked on the global under a registered symbol, like `RequestContext`; the server-functions handler commits a response whose body read the trace through the main entry). A render outside a request scope has its own; `undefined` outside both and on the client.
- **Every build tier.** Forwarding a trace downstream from a server function must work in prod with no observer installed — this sits beside `getRequestEvent()`/`httpHeader()`, not behind `OBSERVE`.

The trace is handed **down to the browser** (which cannot send a header on the document request) on two carriers, one set of `entries`:

- `Server-Timing: traceparent;desc="00-…"` appended when the response head commits — `createSSRResponse` (string + stream), `commitEventResponse`, `commitResponseStub` (now takes the owning `event` in options; falls back to the ambient event when its `response` is the stub). So frames, RPC responses and redirects carry it too.
- `<meta name="traceparent" content="…">` in the HTML shell — the `</head>` splice, or the `onHead` string for embedded hosts; a headless fragment ships only the header.

**Observe/dev only:** `OBSERVE.server.trace.provide(provider)` — one global provider whose answer merges over the derivation (fields replace, `entries` merge by name). This is how Sentry's server SDK contributes OTel's active span and its `sentry-trace`/`baggage` entries once, with no per-request entry point into the host; its browser SDK reads exactly those two carriers (verified: 10.45 reads `Server-Timing` `sentry-trace`/`baggage`; it does not read `traceparent`).

## Rulings worth a look

1. **The browser is told only when something is recording the trace** — continued from upstream, or a provider answered. A trace the runtime originated alone stays available to `getTraceContext()` but emits nothing: there is no recorded server span to attach to, and a parent with flags `00` would make OTel-web's default parent-based sampler *drop* the pageload. Net: zero wire change for apps with no APM and no upstream trace.
2. **Incoming `baggage` is not echoed** to the page — upstream/proxy context; a provider adds its own.
3. **Metas join the `</head>` splice, not the prelude** — the prelude splices right after `<head>`, ahead of a shell's static `<meta charset>`.
4. **`commitEventResponse` with an event that has no response stub** (the server-function handler's default event) now hands the trace on by rebuilding the response; previously that exit was a pure pass-through. Identity is preserved when there is nothing to emit.
5. `Server-Timing` is now folded **entry by entry** wherever a stub meets a base (`responseInit`, a response's own `db;dur=53`), with a quote-aware splitter (a `baggage` desc carries commas). App-written names win; the runtime never duplicates one.

## Tiers

Prod artifacts (`dist/server.js`, `server-functions/dist/server.js`, `frames/dist/server.js`) carry the W3C half; the provider slot folds out behind `"_SOLID_OBSERVE_"` and lives only in `server.observe.js`/`server.dev.js` — pinned by an artifact scan in the spec.

## Size

Client bundles unchanged (the client stub is the shared `voidFn`); observe+engine scenario 26887 → 26892 B (gzip noise, within limit). Server bundles are not gated.

## Tests

`test/server/server-trace.spec.tsx` (34): derivation + malformed table, memo identity across reads/derived events/bundle copies, render-scoped lifetime (cleared with dispose), provider merge/uninstall/replace/throw, prod artifact ignores the provider, artifact tier scan, `Server-Timing` on string/stream/`commitEventResponse` (gap-fill + fold)/no-stub event/RPC dispatch/ambient fallback, dedupe + quoted commas, metas in document/`onHead`/headless/bare render (provider asked with `undefined`), attribute escaping + non-token names skipped. Plus `test/trace-client.spec.tsx` and type-test additions (`TraceSlot` augmentation lands from a second module; `Partial<TraceContext>` rejects a bad answer). Full web suites, `test-types`, size gate green.

## Docs

RFC 12: new *"The trace the request belongs to: `getTraceContext()`"* section + lifecycle notes. RFC 08: the `OBSERVE.server.trace` slot. RFC 10: one sentence on invocations + forwarding. Changeset: `@solidjs/web` patch.


ryansolid:
> Ruling 1 tightened in 51202b78, per review: the browser is told about a continued trace **only when it was sampled**.
>
> The rule said "only when something is recording the trace" but the code tested for "a `traceparent` came in". Flags `00` is the caller saying it did *not* record — and GCP's LB, Envoy/Istio and Azure Front Door stamp `00-…-00` on every request they forward, so every app on that infra with no APM would have shipped `Server-Timing: traceparent` plus a `<meta>` in prod, and handed OTel-web's default parent-based sampler an unsampled parent that drops the pageload it records today.
>
> ```ts
> emit: (incoming !== undefined && incoming.sampled) || answered
> ```
>
> - sampled upstream → advertised (something recorded it; head-consistent for the browser)
> - unsampled upstream → continued, exposed by `getTraceContext()`, forwarded downstream as correct W3C propagation; zero wire change to the browser
> - provider answered → advertised regardless (propagating its own not-sampled decision is the provider's call — Sentry wants `sentry-trace …-0` to reach the browser)
>
> Pinned: the `-00` derivation keeps its formatting assertion; a new case runs an infra-shaped `00-…-00` through `createSSRResponse` (no header, no meta), the no-stub `commitEventResponse` exit (original `Response` returned), and then with a provider installed (emits). RFC 12 and the changeset reworded to match. Web suites, `test-types`, size gate green.
>
> — *Claude via Cursor*

PR #3403, opened 2026-09-13
perf(signals): shared traps for merge()/omit() proxies, state on the target

ryansolid:
Port of #3391 (`main`, `mergeProps`/`splitProps`) to the 2.0 primitives in `@solidjs/signals`.

## What changes

Each `merge()` / `omit()` proxy used to allocate a target holding three fresh closures (`get`/`has`/`keys`) over that instance's sources, with the shared `propTraps` forwarding every trap call through them. Now the per-instance state lives on the target under symbol keys — `{ [$SOURCES]: flattened }` for merge, `{ [$OMIT_PROPS]: props, [$OMIT_KEYS]: keys }` for omit — and each primitive has one shared handler that reads it directly.

```ts
// before: 1 proxy + 1 target + 3 closures per instance, trap → closure → sources
new Proxy({ get(p) {…}, has(p) {…}, keys() {…} }, propTraps)
// after: 1 proxy + 1 one/two-slot object, trap → sources
new Proxy({ [$SOURCES]: flattened }, mergeTraps)
```

Semantics are unchanged: same right-to-left `in` lookup, same `has`/`ownKeys` filtering, `$SOURCES` still answers the flat list on a merge proxy (what `mergeSources()`/`spread()` read) and `undefined` on an omit proxy (#3014). The state keys are never reported by `ownKeys` nor answered by `get`/`has`, so they're invisible through the proxy; a string `Object.defineProperty` on the proxy lands on the target without touching them (before, defining `"get"` would have overwritten the target's method and broken the proxy). Proxy invariants hold: the target is extensible and its state keys configurable.

## Tests

New test: state symbols not visible, `Object.defineProperty(merged, "sources", …)` / `(rest, "props", …)` don't affect reads or `in`/`keys`, live reads after a store write, descriptor getters work. Existing #3014 test and the whole merge/omit suite unchanged. Suites: signals 1768, solid 595, html 200, web dom 735 / server 795 / hydrate 169.

## Performance

Measured on battery, so only directions are reliable (±20% run-to-run). Interleaved A/B, old vs new `dist/prod` in one process, 31 alternating rounds:

| path | ratio new/old |
|---|---|
| retained 200-instance `omit(merge(defaults, store), "id")` batch | 0.87 / 0.93 (two runs) |
| nested `merge(mergeProxy, {x})` | 0.65 / 0.75 |
| `Object.keys` / `{...proxy}` | 0.90–0.98 |
| single construct, single reads, `in` | within noise (process-isolated alternating runs show no regression) |

A computed-symbol literal costs ~13 ns vs ~46 ns for the three-closure literal in isolation, so the construct win is real but small next to the Proxy itself; the retained batch is where it shows. CodSpeed's `merge-proxy-keys` / `omit-proxy` construct+read benches in `utilities.bench.ts` cover this on the PR.

### DIFF
--- a/.changeset/perf-shared-prop-proxy-traps.md
+++ b/.changeset/perf-shared-prop-proxy-traps.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/signals": patch
+---
+
+`merge()` and `omit()` proxies keep their per-instance state on the proxy target under symbol keys and share one handler each, instead of allocating a target with three closures (`get`/`has`/`keys`) per instance and forwarding every trap through them. Creating a props proxy is now a proxy plus a one- or two-slot object, and reads go straight from the trap to the sources. The state keys are never reported or answered through the proxy, and a string `defineProperty` on the proxy can no longer clobber its internals. Port of solidjs/solid#3391 to the 2.0 primitives.
--- a/packages/signals/src/store/utils.ts
+++ b/packages/signals/src/store/utils.ts
@@ -7,34 +7,101 @@ function trueFn() {
   return true;
 }
 
-const propTraps: ProxyHandler<{
-  get: (k: string | number | symbol) => any;
-  has: (k: string | number | symbol) => boolean;
-  keys: () => (string | symbol)[];
-}> = {
-  get(_, property, receiver) {
+// The merge() and omit() proxies keep their per-instance state on the proxy
+// TARGET under symbol keys and share one handler each, so creating one costs
+// a proxy plus a one- or two-slot object — no per-instance trap closures —
+// and a read goes straight from the trap to the sources. The state keys are
+// never reported by `ownKeys` and never answered by `get`/`has`, so they are
+// invisible through the proxy.
+
+const $SOURCES = Symbol(__DEV__ ? "MERGE_SOURCE" : 0);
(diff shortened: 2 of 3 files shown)


ryansolid:
> Closing: the premise doesn't hold up under CodSpeed. Against \`next\`, every shape this was meant to speed up regresses — \`omit-proxy-store\` construct −11% to −15%, \`merge-proxy-keys-store\` construct −7.5% to −9%, and \`merge-merge-signal(deep 0,100,3,2)\` at −88% (275 µs → 2.36 ms). The one win (\`merge-static(0,100)\` +27%) is on a path this PR doesn't touch, so it's noise. The local wall-clock A/B in the description was measured on battery at ±20%, which is wider than the effect it was trying to detect; the instruction-count numbers are the ones to trust here.
>
> The plausible reason: three closures over locals in a fresh object literal are cheap and monomorphic, while symbol-keyed state read off \`target\` inside a shared handler adds a keyed load per trap and turns the target's shape into something V8 handles worse than the closure literal. The nested-merge outlier suggests the flatten path over \`$SOURCES\` became a slow-property read on every hop.
>
> If someone wants to revisit, it needs to start from a CodSpeed-visible win, not a local timing run.
>
> — *Claude via Cursor*

Issue #3404, opened 2026-09-13
2.0.0-rc nested render effect cleanup run too early


GabbeV:
### Describe the bug

When a render effect reading async value is nested within another effect reading another async value cleanup sometimes run too early.

### Your Example Website or App

https://s.olid.uk/id/oHOnRtMeTDKAY2kJ-DQQ3w

### Steps to Reproduce the Bug or Issue

1. click button
2. observe how the portal disappear before later reappearing instead of atomically swapping.

### Expected behavior

The result is observably atomic once the entire update is done.

### Platform

- OS: Windows
- Browser: Firefox

ryansolid:
> Fixed by #3405 — merged into `next`.
>
> — *Claude via Cursor*

ryansolid:
> Fixed by #3405 — merged into `next`.
>
> — *Claude via Cursor*

Issue #3352, opened 2026-09-10
2.0.0-rc.7 (performance): a projection derive that writes one root key of a large draft clones the whole raw on every derive, O(keys); plain stores use a prototype overlay


thedanchez:
### Summary

`@solidjs/signals@2.0.0-rc.7`, prod and dev builds. Performance, not correctness: the output is right, the cost is the surprise.

When a `createProjection` derive sets or deletes a root-level key of its draft, the engine clones the whole raw object before the write (`ensurePB` → `cloneRaw`, which walks every own property descriptor). It does this on the first root-level write of every derive, so a projection over a large keyed record pays O(keys) per derive no matter how small the change is. A nested write on the same projection is O(1), because the nested target is small. A plain `createStore` doing the identical root delete and set is O(1) after its first write: `ensurePB` gives plain stores a prototype overlay (`Object.create(raw)`) and only falls back to the clone for projection targets.

At 20 000 root keys that is about 17 ms per derive for touching a single key.

### Reproduction (headless, only `@solidjs/signals`)

```bash
mkdir solid-projection-root-clone && cd solid-projection-root-clone
npm init -y >/dev/null && npm i solid-js@2.0.0-rc.7 >/dev/null
cat > repro.mjs <<'JS'
// A projection whose draft has many root keys: touching ONE root key per
// derive costs O(keys): the engine clones the whole raw on the first
// root-level write of a derive. A plain store deleting the same key is O(1).
// node rootclone.mjs [keys]
import { createProjection, createRoot, createSignal, createStore, flush } from "@solidjs/signals";
const KEYS = Number(process.argv[2] ?? 20000);
const ms = (fn) => { const t0 = performance.now(); fn(); return performance.now() - t0; };
const time = (label, run) => {
  const samples = [];
  for (let i = 0; i < 10; i++) samples.push(ms(() => { run(i); flush(); }));
  samples.sort((a, b) => a - b);
  console.log(`${label.padEnd(58)} median ${samples[5].toFixed(2).padStart(7)} ms   max ${samples[9].toFixed(2).padStart(7)} ms`);
};
const seed = () => Object.fromEntries(Array.from({ length: KEYS }, (_, i) => [`k${i}`, { n: i }]));
createRoot(() => {
  // 1. projection: one ROOT key deleted + re-added per derive
  const [tick, setTick] = createSignal(0, { ownedWrite: true });
  const rootProj = createProjection((draft) => {
    const i = tick();
    delete draft[`k${i}`];
[...]

PR #3431, opened 2026-09-14
fix(signals): held-input rules — a staged read enters the transaction, dependencies are the committed frame's (#3408, #3410)

### LINKED ISSUE #3408: 2.0.0-rc.8 conditional memo reveals a held signal value early
### Describe the bug

A conditional memo starts reading a signal while its update is held by an async memo. Panel shows 1 while Count and Details still show 0.

https://s.olid.uk/id/Xz1kCVNoSumAVHUJSIQJ5Q

1. Wait for initialization.
2. Click Run.
3. Panel becomes 1 while Count and Details still show 0.

### Expected behavior

Panel and Count should agree whenever Panel is visible.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Replacing `createMemo(() => show() ? count() : "hidden")` with a plain getter removes the mismatch.

Possibly related to the held-input reveal rules discussed in #3334/#3370.

Reproduced on next at 14ded248 and #3405 at 323f9adb, in development and production.

### LINKED ISSUE #3410: 2.0.0-rc.8 conditional memo disagrees with its inputs during a held branch change
### Describe the bug

`selected = createMemo(() => fixed() ? 2 : count())` shows 0 while Fixed is false and Count is 1. With those inputs, Selected should be 1.

https://s.olid.uk/id/dG5ungjvREetJHG8iUxFLA

1. Wait for initialization.
2. Click Run.
3. Count becomes 1 while Fixed remains false and Selected remains 0.

### Expected behavior

Selected should equal Count while Fixed is false.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

The async details memo holds the branch change. Updating Count during that hold exposes the mismatch. Keeping Count as an unconditional dependency removes it.

Reproduced on next at 14ded248. The earlier equivalent repro also reproduces on #3337 at 20f31433 and #3405 at 323f9adb.

ryansolid:
Fixes #3408, fixes #3410.

Two sides of one gap. The engine already says "the staged world belongs to its transaction" for **writes** (`setSignal` on a stamped node enters it) and for **stamped recomputes** (`recompute`'s head enters it). Reads and dependency trimming did not.

## #3408 — a memo that starts reading a held signal revealed its staged value

```ts
const details = createMemo(() => delay(1000, count()));       // count write → held
const panel   = createMemo(() => (show() ? count() : "hidden"));
```

`setCount(1)` is held by `details`. `setShow(true)` flips `panel` mainline; it now reads `count`, and non-stale readers are served the staged value (readers keep speculation — by design). But nothing joined `panel` to the hold, so it published a value derived from the held world into the mainline frame: `Panel: 1` beside `Count: 0`.

**Rule:** a tracked computation served a live transaction's `_pendingValue` enters that transaction (`enterStagedRead`), on each of `read()`'s value selections (both fast paths and the slow path). No-op for the ambient batch, the active transaction, and probe reads (`pendingCheckActive`). It's the read twin of the existing write-side and recompute-side entries.

## #3410 — a memo whose held pass trimmed an input disagreed with its committed inputs

```ts
const selected = createMemo(() => (fixed() ? 2 : count()));
const details  = createMemo(() => delay(1000, selected()));
```

`setFixed(true)` is held. `selected`'s held pass stops reading `count` and `trimStaleDeps` dropped the edge — so `setCount(1)` never reached a memo whose *committed* value still derived from `count`: `Count: 1` beside `Selected: 0` / `Fixed: false`.

**Rule:** dependencies are the committed frame's until the frame is replaced — the deps twin of held children (#3404). A pass that staged its value leaves the previous pass's dependency tail linked; `commitPendingNode` trims it (after a clean pass — `_error` set means the last pass threw and `_depsTail` marks where it stopped). The `count` write then dirties the stamped memo, which enters the hold, exactly as an unconditional `count()` read does. Decided at commit rather than recompute because a plain flush knows nothing at recompute time: the transaction that holds the pass may open later in the same flush. `__OBSERVE__` fan-in counting now walks only the validated prefix so held tails don't inflate distinct-source counts.

Both scenarios now reveal atomically at the landing (`tests/held-conditional-memo.test.ts`).

## Verification

- `@solidjs/signals`: 167 files / 1785 tests green; full monorepo `turbo test --concurrency 1` green for every package except 3 `@solidjs/compiler` parity tests that fail identically on `origin/next` (compiler/babel-plugin trees are untouched here).
- Core floor 23,058 → 23,181 (**+123 B, conscious bump**, noted in `treeshake.test.ts`); eight brotli caps ratcheted by 0.05 KB with notes in `scripts/size/.size-limit.js`.
- `src/` typechecks clean. No existing expectation moved.
- Docs: the two rules are user-observable semantics, so they are spec rules — **A29** (a tracked read served a live transaction's staged value enters that transaction) under *Reads and visibility* and **A30** (a memo's dependencies are the committed frame's until the frame is replaced) under *Transactions and holds* in `SPEC-ASYNC-SEMANTICS.md`, each with Status / Pinned by / Mechanism; cited by ID from `enterStagedRead`, `recompute`'s tail, `commitPendingNode` and the two pins, so the rules-index gate covers them (`RULES-INDEX.md` regenerated: 307 rules, every src citation resolves, every live A-rule test-cited). A28 is #3337's. The `INTERNALS-ASYNC-STATE.md` §3 bullets are reduced to mechanism notes pointing at the rules. Changeset included.

Rebased on `next` after #3432 / #3433.

Co-authored-by: Claude <noreply@anthropic.com>


ryansolid:
@codspeedbot fix this regression

> Requesting the flame-graph diff for `projection derive: delete + set one ROOT key (#3352)` (455.5 → 500 µs, reproduced at the same magnitude across two runs, pre- and post-rebase). Local optimized runs of base vs head show no difference on this benchmark, so the per-function instruction counts are what will settle whether it is the added `commitPendingNode` branch, `read()` losing inlining from the `enterStagedRead` calls, or the base run's GC state (the sibling `write one NESTED field` benchmark, untouched by this PR, moved ×2.4 in the same report).
>
> — *Claude via Cursor*

PR #3436, opened 2026-09-14
fix(web): dynamic() honors xmlns when creating a tag-name element (#3386)

### LINKED ISSUE #3386: `dynamic()` / `Dynamic` with a string tag: no `runHydrationEvents()`, namespace ignores parent

ryansolid:
Closes #3386 (part 2; part 1 landed in #3396).

## Problem

The compiler resolves a tag's namespace from its parent at build time. The `dynamic()` runtime path creates the element before it has a parent, so it can only infer from the tag name — and a tag that exists in both HTML and SVG (`a`, `script`, `style`, `title`) was always created as HTML. `<svg><Link href=…/></svg>` with `Link = dynamic(() => "a")` produced an HTML anchor inside the SVG tree. `@yak/solid` carries an `ambiguousSvgTags` set to route around this.

## Fix

The instance says which namespace it means with the attribute compiled JSX already uses for exactly this purpose:

```tsx
const Link = dynamic(() => "a");
<svg><Link xmlns="http://www.w3.org/2000/svg" href="/x">…</Link></svg>
```

`xmlns` joins `is` as an attribute of the element that also decides how the node is created: read once, untracked, at creation (the DOM can't re-namespace a node), then applied through `spread()` as an ordinary attribute. That keeps a client-rendered element byte-for-byte the same as its own SSR output: the server already serializes `xmlns` (as compiled SSR does for `<a xmlns>`), the parser keeps it, and hydration claims the parser-namespaced node — so the server is untouched and hydration was never affected. Without `xmlns` inference from the tag name still applies.

Deliberately *not* inferring from the mount parent: there is no parent at creation time, and any deferred scheme (create-then-reparent, a context) would tax every `dynamic()` instance to fix four tags.

## Types

`dynamic()` returns `Component<DynamicComponentProps<T>>`, which adds `xmlns?: string` for tag-name targets only. (The ambiguous tags' `IntrinsicElements` types are the HTML attribute sets, which don't include `xmlns`; SVG/MathML attribute sets already do.)

## Tests

- `test/dynamic-namespace.spec.tsx`: `namespaceURI` with/without `xmlns` inside `<svg>` and `<math>`; unambiguous tags still inferred; `xmlns` overrides inference in both directions; `<Dynamic>` parity; read-once (a later value updates the attribute, never recreates the node).
- `test/server/dynamic-xmlns.spec.tsx`: server output equals compiled `<a xmlns>` output.
- Parity-harness scenario `dynamic-xmlns-svg-link`: SSR → hydrate a `dynamic()` `<a xmlns>` inside `<svg>`, adopt-all, stable identity through an update.
- Also pins ref arrays through a spread into a `dynamic()` tag — the shape Kobalte's `solid2` branch works around with a `TODO` at rc.3 — as working on `next`.

All three `@solidjs/web` suites green (778 / 895 / 181).


ryansolid:
> Narrowed the `xmlns` typing in 919b2ab9a. The previous commit put it on `ElementAttributes`, which made every HTML tag accept it; it now lives on `SVGAttributes`/`MathMLAttributes` (as before) plus a small `AmbiguousNamespaceAttributes` mixin on the four tags that exist in both HTML and SVG — `AnchorHTMLAttributes`, `ScriptHTMLAttributes`, `StyleHTMLAttributes`, and a new `TitleHTMLAttributes` (`title` was previously bare `HTMLAttributes`). Those are the only tags where the attribute changes what gets created, in both the compiler and `dynamic()`; `<div xmlns>` is rejected again.
>
> Verified locally: `@solidjs/web` `test-types` clean, `@solidjs/h` types build, both xmlns specs pass. Changeset wording updated to match.
>
> — *Claude via Cursor*

PR #3437, opened 2026-09-14
chore(web): deprecate <Dynamic> in favor of dynamic()

ryansolid:
Deprecates the `<Dynamic>` JSX wrapper (and `DynamicProps`) via `@deprecated` JSDoc. No behavior change, no runtime warning. `<Dynamic>` stays in 2.0 — the RCs are past public API removals — but new code should use `dynamic()`.

## Why

`<Dynamic component={…}>` is the same primitive as `dynamic()` with a worse shape. Every other flow component matches its primitive structurally (`<Show>` ↔ `createMemo`, `<For>` ↔ `mapArray`) — the JSX form costs nothing. `<Dynamic>` is the exception: the tag travels in the props bag, so per instance it pays

- a `merge({ component }, rest)` the compiler emits at the call site,
- an `omit(props, "component")` inside to get the tag back out,
- a fresh `dynamic()` factory (with its factory memo) because a JSX wrapper has nowhere to hoist one.

None of that is fixable by runtime work; it's the shape. It also induces exactly this pattern in the libraries most sensitive to it: Kobalte's `solid2` `Polymorphic` does `omit(props, "as")` and then `<Dynamic {...others} component={props.as}>` — omit the tag, merge it back in under another name, omit it again. Six proxy layers per part instance, two of them this.

```tsx
// before
<Dynamic component={multiline() ? RichTextEditor : "input"} value={value()} />

// after — hoist per component instance, or per module for a constant tag
const Field = dynamic(() => multiline() ? RichTextEditor : "input");
<Field value={value()} />
```

The factory form also makes the lifetime explicit — where the memo lives, that `source` is a function — which is what `<Dynamic component={comp}>` with an uncalled accessor silently gets wrong.

## Changes

- `Dynamic` / `DynamicProps` marked `@deprecated` in `index.ts` and `index.server.ts` with the migration in the JSDoc.
- `documentation/solid-2.0/03-control-flow.md` and `MIGRATION.md`: `<Dynamic>` no longer described as "unchanged"; moved from the removals table to a new *Deprecated (kept in 2.0)* table, summaries updated.
- `documentation/tsrx/frontend-notes.md`: records that the experimental tsrx frontend is the one compiler path that *emits* `<Dynamic>` (for `<{expr}>` dynamic tags) and should retarget to `createComponent(dynamic(() => expr), props)` in both compilers so compiled output does not depend on a deprecated API.
- README package table.

Not in this PR: the tsrx lowering retarget (follow-up; no removal is planned).


ryansolid:
> Corrected in 461cab07a: this PR had claimed `<Dynamic>` would be removed before 2.0 stable. That was wrong — the RCs are past public API removals. `<Dynamic>` is deprecated (`@deprecated` JSDoc, no runtime warning) and stays in 2.0; new code should use `dynamic()`.
>
> Dropped the removal claim from the JSDoc on `Dynamic`/`DynamicProps` (client and server), the changeset, `MIGRATION.md` (three sites), the control-flow RFC (moved from the *Removals* table into a new *Deprecated (kept in 2.0)* table), and the tsrx frontend notes (the lowering retarget is now a recommendation rather than a prerequisite for removal). PR body updated to match.
>
> — *Claude via Cursor*

Issue #3407, opened 2026-09-13
2.0.0-rc.8 shared render effect entangles a synchronous action with an unrelated pending update


GabbeV:
### Describe the bug

A synchronous action gets held by an unrelated pending update when both are read in the same render effect.

Splitting the expression into separate JSX holes, or calling setB directly instead of through action, removes the wait.

### Your Example Website or App

https://s.olid.uk/id/P_RewMzJQ9ylqeKOIsYEmw

### Steps to Reproduce the Bug or Issue

1. Click Start.
2. B is set after 500ms, but the output stays at `0:0` until becoming `1:1` at 2 seconds.
3. Replace the paragraph with `<p>{b()}:{detailsA()}</p>` and rerun. It now shows `1:0` at 500ms.

### Expected behavior

B should publish at 500ms, showing `1:0`, then `1:1` when A finishes. Sharing a render effect should not entangle the updates.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

### Additional context

Possibly related to #3322.

The merge happens when `recompute` re-enters the shared effect's `_transition` while another transition is active. Excluding effects from that re-entry fixes this repro, but breaks existing reveal-completion tests, including cases covered by #3305/#3334. Those suspended reveals still need to resume correctly.

ryansolid:
> Confirmed and fixed in #3434 — thanks for the plain-vs-action contrast, it pointed straight at the cause.
>
> Root cause: `recompute` re-entered a stamped **effect**'s transaction whenever any other transaction was active — at the head of the pass, before anything could have been observed. The counter's render effect was stamped by the held `detailsA` flight, so the `action`'s sync write merged into that flight's transaction and waited; the plain write had no transaction to merge and passed through.
>
> Rule now: a render effect's pass belongs to whatever dirtied it. The stamp re-entry is memo-only (a memo's value *is* its transaction's work); an effect dirtied by another writer runs that writer's pass, reads the held flight as a stale reader (its committed value, coherent with the flight's still-committed inputs) and publishes with the writer. Only a pass that actually observes a flight pending joins it. The compiler's grouping of bindings into one effect is never a reason for one binding to hold another. Both variants of your repro now publish `1:0` at the write and `2:1` at the landing.
>
> — *Claude via Cursor*

ryansolid:
[...]

PR #3439, opened 2026-09-14
fix(signals): an effect's dependencies are the committed frame's until its run applies (#3438)

### LINKED ISSUE #3438: Conditional JSX stays stale during a held branch change
### Describe the bug

While an async memo holds show at true, updating count displays Count: 1 but leaves `show() ? count() : "hidden"` displaying 0.

https://s.olid.uk/id/LTuZMI4_S8SJuZxMpv8QSQ

1. Wait for initialization.
2. Click Run.
3. After 500ms, Count becomes 1 while Show remains true and Panel remains 0.
4. After another 500ms, Show becomes false and Panel becomes hidden.

### Expected behavior

Panel should agree with Count while Show is true.

### Screenshots or Videos

_No response_

### Platform

### Additional context

Reproduced on next at efcaebad, after #3431 merged, and on #3337 at 20f31433.

Related to #3410, but this is an inline JSX conditional. Wrapping the conditional in createMemo avoids the mismatch on next.

Unlike #3412, this needs no Loading boundary or reset; Show still correctly displays true during the mismatch.

brenelz:
Fixes #3438.

The effect arm of A30 (#3410, landed in #3431).

## The gap

```tsx
const delayedShow = createMemo(() => delay(1000, show()));   // show write → held
<p>Panel: {show() ? count() : "hidden"}</p>                   // an insert effect
```

The compiler emits the ternary as an insert effect (only `!!show()` is memoized; the branch body runs in the effect's compute). `setShow(false)` is held by `delayedShow`. The effect's pass on the plain flush direct-committed its value and `trimStaleDeps` dropped the `count` edge right there — while the same flush then pended the async memo, became a hold, and stashed the effect's run. The DOM kept the count-derived frame with `count` no longer an input, so the later mainline `setCount(1)` never reached it: `Count: 1` beside `Panel: 0` while `Show` still read `true`.

#3431 covered memos only: a memo always stages `_pendingValue`, so its trim was moved to `commitPendingNode`. An effect on the plain path writes `_value` directly and owes a run instead — its frame is the run, not the value slot.

## Rule

An effect's dependencies are the committed frame's until its run applies. `recompute`'s tail leaves an effect's dependency tail linked while a run is owed (`_modified`), and `runEffect` trims it once the run applies (after a clean compute pass, as the commit does) — the twin of `commitPendingNode`'s trim for a staged memo pass. A pass that changed nothing owes no run and trims at its tail as before.

The `count` write now reaches the effect. A render effect is a mainline reader of the foreign hold (it is served the committed `show`), so it re-derives `Panel: 1` beside `Count: 1`, and the hold's landing reveals `hidden` with `Show: false`:

```
2500: Count: 1 | Panel: 1
3000: Delayed: false | Panel: hidden | Show: false
```

That differs from the `createMemo`-wrapped conditional, where the count write joins the hold (A29); both frames are coherent, and the shapes differ because effects render mainline by design. Documented under A30 in `SPEC-ASYNC-SEMANTICS.md`.

The attribution engine's subscription diff (`captureDeps`) and wide-scope check walk the validated prefix `[_deps.._depsTail]` only, so a deferred tail is not reported as a dropped or extra dependency (the `__OBSERVE__` fan-in count already did this).

## Verification

- `@solidjs/signals`: 171 files / 1797 tests green, including the new `tests/held-conditional-effect.test.ts` (pins the frames and the dependency list through the hold) and the existing #3408/#3410 pins.
- `@solidjs/web`: 84 files / 770 tests green, including the new `test/conditional-jsx-held-branch-3438.spec.tsx` — a jsdom port of the issue's playground (real click, DOM sampled through the hold) plus the memo-wrapped control; fails on `next` under both the native and Babel compilers, passes here.
- Core floor 23,318 → 23,353 (+35 B, noted, no bump). Three brotli caps ratcheted by 0.05 KB with measurements against a pristine `next` build (`+ createStore` +56, hydrating-no-stores +64, CSR +6); the other scenarios are flat or smaller.
- Rules index regenerated (`--check` clean; the one unresolved test id is pre-existing on `next`). Changeset included.


ryansolid (approved):
> Approved — this is the right fix, and the predicate is the important part.
>
> I had the same diagnosis from the compiled output (the ternary is an insert effect; A30's deferred trim keyed on `_pendingValue`, which an effect doesn't have) and was about to gate the deferral on `isEffect && activeTransition !== null` at the pass. That would have missed this exact case: the effect's pass runs mainline and the transaction opens *later in the same flush*, when the render reader of `delayedShow` observes it pending and the batch is adopted — which is why the run was stashed even though the pass wasn't. "A run is owed" (`_modified`) with `runEffect` trimming once the run applies is the correct statement of "the frame is the run, not the value slot," and the exact twin of `commitPendingNode`'s trim for a staged memo pass. The `captureDeps` / wide-scope walk on the validated prefix matches what the fan-in count already did.
>
> Resulting shapes, for the record: inline effect → `Count: 1 | Panel: 1` at the write, `hidden | Show: false` at the landing (a mainline reader of a foreign hold, A15's stale-reader term); memo-wrapped → the `count` write joins the hold (A29) and reveals at the landing. Both coherent; effects render mainline by design (#3407).
>
> One housekeeping commit pushed to the branch: `docs/RULES-INDEX.md` had been regenerated without `pnpm format`, which strips prettier's table alignment (~360 lines of separator churn that the next regeneration would flip back). Formatted; no other file was affected.
>
> — *Claude via Cursor*

PR #3450, opened 2026-09-15
fix(signals): two flights through one memo settle as one unit; lane-dirtied zombies run (#3443, #3444)

### LINKED ISSUE #3443: Memoized sum disagrees with its async inputs during overlapping updates
### Describe the bug

Two async memos feed a synchronous sum memo. Updating their inputs 500ms apart makes A display 1 while B and Sum both display 0.

Replacing the sum memo with a plain getter removes the mismatch.

https://s.olid.uk/id/WXOvPfyQQ6qoLawfAJMMSw

1. Wait for initialization.
2. Click Run.
3. After one second, A becomes 1 while B and Sum remain 0.
4. After another 500ms, B becomes 1 and Sum becomes 2.

### Expected behavior

Sum should equal the displayed A + B throughout the update.

### Screenshots or Videos

_No response_

### Platform

### Additional context

Reproduced on next at c452850d, #3439 at c2c9789b and #3337 at 20f31433.

Both updates use ordinary signals. No isPending reads, conditional dependencies or observer changes are involved.

Writing both inputs in the same event, without the await between them, removes the mismatch.

### LINKED ISSUE #3444: latest inside Show stays stale while removal is pending
### Describe the bug

Two paragraphs read `latest(count)`, one inside Show and one outside. While an action is pending, Outside becomes 1 but the still-visible Inside remains 0.

https://s.olid.uk/id/r5C3uxJ0RRCPg0eB-PXsWA

1. Click Run.
2. Outside becomes 1 while Inside remains visible showing 0.
3. After one second, Inside disappears.

### Expected behavior

Both paragraphs should show the same latest value while they are visible.

### Screenshots or Videos

_No response_

### Platform

### Additional context

Reproduced on next at a0d6dd24, after #3439 merged. Also reproduces on #3337 at 20f31433.

Possibly related to #3438, but replacing Show with `{count() === 0 ? <p>Inside: {latest(count)}</p> : null}` avoids this mismatch. Keeping Show mounted with `when={true}` also makes both values update correctly.

ryansolid:
Fixes #3443. Fixes #3444.

## #3443 — a shared memo disagrees with its async inputs during overlapping updates

```tsx
const a = createMemo(() => delay(1000, inputA()));
const b = createMemo(() => delay(1000, inputB()));
const sum = createMemo(() => a() + b());
// setA(1); await delay(500); setB(1);
```

Before: `3000: A: 1` alone, `3500: B: 1 | Sum: 2`. Traced with the reporter map: `setB` opens T3 and `b`'s pending **propagates** onto `sum` — a memo T2 holds — without recomputing it (its inputs' values are unchanged), so the memo's stamped re-entry never runs and T2 never learns it is waiting on `b`. `sum`'s reader registers on `b` in *T3* (INV-3 is keyed by transaction); T2's verdict at `a`'s landing sees only its own reporters, commits, and reveals `A: 1` beside the committed `Sum: 0`.

Fix: `notifyStatus`'s dependent walk, on a pending propagation reaching a memo stamped by another live transaction, enters that transaction (`initTransition(sub._transition)` — a merge, or with nothing active an entry that adopts the ambient batch). This is A15's headline case (a shared derivation of both flights settles as one unit) applied at the one moment it is known. Effects are skipped: an effect entangles nothing by itself (A15 shared-hole corollary, #3407) — the plain-getter shape stays parallel (`3000: A: 1 | Sum: 1`, `3500: B: 1 | Sum: 2`), pinned alongside.

Result: `3500: A: 1 | B: 1 | Sum: 2`.

**Semantic to note:** the write that starts the second flight is held with the first when its async work flows into a held memo — `page=1` waits with `count=1` while `details` re-asks, even though `Sum` reads `page` plainly. Pinned. The existing #3375 boundary-reset pin moves accordingly: the reset now publishes `Sum: 2` once instead of the two-pass `Sum: 1 | Sum: 2` its comment described as a drain artifact; every other frame in that pin is unchanged.

## #3444 — `latest()` inside a `Show` stays stale while removal is pending

When the parking batch *is* the transaction, `flush` cancels the zombie recomputes its staged writes queued (`cancelZombieRecompute`) — a zombie renders mainline until the commit disposes it, and the staged world is one it never displays. A zombie dirtied through the **lane channel** (`REACTIVE_OPTIMISTIC_DIRTY`: an override, or the `latest()` companion the write synced) is the exception: the lane's values are the mainline frame, so the still-visible branch showed `latest(count)` at 0 beside the same read outside at 1. Such a zombie now runs; its pass runs under the lane and its run lands on the lane's effect queue, so a held lane defers it as it defers every other reader's.

Result: `500: Inside: 1 | Outside: 1`, `1500: Inside gone`.

## Verification

- `@solidjs/signals`: 173 files / 1804 tests green on top of #3445. New `tests/overlapping-flights.test.ts` (#3443 shape, the effect-arm control, the held-second-write consequence) and a #3444 pin beside #3404 in `tests/nested-render-effect-async-cleanup.test.ts`.
- Docs: A15 status/pins/mechanism amended for #3443 with the consequence stated; INTERNALS bullets for both (§3 Transitions, held-children addendum); rules index regenerated (`--check` clean). Changeset included.
- Floor 23,365 → 23,419 (+54 B, conscious bump to 23,450); three brotli caps ratcheted 0.05 KB with measurements (`+ createStore` 15679, `isPending/latest` 11066, observe tier 15681); every other scenario within cap.


ryansolid:
> Pushed a cap bump on top of 42632d57c: the holds carve-out tips the `signals: core floor` scenario to 8718 B against `next`'s 8691, 18 B over the 8.70 KB cap (`check` failed on it). Ratcheted to 8.75 KB with the measurement in the note; the comment's "no bump, measured at 8690" described the stamp-only first cut. All ten scenarios pass locally.
>
> Reviewed the carve-out itself: keying on `STATUS_PENDING` / a staged `_pendingValue` rather than the stamp matches the #3334 precedent, and the check on `sub` runs before the recursive `notifyStatus(sub, …)` that would mark it, so it reads the memo's pre-propagation state. Locally on this head: signals 174/1895, web 86/783 (args-switch-gate 17/17), solid 27/618, rules-index current.
>
> — *Claude via Cursor*

PR #3479, opened 2026-09-15
fix(signals): five hold-consistency fixes — re-park sweep, first observer, lane outside view, zombie readers, unchanged-pass trims

### LINKED ISSUE #3456: Conditional JSX stays hidden after an async branch is cancelled
### Describe the bug

Switching away from a pending async branch leaves the panel displaying `"hidden"`, even after all async work finishes and `count()` displays `2`.

Removing `async` from `selected` avoids the issue.

https://s.olid.uk/id/rBKUu262Tm2wuvqomN-GUA

1. Wait for initialization.
2. Click Run.
3. Wait for the async work to finish. Count shows `2`, but Panel stays `"hidden"`.

### Expected behavior

Panel should show `0`.

### Screenshots or Videos

_No response_

### Platform

Verified with compiled JSX in jsdom, in development and production builds.

### Additional context

Reproduces on next at d80cd1f6, including the merged fixes in #3371 and #3439. Might be related to their pending-recovery or conditional-dependency handling.

### LINKED ISSUE #3458: Revealing a pending async memo displays its old value alongside updated inputs
### Describe the bug

Revealing B while A is pending lets Count and A publish `1` while the visible B still shows `0`. Both async memos derive directly from Count.

https://s.olid.uk/id/Laazgf4sR6axjK-j7u678g

1. Wait for initialization.
2. Click Run.
3. After about one second, Count and A show `1` while B shows `0`.

### Expected behavior

B should remain hidden until it can display an answer consistent with Count, or the update should remain held.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at 0d8347ad.

### LINKED ISSUE #3460: Existing and newly revealed latest readers display different values
### Describe the bug

A newly revealed call to `value()` displays `1` while an existing call to the same getter still displays `0`. `value` calls `latest(count)`; its async derivation is still pending.

https://s.olid.uk/id/vELCzqgoTuqSuxG63N9TwQ

1. Wait for initialization.
2. Click Run.
3. After half a second, Late shows `1` while Value and Details still show `0`.

### Expected behavior

The existing and newly revealed calls to the same latest getter should display the same published value.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at 0d8347ad.

### LINKED ISSUE #3463: Pending removal releases latest while its async reader is still visible
### Describe the bug

Requesting removal of Details during an action lets Value publish `1` while Details still displays `0`. The removal has not committed: Show remains `true` and Details is still visible.

https://s.olid.uk/id/1p837acQT8GtS13XyPDEfA

1. Wait for initialization.
2. Click Run.
3. Value shows `1` while Show is still `true` and Details still shows `0`.

### Expected behavior

While Details remains visible, it should stay consistent with Value. Requesting its removal should not stop it from blocking publication before the removal actually becomes visible.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Reproduces on next at 0d8347ad.

### LINKED ISSUE #3469: Same-value branch switch leaves a synchronous memo inconsistent with its inputs
### Describe the bug

While B is held at `0`, updating A to `2` leaves Selected showing `1`.

Selected is `b() ? b() : a()`, so with the displayed inputs A=`2` and B=`0`, it should select A.

https://s.olid.uk/id/qbPG6OYNRoG2FZgYcpia-Q

1. Wait for initialization.
2. Click Run.
3. After half a second, A shows `2`, B shows `0`, and Selected still shows `1`.

### Expected behavior

Selected should remain consistent with the published inputs. Either A should remain held, or Selected should update alongside it.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox

### Additional context

Related to #3461, fixed by #3464. This version uses a synchronous memo, and switching branches initially produces the same value (`1`). A subsequent update to the previously selected source exposes the inconsistency.

Reproduces on next at 347a5ca7.

ryansolid:
Batch 2 of @GabbeV's hold-consistency reports, audited against `next` after #3473. Each reproduced at the core level and pinned; every issue pin fails on the base commit (verified), the twin pins (existing behavior) pass on both.

Closes #3456, closes #3458, closes #3460, closes #3463, closes #3469.

## Ruling applied (#3460)

Maintainer: *"a held lane is basically a micro transition from the outside… we wouldn't hold a sync write on a transition. Lanes are the same."* A render effect **off** a held lane is a stale reader of it: it sees the committed value (which is what is on screen — the lane defers its own readers' runs), publishes now, entangles nothing, and re-runs at the release. `latest()` is not special — same rule for a `createOptimistic` source. Inside, a reader on the lane sees the lane's values and the parent transaction's landings as before.

## Fixes

| Issue | Rule | Mechanism |
|---|---|---|
| #3456 conditional stays hidden after its async branch is cancelled | pending-source model | `recompute`'s catch settles every source the earlier pass carried that the re-parked set no longer contains (the re-park twin of the unchanged-value recovery sweep) |
| #3458 revealing a pending memo shows its old value beside updated inputs | A15 reveal corollary, first-observer arm | `heldFromStale`: when the transaction has no entry for the flight, the reader (its first observer) notifies the pending status up its queue chain under the transaction — INV-3, the one registration site |
| #3460 existing and newly revealed `latest` readers disagree | A15 lanes corollary, A17 | `readsHeldCommitted` (lanes.ts) from `read()`'s override arm (`CONFIG_HAS_LANE`-gated hook) and `latestRead`; the release re-run rides the lane's own render queue |
| #3463 pending removal releases `latest` while its reader is visible | A15 | `reporterBlocksSource(reporter, source, verdict)`: a zombie is live until the commit that disposes it; moot only for the verdict of the transaction staging its removal. `sourceObserved` keeps a zombie the verdict passed over |
| #3469 same-value branch switch leaves a memo inconsistent | A30 | `heldTrims`: an unchanged pass's dep trim waits on the flush's verdict (drained at commit, dropped on park); creation / OPT-dirty / tracked-effect passes trim at once |

## Semantics check

- #3469's memo arm resolves to the *held* outcome (`A: 2` reveals with `B: 1`) — the memo is served the staged `b` and enters the hold (A29). This is the pre-existing `sum = a() + b()` behavior on `next`, not a new shift; the report accepts either outcome. The render-effect arm follows its inputs at once (`Selected: 2` on the write, `1` at the commit).
- The body-end supersession pin (A18) is unaffected: a demoted lane is not held, so the override is the visible value.
- No existing test expectation changed.

## Docs

A15 (lanes corollary extended; first observer; zombie readers), A17 (every render effect off the lane sees the committed view), A30 (#3469 paragraph); INTERNALS §2/§3 bullets; RULES-INDEX regenerated.

## Size

Core floor 24,478 → 24,873 (+395; ceiling bumped to 25,000 with the seams listed in `treeshake.test.ts`). All five are core-retained; the #3460 rule itself lives in `lanes.ts` and sheds with the engine. One golf pass applied (`verdict` passed through as the transaction, no ternaries).

## Tests

signals 190/190 files, solid 28/28, web 93/93 against the rebuilt `dist/`.

New pins: `tests/pending-source-repark.test.ts`, `tests/first-observer-stale-reader.test.ts`, `tests/lane-outside-view.test.ts`, `tests/held-frame-dependencies.test.ts`.

## Review follow-up: lanes stage (third commit)

@GabbeV's differential fuzzing on the first commit found `1:0` — an optimistic value that never finished preparing revealed beside its old derivation — and, once that was fixed, a torn-tuple class in the `latest` cohort: an outsider mounted mid-hold saw the source's committed `0` beside a derivation's speculative `1`. The #3460 rule needs the committed view to be a *whole frame*, and the lane mechanism couldn't serve one because lane members direct-committed their speculative results into `_value` (INV-11's "lane direct commit").

Ruling: **an optimistic derivation is an override.** A lane pass on a memo now publishes into the override slot as a *derived override* (`laneOverride`, `CONFIG_DERIVED_OVERRIDE`); `_value` stays committed. The lane's own readers and direct reads see the optimistic frame, a reader off the lane sees the committed one, neither torn. The revert *promotes* a derived override the truth confirmed (no re-ask waterfall after the reveal) and drops one it superseded. A derived override is a plain member everywhere a written override means intent (merging, pending propagation, body-end supersession, the authoritative-blockage census, the hold census).

Two corollaries the fuzzer forced: off the lane is *provenance, not membership* (a pass under the lane's own transaction — its async's landing re-entering it — is the lane's work), and every pass over a live lane member carrying a derived override is the lane's pass whatever channel dirtied it (a boundary reset re-ran a lane-born memo plainly; A18's sync twin took its re-derived lane view for a differing truth and demoted the lane — `latest-1 #2481`, the boundary published `undefined`).

Fuzzer (8 cohorts × 2 seeds × 5000, this tree vs `next`): `latest` S1 published-derivation disagreements 57→5 / 52→6, torn tuples to an effect 26→0; `branches` S1 13→2 / 16→2, required-reader L1 6→0 / 10→2; everything else flat. Six new O1 findings in `latest` are the fuzzer's model disagreeing with the #3460 ruling (a ready optimistic source not publishing to an *outsider* while the action is open) — that is the rule, not a regression.

Size: isPending/latest 11.90→12.05 KB, store app 29.60→29.75, hydrating 19.60→19.70 (brotli scattered a ~40 B branch ±50 B across scenarios). A17 and INV-11 amended; two new pins in `lane-outside-view.test.ts`, plus the #2481 one.


GabbeV:
> AI-assisted differential fuzzing found a regression on `3fb38bfe` compared with `next` at `1de80f40`. Independently reproduced outside the fuzzer in development and production, including compiled JSX in jsdom.
>
> **An optimistic value that never finished preparing becomes visible when the action ends, alongside its old derived value.**
>
> ```tsx
> import { render } from "@solidjs/web";
> import { action, createMemo, createOptimistic, createSignal } from "solid-js";
>
> const delay = <T = void,>(ms: number, value?: T) =>
>   new Promise<T>(r => setTimeout(r, ms, value));
>
> export default function App() {
>   const [source, setSource] = createSignal(0);
>   const [value, setValue] = createOptimistic(source);
>   const fast = createMemo(async () => value());
>   const details = createMemo(() => delay(1500, fast()));
>
>   const save = action(function* () {
>     setValue(1);
>     yield delay(500);
>     setSource(0);
>   });
>
>   return <>
>     <button onClick={() => save()}>Run</button>
>     <p>{`${value()}:${details()}`}</p>
>   </>;
> }
>
> render(() => <App />, document.getElementById("root")!);
> ```
>
> Wait for `0:0`, then click:
>
> - **next:** stays at `0:0`.
> - **This PR:** shows `1:0` when the action finishes, then returns to `0:0` when the corrective details resolve.
>
> The optimistic `1:1` never committed. Its details are still pending when the action ends, so revealing only the optimistic input produces a mixed frame. This persists across tasks; it isn't just an intermediate state within a flush.
>
> In a diagnostic build, making `readsHeldCommitted()` return `false` removes this regression. It also removes 75 of the 77 differential findings from the 95,000-case campaign, which suggests this path is a useful starting point. That disables both the committed-view selection and release re-enqueueing, so it doesn't yet distinguish which part needs adjusting.
>
> This isn't a suggestion to remove the new outside-lane rule. The intended stale-reader behavior makes sense; the problematic consequence here is the extra reveal at action completion exposing an optimistic frame that never became ready.

ryansolid:
> Fixed in 3e80a2559 — thank you for the fuzzer, it earned its keep twice here.
>
> Your `1:0` frame: the landing of `fast` re-entered the lane's transaction with no ambient lane, and the #3460 outsider rule read that pass as a stale reader — it published the committed view and queued a replay that later revealed `1` beside the stale `0`. Off the lane is now provenance, not membership: a pass under the lane's own transaction is the lane's work.
>
> With that fixed, the `latest` cohort surfaced the deeper issue: an outsider mounted mid-hold saw `latest(count)`'s committed `0` beside a memo's speculative `1`, because lane members direct-committed their results into `_value`. The committed view #3460 promises has to be a whole frame, so lanes now stage — a lane pass on a memo publishes a *derived override*, `_value` stays committed, and the revert promotes it when the truth confirmed it. Your cohort's S1 count went 57 → 5 on seed 1 and 52 → 6 on seed 2, torn tuples to an effect 26 → 0. The last one it found (latest-1 #2481, a boundary reset re-running a lane-born memo, publishing `undefined`) is fixed and pinned too.
>
> Six new O1 findings in `latest` are the model disagreeing with the #3460 ruling itself — a ready optimistic source does not publish to a reader *off* the lane while the action is open. If you think the model has the better of that argument, say so; it is the ruling I'd want challenged, not the mechanism.
>
> — *Claude via Cursor*

PR #3481, opened 2026-09-15
feat(solid,web): the server error hook — configureServerErrors / onServerError (#3468 part 3)

### LINKED ISSUE #3468: SSR serializes <Errored> errors verbatim: production page load ships what the server-function wire sanitizes
## Summary

A plain error caught by `<Errored>` during SSR is serialized to the client verbatim (`message`, `cause`, own enumerable props; only `.stack` is stripped). The same error thrown by a server function over HTTP is sanitized to `"Internal Server Error"` outside the dev build (#3116). A `"use server"` function called in-process during SSR never touches dispatch, so the production page load leaks what the RPC wire withholds. Reported with a wire-level reproduction in #3415.

## Where

`packages/solid/src/server/signals.ts`, `createErrorBoundary`:

```ts
const serializeError = (err: any) => {
  if (ctx && boundaryId && !runWithOwner(owner, () => getContext(NoHydrateContext))) {
    ctx.serialize(boundaryId, err);
  }
};
```

`hydratedCreateErrorBoundary` (`packages/solid/src/client/hydration.ts`) loads the record and rethrows it so the fallback hydrates against the same `err()`. The SSR codec (`packages/web/serialization/src/serializer.ts`) strips only `Feature.ErrorPrototypeStack`, keyed on `NODE_ENV`.

Other routes with the same leak: a `createAsync` rejection serialized into the stream (`ctx.serialize(id, deferred.promise)` rejects with the raw reason), and Errors nested in serialized data.

## Constraint

The fallback is rendered on the server *with* the error. Whatever the client receives must be what the server rendered with, or the fallback mismatches on hydration. So the boundary must sanitize before `renderFallback`, not just before `serialize`.

## Proposed fix

1. **Boundary**: outside the dev build, replace an unbranded error with a generic `Error` before rendering the fallback and serializing. `markSafeError` (`Symbol.for("solid.SafeError")`, a registered symbol — no `@solidjs/web` dependency needed) passes through. Dev/prod line is the build variant (`server.dev.js` behind `development`), matching `sanitizeServerError`. The original stays server-side: `SSR_RENDER_ERROR_CONTAINED` already carries it; add an `SSR_ERROR_SANITIZED` finding mirroring `SERVER_FN_ERROR_SANITIZED`.
2. **Serializer backstop**: apply the same policy to every `Error` the SSR serializer encodes (rejected promises, nested values), so the boundary is not the only guard — the shape of the result-graph walk in #3113.
3. **Policy hook (follow-up, separate issue if preferred)**: a single server-side error hook, configured once, that every runtime-handled error passes through — contained SSR render error, rejected async source, server-function throw (direct or HTTP), uncontained render error — receiving `{ error, kind, boundary | functionId, event }` and returning the wire value (undefined → default sanitization). This is the reporting seam as well as the mapping seam: today "report" exists only on the observe tier's diagnostics channel and "map" only for server functions via `wrapInvocation`. Needs once-per-error semantics (a direct server-function call that fails during SSR is both a server-function error and a contained render error; a `<Loading>` re-pull recurs the same throw). Two tiers, matching `wrapInvocation`: an ambient once-per-process registration (`configure*` in `@solidjs/web` — the only tier that sees direct in-process calls during SSR, per the existing entry-only caveat on the per-handler `wrapInvocation`), overridable per request. For `vite-plugin-solid` that means the ambient call goes in the `serverFunctions.configure` module and the per-request form rides the generated handler's existing `handleRequest(request, { nonce, serverFunctions, ... })` options bag, threaded into the generated `render(request, context)` via `context` like `clientEntry` is today — no new plugin option needed. Note the plugin's generated `DefaultErrorBoundary` does `console.error(props.error())` in the fallback, which today runs in the browser during hydration with the raw serialized error.

## Tests

- SSR document for an `<Errored>` catching `new Error("secret")` in the prod build carries no `"secret"`; fallback markup and serialized record agree; hydration is clean.
- `markSafeError` passes through on the SSR path with own-properties.
- Rejected `createAsync` serialized into the stream carries the sanitized reason.
- Dev build keeps full fidelity.

Refs: #3415, #3116, #3113, #3152, #3414.

ryansolid:
Closes #3468 (part 3). Resolves the Sentry plan's open question 4 (C6). Follows #3477.

## The gap

A production build has no `OBSERVE`. Everything the server runtime *handles* — an `<Errored>` fallback rendered, a `<Loading>` fragment rejected and re-rendered by the client, a server-function throw sanitized onto the wire — was recorded only as observe-tier findings; the one failure a prod consumer could hear was the one that fails the request (`renderToStream`'s `onError`). #3477 made the *mapping* uniform; this makes the *reporting* uniform on the same once-per-error spine, and makes the mapping overridable.

## Surface

```ts
import { configureServerErrors } from "@solidjs/web";

configureServerErrors({
  onError(error, { kind, handling, boundary, functionId, direct, ownerPath, event }) {
    Sentry.captureException(error, {
      mechanism: { type: `solid.${kind}.${handling}`, handled: handling !== "failed" }
    });
    // return nothing → default wire policy; return a value → the client receives it
  }
});
// per request, overriding the ambient hook:
renderToStream(code, { onServerError });          // renderToString too
handleServerFunctionRequest(request, { onServerError });
```

- **Once per error object, at first sight.** `kind: "render"` → `fallback` (the boundary's hydration id, `ownerPath` when compiled in), `client` (a fragment rejected; client re-renders), `failed` (request fails; `onError` hears it too). `kind: "server-function"` → `thrown` (`functionId`; `direct: true` for an in-process call during SSR), `channel` (a failure escaping through the result graph). `event` is the request when there is one.
- **Return = wire value**: rendered into the fallback, serialized for hydration, sent as the RPC error. `undefined` keeps the default (generic outside dev, fidelity in dev). A returned value is intent — like a `wrapInvocation` mapping — and is not sanitized again. Ignored for `failed`.
- **Once means once across roads.** A direct server-function call that throws during SSR is met first by the invocation (`server-function`, `direct: true`), then by the `<Errored>` that contains it — which reuses the verdict (same replacement in fallback and record) and does not report again. One `captureException` per failure; the observe tier keeps the multi-event picture.
- **Two tiers, as `wrapInvocation` has.** Ambient, registered on `globalThis` under `Symbol.for("solid-js/server/errors")` so a bundled server build and an `--import`ed module share it (J.24's lesson) — the only tier that sees direct calls. Per request, winning over it. A throwing hook is reported and treated as silent. Fires in every tier; `onError` keeps its meaning.

## How it's wired

- `solid-js`'s server entry owns the **verdict engine** — `reportServerError` (hook call, once; per-request `errorPolicy` on the SSR context → ambient slot; `ownerPath` from the owner chain, `event` from the request store) and `ssrSanitizeError` (mapping → cached verdict → default) — reached by the runtimes through `solid-js/internal`. The SSR context gains `errorPolicy` and `flushed`.
- Sites: `<Errored>` `handleError`; `<Loading>` `finalizeError` (post-flush reports `client` *before* `done`, so the `_fr` rejection and a transport sink's error chunk carry the decided verdict; pre-flush the parent boundary or `failed` is the sight); `failRootRender`; server functions' thrown tail, the channel guard sites and the encode trailer; the direct-call leg (`reportDirectFailure`: report, rethrow the original).
- **One ordering the design didn't foresee**, fixed: the `_fr` rejection reads its verdict at delivery, so a pre-flush failure met next by the parent boundary carries the parent's answer. A rejected async source's *own* serialized rejection is encoded in the rejection's microtask, ahead of the boundary, and carries the default policy's value — documented on `ServerErrorHook`; no tick is added to the error path to make the two agree (a first cut deferred it a macrotask; reverted in `10401f1af`).
- `SSR_ERROR_SANITIZED` carries `data.wire`. Client entry exports the types and a no-op `configureServerErrors` for isomorphic setup code.

## Tests

`server-error-hook.spec.tsx` (12): fallback with full context; no-return default; per-request over ambient (render and handler); throwing hook; non-function refused; post-flush fragment → one `client` call with the `_fr` rejection carrying the mapping; sync failure under `<Loading>` inside `<Errored>` → one `fallback`; root failure → one `failed`, `onError` still called, return ignored; HTTP `thrown` with `functionId`/event; `channel`; **direct call inside `<Errored>` → exactly one call as the function's failure, the boundary rendering the hook's answer**. The #3477 artifact fixture gained a hook scenario asserted in prod, observe and dev (22 total) — `ownerPath` present only where owners are labelled (not in prod; correct by construction and worth a line in the adapter's expectations).

All suites green (web 823/980/185, solid 620), type tests green, size caps unchanged.

## Not in here

The client twin — `render`/`hydrate` `{ onError(error, { kind: "caught" | "uncaught", ownerPath }) }` for the `<Errored>` catches the browser swallows in prod — same shape, next.

RFC 12 (new subsection), RFC 10, RFC 08, the reactivity-diagnostics skill, changeset.


ryansolid:
> Pushed `10401f1af`: the channel guard resolves in the rejection's own microtask again (as #3477 had it). The first cut deferred a rejected async source's wire verdict a macrotask so the hook's mapping could reach the channel — a tick on the error path (rejection script, `deferStream` pre-shell hold, completion) that is not acceptable for a road the hydrating client does not render from. The limit is documented on `ServerErrorHook`; the boundary's record and fallback carry the mapping. The PR body's "two orderings" paragraph is now one: only the lazy `_fr` verdict stands.
>
> — *Claude via Cursor*

Issue #3389, opened 2026-09-12
Perf tracker: Solid primitives vs the hand-rolled `@yak/solid` runtime


ryansolid:
[DigitecGalaxus/next-yak#644](https://github.com/DigitecGalaxus/next-yak/pull/644) rewrote the `@yak/solid` `styled()` runtime and made it 12× faster on SSR and ~2× on hydrate/mount over its own base. Reading the PR against our code, almost every piece of the rewrite is a reimplementation of a Solid primitive that was too slow or too generic to use directly:

| yak hand-rolls | instead of | because |
|---|---|---|
| `serializeElement` | `ssrElement` | needs a finished props object; walks getter-backed merged props again; emits `style=""`/`class=""` (#3382) |
| `copyProps` / `proxyProps` | `merge` / `omit` | proxy layers when any source is `$PROXY`; `omit` has no predicate/prefix form; plain path leaks `$SOURCES` (#3384) |
| `createElementRenderer` (template + `getNextElement` + `spread` + `insert`) | `Dynamic` / `dynamic()` | per-instance memo for a constant tag; no static class in the template; no `runHydrationEvents`; tag-only namespace (#3386) |
| shared `withTheme` proxy | `merge(props, { theme })` | one merge object per instance is measurable |
| `once()` on the server | `createMemo` | server memos are real nodes held until the deferred dispose (#3385) |

This issue tracks closing those gaps so a library can call Solid and get the same numbers. Baseline harness: [ryansolid/yak-bench](https://github.com/ryansolid/yak-bench) — the 14 `css-in-js-bench` workloads, `@yak/solid` at the PR's base and head, both on `2.0.0-rc.8`, with `next-yak`/React as the reference lane. Acceptance for every item below: a variant of the yak runtime using our primitive lands within noise of the hand-rolled one in that harness.

## Where the remaining time goes

Solid 2.0.0-rc.8, `@yak/solid` at PR head. Full tables: [SSR](https://github.com/ryansolid/yak-bench/blob/main/result/ssr-latest.md) · [Chromium](https://github.com/ryansolid/yak-bench/blob/main/result/browser-latest.md).

Static styled intrinsics are fast (SSR 25× React, hydrate/mount 2–3.5× React). The cases still **behind React** are the ones that go through our generic paths — `styled(Component)` with `{...rest}` spreads, dynamic `$props`, imported primitives:

| case | SSR PR/React | hydrate PR/React | mount PR/React |
|---|---:|---:|---:|
[...]

PR #3487, opened 2026-09-16
perf(signals): omit() copies folded filters by exact concat while short, chains them past 8 keys

ryansolid:
Part of the yak/Kobalte convergence work tracked in #3389 (the "slimmer omit/merge records" item). Companion to #3486, split so CodSpeed attributes each.

## What

When an omit folds over another view — an omit of an omit, or of a merge with omit leaves — the two filters were combined into one key list per leaf, per layer, by `slice()` + `push`. V8 grows the backing store to `1.5n + 16` slots on that push, so each copy allocated 2.4–2.7× an exact one (that was #3475's final form; it's where an 11% allocation increase on the `polymorphic-chain` SSR case between that PR's morning build and its merged form came from, found while isolating #3486). On a component chain (defaults → omit → statics → omit …) the copies held every key hidden so far, growing with the depth.

Two lists now combine with `concat` — one builtin call, exact-size — while the result is within 8 keys. Past that, or with a predicate on either side, they **chain**: one two-field link over the filter folded, no copy. `isHidden` walks the links in one loop. A predicate filter no longer needs a closure to combine; a filter with nothing to hide adds nothing.

## Why the limit (first push vs this one)

The first push chained from the first fold. It cut allocation −23% at depth 7 and TurboFan build −32%, but CodSpeed read the tier-1 `polymorphic-chain` SSR bench **−10%**: the walk per key of every leaf cost the bytecode tiers more than the copies had. That bench's four omits hide 1 + 1 + 2 + 1 keys, so under the limit every leaf on that shape stays one flat list and a check is one `includes`; a deep leaf on a longer chain is one short list plus a few links. A limit of 12 was measured and is worse in every tier.

## Numbers

Tier-1 `polymorphic-chain` SSR bench (Kobalte shape, `renderToString`, 200 rows), harness driven in-process under each tier, min-of-5 vs `next`:

| tier | chain | chain-static | compiled (control) |
|---|---:|---:|---:|
| interpreter | **−7%** | **−7%** | flat |
| Sparkplug | **−12%** | **−11%** | flat |
| TurboFan | **−17%** | **−16%** | flat |

Depth-7 props chain (15 hidden keys on the deepest leaf, so it does chain): TurboFan build −14%, build + consume −9%; interpreter/Sparkplug +3–5% — that's the residual cost of the walk past the copy limit, on the artificial shape only. Depth 3 is flat in every tier.

The `omit-static(5, 1, 2)` reading on the first push was on a construct-only path that is byte-identical to `next`; same class of runner noise as the `merge-proxy-keys-store` readings on #3475.

2370 signals tests pass; a test pins the copy/chain rule across lists, predicates, empty omits and the 8-key boundary.


ryansolid:
> **Correction — the numbers in the description are wrong, and CodSpeed is right.**
>
> Every local "vs `next`" measurement behind this PR resolved `@solidjs/signals` through the main worktree's `dist`, which was built at 15:30 — before #3475 landed on `next` at 15:53. So the baseline was the pre-#3475 code (`concat` + closure), not `next` (`slice()` + `push`). Against a freshly built `next` (5fa224a, both trees rebuilt, tier-1 harness driven in-process under each V8 tier, min-of-5):
>
> | variant | tier-1 `chain` TurboFan | tier-1 interp | depth-7 build TurboFan | depth-7 interp | depth-7 alloc |
> |---|---:|---:|---:|---:|---:|
> | `next` (`slice`+`push`) | 0.95 ms | 4.75 ms | 6.2 µs | — | 40.6 KB |
> | this PR (`concat` ≤ 8, chain past) | **+34%** | +5% | +16% | +5% | 24.1 KB |
> | `slice`+`push` ≤ 8, chain past | ≈ | +4% | −8% | +4% | 30.7 KB |
> | exact preallocated copy, no chain | ≈ | +3% | −9% | +10% | 28.8 KB |
>
> Two findings:
>
> 1. `concat` is the regression, not the chain. At depth 3 no chain is ever built and build is still +16%: `Array.prototype.concat` is a C++ builtin TurboFan cannot inline, while `slice`/`push` are CSA builtins it inlines. My standalone microbench had shown `concat` faster because neither inlines in an isolated loop.
> 2. `slice()` + `push` is the instruction-count floor for this operation in the bytecode tiers, which is what CodSpeed's simulation weights. Every alternative — a chain walk, an exact-size preallocated copy, `concat` — costs more instructions per fold or per check, even where it allocates 25–40% less and runs faster under TurboFan. On the tier-1 Kobalte shape the TurboFan gains are nil (its leaves hold ≤ 5 keys); the only TurboFan wins are on the artificial depth-7 chain.
>
> Recommendation: close without merging. The allocation motivation does not turn into throughput on realistic shapes in any tier, and the description's tier table should be disregarded.
>
> — *Claude via Cursor*

ryansolid:
> Closing without merging per the measurements above: against a correctly built `next`, no variant of the fold (chain, exact copy, `concat`) beats `slice()` + `push` on instruction count, and the TurboFan wins only appear on the artificial depth-7 chain. The allocation growth it targeted is real but costs no throughput in any tier.
>
> — *Claude via Cursor*

PR #3490, opened 2026-09-16
fix(signals): a direct-commit reader is neither entered nor born held — until() after a staged frame (#3482)

### LINKED ISSUE #3482: until() inside an action deadlocks when the confirming frame was staged before until() was called (regression from #3451)
### Describe the bug

`yield until(predicate)` inside an `action()` never resolves when the predicate reads the same `createOptimisticStore` a live stream feeds **and the confirming frame landed before `until()` was called**. With a `timeout` it rejects with `TimeoutError`; without one the action hangs forever.

This is the single-primitive shape the `until()` docstring recommends (optimistic row + socket echo), and it works when the frame lands *after* `until()` subscribed. The ordering that breaks is the natural one for a server that broadcasts to subscribers before it answers the mutation.

**Regression.** Bisected to d80cd1f6 "fix(signals): nodes created mainline during a hold are born held (A29 creation-time form) (#3451)". Its parent 61a114c8 passes, published `@solidjs/signals@2.0.0-rc.8` passes, current `next` (37ea8855) fails.

**Mechanism.** The action's transaction holds the staged frame under its optimism. `until()` creates its predicate effect from mainline code (after the `await`), so with #3451 that effect is *born held*: its synchronous first run is skipped and it is added to the transaction's `_gatedSubs` to be replayed at commit. But the commit is the action's settle, which is exactly what the `until()` promise is holding open. Deadlock.

`CONFIG_DIRECT_COMMIT` exists for this case (its comment: "a staged value with an immediate apply ... deadlocks the hold (until)"), but `recompute`'s direct-commit branch is gated on `bornHeld === null`, so the born-held path pre-empts it.

Vitest test against `packages/signals` (below). No app or extra dependencies needed. I hit it in a real app: a seat-reservation demo where a `live()` server function streams the seat map into an optimistic store and a reserve action does `await reserveSeat(); yield until(() => data.reservations.some(r => r.id === saved.id))`.

1. Check out `next`.
2. Save the test below as `packages/signals/tests/until-held-frame-before-call.test.ts`.
3. `cd packages/signals && pnpm vitest run tests/until-held-frame-before-call.test.ts`
4. It fails with `expected TimeoutError ... to not be an instance of TimeoutError`. Push the confirming frame *after* `answer(...)` instead and it passes.

```ts
import { expect, test } from "vitest";
import { action, createOptimisticStore, createRenderEffect, createRoot, flush, TimeoutError, until } from "../src/index.js";

type Row = { id: string; status: "pending" | "confirmed" };
type Snapshot = { rows: Row[] };

const settle = async (n = 3) => {
  for (let i = 0; i < n; i++) {
    await new Promise(r => setTimeout(r, 0));
    flush();
  }
};

// A manually pumped AsyncIterable — what a live() server-function stream
// materializes as on the client.
function stream<T>() {
  const buffered: IteratorResult<T>[] = [];
  let waiter: ((r: IteratorResult<T>) => void) | null = null;
  const iterable: AsyncIterable<T> = {
    [Symbol.asyncIterator]: () => ({
      next: () =>
        new Promise<IteratorResult<T>>(res => {
          if (buffered.length) res(buffered.shift()!);
          else waiter = res;
        }),
      return: () => Promise.resolve({ done: true as const, value: undefined })
    })
  };
  return {
    iterable,
    push(value: T) {
      const r = { done: false as const, value };
      if (waiter) {
        const w = waiter;
        waiter = null;
        w(r);
      } else buffered.push(r);
    }
  };
}

test("until() sees a confirming frame that landed on the held store before until() was called", async () => {
  const feed = stream<Snapshot>();
  let answer!: (row: Row) => void;
  const mutation = new Promise<Row>(res => (answer = res));
  const views: string[] = [];
  let reserve!: () => Promise<unknown>;

  createRoot(() => {
    const [store, setStore] = createOptimisticStore<Snapshot>(() => feed.iterable, { rows: [] }, { key: "id" });
    reserve = action(async function* () {
      setStore(d => {
        d.rows.push({ id: "temp", status: "pending" });
      });
      const saved = await mutation;
      yield until(() => store.rows.some(r => r.id === saved.id), { timeout: 200 });
    });
    createRenderEffect(
      () => store.rows.map(r => `${r.id}:${r.status}`).join(",") || "empty",
      v => {
        views.push(v);
      }
    );
  });

  feed.push({ rows: [] });
  await settle();
  expect(views.at(-1)).toBe("empty");

  const done = reserve().then(
    () => "settled",
    e => e
  );
  await settle();
  expect(views.at(-1)).toBe("temp:pending");

  // Server mutates and broadcasts first — the frame lands while the action
  // is still awaiting the mutation's answer, and is held under its optimism.
  feed.push({ rows: [{ id: "res_1", status: "confirmed" }] });
  await settle();
  expect(views.at(-1)).toBe("temp:pending");

  // Then it answers, and the action reaches its until().
  answer({ id: "res_1", status: "confirmed" });
  await settle();

  const outcome = await done;
  expect(outcome).not.toBeInstanceOf(TimeoutError);
  expect(outcome).toBe("settled");
  await settle();
  expect(views.at(-1)).toBe("res_1:confirmed");
});
```

### Expected behavior

The held frame is the confirmation the predicate names, so `until()` should flip and the action should settle, revealing the frame with the overlay revert (the same outcome as when the frame lands after `until()` subscribed).

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Node: 25.1.0
- `@solidjs/signals`: `next` @ 37ea8855 (fails), 3fc2c549 (fails), d80cd1f6 (first bad), 61a114c8 (passes), 2.0.0-rc.8 published (passes)

### Additional context

**Proposed fix** (one early return in `enterStagedRead`, mainline branch, after the `_verdictPull` check). A direct-commit reader is the documented tunnel through a hold and applies on its own microtask, so it should be neither entered nor born held, the same exemption verdict pulls and optimistic nodes already get:

```diff
     if (GlobalQueue._verdictPull) return;
+    // A promise-delivery effect (resolve()/until(), CONFIG_DIRECT_COMMIT)
+    // reads staged truth by contract — it is the tunnel that keeps a hold
+    // deadlock-free — and applies on its own microtask, so it is neither
+    // entered nor born held. Born held, an until() created mainline after
+    // the confirming frame was staged would replay only at the commit its
+    // own promise is holding open (#3451 follow-up).
+    if (ctx._config & CONFIG_DIRECT_COMMIT) return;
     if (
       ctx._flags & REACTIVE_RECOMPUTING_DEPS &&
```

With that change the test above passes, and so do `until`, `until-entanglement`, `born-held`, `resolve`, `refresh-await`, `superseded-before-first-commit`, `visibility-oracle` and the rest of the signals suite (including `dist-artifacts` after a rebuild). Happy to open a PR with the test and the fix.

**Workaround** in the meantime: settle the action with `await mutation; yield refresh(store)` instead of `until()`; the refreshed answer stages into the held transaction and reveals atomically with the overlay revert.

ryansolid:
Closes #3482.

## Bug

`yield until(predicate)` inside an `action()` never resolved when the predicate read the optimistic store a live stream feeds **and the confirming frame landed before `until()` was called** — the natural ordering for a server that broadcasts to subscribers before it answers the mutation. With a `timeout`, `TimeoutError`; without one the action hung forever. This is the single-primitive shape the `until()` docstring recommends.

Regression from `d80cd1f6` (#3451, A29 creation-time born-held). `2.0.0-rc.8` is unaffected; `next` fails through `c54e4c0ca`.

## Cause

A reader created mainline under a hold is born held (`enterStagedRead` sets `stagedEntry`; `recompute` stamps it, skips its synchronous first run and adds it to the transaction's `_gatedSubs` to replay at the commit). For a promise-delivery effect the commit *is* the action's settle — which its own promise is holding open. `recompute`'s `DIRECT_COMMIT` arm exists for exactly this case ("deadlocks the hold (until)") but is gated on `bornHeld === null`, so the born-held path pre-empted it.

## Fix

`enterStagedRead`'s mainline arm exempts `CONFIG_DIRECT_COMMIT` readers — `resolve()`, `until()`, awaitable `refresh()`'s waiter — beside the existing verdict-pull exemption. They read staged truth by contract (the tunnel that keeps a hold deadlock-free) and apply on their own microtask, not the transaction's stashed queues, so they neither enter the transaction nor are born held. `stagedEntry` never gets set for them; every downstream `bornHeld === null` holds and they take the existing direct-commit path. The predicate sees the held frame, the action settles, the commit reveals the frame with the overlay revert.

```diff
-    if (GlobalQueue._verdictPull) return;
+    if (GlobalQueue._verdictPull || ctx._config & CONFIG_DIRECT_COMMIT) return;
```

## Tests

`tests/until-held-frame-before-call.test.ts` — the issue's reproduction (frame lands while the action awaits its answer; `until()` reached after) and the control ordering (`until()` subscribed first, then the frame). On `next` the first fails with `TimeoutError` and the control passes; with the fix both pass.

signals 2,735 (+2) / 2 expected fails unchanged; solid 620; web 825 / 986 / 185 against the rebuilt dist. `until`, `until-entanglement`, `born-held`, `resolve`, `refresh-await`, tree-shake floor and rules index green.

## Size

~+20 B minified on a core-retained read path; read +20..+40 B brotli across the scenarios (−30 on observe — layout). Two app caps (`hydrating (no stores)`, `CSR with …`) ratcheted 0.05 KB with notes; the in-package tree-shake floor did not move.

Reproduction, bisect and analysis by @brenelz.


ryansolid:
> Closing unmerged — superseded by #3491. Not a bug: the `until(...)` in `yield until(...)` is evaluated in the `await` continuation, before the `yield` re-enters the transaction, so the effect is created mainline and born held as A29 specifies. A bare `yield;` after the `await` puts the call inside the transaction, where `recompute`'s existing `DIRECT_COMMIT` arm — scoped to the effect's *own* held transition — already handles it; verified on `next` in #3491.
>
> The exemption here was also too broad: `CONFIG_DIRECT_COMMIT` covers `resolve()` and awaitable `refresh()`'s waiter, which are routinely called outside actions. From mainline against a *foreign* hold they would have served that action's uncommitted frame instead of the committed view — a leak, not a fix. The gap it exposes (nothing pins that posture for the direct-commit readers) gets its own tests-only PR.
>
> — *Claude via Cursor*

Issue #3511, opened 2026-09-17
Compiler: props literals with accessors are the largest SSR self-time bucket on composition cases (computed keys + closure-per-getter)


ryansolid:
> **Status (Sep 17):** finding 1 landed in #3514. Finding 2: options (a)/(b) below are withdrawn — prototype getters break `{...props}`/destructuring in reactive expressions, which is supported. The plan is (c) re-measured: per-site class behind a `Proxy`, **SSR generate only**, with a literal fallback for sites whose captured bindings aren't provably constant. Verdict, microbenchmarks, and the measured fallback rate (≤0.4% library / 2–4% apps): https://github.com/solidjs/solid/issues/3511#issuecomment-5719473252

Tracking item 6 of #3389. Profiling the composition cases in yak-bench (`polymorphic-chain`, `tabs`) after #3497/#3509 shows the compiled component code itself — not `merge`/`omit`, not the serializer — as the largest single self-time bucket on the server: **24–26% of SSR time** on both cases, and **~18% of allocation**. Almost all of it is the props literal the compiler emits at each component call site.

## Exhibit

`ButtonRoot` from `polymorphic-chain`, as compiled (SSR, `@solidjs/compiler`; babel output is identical by parity):

```js
function ButtonRoot(props) {
  const merged = merge({ type: "button", disabled: false, variant: "solid" }, props);
  const others = omit(merged, "as", "type", "disabled", "loading", "variant");
  return Polymorphic(merge({
    get as() { return merged.as ?? "button"; },
    get type() { return memo(() => merged.as === "a")() ? void 0 : merged.type; },
    get disabled() { return merged.disabled; },
    get ["aria-disabled"]() { return merged.disabled || void 0; },
    get ["data-disabled"]() { return merged.disabled ? "" : void 0; },
    get ["data-loading"]() { return merged.loading ? "" : void 0; },
    get ["data-variant"]() { return merged.variant; }
  }, others));
}
```

Every render of every `ButtonRoot` allocates seven closures and an accessor-bearing object literal. React's equivalent is `{ ...rest, type, disabled, "aria-disabled": ... }` — one plain object.

## What it costs

Microbenchmark of exactly this shape (7 getters over a captured `merged`; build the object, read all 7 keys once — what `ssrElement` does), min of 7 × 200k, Node 26:

| form | TurboFan | `--no-opt` | retained |
|---|---:|---:|---:|
| literal, computed keys — **today** | 966 ns | 715 ns | 1274 B |
[...]

PR #3514, opened 2026-09-17
perf(compilers): emit getter keys as string literals, never computed

ryansolid:
Finding 1 of #3511. Both compilers emit non-identifier getter keys in compiled props literals as **computed** accessors:

```js
// today
Polymorphic(merge({
  get as() { return merged.as ?? "button"; },
  get ["aria-disabled"]() { return merged.disabled || void 0; },
  get ["data-loading"]() { return merged.loading ? "" : void 0; },
  …
}, others));
```

Babel does it via `t.objectMethod("get", id, [], body, !t.isValidIdentifier(key))`, and `packages/compiler/src/shared/ast.rs` mirrors it on purpose ("Babel: … non-identifier getter keys are computed"). `get "aria-disabled"()` is the same property, but a computed key in an object literal takes V8 off the literal boilerplate path into per-property runtime definition for the whole object.

Measured on `ButtonRoot`'s seven-getter `<Polymorphic>` props literal from yak-bench's `polymorphic-chain`, build + read all 7 keys, min of 7 × 200k:

| form | TurboFan | `--no-opt` |
|---|---:|---:|
| literal, computed keys (today) | 966 ns | 715 ns |
| literal, string-literal keys (this PR) | 525 ns | 485 ns |
| literal, all identifier keys (control) | 482 ns | 482 ns |

So the string-literal form is within 10% of what the same object costs with plain identifier keys — the computed key was the cost, not the hyphen. Every `aria-*`, `data-*`, and `class` prop paid it (`class` is a reserved word, so `isValidIdentifier` rejects it too).

### Change
`id` at every site is already an `Identifier` or a `StringLiteral` (from `convertJSXIdentifier`, or `t.identifier` for locked DOM properties), so `computed` is just `false`. Six Babel sites — `shared/component.ts` ×3, `dom/element.ts`, `ssr/element.ts`, `universal/element.ts` — and the single Oxc funnel `object_getter_property_with_statements`. Nothing else in the emitted code changes.

Fixtures regenerated in both trees (`vitest -u` for Babel, `UPDATE_OXC_FIXTURES=1 UPDATE_PARITY=1` for Oxc, including the three cross-mode parity `.diff` files). Audited: every changed line is a getter key. Babel's generator prints a reserved-word string key bare (`get class()`), which is valid and the same property.

### Verified
- `@solidjs/babel-plugin` 242/242; `@solidjs/compiler` vitest 5785/5785 and `cargo test` across all three feature sets.
- `@solidjs/web` DOM (825), hydrate (185), server (1001) suites run end-to-end through the rebuilt native compiler; harness artifacts unchanged.

Case-level SSR A/B in the yak-bench harness is pending a quiet machine — tonight's runs had a load average of 11 from a remote-desktop session and the raw numbers disagreed by 10× within a tag, so I'm not quoting them. The microbenchmark is deterministic and the emitted code is otherwise identical; I'll add the harness number when I have one worth trusting.


ryansolid:
> Harness measurement, second attempt. The machine is still not quiet (load average 17–19: Spotlight indexing the new worktrees plus a remote-desktop session), and both wall-clock and `process.cpuUsage` per render swing ±40% within a single tag, so no time-based number is quotable yet.
>
> What is stable under load is allocation. Scavenge count over a fixed render batch (`--max-semi-space-size=1`, `--trace-gc` enabled at runtime after warm-up; two reps each, reps agree to ±2%):
>
> | case | baseline | this PR | Δ |
> |---|---:|---:|---:|
> | `polymorphic-chain` (n=400 × 100) | 807 / 770 | 733 / 734 | **−7%** |
> | `tabs` (150 × 100) | 874 / 865 | 862 / 854 | −1.5% |
> | `multifile-composition` (150 × 60) | 507 / 505 | 514 / 513 | +1.5% (noise) |
> | `btn-variant` (1000 × 150) | 459 / 461 | 460 / 461 | 0 |
>
> Consistent with the key mix: `polymorphic-chain` is the case whose props are mostly `aria-*` / `data-*`; the others are mostly identifier keys and never took the computed path. Allocation is a lower bound here, not the effect size — in the microbenchmark the computed form's extra cost is mostly the per-property runtime definition, and only ~14% of it is bytes. Time-based harness numbers still to come when the machine is quiet.
>
> — *Claude via Cursor*

ryansolid:
> Harness SSR A/B on a quiet machine (load 5.5, Spotlight and the remote session gone). Baseline `next` at #3509 vs this branch, `solid-mprim` lane, both compiled with the respective checkout's Babel plugin; 8 interleaved pairs per case, fresh process each, 1 s warm + 2 s measure, instances/sec:
>
> | case | median vs median | paired-ratio median | pairs favoring PR |
> |---|---:|---:|---:|
> | `polymorphic-chain` (400) | **+7.5%** | +8.3% | 6 / 8 |
> | `tabs` (150) | +2.6% | +6.4% | 6 / 8 |
> | `multifile-composition` (150) | +2.2% | +2.8% | 4 / 8 — noise-level |
> | `btn-variant` (1000) | −0.9% | +0.2% | flat |
>
> Baseline spread within a tag is ±3% on `polymorphic-chain`, ±12% on `tabs`, so the first row is a real effect and the second is suggestive. Same shape as the allocation proxy above: the win lands where the props are `aria-*` / `data-*` / `class`, and there is nothing to gain where every key is an identifier. `polymorphic-chain` SSR moves from 0.29× React to roughly 0.31×; the rest of that gap is finding 2 in #3511.
>
> — *Claude via Cursor*

Issue #3075, opened 2026-08-27
2.0.0-rc.3 | `latest()` does not work on projected value `createProjection()`


mizulu:
### Describe the bug

`latest` on projected signal does not work, shows uncommitted value 

### Your Example Website or App

https://s.olid.uk/id/sAPh460QTq-eozQTHk06Qg

### Steps to Reproduce the Bug or Issue

playground 

### Expected behavior

latest on projected value `createProjection` should work like it works for derived signal or memo 

### Platform

.

### Additional context

broken since 2.0.0-rc1, but works in [2.0.0-rc.0](https://s.olid.uk/id/UEBRosiAQGGL9zQbrulESg)

```tsx
import { render, untrack, memo } from '@solidjs/web';
import { createSignal, createProjection,  latest, createMemo } from 'solid-js';

export default function App() {
  const [count, setCount] = createSignal(0);

  const [derived, setDerived] = createSignal(()=>{
    return count()
  })

  const m = createMemo(()=>{
    return count()
  })



  const p = createProjection(() => {
    return { a: count() }
  }, {})



  setTimeout(() => {
    untrack(() => {

      setCount(v => v + 5)
    
      console.log("count", count(), latest(count)) // 0 5  latest on source   (success)
      console.log("p.a", p.a, latest(()=>p.a))     // 0 0  latest on projected count  (Fail)
      console.log(derived(), latest(derived))     // 0 5   latest on derived signal (success)
      console.log(m(), latest(m))                 // 0 5   latest on memo  signal (success)

    })
  })



  return (
    <div class="p-2">
     latest does not work for projection, but works for derived signal, or memo
    </div>
  );
}


if (typeof document !== 'undefined') {
  render(() => <App />, document.getElementById('root')!);
}

``

ryansolid:
Fixed in 0e37f909 (same commit as #3074 — two directions of one isolation gap).

Why it never worked: `latest()` is implemented in core `read()`, which routes the read through a companion computed that re-evaluates speculatively against pending values — that's what makes it work for signals, memos, and derived signals. But store property reads go through the proxy traps, which only reach core `read()` when there's a tracking observer; an untracked `latest(() => p.a)` served the committed backing straight from the trap, so the latest window was invisible to stores. (The rc.0 → rc.1 regression window you bisected is the store rewrite — the legacy store happened to route these reads through core.)

[...]

PR #3516, opened 2026-09-17
refactor(signals): store node reads select by readerSeesCommitted (Rule 1, one implementation)

ryansolid:
Stacked on #3515. Step 2 of DESIGN-CONSOLIDATION move 3b.

## What moved

`nodeValue` — the store's untracked node-read selection — restated core `read()`'s committed-vs-staged arm by hand (`inOwnerContext`, a stale-of-foreign `foreignHold` twin, the HELD-truth mask). It now calls `readerSeesCommitted` (from #3515) with core's context selection (`readerContext()`: a root reads as its parent computed, `context` persists under untrack) behind the two store-only tunnels that stay in the store: `latest()` reaching this path for store keys (#3075) and `authoritativeServe()` (truth authors: the projection derive's draft, write-override).

Tracked store reads already went through `readNode` → core `read()` → `readerSeesCommitted`, so after this every store **node** read — tracked or untracked — selects by one Rule 1 implementation.

## Not zero-change: the twin was missing a precondition

A stale reader (render effect) whose **untracked** read is of a store key held by a foreign action was served committed (correct, A15 / A26) but never recorded for replay at the commit (`heldFromStale`'s `_gatedSubs` contract, which the signal path performs). The effect stayed on the pre-action value **permanently** after the action settled. Signal twin: `[0, 0, 1]`; store on `next`: `[0, 0]`.

Pinned as **S4** in `posture-store-parity.test.ts` (signal and store now both `[0, 0, 1]`; the store pin fails on `next`).

This is exactly the pattern the consolidation is for — one rule, N sites, one site missing a precondition — and it was invisible to the suite, the oracles and the 621-cell matrix (no reader kind there is "untracked inside a stale reader"). Found only by handing the store the shared predicate and asking what changed.

## Remaining gaps, pinned / noted

- **Backing twin (step 3):** a key with **no node** goes through `pendingBackingVisible` / `heldFromReader` — same committed value, but there's no node to record the reader on, so no replay. Pinned `it.fails` beside S4. Rule 1's backing-level form is the next item.
- **Lane arm:** `laneReadsCommitted` (#3460) now also applies to untracked store node reads as it did to tracked ones. No matrix cell or test distinguishes; the matrix's `foreignLane × untracked` store cells are identical.

## Gate

| check | result |
|---|---|
| signals suite | 2748 pass, 1 expected fail (the backing-twin pin) |
| visibility oracles + 621-cell posture matrix | byte-identical to `next` |
| fuzzer seed 3289 × 1000 | 994 / 0 / 6 — same as `next` (signals-only, so expected) |
| size | minified store −5 B, full bundle brotli −44 B vs #3515; no caps exceeded |


ryansolid:
>
> — *Claude via Cursor*

Issue #3074, opened 2026-08-27
2.0.0-rc.3 |  projection leaks after flush while in transition 


mizulu:
### Describe the bug

A projection of `count()` leaks the set value after a `flush` while still being in the transition 
while count() still shows the uncommitted value. 

logs
```
[log] action:start
[log] p.a 0
[log] count 0
[log] p.a 5       <--- leak
[log] count 0   <-- no leak 
[log] action:done
```

### Your Example Website or App

https://s.olid.uk/id/ij_hsNqcRVKCqfWdatIU0Q

### Steps to Reproduce the Bug or Issue

playground 

### Expected behavior

expecting the expected. no leak

### Screenshots or Videos

<img width="366" height="390" alt="Image" src="https://github.com/user-attachments/assets/aa825576-7697-4135-9a8f-e21e71cf77ad" />

### Platform

.

### Additional context

```tsx
import { render, untrack } from '@solidjs/web';
import { createSignal, createProjection, action, flush } from 'solid-js';

export default function App() {
  const [count, setCount] = createSignal(0);



  const p = createProjection(() => {
    return { a: count() }
  }, {})


  const act = action(function* () {
    console.log("action:start")
    yield new Promise(r => setTimeout(r, 3000))
    console.log("action:done")

  })

  setTimeout(() => {
    untrack(() => {

      act() // holds transition for 3s


      setCount(v => v + 5)
      console.log("p.a", p.a)        // 0
      console.log("count", count())  // 0 

      flush()

      console.log("p.a", p.a)        // 5      
      console.log("count", count())  // 0 

    })
  })



  return (
    <div class="p-2">
      Projection leak after flush, see logs
    </div>
  );
}


if (typeof document !== 'undefined') {
  render(() => <App />, document.getElementById('root')!);
}

```

ryansolid:
Fixed in 0e37f909 — thanks for the tight repro, and for filing #3075 alongside it: they turned out to be the two halves of the same isolation gap, in opposite directions.

What was happening: a projection recompute commits its output through the store's adoption channel, which swaps the committed backing eagerly by contract (that's what makes reconcile immediately visible outside transitions). Node-level notifications ride core `setSignal` and park correctly under the transition — which is why subscribed effects were properly held — but an untracked read serves the backing directly, so `p.a` showed the speculative value after `flush()` while `count()` (whose write parked in the signal's pending slot) correctly stayed committed.

[...]

PR #3517, opened 2026-09-17
fix(signals): store backing paths record the stale reader's replay (Rule 1, backing-level form)

ryansolid:
Stacked on #3516 (← #3515). Step 3 of DESIGN-CONSOLIDATION move 3b — closes the `it.fails` pin #3516 left.

## The gap

The stale-of-foreign clause has two halves: **serve committed**, and **record the reader for replay** at the hold's commit (A15 / A26, `_gatedSubs`). Core `heldFromStale` does both. The store's backing-level twins did only the first:

- `heldFromReader` — a key with **no node** under a fold hold (setter write inside an action);
- `readSource`'s adoption-hold view — a `reconcile` inside an action (`ht`, #3074), with or without a node.

A render effect whose untracked read hit either was served the committed value (correct) and then stayed on the pre-action value after the action settled. Signal: `[0, 0, 1]`. Store on `next`: `[0, 0]`.

## The fix

One exported helper, `recordStaleReplay(txn, c)` (the two lines lifted out of `heldFromStale`), called by `heldFromStale` and by the store's two backing sites through `staleReplay` (reader in context, not children-forbidden — the same gate core applies before it reaches `heldFromStale`). A `latest()`-pull `PLAIN_HOLD` is not a transaction and records nothing.

## Pins

`posture-store-parity` S4 grows to the four store shapes — fold hold × adoption hold × with/without a node — beside the signal. **All four fail on `next`**; the step-2 `it.fails` flips to a pass.

## Gate

| check | result |
|---|---|
| signals suite | 2751 pass, no expected fails left |
| visibility oracles + 621-cell posture matrix | byte-identical to `next` |
| fuzzer seed 3289 × 1000 | 994 / 0 / 6 — same as `next` |
| size | +31 B minified core (the helper), +123 B minified store (the fix); three brotli caps ratcheted (+30 / +2 / +34 B) with notes |

## Where move 3b stands after this stack

Rule 1 (committed vs staged) now has one implementation for every **node** read, signal or store (`readerSeesCommitted`), and one replay registration for every stale-of-foreign serve, node or backing (`recordStaleReplay`). Still store-shaped and not yet folded: the backing-level *selection* itself (`pendingBackingVisible`'s arms — draft, write-override, the projection authoritative-elect arm) and the two store tunnels (`latest()` for store keys, `authoritativeServe`). Those are the candidates for a `serve(el, reader)` in the design doc, and each will need the same "hand it the shared predicate and see what changes" treatment.


ryansolid:
>
> — *Claude via Cursor*

Issue #2141, opened 2024-04-22
`this` inside jsx-template inside a constructor of a derived class causes transformation-bug


bigmistqke:
### Describe the bug

When a template is written inside the constructor of a derived class and `this` is accessed inside the template, the following error is thrown: `Uncaught ReferenceError: Must call super constructor in derived class before accessing 'this' or returning from derived constructor`.

The following code

```tsx
abstract class Root {
}

class Derived extends Root {
  constructor() {
    super();
    <div>{this.test}</div>;
  }
}

new Derived();
```

will become

```tsx
import { template as _$template } from "solid-js/web";
import { insert as _$insert } from "solid-js/web";
var _tmpl$ = /*#__PURE__*/_$template(`<div>`);
class Root {}
class Derived extends Root {
  constructor() {
    const _self$ = this;
    super();
    (() => {
      var _el$ = _tmpl$();
      _$insert(_el$, () => _self$.test);
      return _el$;
    })();
  }
}
new Derived();
```

`const _self$ = this;` should be set after `super()` when the class is extended.

### Your Example Website or App

https://playground.solidjs.com/anonymous/51a44967-beca-46ba-9332-a9cfa8d81c5d

### Steps to Reproduce the Bug or Issue

1. open link
2. look at error in console
3. look at transformed output

### Expected behavior

using jsx-template inside a class-constructor to not cause issues.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

titoBouzout:
This issue should be in dom-expressions 😊

ryansolid:
Yeah this might be fun. This issue was introduced with this fix: https://github.com/ryansolid/dom-expressions/pull/295

I'm imagining more specifically the fix for https://github.com/ryansolid/dom-expressions/issues/292.

PR #3519, opened 2026-09-17
fix(signals): a write is a proposal (A34) — held-node writes entangle the tick, net-equal writes propose nothing (#3494)

### LINKED ISSUE #3494: Remaining publication and observer-release regressions from #3473
AI-assisted follow-up to the [review feedback on #3473](https://github.com/solidjs/solid/pull/3473#issuecomment-5687335556), tracking what remains after merge and subsequent fixes.

Rechecked on **next at 59c4192a**, including #3479 and #3488. Four independently reproduced regressions remain:

- [ ] **Torn effect input:** an effect receives `[1, 0, 1]` before `[1, 1, 1]` (original comment A).
- [ ] **Stale visible derivation:** Count publishes `1` while Copy displays `0` and its request is still pending (original comment C).
- [ ] **Lost visibility update:** coalesced visibility writes leave content visible after every request has resolved. This is a different sequence from the now-fixed original comment B.
- [ ] **Delayed release:** an ordinary write continues waiting on an unobserved request after removing its last reader and reading `latest`. This eventually completes; it is an unnecessary wait, not permanently lost state.

The repros below cover those four behaviors. Grouping them here carries forward the PR feedback; it does not assume one root cause. The same-value update-grouping change is listed separately as a contract question.

## Representative playground: a stale derivation becomes visible

Paste this TSX into s.olid.uk using the current next build. Wait for all three values to show `0`, then click Run. After one second, Direct disappears and **Count shows 1 while Copy shows 0**. Copy catches up about a second later. Count and its visible derivation should publish consistently.

```tsx
import { render } from "@solidjs/web";
import { createMemo, createSignal, latest, Show } from "solid-js";

const delay = <T = void,>(ms: number, value?: T) =>
  new Promise<T>(r => setTimeout(r, ms, value));

  const [show, setShow] = createSignal(true);
  const [direct, setDirect] = createSignal(true);
  const details = createMemo(() => delay(1500, count()));
  const copy = createMemo(details);

  return <>
    <button onClick={async () => {
      setShow(false);
      await delay(500);
      setCount(1);
      latest(copy);
      await delay(500);
      setShow(true);
      setDirect(false);
    }}>Run</button>
    <p>Count: {count()}</p>
    <Show when={direct()}><p>Direct: {details()}</p></Show>
    <p>Copy: {show() ? copy() : "hidden"}</p>
  </>;
}

render(() => <App />, document.getElementById("root")!);
```

## Torn tuple delivered to an effect

This is original comment A. The signal and its synchronous identity memo disagree in a real effect callback: `[1, 0, 1]`, then `[1, 1, 1]`. The pre-PR next control (`3ae9e921`) only delivers `[1, 1, 1]`.

<details>
<summary>Standalone core reproduction (TypeScript; top-level await)</summary>

```ts
import { createMemo, createRenderEffect, createRoot, createSignal, flush } from "@solidjs/signals";

const tick = () => new Promise<void>(r => setTimeout(r, 0));
const requests: (() => void)[] = [];
const log: number[][] = [];

const state = createRoot(dispose => {
  const [show, setShow] = createSignal(false);
  const details = createMemo(() => {
    const value = count();
    return new Promise<number>(r => requests.push(() => r(value)));
  });
  const copy = createMemo(count);

  createRenderEffect(
    () => show() ? [count(), copy(), details()] : null,
    value => { if (value) log.push(value); }
  );
  createRenderEffect(details, () => {});
  return { setCount, setShow, dispose };
});

flush();
while (requests.length) requests.shift()!();
await tick();

state.setCount(1);
state.setCount(0);
await tick();
state.setShow(true);
state.setCount(1);
await tick();
while (requests.length) requests.shift()!();
await tick();

console.log(log); // [[1, 0, 1], [1, 1, 1]]
state.dispose();
```

</details>

## Final hide is lost after coalesced visibility writes

Wait for Data=0, click Run, then wait for all work to finish. **Data remains 0 instead of becoming hidden.** The pre-PR control eventually hides it. The independently controlled core version also confirms every request has resolved while the effect still displays 0.

The plain `panel` getter preserves the failing effect dependency graph. Inlining the conditional into JSX introduces a compiler-generated memo and masks this case. Adding a separate displayed reader of `show` also masks it.

<details>
<summary>Playground reproduction (TSX)</summary>

```tsx
import { render } from "@solidjs/web";
import { createMemo, createSignal } from "solid-js";

const delay = <T = void,>(ms: number, value?: T) =>
  new Promise<T>(r => setTimeout(r, ms, value));

  const [show, setShow] = createSignal(true);
  const data = createMemo(() => delay(2000, count()));
  const panel = () => show() ? data() : "hidden";

  return <>
    <button onClick={async () => {
      setShow(false);
      setShow(true);
      setCount(1);
      await delay(500);
      setShow(false);
    }}>Run</button>
    <p>Data: {panel()}</p>
  </>;
}

render(() => <App />, document.getElementById("root")!);
```

</details>

## Removing the last reader does not release the write

Count remains `0` after the only Details reader is removed and `latest(details)` is read. It only becomes `1` when the unused replacement request resolves. The same reduced code on pre-PR next (`3ae9e921`) releases Count to `1` without waiting for that request.

There are no actions in this repro. The pending work is no longer displayed or tracked by a reader; the subsequent `latest` call is untracked. This is the remaining latest-related release difference mentioned briefly in the original comment.

<details>
<summary>Standalone core reproduction (TypeScript; top-level await)</summary>

```ts
import { createMemo, createRenderEffect, createRoot, createSignal, flush, latest } from "@solidjs/signals";

const tick = () => new Promise<void>(r => setTimeout(r, 0));
const requests: (() => void)[] = [];

const state = createRoot(dispose => {
  const details = createMemo(() => {
    const value = count();
    return new Promise<number>(r => requests.push(() => r(value)));
  });
  createRenderEffect(count, value => console.log("Count:", value));
  const removeReader = createRoot(remove => {
    createRenderEffect(details, () => {});
    return remove;
  });
  return { count, setCount, details, removeReader, dispose };
});

flush();
requests.shift()!();
await tick();

state.setCount(1);
await tick();
requests.shift()!();
state.setCount(0);
state.setCount(1);
await tick();
state.removeReader();
latest(state.details);
await tick();

console.log("After removal:", state.count()); // 0
requests.shift()!();
await tick();
console.log("After unused request resolves:", state.count()); // 1
state.dispose();
```

</details>

## Separate contract question: repeating an already-held write

This behavior also changed in #3473 and remains changed:

```ts
// A and B initially display 0; an observed async memo derives from B.
setB(1);
await tick(); // B=1 is held by the pending async memo.

setA(1);
setB(1); // Repeat the already-held value in the same task as A's write.
await tick();
```

Pre-PR next holds A with B. Current next publishes A separately while B remains held. This is a verified grouping difference, but not classified here as a correctness bug: if repeating the same pending value is intended to be a no-op for grouping, that would explain it. Including it so the original feedback is accounted for rather than silently dropped.

## Cases that are now fixed

- Original comment B's exact sequence—unobserved Count update, show, hide alongside request resolution, then show again—now publishes the final show correctly. That does not fix the coalesced-toggle sequence above.
- The ordinary show-then-hide release repro, without the extra latest/rewrite sequence, now releases the unobserved work.
- The separate optimistic `1:0` action-end regression reported on #3479 is also fixed; it is not being re-reported here.

## Verification and diagnostic context

- All four behaviors independently reproduce outside the fuzzer on `59c4192a`, in development and production with `__TEST__=false`.
- The two TSX repros were compiled and executed in jsdom; this is not a claim of interactive browser verification. The two standalone core snippets were rerun as fresh processes on current next and the pre-PR next control.
- Restoring only `setSignal`'s previous outside-flush adoption of an existing transition removes the torn tuple and the coalesced visibility failure. The stale-derivation and latest-related release cases still fail with that diagnostic change. This is causal evidence, **not a proposed fix**; the new adoption condition also serves the intended mainline isolation semantics.
- The original campaign retained additional candidates, including mount/control mismatches. These four examples are the independently reproduced behavior groups, not a claim that every retained candidate has the same cause or is covered by one fix. The fuzzer in #3446 can be used to check the wider consequences while iterating.

ryansolid:
Fixes #3494.

## Ruling (A34 — new)

> A write to the same signal already set to that value would entangle, I think. Unless it's committed, both are suggesting a value. If one finished before the other that would be odd.

- **A signal committed at `v`:** writing `v` again is nothing — stages nothing, stamps nothing, pends nothing.
- **A signal held (staged, uncommitted, in transaction T):** any mainline write to it — `v` again or `w` — is a second proposal on a contested node. It joins T, and so does the rest of that tick's batch.

This reverses one mechanism choice from #3473, which removed `setSignal`'s eager entry from mainline to stop the `activeTransition` leak (a memo created after the write was born the transaction's, A29) — and dropped the grouping with it. The entry is now **deferred**: `setSignal` pushes the held node's transaction to `batchJoins`; `flush()` enters each at its start, inside the running flush, adopting the tick's ambient batch. Recorded before the equality gate, so a repeat of the held value proposes too.

## The four #3473 regressions (all pinned in `tests/write-proposals-3494.test.ts`)

| case | cause | fix |
|---|---|---|
| torn `[1,0,1]` effect input | the tick's batch no longer grouped with the hold | deferred entry (`batchJoins`) |
| lost hide after `setShow(false); setShow(true)` | `show` stamped into `count`'s hold with nothing to reveal; next `setShow(false)` routed to the stamp and held | no-proposal drop in the adoption loop + `computePendingState` equality gate |
| stale visible derivation (`Count: 1` beside `Copy: 0`) | a stale reader served a held memo's committed value never turned pending itself, so liveness saw no trace of the flight | `reporterBlocksSource` follows a dep's `_pendingSources` one hop |
| delayed release after mainline `latest(details)` | the `latest()` shadow, backfilled under the transaction, blocked the settle | `transitionBlocked` skips companions |

## From differential fuzzing (gabbev's fuzzer, #3446)

4 cohorts × 2 seeds × 5,000 cases, base = `origin/next`. **124 fixed / 7 new.** `branches` 30→5 and 26→3; `optimistic` 0/0 both sides.

Two regressions the first cut introduced, both fixed and pinned:
- **S3 (latest-2 #1470):** the no-proposal drop hit a *stamped* node when a parked batch folded into a merge — a held proposal rewritten to the committed value is held, not proposal-free. Drop applies to `_transition === null` only.
- **O2 (latest-1 #2141):** a hide joined to a parked action stopped reading a memo, unchanged, so A30 kept its dep tail; the action's truth re-asked the memo and the *pending mark* rode the kept link, registering the hidden reader as a reporter on a fetch nobody displays. **A30 amended:** a pending mark rides only the links a pass made — `notifyStatus` skips `STATUS_PENDING` over links outside the pass's validated prefix (`link._gen !== sub._depGen`, the stamp `link()` already maintains; O(1)). Clears and errors still ride every link. (Gating `reporterBlocksSource` instead was wrong — it runs mid-pass — and the fuzzer caught it as an S1 flood.)

The 7 remaining "new":
- **6 × O1** (816/2141/4826/618/3875/4847) — one shape, and a **semantic consequence of the ruling** to be aware of: a `show`/`mount` written in the *same tick* as an action's proposal is held by the action; toggling it back from mainline is a write to a held node → joins the action → the hide waits for the action to end. Base (post-#3473) publishes the hide immediately; pre-#3473 behaved like this PR. If a mainline write that nets a held node back to its *committed* value should escape the hold, that's a carve-out to A34(1) needing a ruling.
- **1574 P1** — pre-existing (minimal case fails on base): the disposal wake only wakes the disposed node's own `_transition`. Follow-up.

## Size

Core floor +175 B minified (25,249 → 25,424; per-piece: `batchJoins` 71, drop 64, `_gen` skip 24, one-hop 16). Brotli vs a fresh `next` build: +56 core, +75 store, +70 isPending/latest, +45 simple app, +80 hydrating, +36 hydrating+stores, +65 CSR, +93 observe, +40 observe+attribution, 0 frames. Caps ratcheted with notes. A golf pass replaced an O(kept-tail) link walk with the existing generation stamp and dropped a redundant companion snap; the fuzzer differential re-run on the golfed tree is case-for-case identical.

## Docs

A34 new; A15, A28, A30 amended; INTERNALS updated (`batchJoins`, no-proposal drop, one-hop liveness, companion skip, pending-mark prefix rule); rules index regenerated; changeset.

---
*Claude via Cursor*


ryansolid:
> I rebased this against current `next` and found blockers that should go back through the async-semantics pass before merge. Ryan has now made the held-to-committed ruling explicit: if a node is committed at `0`, a transaction holds `1`, and mainline writes `0`, that write still contends with the future and entangles. The six O1 fuzzer cases are therefore accepted consequences of A34, not a carve-out.
>
> The blocking implementation issue is the A30 `_gen` gate in `notifyStatus`. It cannot distinguish an obsolete retained tail from a retained dependency that still belongs to the displayed committed frame. I confirmed this dynamically with: committed `selected = remote() = 0`; a held pass switches `selected` to constant `0` and stops reading `remote`; then `query = 1` starts a new `remote` flight. The PR publishes `query=1, selected=0` and leaves `isPending(selected) === false` while the flight remains unresolved. Removing only the `_gen` early return holds the view and reports pending. Current output was `pending: [false, false]`, `views: [[0, 0, false], [1, 0, false]]`; without the gate it was `pending: [false, false, true]`, `views: [[0, 0, false]]`. I recommend removing this A30 amendment from #3519 and tracking the #2141 stale-tail case separately unless we add explicit committed-frame provenance.
>
> Additional issues found during the rebase:
>
> - A same-value held write pushes `batchJoins` and returns through the equality gate without scheduling. The join then leaks into a later tick and captures unrelated work. The record site must schedule; the full flush should also be forced while joins remain, including chased-dead references.
> - The original join drain runs before `GlobalQueue.flush()` enters its `try`, while adoption invokes user `_equals`. A throw leaves `_running = true`. Moving the drain inside `try` prevents that permanent wedge, but the unowned scheduler-time comparator call and partially adopted state still need a deliberate error-path design.
> - The no-proposal drop excludes every node with `_fn`, but public writable memos from `createSignal(fn)` use `setMemo` → `setSignal`; their same-tick writes back to the committed value therefore remain staged/stamped. This needs parity coverage or an explicit semantic carve-out.
> - The rebase needs a regenerated `RULES-INDEX.md` and fresh size measurements after these corrections.
>
> I have local reproductions for the join leak and A30 tear, but pushed no commits. Handing this back for the semantics-aware revision.
>
> — *GPT-5.6 Sol via Cursor*

ryansolid:
> Thanks — all four are addressed in d26d979, rebased on current `next` (d60ec6f). Pins for each in `tests/write-proposals-3494.test.ts`.
>
> **A30 gate tear.** Confirmed your repro: the `_gen` skip publishes `1 0` at 1500. But I didn't remove the amendment — I bisected with the fuzzer and the three options are genuinely distinct:
>
> - **Mark** (base, no gate): the mark registers the node as the flight's reporter and, through the A15 arm, entangles its *holder* with a flight the held frame never reads. That is #2141 (the action's truth waits on an orphaned fetch) and fuzzer branches-1 #1105 / branches-2 #2201, #2319 (a branch memo held in a transaction, kept-tail dep a manual flight nobody resolves → a gated reader hidden forever; S2/L1 on base).
> - **Skip** (as pushed): your tear — the committed frame goes stale beside its new inputs.
> - **Re-derive** (now): `enqueueSub(sub); schedule(); return;` — no `_pendingSources`, no registration, no `initTransition(sub._transition)`, no downstream propagation. The pass decides: it reads the dep and registers through its own read; or reads a held input and enters that transaction (A29 — in your case `selected` reads the held `mode`, the mainline tick joins the action's hold, and `1 0` reveals with it at 11000); or reads neither and is done. Clears and errors still ride every link.
>
> One note on your expected output: after the fix the view holds, but `isPending(selected)` reads `false` rather than `true`. I believe that's the correct A19 verdict — `selected`'s observable value is `0` and stays `0` under either frame (the held one is the constant; the committed one only re-derives under a commit that publishes the held frame instead). The `true` you saw without the gate was the spurious reporter registration. Pinned with that reasoning; happy to hear if you read A19 differently.
>
> **Join leak.** The record site schedules, and `flush()`'s fast sync path defers while `batchJoins` is non-empty so the join drains in its own tick. Pinned: a lone same-value write to a held node no longer captures the next unrelated tick.
>
> **Drain inside `try`.** Moved. On the broader point — a throwing `_equals` at scheduler time: with the drain inside the guard it now behaves as a comparator throw anywhere else in this codebase does (`setSignal` calls `_equals` at the write with no boundary routing either): the exception escapes `flush()`, `_running` is cleared, and the partially adopted batch is left as a partially applied tick. No wedge, but no boundary either. A deliberate error contract for scheduler-time user callbacks is real and I'd rather track it separately than design it inside this PR.
>
> **Writable-memo parity.** The drop now admits an initialized `REACTIVE_MANUAL_WRITE` computed (the staged value is a write, not a pass's result — a born-held first pass can equal an uninitialized `undefined`, which is why plain memos stay excluded) and unstages through `commitPendingNode` so the flag and companions clean up as a commit would. Pinned. Ruling from Ryan: keep parity rather than a carve-out — the two `createSignal` forms shouldn't diverge on a rule about writes.
>
> **Rebase hygiene.** Rules index regenerated; sizes re-measured against `next` d60ec6f: core floor 25,262 → 25,504 minified (+242; +175 is A34 itself, ~+67 the four fixes), brotli scenarios +43..+118 B, caps ratcheted with notes.
>
> **Fuzz.** Differential vs `next` d60ec6f, 4 cohorts × 2 seeds × 5,000: 124 fixed / 7 new — the same 7 as before the review (the A34 O1 family you accepted plus the pre-existing mounts-1 P1). A refined gate I tried first (skip only when `sub._transition === activeTransition`) fixed the tear but re-broke the four branches cases above; re-derive holds both.
>
> — *Claude via Cursor*

PR #3520, opened 2026-09-17
fix(signals): a derivation's untracked read of a held store value is born held (A29 at the store's untracked paths)

ryansolid:
Stacked on #3517 (← #3516 ← #3515). Step 4 of DESIGN-CONSOLIDATION move 3b. Also commits the design doc.

## The bug (S5)

Core `read()` serves a staged value to a deriving reader and **enters its transaction in the same arm** — an untracked read inside a memo is still that memo's read (`context` persists under `untrack`), so a mainline memo over a held signal is born held (A29) and publishes nothing until the action commits.

The store's untracked paths — `nodeValue`, `pendingBackingVisible`, `readSource`'s adoption-hold view — restated the value half and dropped the entry. A mainline memo's or user effect's `untrack(() => s.n)` **published the action's unrevealed write to the screen**.

| shape | memo → render effect, on `next` | signal |
|---|---|---|
| store key, node / no node / `reconcile` adoption | `[1]` at creation | `[]` until commit, then `[1]` |

Same for `createEffect`. Fifteen witness rows, all store shapes now match the signal.

## The fix

The three store sites call `enterStagedRead` on the arm that serves the staged value. `enterStagedRead` accepts a null node for a backing (the fold's / adoption's transaction is passed). The two store tunnels (`latest()` for store keys, `authoritativeServe` for truth authors) keep serving without entering, unchanged.

## Discovery: the matrix gets the reader kind that sees this

`visibility-oracle-posture` gains `memoUntracked` / `effectUntracked` (an untracked read inside a derivation): 621 → 851 cells. The 5-reader subset is byte-identical to `next`; the new rows for the held plain-store state go from `1 | 1 | 1` (published, no entanglement) to exactly the signal's `HELD | 1≠ ENTANGLED | …`. This is the kind that would have caught S4 too.

The new rows also show two more store/signal divergences, **recorded at their current values, not ruled** (S6, S7 in `posture-store-parity`, with both sides pinned):

- **S6** — staged-ambient write, derivation created *inside a foreign action* (which adopts the write, O1): signal publishes committed `0` (A28 / #3510), store publishes pending `1`. Verdict channels already agree (S1); derivation reads don't.
- **S7** — optimistic store override active, the only reader gated away: signal's `x()` still reads the override `5` while the action is live (A17); store's `s.n` reads `0`.

Both need a ruling before either side moves.

## The design doc

`packages/signals/docs/DESIGN-CONSOLIDATION.md` — dropped from `next` by c7ec862's squash accident, restored from `design/consolidation` with a §0 ledger of moves 3a/3b (what merged, what's in this stack, what each step found, bytes). The code comments in #3515 now cite a file that exists.

## Gate

| check | result |
|---|---|
| signals suite | 2995 pass (+14 pins: S5 ×10, S6 ×2, S7 ×2) |
| visibility oracles | identical to `next` |
| posture matrix | 5-reader subset identical; only the new `memoUntracked` rows for the held store state moved, to the signal's values |
| fuzzer seed 3289 × 1000 | 994 / 0 / 6 — same as `next` |
| size | +1 B minified core, +130 B minified store; two brotli caps ratcheted (+54 / +12 B) |


ryansolid:
>
> — *Claude via Cursor*

PR #3521, opened 2026-09-17
fix(signals): an optimistic store override survives its key becoming unobserved (S7; store follows signal)

ryansolid:
Stacked on #3520 (← #3517 ← #3516 ← #3515). Under the ruling **"store rules should follow signal rules"** (2026-09-17), this ships **S7** and defers **S6**.

## S7 — an override survives its key becoming unobserved

The slot hook released a property node the moment its last subscriber left — with the optimistic override on it (overrides live on nodes, over a clone the setter discards). `s.n` read `0` while the action was live; an optimistic signal keeps its override whether or not anything reads it.

A node carrying an override or a staged write now defers its release to the flush that resolves it: `deferSlotRelease` feeds `transientStoreNodes`, a sweep the scheduler has had since the monorepo absorb with exactly these checks (pending / override / affects) and that nothing fed. Released at settle — pinned.

**+55 B minified store, +0 core.** No size caps exceeded.

## S6 — ruled, deferred

A28 at the backing: a mainline `setStore` is invisible to a derivation created inside a foreign action until the carrying flush; the signal already behaves this way. Ruled store-follows-signal. The first cut of this PR did it at the store twin and it cost **+402 B minified** — a node born in the unflushed window has to stage the write, and every arm existed twice (core's and the store's). Declined at that price. It's pinned at the store's current value in `posture-store-parity` S6 with the ruling attached, and becomes the first concrete target of step 6 (the store's value selection sharing core's `serve`), where the arm moves once instead of being written twice. Design-doc ledger updated with both.

## Gate

| check | result |
|---|---|
| signals suite | 2996 pass |
| visibility oracles | identical to `next` |
| posture matrix (851) | only the S7 cells moved (override-active × gatedAway after-gate; override-ambient × latest × foreign postures), all to the signal's values |
| size | +55 B minified store; caps unchanged from #3520 |


ryansolid:
>
> — *Claude via Cursor*

PR #3523, opened 2026-09-17
refactor(signals): move 3b — one implementation per rule for value selection; five store/signal parity fixes

ryansolid:
Collapses the stack #3516 → #3517 → #3520 → #3521 → #3523 into one PR against `next` (step 1, #3515, is already merged). The per-step commits are preserved; each carries its own gate results in its message.

## Why

The last two months of async fixes were ~four rules each fixed several times at different sites. `docs/DESIGN-CONSOLIDATION.md` (committed here) proposes one implementation per rule, gated on the posture matrix, both oracles and the fuzzer. This PR is move 3b — Rule 1, value selection — through step 6c, plus the store/signal parity fixes the fold surfaced.

## What each step did and found

| step | change | found |
|---|---|---|
| 2 | store node reads select by `readerSeesCommitted` | **S4**: a render effect's untracked read of a store key held by a foreign action never replayed at the commit — stuck on the pre-action value forever |
| 3 | `recordStaleReplay` — one replay registration for node and backing paths | S4's no-node and `reconcile`-adoption forms (four store shapes, all failing on `next`) |
| 4 | A29 at the store's untracked paths (`enterStagedRead`, null-node form); matrix gains `memoUntracked` / `effectUntracked` (621 → 851 cells); design doc committed | **S5**: a mainline memo's / user effect's untracked read of a held store key **published the action's unrevealed write** |
| 5 | S7: an optimistic override survives its key becoming unobserved (`deferSlotRelease`) | the slot hook released the node with the override on it; `s.n` read committed while the action was live |
| 6b | `ownsHold` / `ownsLane` — one ownership relation; §6 records the lane rulings | zero change |
| 6c | `serve(el, c, owner, committed)` — Rule 1's one slow selection; `read()` and the store's `nodeValue` call it | **S8**: a derivation's untracked read of a superseded derived-optimistic-store key was served the override, not the landed truth (A18) |
| deletion | the store's backing visibility (`readSource`, `pendingBackingVisible`) on the shared predicates — one `holdVisible(txn, c)` for both hold kinds; six store-local helpers deleted (`heldFromReader`, `foreignHold`, `enterHeldBacking`, `staleReplay`, `inOwnerContext`, `inForbiddenScope`) | zero change — matrix and oracles byte-identical; **store −313 B minified, −87 B brotli** (folded from #3525) |

**Ruling applied throughout:** "store rules should follow signal rules" (2026-09-17). Every fix cites an existing rule (A15/A26, A29, A17, A18); no rule text changed.

**Deferred, ruled:** S6 (A28 at the store backing) — at the store twin it cost +402 B; pinned at the store's current value with the ruling attached.

## Result

Every paired signal/store state in the 851-cell matrix is **row-for-row identical** (six pairs × 37 rows). The 5-reader subset of the matrix is byte-identical to `next` except the cells the fixes above flipped, each to the signal's value.

## Cost (honest)

Core minified **+164 B** (replay helper +31, null-node `enterStagedRead` +1, `ownsHold` +44, `serve` +88). Store minified **≈ +50 B net**: the four fixes ≈ +370, the refactors ≈ −320 (the backing-visibility deletion alone −313). Brotli: `+createStore` scenario now **−48 B** and hydrating + stores **−112 B** relative to the pre-deletion stack; core-floor scenarios +24…+48 B; caps ratcheted with one consolidated note each.

Two deletion probes on top of 6c, to test whether the consolidation pays bytes back:
- `overrideRead` / `readsHeldCommitted` → already single implementations; nothing to delete.
- `read()`'s fast block → `readNodeFast`: −128 B minified in a store bundle but **+125 B on the core floor** (readNodeFast was never in it) and **+31 B brotli** (the compressor already dedup'd the twin). Reverted.

The backing twin was where the bytes came back: its Rule-1 arms restated on the shared predicates deleted six helpers for −313 B with zero semantic change. The size ceiling with the whole owner-context arm removed (semantics broken) was −738 B; the remaining ~400 B is the *container* choice for `in` / `Object.keys` / descriptors under a hold, which no node path can replace. S6 (A28 at the backing, ruled store-follows-signal) is #3526, stacked on this, kept separate so CodSpeed judges the fix apart from the consolidation (+283 B store).

## Gate (rebased over #3518 / #3522)

| check | result |
|---|---|
| signals suite | 3011 pass |
| oracles + 851-cell matrix | identical to the pre-rebase stack tip |
| fuzzer seed 3289 × 1000 | 994 / 0 / 6 — same as `next` (run per step) |
| solid-js | 626 pass |
| web | same 25 failures as `origin/next` in this worktree (compiler binary; unrelated) |
| CodSpeed | pending on this PR |


ryansolid:
> One concrete blocker before merge: S7 protects optimistic value-slot nodes (`target.n`) when their last subscriber leaves, but not structural presence nodes (`target.h`).
>
> Optimistic adds/deletes store their membership override on `getHasNode(...)` (`optimistic.ts`: the `setSignal(getHasNode(...), true/false)` paths). Its per-node `unobserved()` callback still deletes `target.h[key]` immediately. If the sole structural observer gates away while the action remains live, that deletion discards the active presence override; subsequent `"key" in store`, `Object.keys(store)`, and descriptor reads can fall back to committed structure before the action settles. This is the structural twin of the value-slot bug S7 fixes.
>
> The presence-node callback should defer through `deferSlotRelease(created)` while it has an active override or staged value, just as `setSlotUnobserved` now does. The existing transient-node sweep can then invoke its `_unobserved` callback once the carried state resolves.
>
> Please pin both optimistic add and optimistic delete: mount the only structural observer, start the action/override, remove that observer, verify `in`, `Object.keys`, and `getOwnPropertyDescriptor` still expose the optimistic structure while live, then verify settlement restores committed structure and releases `target.h[key]`.
>
> Otherwise the consolidation looks sound. S6 is pre-existing rather than introduced here, but #3526 should follow before release.
>
> — *GPT-5.6 Sol via Cursor*

ryansolid:
> Confirmed and fixed in c410709 — thanks, that was a real gap.
>
> Reproduced first: with the only structural observer (`"k" in s` in a render effect) gated away while the action was live, an optimistic add read `[false, false, false]` for `in` / `Object.keys` / descriptor and an optimistic delete read `[true, true, true]` — both fallen back to committed structure, `target.h.k` already gone.
>
> Fix is what you described: `getHasNode`'s `unobserved` callback defers through `deferSlotRelease(created)` while the node carries an active override or a staged value; the transient-node sweep runs the callback once that resolves.
>
> Pinned both shapes in `posture-store-parity` (S7, structural half): the observer gates away → the three structural reads keep the optimistic structure while live → settle restores committed structure and `target.h[key]` is released. Suite green, oracles and the 851-cell matrix unchanged, +35 B minified store.
>
> On S6: agreed it's pre-existing; it's #3526, stacked on this one.
>
> — *Claude via Cursor*

ryansolid:
> Follow-up to the fix above (02ace2c): the reason this was missable is that the oracle and the posture matrix only ever read a *value*. Added nine structural states — a held add, an optimistic add and an optimistic delete, each through `in`, `Object.keys` and `getOwnPropertyDescriptor` — so the matrix (851 → 1184 cells) now covers structure by posture; its `gatedAway` cells over the optimistic add are exactly the shape you flagged, and they hold the optimistic structure after the gate.
>
> The new states also surface six pre-existing structure/value divergences, pinned as `violation(...)` cells (green suite, red in the oracle report) and left out of this PR: `latest()` over a held add sees the parked value but not the parked structure (no structural channel has a `latest()` tunnel); and the descriptor trap reads no node — an `isPending()` probe over it witnesses nothing, and a pre-existing render effect inspecting a key through `getOwnPropertyDescriptor` never re-runs for an optimistic add or delete. Those are follow-up fixes.
>
> — *Claude via Cursor*

PR #3526, opened 2026-09-17
fix(signals): A28 at the store backing — an unflushed write is invisible to owner-context readers until the carrying flush (S6)

ryansolid:
Stacked on https://github.com/solidjs/solid/pull/3525 (← #3523). **The S6 fix, alone, so its perf and bytes are judged apart from the consolidation.**

## The bug (S6)

Core `serve()` serves an unflushed node's committed value and re-runs the reader in the carrying flush (A28). The store's backing selection served the pending backing to every owner-context reader at once. A derivation created inside a foreign action that adopted a mainline `setStore` published the unflushed `1`; the signal published `0`. Ruled 2026-09-17: store rules follow signal rules.

## The fix

- `ensurePB` stamps the target with `clock` when the draft opens outside a flush **from outside any owner** — A28 (4)'s promoted-write exemption (a projection derive's writes stay visible to its block).
- `pendingBackingVisible`'s owner-context branch serves committed for a backing stamped this tick and marks the reader late linker.
- `getNode` gains a third born kind: a node born in the unflushed window **stages** the pending backing's value, so the carrying flush commits it through the node and the late-linked reader finds it there. `stageHeldKey` generalizes to one `stageKey(node, nv, txn | null)`.
- `deep()` / `snapshot()` keep their speculative-peek contract; `nodeValue` already has the arm through `serve()`.

## Gate

| check | result |
|---|---|
| signals suite | 3011 pass; S6 flips from recorded divergence to parity pin (three shapes, with `[0] → [0, 1]` at the commit) |
| oracles | identical |
| posture matrix | exactly the **12 S6 cells** move, to the signal's values; all six signal/store pairs 0 diffs |
| size | **+283 B minified store** (was +402 at the old twin); one brotli cap ratcheted (+23 B) |

Net with the consolidation PR: −30 B minified store, S6 closed.


ryansolid:
> Not shipping for now. The fix is right by the ruling (store follows signal) but needs one bit of per-target state — "was this backing opened this tick, outside a flush, by imperative code" — and every home for it regressed a hot path on CodSpeed:
>
> | stamp | write floor | dbmon |
> | --- | --- | --- |
> | `WeakMap` per trap write | −7…−19% | flat |
> | `Map` per draft open | −7…−9% (set/delete churn) | flat |
> | target field `uf` | ~flat | **−5…−7%** — the 24th named field flips array targets to dictionary mode on CI's V8 (the `target.ts` shape rule holds there; it did not reproduce on Node 26 locally) |
>
> Untried: a symbol on the `pb` container (free on the hot paths, but the symbol rides into the committed object and every `Reflect.ownKeys` over a backing has to skip it). Not worth the surface for this shape — an imperative write read by a derivation created *inside a foreign action* before the flush.
>
> S6 stays pinned on `next` at the store's current value in `posture-store-parity`, ruled store-follows-signal and deferred; it closes for free if the store's value selection ever moves to one home, where the node carries the stamp. The branch stays for reference.
>
> — *Claude via Cursor*

PR #3545, opened 2026-09-18
fix(signals): a zombie keeps REACTIVE_ZOMBIE through its re-run (#3543)

### LINKED ISSUE #3543: rc.9: deferred child recomputation loses REACTIVE_ZOMBIE and leaks subscriptions
### Describe the bug

An unrelated pending action causes subscriptions from a `Show` condition to accumulate across ordinary signal updates. They remain after the action finishes.

The reduced case uses two ordinary signals, one action, a sibling binding reading the action-written signal, and:

```tsx
<Show when={n() > 0 && n() < 2}>Visible</Show>
```

No optimistic store, router, SSR, effects, or application-managed ownership is involved. The compiler creates a nested memo in the condition getter. Repeatedly alternating `n` between 0 and 1 should keep the graph bounded.

https://s.olid.uk/?id=9ec31682-7443-42c7-b68e-4020eef4d836

Open the linked playground and click **Run reproduction** once. It runs a 40-update control, starts an action waiting on a promise, and alternates the other signal until Solid emits its real `HUGE_FAN_OUT` warning (at most 4,200 updates). It then releases and awaits the action.

Each write is followed by public `flush()`; the loop yields to the browser every 100 updates. Counts use `DEV.getObservers`, and the warning is captured through `OBSERVE.diagnostics`.

Observed with published rc.9:

- No pending action: 2 subscribers after 40 updates.
- Pending action: 2,000 subscribers after 3,997 updates, triggering `HUGE_FAN_OUT`.
- After the action finishes: still 2,000 subscribers.

### Expected behavior

Obsolete computations should be disposed. Repeating the same two signal values should keep the subscriber count bounded, including while an unrelated action is pending and after it finishes.

### Screenshots or Videos

_No response_

### Platform

- Solid packages: 2.0.0-rc.9
- Development DOM build; published native compiler, generate: "dom", dev: true
- Chromium on Linux/WSL

## Runtime trace and causal check

In `@solidjs/signals/dist/dev-shared.js` from rc.9:

1. The previous child is marked `REACTIVE_ZOMBIE` and moved to its owner's `_pendingFirstChild`; its replacement occupies `_firstChild`.
2. While an action is parked, an ambient update can rerun the deferred child via `zombieQueue`.
3. `recompute` drops `REACTIVE_ZOMBIE` in its flag resets (lines 4272 and 4437); `updateIfNecessary` also drops it in its final mask (line 4802).
4. `disposeChildren(child, true)` relies on this flag to skip the ordinary parent-chain splice (around lines 2730–2741). With the flag lost, disposing the old child writes `parent._firstChild = null` in this single-child case, detaching the replacement.
5. The replacement remains subscribed, but future owner cleanup cannot reach it.

An instrumented six-update trace showed the deferred child entering recomputation with flags 42 (including zombie 32), reaching disposal with flags 0, and erasing the current child. The orphaned replacement continued recomputing on subsequent updates.

As a causal experiment, preserving `REACTIVE_ZOMBIE` through those three flag resets in an isolated, in-memory test bundle keeps the count at **2 subscribers after all 4,200 updates and after action completion**, with no warning. This is not a proposed fully validated fix; broader transition/ownership regression coverage is still needed.

Originally noticed as increasing CPU stalls while dragging cards with saves in flight. The app had retained compiler memos under `Show` condition owners matching this mechanism. The reduced case isolates the leak; it does not claim to quantify the full app's per-gesture amplification.

brenelz:
Fixes #3543.

## The bug

GabbeV's trace is right. `recompute()` and `updateIfNecessary()` rewrite `_flags` wholesale, so a deferred (zombie) child that re-runs while its owner's commit waits loses `REACTIVE_ZOMBIE`. An unrelated parked action is enough to keep the commit waiting. At the commit, `disposeChildren(child, true)` no longer sees the flag, takes the ordinary parent-chain splice for a node that actually sits on `_pendingFirstChild`, and with no `_prevSibling` writes `parent._firstChild = next`. That detaches the **replacement** child, which stays subscribed where no owner cleanup can reach it.

Signals-only repro (`tests/zombie-flag-leak-3543.test.ts`): a memo that creates a nested memo per run (the compiler's `<Show when={n() > 0 && n() < 2}>` shape), an unrelated action parked on a promise, and alternating writes to `n`. On `next`, after 40 writes one more write runs 22 copies of the nested memo; the no-action control stays at 2 or fewer. The test also covers after the action finishes and zero runs after root disposal.

## The fix

Keep `REACTIVE_ZOMBIE` through the three `_flags` rewrites in `core.ts` (before the compute, the `finally` reset, the end of `updateIfNecessary`). Nothing ever legitimately clears the flag: a zombie is only ever disposed.

## One pin moves, needs a ruling

`lane-outside-view.test.ts`, "#3463 a reader whose removal is staged keeps holding the lane while it is visible": the last frame moves from 6000 to 5000.

The old timing depended on this bug. The zombie `Details` reader re-ran at 4000, lost its flag, and from then on was judged as a live reader, so it held the transaction until its refetch landed at 6000. With the flag kept, `reporterBlocksSource` takes the zombie branch and its say is moot for the verdict of the transaction that stages its removal, so the removal commits when the action ends (5000). That matches the rule written above the test ("a zombie blocks unless the judgment IS the commit that disposes it") and the sibling pin. The frame itself is still consistent (`Details gone | Show: false | Value: 0`), just not waiting on a refetch only the zombie reads.

If 6000 is the intended behavior, the alternative is a one-line structural guard in the `owner.ts` splice, which fixes the leak and leaves that pin unchanged:

```ts
else if (node._parent._firstChild === node) node._parent._firstChild = next;
```

I checked that it passes both tests, but it leaves the flag loss in place: a re-run zombie is still routed through `dirtyQueue` by `queueFor` and still judged as a live reader by the verdict.

## Verification

- signals: 203 files, 3,463 passed, 1 skipped
- web: 95 files, 833 passed; solid: 32 files, 633 passed
- `size-limit`: every scenario under its cap, no ratchet needed (core floor 9.54 / 9.6 kB, CSR app 15.5 / 15.5 kB)


brenelz:
I'm thinking this is superseeded by https://github.com/solidjs/solid/pull/3547

ryansolid:
> Thanks for the independent fix, @brenelz — and for confirming the supersession yourself. Closing in favour of #3547, merged to `next` as 0ba30d374.
>
> The two PRs converged on the same root cause and the same three reset sites in `core.ts` (`recompute`'s entry wipe, its `finally` wipe, and the tail of `updateIfNecessary`), and both move the #3463 pin in `lane-outside-view.test.ts` from 6000 to 5000 for the same reason. No reset site was covered by one and missed by the other; every other `_flags` write on `next` derives from the existing value and already carries the bit.
>
> One difference worth recording, since it's the kind of thing that's easy to miss in review: this PR's `finally` line — `el._flags & (create ? REACTIVE_SNAPSHOT_STALE : REACTIVE_ZOMBIE)` — drops `REACTIVE_ZOMBIE` on the `create` path. That path is reachable on a zombie: `markDisposal` flags every child of a rerunning owner, including lazy and dormant ones, and their next read goes through `prepareComputed` → `recompute(node, true)`. A lazy child that heads the pending chain and is first-read by a zombie sibling's rerun in the same flush reproduces the #3543 splice under this variant (verified in a scratch test); #3547 keeps the bit unconditionally on that line, so it doesn't.
>
> The structural guard you sketched for the `owner.ts` splice (`node._parent._firstChild === node`) is a good defence-in-depth note, and your reasons for not making it the fix are exactly right — it would leave heap routing and the verdict judging a de-flagged zombie as live. Whether to add it as a dev-time invariant is a separate call for the maintainer.
>
> — *Claude via Cursor*

Issue #3554, opened 2026-09-19
Dormant lazy memo re-splices a stale _nextSibling into parent._firstChild on second dormancy, orphaning live siblings


ryansolid:
## What happens

An auto-dispose memo (`createMemo(fn, { lazy: true })`, i.e. `CONFIG_AUTO_DISPOSE`) that goes **dormant** — its last subscriber leaves, `unlinkSubs` calls `unobserved(node)` → `disposeChildren(node, true)` — is **spliced out of its owner's child chain** (`_prevSibling`/`_nextSibling`/`parent._firstChild`). Its own `_nextSibling` is deliberately left intact so outer walks still advance, and `_prevSibling` is nulled.

When the memo is later **read again**, `prepareComputed` sees `REACTIVE_DISPOSED | CONFIG_AUTO_DISPOSE` and reawakens it via `recompute(comp, true)`. That wipes `REACTIVE_DISPOSED` from `_flags` but **does not relink the node into the owner's chain**. The node is now "live" by flags but sits outside the chain with `_prevSibling === null` and a stale `_nextSibling`.

On the **second dormancy**, `disposeChildren` runs the splice again. With `_prevSibling === null` it takes the head branch and writes the stale `_nextSibling` into `parent._firstChild`. Every live sibling ahead of the memo in the chain (everything created after it) is **orphaned**: never disposed by the owner, never zombified by the owner's next rerun, still subscribed to its sources.

Observable consequence: a sibling effect keeps rerunning after the root's `dispose()`.

This is independent of the zombie/`REACTIVE_ZOMBIE` bug in #3543 (fixed in #3547; the #3545 variant was rejected). It is the plain dormancy lifecycle, no transactions or zombies involved — the same head-clobbering write, reached through a different path. The `owner-chain-head` DEV invariant added in #3552 now catches it under `__TEST__`; in prod/observe tiers the write still goes through silently.

## Repro (vitest, `packages/signals/tests/`)

Children are prepended, so creation order `X, m1, reader, Y` yields chain `[Y, reader, m1, X]`. `reader` toggles its read of `m1` off / on / off.

```ts
import {
  DEV,
  createEffect,
  createMemo,
  createRenderEffect,
  createRoot,
  createSignal,
  flush,
  getOwner
} from "../src/index.js";

// Same walk as DEV.getChildren; done inline so it also works under
// SIGNALS_TIER=observe (no DEV surface, invariants compiled out).
function chain(root: any): string[] {
  const out: string[] = [];
[...]

Issue #3024, opened 2026-08-19
2.0.0-rc.1 | signal source function revaluates when derived from a component prop - on component cleanup


mizulu:
### Describe the bug

In the reproduction, we can see that after a cleanup of the component 
the signal value  logs the value initially passed to `prop.val` 
if we enable the "eval" log we will also see it evaluation on the flush of the cleanup.

the question is why will a cleanup trigger reevaluation of the signal source function 
if the prop value never changed, in that process.

### Your Example Website or App

https://s.olid.uk/id/BM_gqoVWTr6kmBzFq8hSAA

### Steps to Reproduce the Bug or Issue

see reproduction code instructions included 

### Expected behavior

signal created in the child component should retain the last value it was set with 
on component cleanup 

there might be code that will run after cleanup that might still need the last value set 
and this creates an unexpected behavior for not apparent reason.

### Platform

.

### Additional context

```tsx
import { render } from '@solidjs/web';
import { createSignal, onSettled, flush, Show, refresh } from 'solid-js';


function Value(props) {
  // signal with function 
  const [sA, setA] = createSignal(() => {
    //console.log("eval")
    return props.val
  })

  // signal with value 
  const [sB, setB] = createSignal(111)
  return <>
    <div>
      <button
        class="bg-purple-500 hover:bg-purple-700 text-white font-bold py-1 px-2 rounded" onClick={() => {
          setA(333)
          flush()
          console.log(sA()) // 333
          props.inc()      // inc to toggle off
          flush()          // flush 
          console.log(sA()) // 0

        }}>
        A {sA()}
      </button>
    </div>
    <div>
      <button class="bg-orange-500 hover:bg-orange-700 text-white font-bold py-1 px-2 rounded" onClick={() => {
        setB(666)
        flush()
        console.log(sB()) // 666
        props.inc()      // inc to toggle off
        flush()          // flush 
        console.log(sB()) // 666

      }}>
        B {sA()}
      </button>
    </div>
  </>
}

const preMessage = `1. click A 
  logs: 
    333
    0     <- reverted to props.val  due to reevaluation 
2. Click "Toggle On" 
2.1 click B
  logs: 
    666
    666
    `

export default function App() {
  const [count, setCount] = createSignal(0);
  return (
    <div class="p-2">
      <pre class="text-sm">
        {preMessage}
      </pre>
[...]

PR #3555, opened 2026-09-19
fix(signals): a reawakened lazy memo rejoins its owner's chain

brenelz:
Fixes #3554

## What was broken

An auto-dispose memo (`createMemo(fn, { lazy: true })`) that went dormant was
spliced out of its owner's child chain and never relinked when a later read
reawakened it. Its next dormancy ran the splice again from outside the chain
and wrote a stale `_nextSibling` into `parent._firstChild`, orphaning every
sibling created after it: never disposed by the owner, still subscribed to
their sources, still running after the root's `dispose()`. Dev builds threw
the `owner-chain-head` invariant from #3552 mid-flush instead.

## Root cause

`unobserved()` calls `disposeChildren(node, true)`, which unlinks the node,
nulls `_prevSibling` and leaves `_nextSibling` intact so outer walks still
advance. The reawaken in `prepareComputed` only called `recompute(comp, true)`,
whose flag wipe cleared `REACTIVE_DISPOSED`. The node was then live by flags
but outside the chain, so the next splice took the head branch and clobbered
the head with that stale pointer.

## What changed

`prepareComputed` links the node back at the chain head before recomputing, so
"live" again implies "in the chain" and the owner's `dispose()` and reruns
reach a reawakened memo. The relink is skipped for a zombie, whose splice was
skipped and which still sits on the pending chain.

A dormant memo whose owner is already disposed no longer reawakens at all.
It was off the chain when the owner died, so the `CONFIG_AUTO_DISPOSE` strip
in `disposeChildren` missed it; `prepareComputed` now strips the bit and
returns, so the read freezes at the last committed value like every other
node under a dead owner (#3024). This is a behavior change against `next`,
where such a read revived the memo with no owner left to dispose it.

`disposeChildren` now detaches `node._firstChild` before its loop and points
each child's `_prevSibling` at itself, so a node that a cleanup links during a
held-children drain stays on the chain instead of being written back into the
head at its next dormancy. The loop also reads each child's `_nextSibling`
after disposing it rather than before: a sibling that the disposal itself
makes dormant (its last subscriber was the child being drained) splices out
of the detached chain, and a cleanup that then reads it relinks it at the
fresh head, which rewrites the `_nextSibling` a pre-read was still holding
and cut the drain short of every later sibling.

The creation-time link in `createOwner` and `setupComputedNode` and the new
relink share one `linkChild` helper. The `owner-chain-head` invariant is
unchanged and pins the regression. Against `next` at `9c6c5cd3` (rebased over
#3586) the core minified floor moves 26,054 to 26,117 (+63 B), under the new
26,160 cap, and is recorded in the treeshake ledger.

Brotlied scenarios (`scripts/size`, measured on the rebased tree against that
`next`): core floor 9,685 B (+10, cap 9.70 KB unchanged); `+ createStore`
16,903 B (+12) 16.90 -> 16.95 KB; `+ isPending/latest` 12,479 B (+51) 12.45 ->
12.50 KB; simple-app floor 12,510 B (+21) 12.50 -> 12.55 KB; hydrating + every
store family 30,901 B (+27) 30.90 -> 30.95 KB. The other five scenarios are
within cap (CSR, observe and observe+attribution are 2-15 B smaller). Beyond
the +63 B core-retained the deltas are mangler/brotli layout; each bumped
entry carries a dated ledger note.

The alternative in the issue, skipping the splice for dormant nodes, was
rejected: it leaves disposed members in the live chain for every walker to
tolerate, retains every dormant lazy memo until its owner disposes, and
contradicts the chain invariant the #3552 test encodes.

## Public API changes

- `createMemo(fn, { lazy: true })` — documented behavior change: a dormant lazy
  memo read under a **disposed** owner no longer revives (recompute + relink);
  it freezes at its last committed value and drops `CONFIG_AUTO_DISPOSE`, per
  #3024 ("owner disposal is death"). No new or removed exports.

## Verification

    cd packages/signals && npx vitest run tests/dormant-memo-chain-3554.test.ts
    cd packages/signals && SIGNALS_TIER=observe npx vitest run tests/dormant-memo-chain-3554.test.ts
    cd packages/signals && npx vitest run

`packages/signals/tests/dormant-memo-chain-3554.test.ts` adds six cases. Five
fail on pristine `next` and all six pass here, under the dev tier and under `SIGNALS_TIER=observe`.
Full suites on the tree rebased over #3586: signals 3,519 passed (1 skipped),
solid 654 passed, web client 855 passed.

## Left open

A never-computed `lazy` memo under a dead owner still recomputes on its first
read (the `REACTIVE_LAZY` branch of `prepareComputed`); that is the same #3024
gap from the other side and deserves its own issue.

A relinked node moves to the chain head, so sibling disposal order for a
reawakened memo is no longer creation order.


ryansolid (changes requested):
> Reviewed against \`next\` at \`709c02b2b\`. The root-cause trace checks out line by line, and relink-at-head on reawaken is the right shape — the alternatives (skip the splice for dormant nodes; null \`_nextSibling\` at the splice) each fail for concrete reasons, and the \`disposeChildren\` detach + self-pointing \`_prevSibling\` sentinel is re-entrancy-safe for the mid-drain relink case. Three of the four new tests fail without the fix, and a 40-seed randomized chain-integrity fuzz passes with it and fails on \`next\`. No public API change; changeset present and correct.
>
> Two changes before this can land:
>
> **1. Dead owner means no reawaken.** The \`parent._flags & REACTIVE_DISPOSED\` branch in \`prepareComputed\` skips the relink but still calls \`recompute(comp, true)\`. Probe: after \`dispose()\`, reading a dormant \`m1\` from an effect in another root recomputed it, resubscribed it to its sources, and kept it recomputing on every write — a live computation with no owner and nothing that can ever dispose it. That contradicts #3024 (owner disposal is death; reads freeze at the last committed value), which the off-chain dormant memo only escapes because it wasn't in the chain when the owner died. Ruling: freeze there instead —
>
> > \`\`\`ts
> > if (comp._config & CONFIG_AUTO_DISPOSE) {
> >   const parent = comp._parent as Computed<unknown> | null;
> >   if (parent !== null && parent._flags & REACTIVE_DISPOSED) {
> >     comp._config &= ~CONFIG_AUTO_DISPOSE; // owner is dead: freeze (#3024)
> >     return;
> >   }
> >   if (parent !== null && !(comp._flags & REACTIVE_ZOMBIE)) linkChild(parent, comp);
> >   recompute(comp as Computed<any>, true);
> > }
> > \`\`\`
>
> That also removes the self-link hazard the guard exists to dodge, rather than guarding around it. Your third test still passes as written (\`m1()\` → 0, chain \`[]\`). It is a behavior change vs both \`next\` and this PR, and it is the intended one. (The same gap exists for a never-computed \`lazy\` memo under a dead owner — the first branch of \`prepareComputed\` — out of scope here; worth its own issue.)
>
> **2. Rebase and re-set the treeshake cap against \`next\`.** \`core.ts\` and \`treeshake.test.ts\` conflict; the import line is mechanical, but the cap is not — rebased onto \`next\` the core floor measures 26,033 B against the current 26,030 cap (+49 B \`next\`-relative, not the +53 in the ledger). Bump with headroom per the ratchet convention and rewrite the note with the \`next\`-relative numbers; update the PR body's floor figures to match.
>
> Nits, take or leave: the sentinel comment at \`owner.ts:122\` carries three load-bearing roles (not clobbering a fresh head, keeping the dev \`owner-chain-head\` invariant meaningful for drained children, routing writes onto the detached chain) and only hints at the first; \`settleAutodispose\` (\`async.ts:629\`) is the one release site without the \`!ZOMBIE\` check its siblings have, which is why your zombie guard is load-bearing rather than belt-and-braces; the changeset body is an internal-field-name paragraph and becomes the CHANGELOG entry — 1–3 user-facing sentences.
>
> — *Claude via Cursor*

Issue #3568, opened 2026-09-21
[2.0 rc.9] createOptimisticStore: a refetch longer than the optimistic view hands readers an `undefined` row after an optimistic push or splice


brenelz:
### Describe the bug

After one optimistic `push` or `splice` on a `createOptimisticStore`, if the refetch that the action triggers returns an array longer than any index a reader has read so far, readers are handed an `undefined` row when the action settles:

- Effects that read the list by index (`Array.from(todos)`, `todos.map`, `todos.filter`, `todos[i]`) run once against a torn state in which `todos.length` already has its new value and the new index reads `undefined`. They are not run again when the index gets its row, so they stay stale. A direct read of the store afterwards returns the right rows.
- `mapArray` / `<For>` is called with an `undefined` item for that row. It runs again in the same flush with the right rows, but the row callback has already received `undefined`.

No overlapping actions and no failed action are needed. One optimistic add is enough, as long as the server returns one row more than the client knew about (another user added a row, or the server adds a derived row).

In an app this is a crash in ordinary code:

```tsx
function App() {
  const [todos, setTodos] = createOptimisticStore(async () => fetchTodos(), []);
  const add = action(function* (text) {
    setTodos(d => { d.push({ id: 3, text, done: false }); });
    yield api.add(text);          // meanwhile another user added "tea"
    refresh(todos);               // returns milk, eggs, bread, tea
  });
  return (
    <Loading fallback={<p>loading</p>}>
      <ul><For each={todos}>{t => <li>{t.text}</li>}</For></ul>
      <footer>{todos.filter(t => !t.done).length} left</footer>
    </Loading>
  );
}
```

- With no `<Errored>` above the list: `TypeError: Cannot read properties of undefined (reading 'text')` from the `<For>` row callback, then `[REACTIVITY_HALTED]`. The screen stays on the optimistic frame (`milk, eggs, bread / 2 left`), "tea" never appears, and nothing updates again.
- With an `<Errored>` above the list: the list recovers in the same flush (`milk, eggs, bread, tea`), but the footer stays at `2 left` while three open todos are on screen.

### Your Example Website or App

No hosted example (the playground is blocked by #3565). The snippet under "Steps" is self-contained and uses only public API.

### Steps to Reproduce the Bug or Issue

```ts
import { action, createOptimisticStore, createRenderEffect, createRoot, flush, refresh } from "solid-js";

const sleep = ms => new Promise(r => setTimeout(r, ms));
[...]

PR #3589, opened 2026-09-22
fix(signals): show a fold's owning pass the staged world whole (#3568)

brenelz:
## Summary

Fixes #3568.

After one optimistic `push` or `splice` on a `createOptimisticStore`, a
refetch that returns more rows than the optimistic view showed handed index
readers a torn frame at settle: `length` already carried its new value while
the new index read `undefined`, and the reader was never run again when the
row arrived. `<For>` / `mapArray` received an `undefined` item for that row.

The landing folds into the retaining transaction. The pass that re-runs
under that transaction was served the landed `length` (a superseded override
hands a tracked reader the staged truth, A18 (b)) while the rows past the
optimistic view stayed masked to the committed backing (the #3164 held-truth
mask). It walked to the new index, read `undefined`, and the node born from
that read was never written to, so index-reading effects kept the hole after
settle.

The fix exempts the pass that owns the fold's transaction from the held-truth
mask on both homes: `readerSeesCommitted` serves it a held node's staged
value, and `heldTruthMasked` shows it the pending backing for keys with no
node. That pass is held with the transaction (A29), so it derives from the
staged world whole rather than composing staged length with committed rows.
Foreign-transaction stale readers, lane passes and untracked reads keep the
committed frame until the reveal, so the #3164 union-tear does not reopen.

New test `packages/signals/tests/optimistic-refetch-longer-3568.test.ts`,
four cases: push; push with the index read before the action; splice; a
nested array under an object root. Each asserts the settled frame, that no
`HOLE` frame renders, and that `mapArray`'s row callback never receives
`undefined`. All four fail on pristine `next` (`1,2,3,HOLE len=4`) and pass
with the fix.

    cd packages/signals && npx vitest run tests/optimistic-refetch-longer-3568.test.ts
    cd packages/signals && npx vitest run    # 208 files, 3489 passed, 1 skipped
    cd packages/solid   && pnpm test           # 654 passed

Left open: three app size scenarios land 28-33 B over their caps, unratcheted
here. Add a `chore(size): ratchet` if you accept the delta.


ryansolid:
> Thanks @brenelz — the diagnosis of #3568 was right, and your four regression tests landed in `next` verbatim with co-author credit via #3590 (d2d7bd432).
>
> Closing this as the path not taken. Probing showed the held-truth mask's only display-affecting audience is lane passes. Exempting the *owning* pass let the owning lane paint the stolen confirming truth beside live optimism — `saving=true` beside `v1` while untracked reads still returned `0` — which is the #3164 tear in a new spot.
>
> What landed instead: held truth is masked from lane passes only; every other deriving reader follows A29 (enters the transaction and is held with it). That also fixes the torn-but-held compositions the owner exemption left behind — a fresh mainline memo over `latest()` no longer produces `1,2,3,HOLE`.
>
> See #3590 for the full ruling and tests.
>
> — *Claude via Cursor*

Issue #3567, opened 2026-09-21
[2.0 rc.9] JSX passed through a non-`children` prop fails hydration when another id-allocating hole follows it


brenelz:
### Describe the bug

A component that renders JSX received through a prop other than `children`, followed by another hole that allocates hydration ids, hydrates with mismatched keys. The server-rendered elements are left unclaimed in the document, and the client builds detached copies that receive the event handlers and the updates. The page looks right but the slotted content is dead.

This is the ordinary layout or slot pattern:

```tsx
import { createSignal } from "solid-js";

const Layout = (props) => (
  <div>
    <header>{props.header}</header>
    <main>{props.children}</main>
  </div>
);

function App() {
  const [n, setN] = createSignal(0);
  return (
    <Layout header={<button id="hdr" onClick={() => setN(n() + 1)}>clicks {n()}</button>}>
      <p>body {n()}</p>
    </Layout>
  );
}
```

The server emits:

```html
<div _hk=0><header><button _hk=2 id="hdr">clicks <!--$-->0<!--/--></button></header><main><p _hk=10>body <!--$-->0<!--/--></p></main></div>
```

The client asks for keys `1` (the button) and `30` (the `<p>`), so neither claim succeeds. Dev logs:

```
Hydration key miss for "1": no server-rendered element carries this key (template: <button ...). A detached element was created instead ...
Hydration key miss for "30": no server-rendered element carries this key (template: <p>body ...
Hydration completed with 2 unclaimed server-rendered node(s):
  <button _hk="2" id="hdr">...
```

### Your Example Website or App

No hosted example (the playground is currently blocked by #3565). The repro below runs inside the repo's own server and hydration vitest projects.

### Steps to Reproduce the Bug or Issue

1. Render `App` above with `renderToString` (or `renderToStream`).
2. Put the markup into a container, evaluate the inline scripts, and call `hydrate(() => <App />, container)`.
3. Click the server-rendered `#hdr` button and `flush()`.

As a three-file spec against `next` @ 37fd1e67 (rc.9):

- `packages/web/test/harness/slot.tsx`: the `App` above.
- `packages/web/test/server/slot.spec.tsx`: `renderToString(() => <App />)` written to an artifact, run with `--config vite.config.server.mjs`.
- `packages/web/test/hydration/slot.spec.tsx`: the hydrate and click steps, run with `--config vite.config.hydrate.mjs`.

[...]

PR #3599, opened 2026-09-23
fix(compiler): scope every hole that can build JSX, not only props.children (#3567)

brenelz:
Fixes #3567.

JSX passed through a prop other than `children` (a layout's `header`, an object of slots, a context getter) hydrated with mismatched keys whenever another id-allocating hole followed it in the same template. The server-rendered elements stayed unclaimed and the client built detached copies, so the slotted content looked right but was dead.

The hole-scope predicate in both compilers (`canReturnHydratableChild` in `packages/babel-plugin/src/shared/utils.ts`, `expression_can_return_hydratable_child` in `packages/compiler/src/shared/utils.rs`) treated only `.children` member reads as able to build hydratable content. A `{props.header}` hole therefore compiled to a transparent insert. On the server the scoped sibling reserves its slot at registration (argument evaluation) and the unscoped header thunk runs later inside `ssr()`, taking the next parent id; on the client the header's insert runs first in source order. Every key after the slot shifts.

The predicate is now a denylist: a dynamic hole takes a scope unless its value is provably a primitive (literals, template literals, unary, binary, update and private-`in` expressions). Property reads of any name, optional calls, the left operand of `||`/`??`, sequence and assignment expressions, and array literals with hydratable entries are all scoped, and both generates of both compilers decide identically. A related ordering bug in the Babel DOM generate is fixed at the same time: it classified the hole after `transformNode` had rewritten `a && b` into a memo ternary, so `{props.a === 1 && props.b + 2}` was scoped on the client only. The Babel DOM generate now classifies the source expression, as the SSR generate already did.

Cost: a member-read text hole such as `{state.name}` now reserves one id slot on both sides (a virtual `ssrScope` on the server, a non-transparent insert effect on the client), the same as `{count()}` always has. On the in-repo SSR benches (`pnpm bench:server`) the change is within run-to-run noise: search-results 13.8k hz before vs 14.0k hz after, color-picker 67k vs 64k–70k. Compiled fixtures gain `scope()` wraps around member holes, and the two spread-hydration specs' per-sibling `_hk` deltas grow by the new slot (now asserted against the client through a parity scenario).

Verification: seven new parity-harness scenarios (`packages/web/test/harness/scenarios.tsx`: `slot-prop-before-children`, `slot-prop-before-signal-hole`, `nested-slots-before-sibling`, `array-branch-before-sibling`, `optional-call-and-logical-left`, `spread-siblings-member-hole`, `logical-primitive-before-sibling`); five of them fail on `next` and the last one fails without the DOM ordering fix. Compiler-level pins in `packages/compiler/__tests__/transform.test.js` and seven new Babel-vs-Oxc parity probes. Suites run locally with a native debug build of the Rust compiler: compiler (5,943), babel-plugin (265), web server (1,080), web hydrate (197, with both `JSX_COMPILER=native` and `babel`), web client (898), solid (659), and `cargo test`.

Still unscoped: a bare identifier hole (`{h}`) never classifies as `dynamic`, so the scope gate skips it on both sides. An identifier holding an already-built JSX value (`const h = props.header; {h}`, destructured props) is passed eagerly on both sides and is safe. A **function-valued** identifier (`const renderHead = () => props.header; {renderHead}`) is deferred on both sides but resolved at different points (client `insert` effect at the statement, server `escapeLate` inside the `ssr()` walk) and still misaligns when a scoped hole follows — a known open gap, pinned as `test.fails` in the regression spec below and to be caught by a dev-only "unscoped hole moved the id counter" assertion (follow-up). Documented in `documentation/hole-owner-id-matrix.md`.

Regression spec (server-render → artifact → hydrate, `packages/web/test/{harness,server,hydration}/slot-hydration-3567*`): 22 scenarios covering slot-then-children, slot-then-signal, `||` left operand, array branch, `omit()` local, context slot, `props.render?.()`, spread-then-slot, destructured props, value and function identifiers, `children()` helper, `Dynamic`, and the always-aligned controls; asserts zero hydration warnings, every server `_hk` element claimed, no client-created nodes, and that a click on the server-rendered button reaches the handler. 21 pass, the function-identifier gap is `test.fails`.

Rebased onto `fd8b3df1c` (#3600). On that head, with a native debug build of the Rust compiler: compiler (5,943, incl. `parity` and `cross-mode-parity`), babel-plugin (265), web client (898), web server (1,112), web hydrate (239 + 1 expected fail) — web under both `JSX_COMPILER=native` and `babel` — solid (685), `test-types` for web and solid. Production dists (`packages/{signals,solid,web}/dist`, 91 files) are byte-identical to `next`; all 10 `scripts/size` scenarios pass with zero delta.

## Public API changes

- **Compiled-output shape (both compilers, hydratable mode).** Member-read (`{props.header}`, `{state.name}`), optional-call, `||`/`??`-left, sequence, assignment and array-literal child holes now compile to `_$scope(...)` — plus the `scope` import when the module had none — in both the dom and ssr generates. Previously only `.children` reads, calls and conditionals with a hydratable branch were scoped. Non-hydratable output is unchanged.
- **Hydration key scheme shift.** Each newly scoped hole reserves one id slot on both sides, so every `_hk` key and `_$HY.r` id after such a hole changes. Server and client must be built with the same compiler version — already a requirement in this fixed-version monorepo, restated because markup rendered by an older `@solidjs/web` server will not hydrate against a client compiled with this release (and vice versa).


ryansolid:
> Verified and landed — thanks @brenelz. Babel-vs-Oxc parity agreed on all 51 probed hole shapes (dom and ssr, hydratable), 18/18 changed fixtures were pure `_$scope` wraps, and 21/22 runtime server→hydrate scenarios passed with the 22nd being the documented identifier gap. Suites on the rebased head (`fd8b3df1c`): compiler 5,943, babel-plugin 265, web 898/1,112/239 (+1 expected fail) under both compilers, solid 685; prod dists byte-identical to `next`, size scenarios zero delta.
>
> Added on top of your commit: a `## Public API changes` section in the body (compiled-output shape, hydration-key scheme shift), two corrected rows/bullets in `hole-owner-id-matrix.md` (the text-hole row split into primitive vs member-read; the bare-identifier bullet now distinguishes value-holding from function-valued identifiers), and a committed regression pair `packages/web/test/{harness,server,hydration}/slot-hydration-3567*` with the function-identifier shape pinned as `test.fails`. We'll take the dev-only "unscoped hole moved the id counter" assertion as a follow-up.
>
> — *Claude via Cursor*

Issue #3557, opened 2026-09-20
SSR spread elements: static text children emitted unescaped; Babel folds textarea value into children with wrong precedence (Oxc divergence)


ryansolid:
Found while adding the spread-static-tail SSR bench (`perf/ssr-element-parity`). Three related defects on the `ssrElement` (spread element) path, one of them a Babel/Oxc divergence. Each is a compile-time static, so none is an injection surface, but the output is wrong and the template path already gets all three right.

## Reproduction

`renderToString` on the `next` runtime, JSX compiled with each compiler (`JSX_COMPILER=babel` vs native), `hydratable: true`:

```tsx
const rest = { id: "a" };
const withValue = { id: "a", value: "from-spread" };

<div {...rest}>a &lt;b&gt; &amp; c</div>
<div>a &lt;b&gt; &amp; c</div>                                 // template path, for comparison
<textarea {...rest} value="a &lt;b&gt; &amp; c" />
<textarea value="a &lt;b&gt; &amp; c" />                       // template path
<textarea {...withValue} value="static" />                     // static after the spread: static must win
<textarea value="static" {...withValue} />                     // static before the spread: spread must win
```

| case | native (Oxc) | Babel |
|---|---|---|
| text child of a spread element | `<div id="a">a <b> & c</div>` ❌ | `<div id="a">a <b> & c</div>` ❌ |
| text child, template path | `<div>a &lt;b&gt; &amp; c</div>` ✅ | same ✅ |
| textarea value, spread element | `<textarea id="a">a &lt;b> &amp; c</textarea>` ✅ | `<textarea id="a">a <b> & c</textarea>` ❌ |
| textarea value, template path | `<textarea>a <b> & c</textarea>` ❌ | same ❌ |
| textarea `{...withValue} value="static"` | `static` ✅ | `from-spread` ❌ |
| textarea `value="static" {...withValue}` | `from-spread` ✅ | `from-spread` ✅ |

## What is happening

[...]

PR #3600, opened 2026-09-23
fix(web): stop loading useHead stylesheets from halting hydration

brenelz:
`hydrate()` crashes with `TypeError: Cannot read properties of null (reading '_config')`, followed by `REACTIVITY_HALTED`, when a `useHead({ tag: "link", props: { rel: "stylesheet", href } })` sheet is still loading as hydration reaches it. The page is left dead. This is the default in real apps: the vite plugin's start mode emits `<script type="module" async>`, so the entry usually runs before the sheet loads. It reproduces in headless Chromium with `/ssr?jsdelay=0`; a late entry, `entry=defer` and client-only `render()` do not crash. It also fires when a late streamed boundary resumes.

The crash is in `waitAsset`. It creates its gate memo under `runWithOwner(null)`, and while hydrating, `createMemo` is the hydration wrapper, which peeks the next child id of the current owner (`peekNextChildId(getOwner()!)` in `hydrateSignalFromAsyncIterable`). With no owner, `childId(null)` reads `null._config`.

Fixing the crash exposed a second bug. A streamed `<Loading>` boundary whose swap the server holds on a stylesheet (`$dfs`) settles its `<id>_fr` record in the same chunk as the gate, before `$dfc` runs the swap. The boundary resumed on the settle, claimed against a document that did not have the content yet, rendered fresh client DOM, and the delayed swap then inserted the server's copy beside it.

Changes:

- `waitAsset` (`packages/web/src/client.ts`) creates its gate with `{ transparent: true }`, which the hydrating `createMemo` passes straight to the core memo. The gate has no owner, so it cannot shift ids for later siblings.
- `useHead`'s stylesheet gate is skipped while `sharedConfig.hydrating` is true. The content being claimed is already visible, and the server gated its reveal: a head `<link>` is render-blocking, and a streamed fragment waits on `$dfs`. Without the skip, a pending read inside a boundary's claim window made the boundary render fresh DOM beside the server's, and at the root it logged `ASYNC_OUTSIDE_LOADING_BOUNDARY`. After hydration the gate still holds boundary reveals and transitions (`use-head-css-gating-client.spec.tsx` is unchanged and passes).
- The fragment ledger (`packages/solid/src/client/hydration.ts`) now resumes a boundary when its swap lands, not when `_fr` settles. `fragmentParked(id)` is true for a record that settled OK while its content template and `pl-*` placeholder are both still in the document. Such a boundary takes the pending path, and `whenRevealed` waits for the ledger's reveal notification before resuming. Only a settled-OK hydrating resume waits: a rejected or truncated fragment keeps its old path. `fragmentPending` reuses the same test.

Size (re-measured after rebasing over #3605/#3606; matches CI byte-for-byte): the web changes are 0 B in every fixture. The ledger change costs +36 B (hydrating, no stores: 20,919 B against `next`'s 20,883) and +10 B (hydrating with every store family: 31,155 B against 31,145), brotli. Caps raised to 20.95 KB and 31.2 KB with notes in `scripts/size/.size-limit.js`. Note: `next` at ed60f054a already exceeds the hydrating (no stores) cap by 33 B and the observe tier by 5 B — #3606 landed over #3605 without a re-measure — so this PR also lifts the observe tier to 17.5 KB (0 B from this PR) with a note attributing it to #3606; otherwise the size gate cannot go green on any PR head.

Verification, all failing on `next` and passing here:

- Parity-harness scenarios `use-head-stylesheet-pending` and `use-head-stylesheet-streamed-boundary`, each in loaded and streamed mode (`packages/web/test/harness/scenarios.tsx`), fail on `next` with the `_config` crash.
- A document-shell case (`StyledIsland`) with a full `<head>`, a shell sheet still loading, and a late boundary whose swap waits on its own sheet. It asserts no crash, no early reveal, exactly one late button (the server's node), live clicks and no warnings. It fails without the ledger change (early reveal and duplicate content) and without the hydrating skip (the boundary never claims).
- `test/hydration/wait-asset-hydration.spec.tsx` calls `waitAsset` directly during `hydrate()`.
- `test/hydration/truncated-stream-template.spec.tsx` and `truncated-stream-late-boundary.spec.tsx` cover a stream cut after the content template arrived, with the boundary registering before and after the cut. Neither fails on `next`; both guard against the first version of the ledger change hanging hydration there.

Suites run locally with the wasm compiler: web client (898), web hydrate (199), solid (659), signals (3,574), size-limit (all scenarios pass). The web server suite has 15 failures in the diagnostics-records and #3557 escaping specs; the same 15 fail on an unmodified `next` checkout with the same compiler build, so I believe the local wasm binary is older than `next`. In headless Chromium, `/ssr?jsdelay=0`, `/ssr?jsdelay=2500`, `/ssr?entry=defer` and `/csr` all hydrate with both buttons live and one late button, for both a shell-registered and a style-gated late sheet.

Still open:

- With an async entry and a stylesheet still loading in `<head>`, `_$HY.r` is empty when `hydrate()` starts. The pending sheet blocks the inline records `<script>` from running but not the async module. In Chromium this shows up as `Hydration key miss for "210"` on the boundary fallback and a duplicate `<link>` for the late sheet. It happens with only the `transparent` change too, and the page still works, so it belongs in its own issue.
- Any other memo created without an owner during hydration still crashes the same way. Treating a null owner like `transparent` in `hydratedCreateMemo` would fix that whole class instead of this one caller.

## Public API changes

- `useHead` stylesheet reveal-gating is skipped while `sharedConfig.hydrating` is true (hydration claims DOM the browser already painted; the server gated the paint). Documented-behavior carve-out; no FOUC introduced. The carve-out is stated in the `useHead` JSDoc.
- No new or removed exports, props, options, or diagnostic codes. `waitAsset`'s gate memo now passes `{ transparent: true }` to `createMemo` — an internal change to an existing seam, not a signature change.


ryansolid:
> Merged as fd8b3df1c — thanks, this was the right fix. We reproduced the 6 red on `next` (the `_config` throw, not just the halt), and the `transparent` gate plus the hydrating skip in `useHead` are what we'd have written.
>
> What we did on top rather than asking for a rebase: rebased over #3605/#3606 and re-measured the caps (+36 B / +10 B on the two hydrating scenarios; `next` itself was already over on two caps from #3606, so those are noted and lifted in the same file); added a `## Public API changes` section to the body; and stated the hydration carve-out in the `useHead` JSDoc.
>
> Two follow-ups we'll file separately: treat a null owner like `transparent` in `hydratedCreateMemo` / `hydratedCreateSignal` (your own note — that's the durable fix for the whole class), and the async-entry empty-`_$HY.r` key miss you described.
>
> — *Claude via Cursor*

Issue #2987, opened 2026-08-10
2.0.0-beta.32 | The case of the silent error.  read after `await` of async memo in and async signal.


mizulu:
### Describe the bug

reading an async computed memo `m` in an async signal `s` 
after an `await`, is not throwing an exception. 

the read has no owner or observer, the `s` signal will never settle 
when `m` settles 

### Your Example Website or App

https://s.olid.uk/id/9YSlwUjhQ5OnkCwwcfy0ZQ

### Steps to Reproduce the Bug or Issue

see playground 

### Expected behavior

a read of an un settled async computed, should throw when not read in an owner scope.

similar to how it throws if we read it in a setTimeout callback 

### Platform

.

### Additional context

```
function App() {

  const m = createMemo(async () => {
    await new Promise(r => setTimeout(r, 100))
  })

 
  const [s, ss] = createSignal(async () => {
    await Promise.resolve()  // yield async 

    m()      // read without owner, still loading,  silently throws 

    // never reached.  and s never settles 
    return Math.random()
  })


  return (
    <div class="p-2">
      <Loading fallback={"Loading..."}>
        Value is : {s()}
      </Loading>

    </div>
  );
}
```

ryansolid:
Fixed on `next` in b37d19a9 — dev builds now fail loud on this instead of hanging.

**What was happening.** Your instinct was right that the read is forbidden, but the error wasn't so much swallowed as *reinterpreted*. When an async computation's flight rejects with `NotReadyError`, the machinery deliberately treats it as *pending*, not an error — because there's a legitimate case: a post-`await` **re-read** of a source you already read before the first `await`. There, the tracked read created a dependency edge, so when the source settles, the settle sweep walks its subscribers, finds your computation, and re-runs it.

Your case is a post-`await` **first** read: untracked, so no edge exists. The node gets marked pending-on-`m`, `m` settles and sweeps its subscribers — your signal isn't among them — and nothing ever happens again. The `Loading` boundary waits correctly on a node that can never be retried. Worse, `isPending` reads `false` while it hangs, so the wedge was invisible to every probe. (The setTimeout case throws loudly only because no machinery wraps the callback to catch and reinterpret the throw.)

**The fix.** At rejection time, dev builds check whether the retry can ever arrive — is the error's source among the node's dependencies, or carried by any dependency's pending chain? If not, the rejection converts to a descriptive error that propagates to `Errored`/logging:

[...]

PR #3602, opened 2026-09-23
feat(signals): warn in dev when an async computation reads after await

brenelz:
## Summary

An async computation only tracks reads made before its first `await`. A signal, memo, or store property first read after an `await` returns its current value, but the computation never re-runs when that value changes, so the result goes stale with no signal to the developer. The existing dev error from #2987 only covers a source that is still loading its first value (that read throws and wedges); a source that is ready, or refetching with an old value, is silent.

```ts
const items = createMemo(async () => {
  const res = await fetchItems();
  return res.filter(item => item.tag === query()); // query() is never tracked
});
```

This adds a dev-only `UNTRACKED_READ_AFTER_AWAIT` warning for that read. JavaScript has no async context, so the read cannot be attributed to its computation directly. Instead, in dev, a native-promise flight is consumed by a uniquely named async function (`__solidAsyncCompute_<id>`) rather than `.then`. V8's async stack traces then carry that frame on every read in the computation's continuation, including reads inside awaited helpers. Solid's settle callbacks run inside that function's `await`, which resumes on the same tick as `.then`, so dev and production settle timing stay identical and `_inFlight` keeps its original identity.

The check runs from core `read()` and the store proxy `get` trap. To keep dev reads cheap, it:

- skips reads that have an owner context (continuations always resume with none, while mount, flush, and effect reads have one);
- skips reads made by code Solid itself runs on the continuation's stack — an effect callback (the continuation called `flush()`), a cleanup (it called `dispose()`), or an `action()` body (it invoked the action) — via two dev-only depth counters in `core.ts`;
- reads V8 call-site objects through `Error.prepareStackTrace` instead of formatting `.stack`;
- caches the negative answer ("no flight on the stack") once per microtask window, so untracked reads elsewhere while a flight is open cost one capture. A positive answer is taken afresh per read inside a flight, because one drain resumes every sibling continuation whose promise settled and the flight on the stack changes between reads in the same window. The reset can land a few jobs late, which can miss a warning but never invent one.

It stays quiet for sources also read before the first `await`, for `untrack()`, for superseded flights, and for sources still loading their first value (the #2987 error covers those). It warns once per computation per signal or memo, and once per computation per store — a row walk (`items.map(i => i.name)`) after an `await` produces one warning naming the first untracked key, not one per row proxy. Store checks only consider own data keys, so `map` and array indices never warn on their own. A `latest()` read after an `await` names `latest(<source>)` (the shadow is named after its source in dev). Flight entries hold the computation and promise through `WeakRef` and are swept as the registry grows, so a flight that never settles does not pin its node.

Production builds are unchanged: every addition is behind `__DEV__`, and the prod and observe trees of `@solidjs/signals` are byte-identical to `next` (`diff -rq`), as are the `solid-js` and `@solidjs/web` dists. The tree-shake floor test passes and every size scenario is within its cap.

**V8 only.** The attribution needs `Error.captureStackTrace` and V8's async frames, so the warning fires on Chromium browsers, Node, Deno, and Bun, and is silent in Firefox and Safari. It is complementary to a lint rule, not a replacement: lint sees the `await` and the read in one function body; this check follows the read into sync helpers, awaited async utilities, `.then` callbacks, and `Promise.all` — places lint cannot see — because the runtime knows which computation resumed.

Known gaps: `async function*` bodies are not covered (async-iterable computations are not wrapped; the engine does not track between-yield reads either); non-native thenables and a helper returned without `await` (`return load()`) are not checked; a read more than ~40 synchronous frames deep is missed (`stackTraceLimit = 50`, justified in a comment); and an unowned read queued in the same microtask window just before a continuation can hide that continuation's read. The last two shapes are pinned as tests.

The alternative that would make these reads work rather than warn is a generator form (`createMemo(function* () { const res = yield fetch(); return filter(res, query()); })`), where Solid resumes the generator and can restore tracking before each step, as `action()` already does. That would be a separate proposal.

## Public API changes

- New dev diagnostic code `UNTRACKED_READ_AFTER_AWAIT` (`kind: "async"`, `severity: "warn"`), added to the `DiagnosticCode` union and emitted through `OBSERVE.diagnostics`. Dev-only; no runtime surface in the prod or observe tiers.
- No other exports, props, options, or signatures change. (`latest()` shadows carry a dev-only `_name` of `latest(<source>)` for diagnostics; `_name` is an internal field.)

- `packages/signals/tests/untracked-read-after-await.test.ts` (30 tests) covers stale signal and store reads (asserting the computation really does not re-run), awaited helpers, projections, refetching sources, arrays and prototype methods, the quiet cases above, cache edge cases, a never-settling flight being released, the known false negatives, and — added in review — the three false-positive shapes (`flush()` → effect callback, `dispose()` → `onCleanup` and effect-returned cleanup, `action()` body) plus recovery from a cleanup that throws mid-teardown; sibling continuations resolving in one drain (distinct sources, and the same source) each attributed to their own computation; once-per-store dedupe across nested paths and row walks; `latest()` naming; and a timer-backed `await` (warns for the read after it, never for a read before it or in a timer callback).
- Full suites on the rebased branch after `pnpm build`: signals 219 files / 3,626 passed; solid 685 passed; web client 898, server 1,112, hydrate 239 passed; solid and web type tests pass. `scripts/size`: every scenario within cap; `@solidjs/signals` `dist/prod` and `dist/observe` byte-identical to `next`.
- Dev read cost against the built dev dist: 100k untracked reads outside a continuation while a flight is open, 2.7 ms (one capture per window); inside a continuation, ~2.5 µs per read (`Error.captureStackTrace` with async frames — the floor; 10k distinct reads after an `await` ≈ 30 ms including 10k emits). Dev dist grows by ~2.9 KB brotli (`dev.js` +325 B, `dev-shared.js` +2,604 B).

🤖 Generated with [Claude Code](https://claude.com/claude-code); reviewed and finished with Claude via Cursor.


ryansolid:
> Landed as eb3d6995 — thank you, this is a genuinely clever mechanism. Verified before merge: `@solidjs/signals` prod and observe trees byte-identical to `next`, every size scenario within cap, signals/solid/web (client, server, hydrate) suites and type tests green on the rebased branch, and the real case (`await sleep(0); return count()` warns; `count()` before the await doesn't; a `setTimeout` callback read doesn't).
>
> Three behavior fixes went in on your branch, each with failing-first tests:
> 1. **False positives from Solid-run callbacks.** A continuation that called `flush()`, `dispose()`, or an `action()` had the effect callback's / `onCleanup`'s / action body's reads blamed on the awaiting memo. Two dev-only depth counters (`callbackDepth`, `disposalDepth`) now mark those scopes; the disposal one is repaired on the next microtask if a cleanup throws.
> 2. **Sibling continuations collapsing.** The per-window cache took the first continuation's id for every memo resuming in one drain (five memos → one warning; two reading the same source → one). The cache now only remembers the negative answer; positive answers are captured per read (~2.5 µs, dev-only, inside continuations only).
> 3. **Store noise.** Dedupe is per root store rather than per row proxy, so `items.map(i => i.name)` after an `await` is one warning instead of N+1.
>
> Nits: the `latest()` shadow is named `latest(<source>)` in dev, `stackTraceLimit = 50` has its justification, and the docs now carry the codes-table row plus the V8-only note (Chromium, Node, Deno, Bun; silent in Firefox/Safari) framed as complementary to lint, and the `async function*` gap. CodSpeed's flagged store/mount benches were all under 10% with the differing-environments caveat.
>
> — *Claude via Cursor*

Issue #3609, opened 2026-09-23
hydration: treat a null owner like `transparent` in `hydratedCreateMemo` / `hydratedCreateSignal`


ryansolid:
Follow-up promised in [#3600 (comment)](https://github.com/solidjs/solid/pull/3600#issuecomment-5791373769). Robustness fix; no public API change.

## Summary

While `sharedConfig.hydrating` is true, `createMemo` and function-form `createSignal` are the hydration facades. Both hand off to `hydrateSignalLike`, which peeks the next child id off `getOwner()!`. With no owner (`runWithOwner(null, …)`, or any detached creation) that is `childId(null, …)` → `TypeError: Cannot read properties of null (reading '_config')`, and the reactive system halts (`REACTIVITY_HALTED`).

#3600 fixed the one caller it hit (`useHead`'s `waitAsset` gate) by passing `{ transparent: true }` at the call site (`packages/web/src/client.ts`, `waitAsset`, ~L153–165). The general fix belongs in the facades: an ownerless node has no id counter to consume, so there is nothing to hydrate positionally and it should take the same non-hydrating path `transparent` already takes.

## Code (`origin/next` @ fd8b3df1c)

`packages/solid/src/client/hydration.ts`:

- `hydratedCreateMemo` (~L1268): `if (!sharedConfig.hydrating || options?.transparent) return coreMemo(...)`, else `hydrateSignalLike(coreMemo, …)`.
- `hydratedCreateSignal` (~L1275): `if (typeof fn !== "function" || !sharedConfig.hydrating) return coreSignal(...)`, else `hydrateSignalLike(coreSignal, …)`. Note this one has no `transparent` check at all.
- `hydrateSignalLike` (~L1154): the hybrid branch reads `peekNextChildId(getOwner()!)` (~L1196); the default path calls `hydrateSignalFromAsyncIterable` (~L764), whose first two lines are `const parent = getOwner()!; const expectedId = peekNextChildId(parent);`.
- `hydratedEffect` (~L1480) already has the `!sharedConfig.hydrating || options?.transparent` short-circuit, so effects have the escape hatch but memos/signals only honor it for `createMemo`.
- `lazyHydrationLookup` (~L1536) already does the null-safe version of this: `const key = o && o.id != null ? peekNextChildId(o) : undefined;`.

`packages/signals/src/core/owner.ts`: `childId` (~L220–226) dereferences `owner._config` unconditionally; `peekNextChildId` (~L260) is a thin wrapper.

[...]

PR #3611, opened 2026-09-23
fix(solid): treat null owner as transparent during hydration

Yahiro025:
## Summary

Fixes #3609

While hydrating, a missing owner should behave like `{ transparent: true }`. Ownerless `createMemo` / function-form `createSignal` were calling `peekNextChildId(null)` and throwing `TypeError: Cannot read properties of null (reading '_config')`. Transparent function-form signals also failed to adopt the parent serialized value.

This change:
- In `hydratedCreateMemo` / `hydratedCreateSignal`, when `sharedConfig.hydrating` and there is no owner, return the core memo/signal (same path as transparent)
- Honor `transparent` on function-form `createSignal` as well
- Adds ownerless hydrate regression coverage and a changeset

```bash
# packages/web
pnpm exec vitest run --config vite.config.hydrate.mjs test/hydration/ownerless-hydration-3609.spec.tsx
# 9 passed

# packages/solid (related hydration suites)
pnpm exec vitest run test/client-hydration.spec.ts test/refresh-hydration.spec.ts test/hybrid-memo-handoff.spec.ts test/hybrid-store-handoff.spec.ts test/shallow-hydration.spec.ts test/hydration-root-snapshot-scope-3504.spec.ts test/multi-root-hydration.spec.ts
# 188 passed
```

Before the fix, the two ownerless `hydrate()` cases threw on `null._config`. After, they pass; transparent function-form `createSignal` adopts `parent-serialized`.


ryansolid:
> Thanks for this fix, Bennett — and sorry, we overlapped you. #3616 (https://github.com/solidjs/solid/pull/3616, `b149cd257`) landed the same change while this PR was open, extended across every hydration facade plus a hybrid-latch change for non-stream shapes. Your PR was open first; not checking open PRs before starting on #3609 is on us.
>
> Your diagnosis and approach matched what shipped exactly: a null owner during hydration takes the transparent path instead of hitting `peekNextChildId(null)`, and `transparent` is honored on function-form `createSignal`. The ownerless `hydrate()` regression coverage you added is the same shape as what's now in the tree.
>
> Closing as superseded, since the squash commit is already merged and can't be amended to co-author you. We'd genuinely welcome your next PR — if you spot something else in hydration, please open it and we'll be more careful to check first.
>
> — *Claude via Cursor*

Issue #3610, opened 2026-09-23
hydration: async entry runs `hydrate()` before the records `<script>` — empty `_$HY.r` makes a streamed `<Loading>` fall through (key miss, refetch, unclaimed fragment)


ryansolid:
Follow-up promised in [#3600 (comment)](https://github.com/solidjs/solid/pull/3600#issuecomment-5791373769); the "still open" item in that PR's body. Reproduced on a clean `origin/next` (fd8b3df1c) worktree; repro below.

## Summary

`hydrate()` can start while `_$HY.r` is still the empty `{}` that `HydrationScript` bootstraps. The serialized records (`_fr` fragment declarations, async-source stubs) are written by a separate inline classic `<script>`, and a `<link rel="stylesheet">` that is still loading blocks every parser-inserted classic script after it — but not an `async` module script. `@solidjs/vite-plugin`'s start mode injects the client entry as `<script type="module" src=… async>`, so in the default setup the entry regularly runs before the records do.

With no records, `hydratedCreateLoadingBoundary` finds neither `id` nor `id_fr` and falls through to the plain `coreLoadingBoundary`, i.e. it hydrates as if the server had rendered the content synchronously. The DOM actually holds the streamed fallback + `pl-<id>` placeholder. Result: a `Hydration key miss` on the boundary fallback, a real client refetch of the async source, and — when the blocked scripts finally run — the server's content arrives as a fragment nobody claimed. The page works, but it is duplicated/inert work and dev logs a mismatch that points users at `renderId`/`<Hydration>`, which is the wrong diagnosis.

In #3600's app (Chromium, `/ssr?jsdelay=0`) this surfaced as `Hydration key miss for "210"` on the boundary fallback and a duplicate `<link>` for the late boundary's `useHead` stylesheet (the boundary re-rendered fresh, so its `useHead` mounted a second link next to the server's head patch).

## Where the key is produced (server) — `packages/web/src/server.ts`

- `generateHydrationScript` (~L4735–4740) emits `_$HY={events:[],completed:new WeakSet,r:{},fe(){}}` followed by the `<!--xs-->` marker. This is the only thing guaranteed to have run before an async entry.
- Records are `_$HY.r[key] = …` assignments inside `<script>` tasks. Pre-shell pending stubs (fragment `_fr` promises, thrown async sources) batch into one write under `STUB_BATCH_KEY` (`"$B"`) plus a spreader task (~L2021–2052, `flushStubBatch`).
[...]

PR #3615, opened 2026-09-23
fix(web): hydrate() waits for the shell records when a stylesheet delays them (#3610)

ryansolid:
Fixes #3610.

## Repro

`hydrate()` can run before the document's records `<script>` has. The serialized records (`_fr` declarations, async-source stubs, `_assets`) are written by classic inline `<script>`s; a still-loading `<link rel="stylesheet">` blocks every parser-inserted classic script after it, but not an `async` module entry, which runs on arrival. `hydrate()` then reads the bootstrap's empty `_$HY.r`: `hydratedCreateLoadingBoundary` finds neither `id` nor `id_fr` and falls through to the core boundary, i.e. hydrates as if the server rendered the content synchronously. In dev: `Hydration key miss for "310"` on the boundary fallback; a real client refetch of the async source; and when the blocked scripts finally run, the server's fragment arrives with no claimant (the ledger holds the `$df` swap — template and `pl-*` placeholder stay in the document inert beside a fresh client copy).

The new spec (`test/hydration/hydrate-wait-for-records-3610.spec.tsx`) uses the issue's recipe verbatim: the `loading-late-fragment` stream fixture with the shell's records script withheld until after `hydrate()`. On `next` it fails at the first assertion (the root starts against empty records); on this branch: no warning, zero client fetches, the server's `<div>item 1</div>` is the one node in the document (adopted through the `$df` swap), template gone.

## Where the race actually lives (and why the server flag is shaped the way it is)

While implementing the approved design I checked where the bootstrap and the records script sit relative to each other, because the pending state has to be *observable* between them:

- **JSX document (`<HydrationScript />`)** — `assembleDocument` splices the shell's records script at `<!--xs-->`, which `generateHydrationScript` emits immediately after the bootstrap's `</script>`. The two are adjacent: a stylesheet can block both or neither, and an `async` entry injected at `</head>` isn't even parsed while the parser is blocked at the bootstrap. There is no window in which a browser sees "bootstrap ran, records didn't" — so in this layout no flag is needed, and none is emitted. The server spec pins the adjacency on a streamed document with a stylesheet in `<head>`.
- **Hand-built document (`generateHydrationScript()` in the host's template)** — the render appends its records script to the end of its output (no marker), so the host's stylesheet(s) and `async` entry sit between the bootstrap and the records. This is the layout where the race is real. The render cannot reach the host's bootstrap to say "records follow", so the flag can't be "set only when a records script will be emitted" as literally stated. What the bootstrap *can* say is "records, if any, are placed apart from me", and the **parser** supplies the other half of the proof: while `document.readyState === "loading"` a parser-inserted script may still be pending; once the parser finishes, none can be (the same reasoning the truncation sweep at `DOMContentLoaded` already rests on).

So the design as built:

## Design

**Server** — `generateHydrationScript()` emits the bootstrap with `p:1` (records may be parser-pending). `<HydrationScript />` emits it without (its records are spliced right after it). The flag's shape and meaning are documented in a comment at the bootstrap (`hydrationBootstrap` in `server.ts`); no records-script bytes change, and no `__artifacts__` change. A hand-constructed `_$HY` (tests, hosts that build their own) keeps today's default: absent means ready.

**Client** — `hydrate()`: if `_$HY.p` is set, the parser is still running, and `_$HY.r` has no key yet, park the start on whichever comes first:
- the first `_$HY.r` write — a `Proxy` on `r` whose `set` trap writes through, hands the bootstrap's own object straight back, and queues the start on a microtask (so the *whole* records script's writes are visible to the claim walk; the records script re-reads `_$HY.r` per statement, so nothing else ever sees the proxy);
- `DOMContentLoaded` — the parser finished, so no records script is coming.

The rest of `hydrate()` moved unchanged into `hydrateRoot`. Later `$df` / record scripts register into a live runtime exactly as before.

## Islands default

Post-load hydration never waits: deferred and module (non-`async`) entries run after parsing, `readyState !== "loading"`, straight through. A JSX document never waits, records or not, even mid-parse. The one case that waits when it didn't before: a hand-built document with **no** records whose `async` entry fires mid-parse — it now starts at `DOMContentLoaded` instead of immediately (the bootstrap can't tell that page from one whose records are stuck behind a stylesheet). Pinned as its own test.

Known narrowing: a page composed of several independent renders (one bootstrap, N records scripts) only gates on "no record yet at all" — if island A's records ran and island B's are still blocked, B doesn't wait. Same as today for B; noted for a follow-up if it shows up.

## Public API changes

- **No signature change.** `hydrate(fn, node, options?) => () => void` is unchanged; the dispose function is **still returned synchronously** and now also cancels a deferred start.
- **Documented behavior:** `hydrate()` may start asynchronously in a hand-built document (bootstrapped by `generateHydrationScript()`) when the records `<script>` has not yet run while the parser is still working; it starts on the first record write or `DOMContentLoaded`. Stated in the `hydrate()` JSDoc and in `documentation/solid-2.0/12-ssr-http.md`.
- **Wire:** `generateHydrationScript()` output gains `p:1` in the `_$HY` literal (4 bytes). `<HydrationScript />` output is byte-identical to before. `_$HY.p` is a new documented key on the page-global.
- No new/removed exports, props, options, or diagnostic codes.

## Tests

- `test/hydration/hydrate-wait-for-records-3610.spec.tsx` (6): (1) records stalled behind `hydrate()` → parked, then resumes on the records write with the microtask ordering pinned, interceptor withdrawn, server fragment adopted, 0 fetches, no warnings; (2) control, records before `hydrate()` → synchronous, same end state; (3) JSX document, no records, mid-parse → synchronous; (4) hand-built, no records, parser finished → synchronous; (5) hand-built, no records, mid-parse → released at `DOMContentLoaded`; (6) dispose before the records → cancels, later scripts start nothing, dispose idempotent.
- `test/server/hydration-bootstrap-3610.spec.tsx` (3): the two bootstrap shapes (with options), and bootstrap/records adjacency in a streamed JSX document with a head stylesheet.
- Suites: web client 898 / server 1115 / hydrate 245 (+1 expected fail) under both the native compiler and `JSX_COMPILER=babel`; solid 685; signals 3626; h/html/element/universal/babel-plugin/diagnostics; `test-types` and `typecheck` for web and solid. The server harness regenerated 7 `__artifacts__` with the known `<script>`-coalescing order nondeterminism (task order inside a chunk only, no content change) — excluded from the commit.

## Size

Measured against a clean `origin/next` build (brotli):

| scenario | next | this PR | Δ |
|---|---|---|---|
| app: hydrating (no stores) | 20,919 B | 21,056 B | +137 B |
| app: hydrating + every store family | 31,155 B | 31,284 B | +129 B |
| render + one signal / CSR / observe / frames / signals | — | — | −8 / +1 / −19 / 0 / 0 (brotli layout) |

Web only, in `hydrate()` itself — the one entry every hydrating app retains. Caps 20.95 → 21.10 KB and 31.2 → 31.30 KB with dated notes in `scripts/size/.size-limit.js`.

— *Claude via Cursor*


ryansolid:
> Reverted in #3618 (`777916a36`). The ordering this guarded against does not occur in a correctly assembled document: the JSX path splices the records script directly after the bootstrap, and the hand-built path's entry is a deferred module that runs after the records regardless of stylesheets. Only an `async` entry in a hand-built document can run early, which is a misconfiguration we do not recommend — now documented beside `generateHydrationScript()` in `12-ssr-http.md` rather than guarded at runtime. #3610 is closed as not planned with the full reasoning.
>
> — *Claude via Cursor*

PR #3622, opened 2026-09-23
fix(web): stop sending undefined as null before a trailing string

brenelz:
## Summary

A server-function call whose last argument is a string sent an earlier `undefined` argument to the server as `null`, so default parameters and `=== undefined` checks broke:

```ts
const impl = async (id: number, limit = 10, q = "") => `id=${id} limit=${limit} q=${q}`;
await impl(1, undefined, "milk");   // "id=1 limit=10 q=milk"
await search(1, undefined, "milk"); // "id=1 limit=null q=milk"
```

`initializeResponse` has a fast path for bound calls that end in a natural HTTP encoding (`action.with(id)` posting a form): the leading arguments go in `?args=` as JSON and the last argument is the body. It maps `undefined` to `null` before the `isJSONSafe` check, and `getHeadersAndBody` counts a plain string as a natural encoding. A string call can only reach that path when a leading argument is not JSON-safe, so in practice it got there only because of the `undefined` mapping. This happened with `enableRichArguments()` enabled too.

The fix keeps strings out of that fast path, so an `undefined` before a string goes to the codec. With rich arguments the function now receives `undefined` and the default applies. Without them the call throws the same "sent as JSON by default" error that `search(1, undefined)` already throws. Bound form actions keep their current shape: before a FormData, URLSearchParams or File body, `undefined` is still sent as `null`, which matches the no-JS action url the router renders with `hashKey(args)`. The existing `server-functions-body-formats` spec pins that behaviour.

Dropping the mapping altogether was not chosen: without rich arguments every bound form action with an `undefined` binding would start throwing. Limiting the fast path strictly to FormData, URLSearchParams and File was not chosen either: a trailing Blob would then go to the codec, which cannot serialize a Blob, so `f("folder", blob)` would break.

The docs paragraph on argument encoding in `documentation/solid-2.0/10-server-functions.md` now lists a top-level `undefined` among the values that need rich arguments.

### Open questions for review

- Without rich arguments, `search(1, undefined, "milk")` used to resolve (with `null`) and now rejects. This follows the `isJSONSafe` rule of refusing rather than silently changing a value, but it is a runtime change for default-config apps that relied on the `null`.
- A call ending in a Blob, Uint8Array or ArrayBuffer still sends a leading `undefined` as `null`. The codec could carry the Uint8Array and ArrayBuffer cases, but not Blob, so this PR leaves all three unchanged.
- `serverFunctionActionUrl(fn, undefined)` throws rather than mapping to `null`, so the JS and no-JS paths of this package already disagree on `undefined` in bound arguments. The router's `hashKey` is what maps it to `null`.

The new `packages/web/test/server/server-functions-undefined-arguments.spec.tsx` covers:

- In the default config, `search(1, undefined, "milk")` rejects and the function never runs. Before the fix it resolved with `limit=null`.
- With rich arguments enabled, `search(1, undefined, "milk")` returns `id=1 limit=10 q=milk` through the codec. Before the fix it returned `limit=null`.
- `search(1, undefined)` still rejects.
- `search(1, 5, "milk")` is still sent as the plain JSON body `[1,5,"milk"]`.
- Bound calls ending in FormData, URLSearchParams and File keep `?args=[...,null]` and their body format tag.

Against the unfixed source, 2 of the tests failed, both with `limit=null`. With the fix, all 7 pass.

From `packages/web`, after `pnpm build`:

```
npx vitest run                                    # 101 files, 898 passed
npx vitest run --config vite.config.server.mjs    # 111 files, 1119 passed, 2 skipped
npx vitest run --config vite.config.hydrate.mjs   # 37 files, 243 passed, 1 expected fail
pnpm run test-types                               # clean
```

`scripts/size`: `npx size-limit` shows all 10 entries passing with no size change (the frames entry does not retain `initializeResponse`). A bundle importing `createServerReference` grows by 24 B minified (21 B gzip, 26 B brotli).


ryansolid:
> Verified locally on a rebase onto `next`: the fix is exactly the trailing-string guard on the bound `?args=` fast path, FormData/URLSearchParams/File controls pin the unchanged bound shape, and all web suites pass under both compilers plus type tests and `size-limit`.
>
> Ruling on the open question: routing `undefined` to the codec is correct. Under the default config it hits the pre-existing "sent as JSON by default" throw, the same as any other `undefined` argument — the fast path was the exception, not the rule. Recorded in `MATRIX.md`.
>
> Thanks, Brenley.
>
> — *Claude via Cursor*

Issue #3414, opened 2026-09-14
[2.0.0-rc.8] Errored fallback rendered during initial SSR has an unresponsive counter


everton-dgn:
An `Errored` fallback renders in the initial SSR response, but clicking its counter button leaves the text at `Count: 0`. I can reproduce this with both a synchronous throw and an asynchronous rejection. The reduced app has no Router, `query()`, `lazy()`, Nitro, or Vite+.

@ryansolid, I found this while adding an error page and reduced it to a local signal and a button whose handler only calls `setCount`.

## Reproduction

Reproduction: [solid-error-repro](https://github.com/everton-dgn/solid-error-repro). The reduced app is in `src/CoreApp.tsx`, with tests in `tests/core.spec.ts`.

```sh
pnpm install --frozen-lockfile
pnpm exec playwright install chromium
REPRO_RUN=bug pnpm repro:bug
```

This builds the app with Vite and serves it with `NODE_ENV=production`. On the pinned versions, the test command exits with code 1: 2 failed, 2 passed. Both failures report `Expected: "Count: 1", Received: "Count: 0"` after a 3-second timeout. The tests have no expected-failure annotations.

The synchronous case, simplified from `src/CoreApp.tsx`:

```tsx
import { httpStatus } from '@solidjs/web'
import { createSignal, Errored } from 'solid-js'

function Failure(): never {
  throw new Error('Synthetic synchronous render failure')
}

function Fallback() {
  httpStatus(500)
  const [count, setCount] = createSignal(0)
  return (
    <main>
      <h1>Something went wrong</h1>
      <button onClick={() => setCount(n => n + 1)}>Count: {count()}</button>
    </main>
  )
}

export default function App() {
  return <Errored fallback={() => <Fallback />}><Failure /></Errored>
}
```

The shared document includes `<HydrationScript />`. In `vite.config.ts`, the plugin uses `start: { app }`, `ssr: true`, and `serverFunctions: true`. Building with `--mode core` selects `src/CoreApp.tsx`; the default build selects the larger comparison app in `src/App.tsx`.

## Actual and expected behavior

| Case | HTTP status | Counter after click | Browser `pageerror` |
| --- | ---: | --- | --- |
| Synchronous throw | 500 | `Count: 0` | None observed |
| Async rejection before the initial document is sent | 500 | `Count: 0` | Original rejection message |
| Async rejection inside an inner `Loading` boundary | 200 | `Count: 1` | Original rejection message |
[...]

PR #3624, opened 2026-09-23
fix(flow): Errored builds a thunk fallback inside its own scope, like Show

ryansolid:
Follow-up to #3620, which surfaced this under "Also surfaced" and pinned it without changing it.

## Problem

`<Errored fallback={() => <F />}>` with a **zero-arity** thunk is type-reachable — `() => X` satisfies `(err, reset) => X` — but the runtime read `f.length === 0` as a value thunk and handed it back **unresolved**. The consuming hole then built it on the enclosing owner's counter: the client's `insert` at the statement, the server's `escapeLate` inside the `ssr()` walk after every scoped sibling of the template had reserved its slot. That is symmetric only when nothing scoped follows the boundary in the same element; otherwise server and client keys permute — the fallback's `_hk` lands on the enclosing counter on the server (`main _hk=6`) while the client expects it under the boundary, so the hydrated fallback goes dead. The server spec pinned both halves (silent with nothing scoped after; reported by `UNSCOPED_HOLE_ALLOCATED_IDS` with a scoped hole after).

## Ruling

**`Errored`'s fallback works like `Show`'s function child.** `Show` calls a function child inside its own value memo, so the content's ids nest under the flow control's scope identically on both sides. `Errored` now calls a function-valued `fallback` **whatever its arity**, inside the boundary's own output scope — the client's `_output` computed, the server's mirroring `outputOwner` — the same node the `(err, reset) => X` form always ran under. The raw thunk never reaches the hole, so zero-arity and two-arity fallbacks allocate identically.

The change is a one-token guard removal on both facades (`typeof f === "function" && f.length ? f(err, reset) : f` → `typeof f === "function" ? f(err, reset) : f`). Preserved: reset/retry (via the two-arity or rest-parameter form), error narrowing, the dev `console.error` for a fallback that cannot see the error (value or zero-arity), fallback reactivity (the fallback runs inside the output computed, tracked exactly as the two-arity form is — a zero-arity fallback that reads a signal re-renders), and the server's `escapeLate`/`$esc` flow for deferred holes (untouched; nothing of the boundary's output is a deferred function any more).

## Audit — flow controls with a function-valued `fallback`

| Flow control | `fallback` type | thunk type-reachable? | Shares the hazard at runtime? | Changed? |
|---|---|---|---|---|
| `Errored` | `Element \| ((err, reset) => Element)` | **yes** (`() => X` assignable) | **yes** — `f.length == 0` handed back raw | **yes** (client + server) |
| `Loading` | `Element` | no (`Element` excludes functions via `call?: never`) | latent from JS / a cast only: `() => props.fallback` returns the value, a thunk would reach the hole | no — the #3620 ruling (no production bytes for a type-rejected shape); `UNSCOPED_HOLE_ALLOCATED_IDS` covers it |
| `Show` | `Element` | no | latent, same (returns `props.fallback` raw; the `show-thunk-fallback-under-errored` parity scenario pins the aligned shape via a cast) | no — same ruling |
| `Switch` | `Element` | no | latent, same | no — same ruling |
| `For` / `Repeat` | `Element` | no | latent, same (`fallback: () => props.fallback` accessor into `mapArray` / `repeat`) | no — same ruling |
| `Match` | no `fallback` | — | — | — |
| `Reveal` | no `fallback` | — | — | — |

Only `Errored` exposes a function type on `fallback`, so only `Errored` has the type-reachable form; the rest reach the shape only the way the bare-identifier hole does, and the dev check already names that.

## Tests

- **#3567 harness** (`test/harness/slot-hydration-3567.tsx`): new `errored-thunk-fallback-followed-by-scoped-hole` scenario — an `<Errored fallback={() => <Fallback />}>` (wrapped in a component, as in the pinned spec) followed by a scoped `{props.title}` hole in the same `<section>`; server render → artifact → hydrate under both compilers: zero client-created nodes, every `_hk` claimed, **no** `Hydration key miss`, **no** `UNSCOPED_HOLE_ALLOCATED_IDS`, and the server-rendered `#hdr` button is live. Failed before the fix on both sides (server `button _hk=20` vs client key `3`; the client diagnostic fired). A `logsCaughtError` scenario flag silences and pins the dev `console.error` both sides emit for a fallback that cannot see the error.
- **Server spec** (`test/server/slot-hydration-3567.spec.tsx`): the "reported when a scoped hole follows it" half flipped to "silent, aligned"; the "silent when nothing scoped follows" half kept; new: zero-arity vs two-arity `renderToString` output is byte-identical (same `_hk` layout for the fallback and for the following `{props.title}`, `{count()}` and `{props.children}` siblings), both through a component and directly under the element.
- **Client** (`test/errored.spec.tsx`): zero-arity fallback renders (dev `console.error` pinned); a zero-arity fallback that reads a signal in its body re-renders on change; the fallback's owner sits at the same depth under the component owner for both arities (not under the consuming insert effect); a rest-parameter fallback (`(...args)`, `length` 0) receives `err` and `reset` and reset recovers; the two-arity form is unchanged (accessor, reset, no dev log).
- Regenerated the four existing `errored-thunk-fallback-*` parity artifacts (the fallback's `_hk` now nests one level deeper on both sides, e.g. `010` → `0010`); hydrate passes on them under native and Babel.

Full `@solidjs/web` client / server / hydrate suites under the native compiler and `JSX_COMPILER=babel`; `solid-js` suite; `test-types` for `solid-js` and `@solidjs/web`. All green.

## Size

`scripts/size` against a `next` (`7599885a9`) build in a sibling worktree, bytes brotli:

| scenario | `next` | this PR | Δ |
|---|---|---|---|
| app: hydrating (no stores) | 20,907 | 20,981 | **+74** |
| app: hydrating + every store family | 31,220 | 31,183 | −37 |
| app: CSR | 15,813 | 15,782 | −31 |
| app: CSR, observe tier | 17,455 | 17,471 | +16 |
| app: CSR, observe + attribution | 28,985 | 28,965 | −20 |
| signals scenarios, simple app, frames | — | — | 0 |

The hydrating scenario breached its 20.95 KB cap by 31 B. The minified bundle is 10 B **smaller** in every scenario (61,334 → 61,324 for hydrating) and an identifier-normalised diff of the two bundles is exactly one statement (the removed `&& f.length`); the swing is esbuild's mangled-name reassignment shifting under brotli, the same phenomenon the 2026-09-22 performance-tracks ledger note documents. Cap bumped 20.95 → 21.00 KB with a dated audit note recording the measurement and the normalised diff.

## Docs

- `documentation/hole-owner-id-matrix.md`: the bare-identifier bullet no longer cites the zero-arity thunk as the silent example; it records the ruling and how `Errored` now aligns it by construction.
- `documentation/solid-2.0/08-dev-diagnostics.md` (`UNSCOPED_HOLE_ALLOCATED_IDS`): the same example updated; the diagnostic's semantics are unchanged.
- Stale comments describing the old hand-back (the #3414 harness block, the server `createErrorBoundary` note) updated.

## Public API changes

**None.** No export, prop, option, signature, type or diagnostic code is added, removed or changed. This corrects the behavior of an already-supported, already-typed form of `<Errored fallback>`: a zero-arity (or rest-parameter, `length` 0) function fallback is now called with `(err, reset)` inside the boundary like the two-arity form, instead of being returned as a value for the hole to unwrap. Observable consequences of that correction: hydration keys of a zero-arity fallback and of siblings after the boundary now match between server and client (they did not when a scoped hole followed); a rest-parameter fallback receives its arguments (it was previously called with none by the hole). Tracking of signal reads in the fallback body is the same as before (the hole's transparent effect tracked them; the boundary's output computed does now).

Changeset: `.changeset/errored-fallback-thunk-scope.md` (`solid-js`, patch).


ryansolid:
> CodSpeed verdict before merging with `--admin` over its red check.
>
> The two flagged benchmarks cannot execute this change: `packages/signals/tests/store/props-chain.bench.ts` runs against `@solidjs/signals` source, which this PR leaves byte-identical (`git diff origin/next..HEAD -- packages/signals` is empty), and `packages/web/test/spread-enumerate.bench.tsx` imports `createStore`/`merge`/`spread` and never renders an `<Errored>`. The only runtime change is the removed `&& f.length` guard in `Errored`'s fallback callback (client and server). CodSpeed's report also carries its "Different runtime environments detected" banner, which applies to the benchmarks with significant changes — these two are the only ones. Per the ruling for this PR, a cross-environment swing on untouched code is not a stop condition; treating it as measurement noise.
>
> — *Claude via Cursor*

